PR #10, opened 2025-10-21
Add docstrings to DatasetQuery

MalteEbner:
## What has changed and why?

Add better docstring to DatasetQuery
The other docstrings are updated in separate PRs.

## How has it been tested? - Screenshot

image

[DatasetQuery - LightlyStudio Docs.pdf](https://github.com/user-attachments/files/23015317/DatasetQuery.-.LightlyStudio.Docs.pdf)

### DIFF
--- a/lightly_studio/src/lightly_studio/core/dataset_query/dataset_query.py
+++ b/lightly_studio/src/lightly_studio/core/dataset_query/dataset_query.py
@@ -35,7 +35,8 @@ class DatasetQuery:
     dataset.match(...) # shorthand for dataset.query().match(...)
     ```
 
-    The object is converted to a SQL query that is lazily evaluated when iterating over it or converting it to a list.
+    The object is converted to a SQL query that is lazily evaluated when iterating over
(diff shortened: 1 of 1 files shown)


lightly_studio/src/lightly_studio/core/dataset_query/dataset_query.py
```diff
@@ -20,7 +20,102 @@
...
+    dataset : Dataset = ...
+    query = dataset.query()
+    ```
+    The `match()`, `order_by()`, and `slice()` methods can be chained in this order.
+    You can also access the methods directly on the Dataset instance:
+    ```python
+    dataset.match(...) # shorthand for dataset.query().match(...)
+    ```
+
+    The query is lazily evaluated when iterating over it or converting it to a list.
```
michal-lightly: Minor: I really think we should mention the database:

```suggestion
    The object is converted to a SQL query that is lazily evaluated when iterating over it or converting it to a list.
```

PR #19, opened 2025-10-22
Multimodal: Link tags to SampleTable

michal-lightly:
## What has changed and why?

To be merged into a feature branch.

Link tags with SampleTable instead of ImageTable.

There are still some TODO's to be addressed in follow-ups.

## How has it been tested?

Existing tests.

- [ ] Yes

### DIFF
--- a/lightly_studio/src/lightly_studio/api/routes/api/sample.py
+++ b/lightly_studio/src/lightly_studio/api/routes/api/sample.py
@@ -34,29 +34,6 @@ from lightly_studio.resolvers.samples_filter import (
 samples_router = APIRouter(prefix="/datasets/{dataset_id}", tags=["samples"])
 
 
-@samples_router.post("/samples")
-def create_sample(
-    session: SessionDep,
-    input_sample: ImageCreate,
-) -> ImageView:
-    """Create a new sample in the database."""
-    image = image_resolver.create(session=session, sample=input_sample)
-    # TODO(Michal, 10/2025): Add SampleView to ImageView and then use a response model
-    # instead of manual conversion.
-    return ImageView(
-        file_name=image.file_name,
-        file_path_abs=image.file_path_abs,
-        sample_id=image.sample_id,
-        dataset_id=image.dataset_id,
-        annotations=image.annotations,
-        captions=image.captions,
-        tags=image.sample.tags,
-        metadata_dict=image.metadata_dict,
-        width=image.width,
-        height=image.height,
-    )
-
-
 class ReadSamplesRequest(BaseModel):
     """Request body for reading samples with text embedding."""
 


lightly_studio/src/lightly_studio/api/routes/api/sample.py
```diff
@@ -24,23 +24,37 @@
...
 samples_router = APIRouter(prefix="/datasets/{dataset_id}", tags=["samples"])
 
 
-@samples_router.post("/samples", response_model=ImageView)
+@samples_router.post("/samples")
 def create_sample(
     session: SessionDep,
     input_sample: ImageCreate,
-) -> ImageTable:
+) -> ImageView:
```
michal-lightly: Need to manually convert the output since ImageTable is not yet compatible with ImageView (I did not want to change the API interface).

lightly_studio/src/lightly_studio/models/image.py
```diff
@@ -56,13 +56,6 @@ class ImageCreate(ImageBase):
     """Image class when inserting."""
 
 
-class SampleTagLinkTable(SQLModel, table=True):
```
michal-lightly, line 59: Moved to sample.py

lightly_studio/src/lightly_studio/api/routes/api/sample.py
```diff
@@ -24,23 +24,37 @@
...
+        file_path_abs=image.file_path_abs,
+        sample_id=image.sample_id,
+        dataset_id=image.dataset_id,
+        annotations=image.annotations,
+        captions=image.captions,
+        tags=image.sample.tags,
+        metadata_dict=image.metadata_dict,
+        width=image.width,
+        height=image.height,
+    )
```
ikondrat: I would just drop this endpoint, we don't have sample creation in UI for now and not sure we will have

lightly_studio/src/lightly_studio/api/routes/api/sample.py
```diff
@@ -70,14 +84,35 @@ def read_samples(
     Returns:
         A list of filtered samples.
     """
-    return image_resolver.get_all_by_dataset_id(
+    result = image_resolver.get_all_by_dataset_id(
```
ikondrat, line 64: Not sure why we need conversion here, can we have proper types in return from get_all_by_dataset_id? 

What is further plan with refactoring? 

I would make new endpoint to fetch images

lightly_studio/src/lightly_studio/api/routes/api/sample.py
ikondrat: I would actually make new endpoint, this way we would be able to merge it to main and switch when everything will be ready

ikondrat (approved):
alright lets maybe move like this

but I would maybe next time make migration without touching frontend dependencies
and make clean up on final step

lightly_studio/src/lightly_studio/api/routes/api/sample.py
michal-lightly: I like the idea of creating new routes for image fetching

lightly_studio/src/lightly_studio/api/routes/api/sample.py
michal-lightly: I'll remove

PR #23, opened 2025-10-23
Leonardo lig-7713 Add tag section to annotation detail view

LeonardoRosaa:
## What has changed and why?

Added `Sample` tags section within the Annotation Details

## How has it been tested?

Screenshot 2025-10-23 at 09 20 16

- [x] Yes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Added a bottom status bar with quick links to Lightly docs and support.
 - Improved class docstrings for the most important user-facing classes.
-- Added `Sample` tags section within the Annotation Details
+- Added `Annotation` tags section within the Annotation Details
 
 ### Changed
 - Updated button text to "View sample" in annotation details panel for better clarity.


mihnea-necsulescu:
if I create a tag for samples, it will be shown on the annotation detail view too (for all the annotations):
image
Screenshot 2025-10-23 at 12 19 05

lightly_studio/src/lightly_studio/models/annotation/annotation_base.py
```diff
@@ -147,6 +153,8 @@ class AnnotationLabel(SQLModel):
     instance_segmentation_details: Optional[InstanceSegmentationAnnotationView] = None
     semantic_segmentation_details: Optional[SemanticSegmentationAnnotationView] = None
 
+    tags: Optional[List[AnnotationViewTag]] = None
```
michal-lightly: This should not be optional if possible (I'm not sure it is). No tags should be represented by an empty list.

```suggestion
    tags: List[AnnotationViewTag] = []
```

Note that pydantic will treat the right hand side correctly (this is not possible with regular classes, because there the default is shared).

CHANGELOG.md
```diff
@@ -10,13 +10,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Added a bottom status bar with quick links to Lightly docs and support.
 - Improved class docstrings for the most important user-facing classes.
+- Added `Sample` tags section within the Annotation Details
```
mihnea-necsulescu: shouldn't this be Annotation tags instead?

PR #27, opened 2025-10-23
feat: introduce button to reset to default view in plot

ikondrat:
## What has changed and why?

When we scroll within the plot we can be lost. We need to have button to switch to initial view.
This PR introduces button below the Plot to reset the view
Also fixed issue with plot panel, there was overlap with "documentation panel"
Screenshot 2025-10-23 at 14 55 28

## How has it been tested?
Run the app
Go to samples page and open plot
Move around the plot
Click "Default view" button 
You should be back to the initial view

- [x] Yes

## Visuals

https://github.com/user-attachments/assets/08c27766-44e5-4fa6-8643-b7d713eea731


LeonardoRosaa:
LGTM :)

lightly_studio_view/src/lib/components/PlotPanel/PlotPanel.svelte
```diff
@@ -315,48 +320,50 @@
...
+        <div class="mt-1 flex items-center gap-4 text-sm text-muted-foreground">
+            <span class="flex items-center gap-2">
+                <span class="legend-dot" style={`background-color: ${categoryColors[0]}`}></span>
+                All samples
+            </span>
+            <span class="flex items-center gap-2">
+                <span class="legend-dot" style={`background-color: ${categoryColors[1]}`}></span>
+                Filtered samples
+            </span>
+            <Button variant="outline" size="sm" onclick={reset}>Default view</Button>
```
michal-lightly: We should call it "Reset view" - it sounds like more descriptive (and is also used in other tools).
```suggestion
            <Button variant="outline" size="sm" onclick={reset}>Reset view</Button>
```

PR #33, opened 2025-10-24
Cache 2d-embeddings in the database.

MalteEbner:
## What has changed and why?

Implements the caching of 2d-embeddings in the DB. Implements it via implementing the `get_twodim_embeddings` in the resolvers.

First gets the current `sample_ids` orders by `sample_id`. Then computes their hash based on the full embeddings in the new `sample_embedding_resolver.get_hash_by_sample_ids`. It does the hash computation in the database itself, making it very fast, as the embeddings stay in the DB.

If there is a cache hit, returns the 2d-embeddings directly.

If not, gets the full embeddings via `sample_embedding_resolver.get_by_sample_ids`, computes the 2d-embeddings 

### Manually in GUI
The name of the endpoint is "default" for "embeddings2d/defaut". The first load of around 7k embeddings takes 1.9s (including compute) the second one only 0.18s.
image
### With a benchmark
A fully vibe-coded benchmark yields:

```
(lightly-studio) malteebnerlightly@MacBook-Pro lightly-studio % python lightly_studio/src/lightly_studio/benchmarks/benchmark_twodim_embeddings.py --samples 20000  
Inserted 20000 samples with embeddings in 4.889s.
First call (compute) : 2.990s for 20000 samples
Second call (cached) : 0.243s for 20000 samples
Third call (cached) : 0.287s for 20000 samples
```

Code: [benchmark_twodim_embeddings.py](https://github.com/user-attachments/files/23120185/benchmark_twodim_embeddings.py)

- [ ] Yes

### DIFF
--- a/lightly_studio/tests/resolvers/test_twodim_embedding_resolver.py
+++ b/lightly_studio/tests/resolvers/test_twodim_embedding_resolver.py
@@ -23,7 +23,54 @@ def test__calculate_2d_embeddings__2_samples() -> None:
     assert projected == [(0.0, 0.0), (1.0, 1.0)]
 
 
-def test__get_twodim_embeddings__cache_hit(
+def test_get_twodim_embeddings__no_samples(
+    test_db: Session,
+) -> None:
+    dataset = helpers_resolvers.create_dataset(session=test_db, dataset_name="no_samples")
+    embedding_model = helpers_resolvers.create_embedding_model(
+        session=test_db,
+        dataset_id=dataset.dataset_id,
+        embedding_dimension=3,
+    )
+
+    x_values, y_values, sample_ids = twodim_embedding_resolver.get_twodim_embeddings(
+        session=test_db,
+        dataset_id=dataset.dataset_id,
+        embedding_model_id=embedding_model.embedding_model_id,
+    )
+
+    assert len(x_values) == 0
+    assert len(y_values) == 0
+    assert sample_ids == []
+
+
+def test_get_twodim_embeddings__no_samples_with_embeddings(
+    test_db: Session,
+) -> None:
+    dataset = helpers_resolvers.create_dataset(session=test_db, dataset_name="missing_embeddings")
+    embedding_model = helpers_resolvers.create_embedding_model(
+        session=test_db,
+        dataset_id=dataset.dataset_id,
+        embedding_dimension=3,
+    )
+    helpers_resolvers.create_sample(
+        session=test_db,
+        dataset_id=dataset.dataset_id,
+        file_path_abs="sample_missing_embedding.jpg",
+    )
+
+    x_values, y_values, sample_ids = twodim_embedding_resolver.get_twodim_embeddings(
+        session=test_db,
+        dataset_id=dataset.dataset_id,
+        embedding_model_id=embedding_model.embedding_model_id,
+    )
+
+    assert len(x_values) == 0
+    assert len(y_values) == 0
+    assert sample_ids == []
+
+
+def test_get_twodim_embeddings__cache_hit(
     test_db: Session,
     mocker: MockerFixture,
 ) -> None:
@@ -35,10 +82,6 @@ def test__get_twodim_embeddings__cache_hit(
         embedding_dimension=3,
     )
 
-    embeddings = [
-        [0.1, 0.2, 0.3],
-        [0.4, 0.5, 0.6],
-    ]
     helpers_resolvers.create_samples_with_embeddings(
         db_session=test_db,
         dataset_id=dataset.dataset_id,
@@ -91,7 +134,7 @@ def test__get_twodim_embeddings__cache_hit(
     assert sample_ids_first_call == sample_ids_third_call
 
 
-def test__get_twodim_embeddings__recomputes_when_samples_change(
+def test_get_twodim_embeddings__recomputes_when_samples_change(
     test_db: Session,
     mocker: MockerFixture,
 ) -> None:


lightly_studio/src/lightly_studio/api/routes/api/embeddings2d.py
```diff
@@ -14,9 +14,9 @@
 from lightly_studio.db_manager import SessionDep
 from lightly_studio.models.dataset import DatasetTable
 from lightly_studio.models.embedding_model import EmbeddingModelTable
-from lightly_studio.resolvers import sample_embedding_resolver, sample_resolver
+from lightly_studio.resolvers import sample_resolver
 from lightly_studio.resolvers.samples_filter import SampleFilter
-from lightly_studio.resolvers.twodim_embedding_resolver import _calculate_2d_embeddings
+from lightly_studio.resolvers.twodim_embedding_resolver import get_twodim_embeddings
```
michal-lightly: Minor: Style
```suggestion
from lightly_studio.resolvers import twodim_embedding_resolver
```

lightly_studio/src/lightly_studio/resolvers/sample_embedding_resolver.py
```diff
@@ -81,6 +82,36 @@ def get_all_by_dataset_id(
         .where(SampleEmbeddingTable.sample_id == SampleTable.sample_id)
         .where(SampleTable.dataset_id == dataset_id)
         .where(SampleEmbeddingTable.embedding_model_id == embedding_model_id)
-        .order_by(col(SampleTable.file_path_abs).asc())
```
michal-lightly, line 84: This should stay `file_path_abs` - it is the fix for e2e tests.

lightly_studio/src/lightly_studio/resolvers/sample_embedding_resolver.py
```diff
@@ -81,6 +82,36 @@ def get_all_by_dataset_id(
...
+
+    The combination is deterministic with respect to the provided sample IDs.
+    """
+    if not sample_ids_ordered:
+        return "empty"
+
+    rows = session.exec(
+        select(
+            SampleEmbeddingTable.sample_id,
+            func.hash(SampleEmbeddingTable.embedding).label("h64"),
```
michal-lightly: Minor: Can we call the column more descriptively?

```suggestion
            func.hash(SampleEmbeddingTable.embedding).label("hash_column"),
```

lightly_studio/src/lightly_studio/resolvers/sample_embedding_resolver.py
```diff
@@ -81,6 +82,36 @@ def get_all_by_dataset_id(
...
+        )
+        .where(col(SampleEmbeddingTable.sample_id).in_(set(sample_ids_ordered)))
+        .where(SampleEmbeddingTable.embedding_model_id == embedding_model_id)
+    ).all()
+
+    # Mypy does not get that 'h64' is an attribute of the returned rows
```
michal-lightly: Optional: Would `row["h64"]` solve the mypy issue?

lightly_studio/src/lightly_studio/resolvers/sample_embedding_resolver.py
```diff
@@ -81,6 +82,36 @@ def get_all_by_dataset_id(
...
+    sample_id_to_hash = {row.sample_id: row.h64 for row in rows}  # type: ignore[attr-defined]
+    hashes_ordered = [sample_id_to_hash[sample_id] for sample_id in sample_ids_ordered]
+
+    hasher = hashlib.sha256()
+    hasher.update("".join(str(h) for h in hashes_ordered).encode("utf-8"))
```
michal-lightly, line 127: Is it okay to concat all the hashes? Can't we instead iterate over them and call ``hasher.update` iteratively - or is it too slow?

lightly_studio/src/lightly_studio/resolvers/sample_embedding_resolver.py
```diff
@@ -81,6 +82,36 @@ def get_all_by_dataset_id(
...
+            SampleEmbeddingTable.sample_id,
+            func.hash(SampleEmbeddingTable.embedding).label("h64"),
+        )
+        .where(col(SampleEmbeddingTable.sample_id).in_(set(sample_ids_ordered)))
+        .where(SampleEmbeddingTable.embedding_model_id == embedding_model_id)
+    ).all()
+
+    # Mypy does not get that 'h64' is an attribute of the returned rows
+    sample_id_to_hash = {row.sample_id: row.h64 for row in rows}  # type: ignore[attr-defined]
+    hashes_ordered = [sample_id_to_hash[sample_id] for sample_id in sample_ids_ordered]
```
michal-lightly: Optional: We could simplify this logic. We could verify that sample_ids_ordered are ordered on the input (or just accept that this is an invariant) and then order in SQL. Then there's no need for this reordering.

lightly_studio/src/lightly_studio/resolvers/twodim_embedding_resolver.py
```diff
@@ -1,14 +1,92 @@
...
+        embedding_model_id: Embedding model identifier.
+
+    Returns:
+        Tuple of (x coordinates, y coordinates, ordered sample IDs).
+    """
+    embedding_model = session.get(EmbeddingModelTable, embedding_model_id)
+    if embedding_model is None:
+        raise ValueError(f"Embedding model {embedding_model_id} not found.")
+
+    sample_ids_ordered = list(
```
michal-lightly, line 43: Please add more inline comments in this function, it is nontrivial and has several early returns (not sure if it some functionality can be slit out to a helper).

michal-lightly: E.g. whether it is possible to split these two parts:
- try get cached
- compute and save to cache

lightly_studio/tests/resolvers/test_twodim_embedding_resolver.py
```diff
@@ -1,13 +1,149 @@
...
+        embedding_model_name="cache_hit_model",
+        embedding_model_hash="cache_hit_hash",
+        embedding_dimension=3,
+    )
+
+    embeddings = [
+        [0.1, 0.2, 0.3],
+        [0.4, 0.5, 0.6],
+    ]
+    for index, embedding in enumerate(embeddings):
```
michal-lightly: Optional: I would prefer not having a for loop in tests which loops only twice. We already have a function to create multiple samples, and `create_sample_embedding` can be just called twice.

lightly_studio/tests/resolvers/test_twodim_embedding_resolver.py
```diff
@@ -1,13 +1,149 @@
...
+    test_db: Session,
+    mocker: MockerFixture,
+) -> None:
+    # Create dataset, embedding model, samples, and embeddings.
+    dataset = create_dataset(session=test_db, dataset_name="cache_hit_dataset")
+    embedding_model = create_embedding_model(
+        session=test_db,
+        dataset_id=dataset.dataset_id,
+        embedding_model_name="cache_hit_model",
+        embedding_model_hash="cache_hit_hash",
```
michal-lightly: Can this be left out? Or it causes issues with the embedding manager singleton?

lightly_studio/tests/resolvers/test_twodim_embedding_resolver.py
```diff
@@ -1,13 +1,149 @@
...
+    )
+
+    calculate_spy.assert_called_once()
+    np.testing.assert_allclose(x_first, x_second)
+    np.testing.assert_allclose(y_first, y_second)
+    assert sample_ids_first_call == sample_ids_second_call
+
+
+def test__get_twodim_embeddings__recomputes_when_samples_change(
+    test_db: Session,
```
michal-lightly, line 138: Sidenote: I just notice that we have two very similar fixtures: `test_db` and `db_session`.

lightly_studio/src/lightly_studio/resolvers/sample_embedding_resolver.py
michal-lightly: I'm taking this back - see slack conversation with Mihnea.

lightly_studio/src/lightly_studio/resolvers/sample_embedding_resolver.py
MalteEbner: Given that all the hashes have the same size, the joining is uniqueness-preserving. This can be seen by the fact that the join can be undone and converted to a list of hashes again.

And calling `hasher.update` all at once is faster.

lightly_studio/src/lightly_studio/resolvers/sample_embedding_resolver.py
MalteEbner: I had thought about that.
My goal is to ensure that `get_hash_by_sample_ids` and `get_by_sample_ids` use the same order order of `sample_ids`. However, they are independent functions, thus it cannot be ensured on the sample level. It can only be ensured on the level of the calling function `get_twodim_embeddings` that passes the same `sample_ids_orderd` to them.
This way, both functions do not assume anything about the order of `sample_ids`, but just return in the order of the `sample_ids` they get passed.

lightly_studio/src/lightly_studio/resolvers/sample_embedding_resolver.py
MalteEbner: I am taking this back ;) 
The change is not necessary in this PR

lightly_studio/src/lightly_studio/api/routes/api/embeddings2d.py
MalteEbner: Done, also at another place.

lightly_studio/src/lightly_studio/resolvers/sample_embedding_resolver.py
MalteEbner: Yes, we can

lightly_studio/src/lightly_studio/resolvers/sample_embedding_resolver.py
MalteEbner: No, that causes a runtime error.

```
    sample_id_to_hash = {row.sample_id: row["hash_column"] for row in rows}  # type: ignore[attr-defined]
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

>   ???
E   TypeError: tuple indices must be integers or slices, not str

lib/sqlalchemy/cyextension/resultproxy.pyx:54: TypeError
```

lightly_studio/src/lightly_studio/resolvers/twodim_embedding_resolver.py
MalteEbner: I tried splitting out some functionality, but it did not improve readability in my eyes. Added more comments instead.

lightly_studio/tests/resolvers/test_twodim_embedding_resolver.py
MalteEbner: I switched to using `create_samples_with_embeddings` created by Lukas previously

lightly_studio/tests/resolvers/test_twodim_embedding_resolver.py
MalteEbner: Yes, I could remove it.

lightly_studio/tests/resolvers/test_twodim_embedding_resolver.py
MalteEbner: I agree

lightly_studio/tests/api/routes/api/test_embeddings2d.py
```diff
@@ -64,7 +64,8 @@ def test_get_embeddings2d__2d(
             dataset_id=dataset_id,
         ).samples
     ]
-    assert sample_ids == expected_sample_ids
+    assert len(sample_ids) == n_samples
+    assert set(sample_ids) == set(expected_sample_ids)
```
michal-lightly, line 68: This change was needed because of the change in embeddings resolver ordering. Feel free to keep it or remove it - I need to introduce it for my PR anyway.

lightly_studio/tests/resolvers/test_twodim_embedding_resolver.py
```diff
@@ -1,13 +1,155 @@
...
+        [0.1, 0.2, 0.3],
+        [0.4, 0.5, 0.6],
+    ]
+    helpers_resolvers.create_samples_with_embeddings(
+        db_session=test_db,
+        dataset_id=dataset.dataset_id,
+        embedding_model_id=embedding_model.embedding_model_id,
+        images_and_embeddings=[
+            (SampleImage(path=f"sample_{i}.jpg"), embedding)
+            for i, embedding in enumerate(embeddings)
```
michal-lightly: Optional:
```suggestion
            (SampleImage(path="sample_1.jpg"), [0.1, 0.2, 0.3]),
            (SampleImage(path="sample_2.jpg"), [0.4, 0.5, 0.6]),
```

lightly_studio/tests/resolvers/test_twodim_embedding_resolver.py
```diff
@@ -1,13 +1,155 @@
...
+    calculate_spy.assert_called_once()
+    np.testing.assert_allclose(x_first, x_second)
+    np.testing.assert_allclose(y_first, y_second)
+    assert sample_ids_first_call == sample_ids_second_call
+
+    # Third call after adding a sample without embeddings - should still use cache.
+    helpers_resolvers.create_sample(
+        session=test_db,
+        dataset_id=dataset.dataset_id,
+        file_path_abs="/sample_2.jpg",
```
michal-lightly: Minor:
```suggestion
        file_path_abs="sample_3.jpg",
```

lightly_studio/tests/resolvers/test_twodim_embedding_resolver.py
```diff
@@ -1,13 +1,155 @@
...
+    x_second, y_second, sample_ids_second = twodim_embedding_resolver.get_twodim_embeddings(
+        session=test_db,
+        dataset_id=dataset.dataset_id,
+        embedding_model_id=embedding_model.embedding_model_id,
+    )
+
+    assert calculate_spy.call_count == 2
+    assert x_second.shape == (2,)
+    assert y_second.shape == (2,)
+    assert len(sample_ids_second) == 2
```
michal-lightly, line 198: I noticed we do not have tests for the two branches that return `empty, empty, []`. Would it be possible to add a test?

MalteEbner: I added them.

lightly_studio/tests/api/routes/api/test_embeddings2d.py
MalteEbner: I'll keep it.

PR #34, opened 2025-10-24
Multimodal: Link embeddings to Sample instead of Image

michal-lightly:
## What has changed and why?

Link embeddings to SampleTable instead of ImageTable.

A change in default embedding sorting is necessary. The embeddings can't be sorted by file_name_abs because a general sample does not have the field.

## How has it been tested?

Existing tests.

- [ ] Yes


horatiualmasan (approved):
LGTM

PR #35, opened 2025-10-24
Multimodal: Link metadata to Sample instead of Image

michal-lightly:
## What has changed and why?

To be merged into a feature branch.

Link metadata to Sample instead of Image.

## How has it been tested?

Existing tests.

- [ ] Yes

### DIFF
--- a/lightly_studio_view/src/lib/schema.d.ts
+++ b/lightly_studio_view/src/lib/schema.d.ts
@@ -1750,14 +1750,14 @@ export interface components {
              * @default []
              */
             captions: components["schemas"]["CaptionView"][];
-            /** Metadata Dict */
-            metadata_dict?: unknown | null;
             /** Width */
             width: number;
             /** Height */
             height: number;
             /** Tags */
             tags: components["schemas"]["ImageViewTag"][];
+            /** Metadata Dict */
+            metadata_dict?: unknown | null;
         };
         /**
          * ImageViewTag


lightly_studio/src/lightly_studio/examples/example_metadata.py
```diff
@@ -120,37 +120,37 @@ def demonstrate_bulk_metadata_filters(dataset: ls.Dataset) -> None:
     # Filter by temperature
     print("\n1. Filter by temperature > 25:")
     filter_temp = SampleFilter(metadata_filters=[Metadata("temperature") > 25])  # noqa PLR2004
-    samples = dataset_table.get_samples(filters=filter_temp)
-    print(f"   Found {len(samples)} samples with temperature > 25")
-    for sample in samples[:3]:  # Show first 3
-        print(f" {sample.file_name}: {sample['temperature']}")
+    images = dataset_table.get_samples(filters=filter_temp)
```
michal-lightly, line 123: This example is the only place (except tests) where we:
* Get and set metadata with `sample_table["key"]` accessors
* Filter metadata with filter expressions

We keep this example for now since it is the only way to filter by metadata. When we implement Sample interface to metadata we can simplify a lot of code here (and remove the getters/setters).

lightly_studio/tests/metadata/test_complex_metadata.py
```diff
@@ -27,7 +27,7 @@ def test_complex_metadata(test_db: Session) -> None:
         session=test_db,
         dataset_id=dataset_id,
         file_path_abs="/path/to/sample1.png",
-    )
+    ).sample
```
mihnea-necsulescu, line 30: this is a bit confusing - shouldn't the helper be renamed to `create_image`?

mihnea-necsulescu (approved):

lightly_studio/tests/metadata/test_complex_metadata.py
michal-lightly: Yes, absolutely! I was planning it for a follow-up.

PR #37, opened 2025-10-24
Improve UX for label picker when adding labels

IgorSusmelj:
## What has changed and why?

- Several issues with the UX for creating or selecting classes when adding new labels to an image have been addressed:
  - We match the classes the user types with the available classes (case-insensitive, "dog" matches "Dog")
  - The user can always create a new class if there is no match (for example, the user can create "car" even if there is already a "caravan" class)
  - We auto-select the highlighted class from the dropdown list if the user presses Enter on the keyboard
  - The dropdown for creating or selecting classes is now wider

Old behaviour

https://github.com/user-attachments/assets/7df91948-0d0c-45a1-a6a0-7119b204dd17

New behaviour

https://github.com/user-attachments/assets/0733ba1f-73fa-49c9-8770-7ef41008628b

## How has it been tested?

- updated tests

- [X] Yes


lightly_studio_view/e2e/pages/annotation-details-page.ts
```diff
@@ -81,7 +81,7 @@ export class AnnotationDetailsPage {
         const input = this.page.getByTestId('select-list-input');
         await input.waitFor({ state: 'visible' });
         await input.fill(label);
-        await this.page.getByText(label, { exact: true }).click();
+        await this.page.getByRole('option', { name: label, exact: true }).click();
```
IgorSusmelj, line 84: Previously we would just click on any text with the `label`. Now, we also filter by ARIA role `option` which is used for the dropdown fields. So we just more precisely tell playwright where to click.

lightly_studio_view/src/lib/components/SelectList/SelectList.svelte
```diff
@@ -52,29 +54,43 @@
...
     };
 
     const handleKeyDown = (event: KeyboardEvent) => {
-        if (event.key === 'Enter' && inputValue) {
+        if (event.key === 'Enter' && !highlightedValue && inputValue) {
             createNewItem(inputValue);
-            inputValue = '';
             event.preventDefault();
+            event.stopPropagation();
+        } else if (event.key === 'ArrowLeft' || event.key === 'ArrowRight') {
```
LeonardoRosaa, line 91: Minor suggestion: create an Enum for event keys.

IgorSusmelj: I switched to use an enum

lightly_studio_view/src/lib/components/SelectList/SelectList.svelte
```diff
@@ -52,29 +60,43 @@
...
+        } else {
+            const newItem = { value: item, label: item };
+            items.push(newItem);
+            handleOnSelect(newItem);
+        }
     };
 
     const handleKeyDown = (event: KeyboardEvent) => {
-        if (event.key === 'Enter' && inputValue) {
+        if (event.key === KeyboardKey.Enter && !highlightedValue && inputValue) {
```
ikondrat: I would rather not use enum, it just makes code less readable

IgorSusmelj: I reverted the last commit

PR #38, opened 2025-10-25
Show number of filtered and total samples to footer

IgorSusmelj:
## What has changed and why?

- Adds information to the footer about currently filtered and total size of the current dataset. We count samples or annotations depending on the view the user is in. (for captions we simply count the samples).
- Modifies the backend to return size of the dataset
- Make footer nicer

Before:
image

After:
image
image
image

## How has it been tested?
- This PR has no functional changes

- [X] Yes

### DIFF
--- a/lightly_studio/src/lightly_studio/resolvers/dataset_resolver.py
+++ b/lightly_studio/src/lightly_studio/resolvers/dataset_resolver.py
@@ -79,9 +79,7 @@ def get_dataset_details(session: Session, dataset: DatasetTable) -> DatasetViewW
     """Convert a DatasetTable to DatasetViewWithCount with computed sample count."""
     sample_count = (
         session.exec(
-            select(func.count(SampleTable.sample_id)).where(
-                SampleTable.dataset_id == dataset.dataset_id
-            )
+            select(func.count("*")).where(SampleTable.dataset_id == dataset.dataset_id)
         ).one()
         or 0
     )


lightly_studio/src/lightly_studio/models/dataset.py
```diff
@@ -32,6 +32,7 @@ class DatasetView(DatasetBase):
     dataset_id: UUID
     created_at: datetime
     updated_at: datetime
+    total_sample_count: int | None = None
```
michal-lightly: This is breaking a pattern that we have. See "SampleViewsWithCount" and "AnnotationViewsWithCount" - here to follow the pattern we'd need "DatasetViewWithCount".

ikondrat: I would support @michal-lightly 

I would also use DatasetViewWithCount or even better
DatasetDetailsView because we probably would have more information

lightly_studio/src/lightly_studio/api/routes/api/dataset.py
```diff
@@ -64,16 +64,17 @@ def read_datasets(
     return dataset_resolver.get_all(session=session, offset=paginated.offset, limit=paginated.limit)
 
 
-@dataset_router.get("/datasets/{dataset_id}")
+@dataset_router.get("/datasets/{dataset_id}", response_model=DatasetView)
```
ikondrat: I would use different model to return  detailed information about dataset e.g. DatasetDetailsView

lightly_studio/src/lightly_studio/resolvers/dataset_resolver.py
```diff
@@ -75,6 +75,23 @@ def get_by_name(session: Session, name: str) -> DatasetTable | None:
     return session.exec(select(DatasetTable).where(DatasetTable.name == name)).one_or_none()
 
 
+def to_dataset_view(session: Session, dataset: DatasetTable) -> DatasetView:
+    """Convert a DatasetTable to DatasetView with computed sample count."""
+    sample_count = (
```
ikondrat, line 80: I would make one thing - get_samples_count 
and just call it in router

I think view conversion is only for API

lightly_studio_view/src/lib/components/Footer/Footer.svelte
```diff
@@ -0,0 +1,65 @@
...
+    const {
+        totalSamples = 0,
+        filteredSamples = 0,
+        totalAnnotations = 0,
+        filteredAnnotations = 0,
+        gridType = 'samples'
+    }: FooterProps = $props();
+
+    const statsText = $derived.by(() => {
+        const isAnnotationView = gridType === 'annotations';
```
ikondrat: I would avoid use gridType because it is artificial

you can use 
export const isAnnotationsRoute = (routeId: string | null): boolean => {
    return routeId ? routeId == APP_ROUTES.annotations : false;
};
https://github.com/lightly-ai/lightly-studio/blob/igor-lig-7488-show-dataset-size-after-filter/lightly_studio_view/src/lib/routes.ts#L38

lightly_studio/src/lightly_studio/models/dataset.py
IgorSusmelj: Addressed

lightly_studio/src/lightly_studio/api/routes/api/dataset.py
IgorSusmelj: Addressed

lightly_studio/src/lightly_studio/resolvers/dataset_resolver.py
IgorSusmelj: Let me know if the current proposal works

ikondrat: The thing is View is rather for API response, we don't return view models in the resolver 

ok let's keep it like this

lightly_studio/src/lightly_studio/resolvers/dataset_resolver.py
```diff
@@ -75,6 +75,25 @@ def get_by_name(session: Session, name: str) -> DatasetTable | None:
     return session.exec(select(DatasetTable).where(DatasetTable.name == name)).one_or_none()
 
 
+def get_samples_count(session: Session, dataset: DatasetTable) -> DatasetViewWithCount:
```
ikondrat:
```suggestion
def get_dataset_details(session: Session, dataset: DatasetTable) -> DatasetViewWithCount:
```

lightly_studio/src/lightly_studio/api/routes/api/dataset.py
```diff
@@ -64,16 +65,17 @@ def read_datasets(
...
     dataset: Annotated[
         DatasetTable,
         Path(title="Dataset Id"),
         Depends(get_and_validate_dataset_id),
     ],
-) -> DatasetTable:
+) -> DatasetViewWithCount:
     """Retrieve a single dataset from the database."""
-    return dataset
+    return dataset_resolver.get_samples_count(session=session, dataset=dataset)
```
ikondrat:
```suggestion
    return dataset_resolver.get_dataset_details(session=session, dataset=dataset)
```

PR #40, opened 2025-10-27
Add AnnotationClassBalancingStrategy

lukas-lightly:
## What has changed and why?
Also handle AnnotationClassBalancingStrategy in the select_via_db() function.

## How has it been tested?
Added a test

- [x] Yes


lightly_studio/src/lightly_studio/selection/select_via_db.py
```diff
@@ -3,25 +3,50 @@
...
+    AnnotationClassBalancingStrategy,
     EmbeddingDiversityStrategy,
     MetadataWeightingStrategy,
     SelectionConfig,
 )
 
 
+def _aggregate_class_distributions(
+    input_sample_ids: list[UUID],
+    annotations_by_sample_id: dict[UUID, list[AnnotationBaseTable]],
```
MalteEbner: Minor: For dictionaries, we often use the naming scheme of `{key}_to_{value}`
```suggestion
    sample_id_to_annotations: dict[UUID, list[AnnotationBaseTable]],
```

lightly_studio/src/lightly_studio/selection/select_via_db.py
```diff
@@ -74,7 +99,26 @@ def select_via_database(
...
+                annotations_by_sample_id[annotation.sample_id].append(annotation)
+
+            class_distributions = _aggregate_class_distributions(
+                input_sample_ids=input_sample_ids,
+                annotations_by_sample_id=annotations_by_sample_id,
+                target=strat.target,
+            )
+            mundig.add_class_balancing(
+                class_distributions=class_distributions,
+                target=list(strat.target.values()),
```
MalteEbner: Two related points for improvement  here:
1. The function `_aggregate_class_distributions` does not need the target values, only the keys. However, it needs them in order.
2. This relies on the iteration over the target keys in `_aggregate_class_distributions` and the iteration over the values in `list(strat.target.values())` to return the same order. If the target was changed inbetween, this would turn out to be wrong.

I would rather have one `target_keys, target_values = map(list, zip(*target.items()))` and pass the `target_keys` to `_aggregate_class_distributions` and the `target_values` to `mundig.add_class_balancing`

lightly_studio/src/lightly_studio/selection/select_via_db.py
```diff
@@ -3,25 +3,50 @@
...
+        # For each target label, count the number of annotations
+        counts_for_sample: dict[UUID, int] = defaultdict(int)
+        for annotation in annotations_for_sample:
+            counts_for_sample[annotation.annotation_label_id] += 1
+
+        counts_as_list = []
+        for label_id in target:
+            counts_as_list.append(counts_for_sample.get(label_id, 0))
+        class_distributions.append(counts_as_list)
+    return class_distributions
```
MalteEbner, line 67: Optional: Appending to lists is slow in python. 
And the intermediate logic of first computing the counts for a sample and then only saving is more complicated than necessary.

I would rather do:
```
) -> NDArray[float]:
    n_samples = len(input_sample_ids)
    n_labels = len(label_id_to_index)
    class_distributions = np.zeros((n_samples, n_labels), dtype=np.int32)

   label_id_to_index = {lid: j for j, lid in enumerate(target.keys())}
   # or if target was a list[UUID]
   # label_id_to_index = {lid: j for j, lid in enumerate(target)}

    for i, sample_id in enumerate(input_sample_ids):
        for annotation in annotations_by_sample_id[sample_id]:
            label_idx = label_id_to_index.get(annotation.annotation_label_id)
            if label_idx is not None:
                class_distributions[i, label_idx] += 1

    return class_distributions
```

lightly_studio/src/lightly_studio/selection/select_via_db.py
```diff
@@ -3,25 +3,50 @@
...
     MetadataWeightingStrategy,
     SelectionConfig,
 )
 
 
+def _aggregate_class_distributions(
+    input_sample_ids: list[UUID],
+    annotations_by_sample_id: dict[UUID, list[AnnotationBaseTable]],
+    target: dict[UUID, float],
+) -> list[list[int]]:
```
MalteEbner: Minor: A class docstring would be nice, especially for explaining that the first two UUIDs are `sample_ids`, while the target UUID is an `annotation_label_id`

lightly_studio/src/lightly_studio/selection/selection_config.py
```diff
@@ -35,3 +35,10 @@ class MetadataWeightingStrategy(SelectionStrategy):
 
     strategy_name: Literal["weights"] = "weights"
     metadata_key: str
+
+
+class AnnotationClassBalancingStrategy(SelectionStrategy):
+    """Selection strategy based on class balancing."""
+
+    strategy_name: Literal["balance"] = "balance"
+    target: dict[UUID, float]
```
MalteEbner: Minor: Best add a docstring that the target keys are `annotation_label_ids`. And/or rename the attribute to `annotation_label_id_to_target`

lightly_studio/tests/selection/test_select_via_db.py
```diff
@@ -295,6 +299,84 @@ def test_select_via_database__tag_name_already_exists(
...
+    sample_ids = _all_sample_ids(test_db, dataset_id)
+
+    label_cat = create_annotation_label(session=test_db, annotation_label_name="cat")
+    label_dog = create_annotation_label(session=test_db, annotation_label_name="dog")
+    label_bird = create_annotation_label(session=test_db, annotation_label_name="bird")
+
+    # Create annotations
+    # * sample 0 cat/dog
+    # * sample 1 dog
+    # * sample 2 dog/bird
```
MalteEbner: Minor
```suggestion
    # * sample 0: cat + dog
    # * sample 1: dog
    # * sample 2: dog + bird
```

lightly_studio/src/lightly_studio/selection/selection_config.py
lukas-lightly: Renamed

lightly_studio/src/lightly_studio/selection/select_via_db.py
lukas-lightly: Changed implementation to yours with some name changes. Also the return type should be float, because that's what Mundig expects.

We'll have to rethink this for input and uniform targets, because you don't know all the annotation IDs, but it's good enough for now.

lightly_studio/src/lightly_studio/selection/select_via_db.py
lukas-lightly: Done, though I used the slightly more verbose
```
            target_keys, target_values = (
                list(strat.annotation_label_id_to_target.keys()),
                list(strat.annotation_label_id_to_target.values()),
            )
```

lightly_studio/src/lightly_studio/selection/select_via_db.py
lukas-lightly: Done

lukas-lightly: Btw, in other languages I would make a type alias to distinguish sample UUID against Annotation UUID.

lightly_studio/src/lightly_studio/selection/select_via_db.py
MalteEbner: Looks good :)

lightly_studio/src/lightly_studio/selection/select_via_db.py
MalteEbner: >  I would make a type alias to distinguish sample UUID against Annotation UUID.

Optional: You can also do that in python:
```python
from typing import TypeAlias

SampleUUID: TypeAlias = UUID
AnnotationLabelUUID: TypeAlias = UUID
```

PR #47, opened 2025-10-27
Added steps about how to set up the project for development

LeonardoRosaa:
## What has changed and why?

The contributing guide also includes instructions on how to set up the project for development.

- [ ] Yes

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -60,7 +60,7 @@ written in Markdown (MyST flavor). For more information regarding formatting, se
 ## SETTING UP YOUR ENVIRONMENT FOR DEVELOPMENT
 
 ### Requirements
-- Python **3.8+**
+- Python **3.8** <= version < **3.13**
 - Node.js **22.11+**
 - Access to **Google Cloud Platform** (request permission from @IgorSusmelj)
 


CONTRIBUTING.md
```diff
@@ -57,6 +57,82 @@ written in Markdown (MyST flavor). For more information regarding formatting, se
 - https://pradyunsg.me/furo/reference/
 - https://myst-parser.readthedocs.io/en/latest/syntax/typography.html
 
+## SETTING UP YOUR ENVIRONMENT FOR DEVELOPMENT
+
+### Requirements
+- Python **3.8+**
```
ikondrat: I think we don't support yet latest python version, I would mention upper limitation as well
here is the task about supporting 3.13 https://linear.app/lightly/issue/LIG-7702/story-support-python-313

ikondrat (approved):
lgtm

CONTRIBUTING.md
LeonardoRosaa: You're right. We aren't supporting Python 3.13+ yet. Would you suggest describing that we have some limitations with Python 3.13+?

horatiualmasan: I would just add "Python  >=3.8, <3.13". There is a fixed version for scikit-learn and it will not work with Python 3.13

CONTRIBUTING.md
```diff
@@ -57,6 +57,82 @@ written in Markdown (MyST flavor). For more information regarding formatting, se
...
+- Node.js **22.11+**
+- Access to **Google Cloud Platform** (request permission from @IgorSusmelj)
```
horatiualmasan, line 65: Why is this required?

LeonardoRosaa: GCP credentials are required to download the `lightly-mundig` package in the development environment.

PR #51, opened 2025-10-28
Rename variables in tests

michal-lightly:
## What has changed and why?

Follow-up to create_sample -> create_image function rename. Rename also variables.

I used VSCode rename feature for every occurrence.

The goal is not for the rename to be 100% perfect, the goal is for the new test writers (and for AI) not to be confused and use new naming.

## How has it been tested?

CI

- [ ] Yes

### DIFF
--- a/lightly_studio/tests/api/routes/api/test_dataset.py
+++ b/lightly_studio/tests/api/routes/api/test_dataset.py
@@ -125,7 +125,7 @@ def test_delete_dataset(test_client: TestClient) -> None:
 def test_export_dataset(db_session: Session, test_client: TestClient) -> None:
     client = test_client
     dataset_id = UUID(create_dataset(client))
-    samples = create_images(
+    images = create_images(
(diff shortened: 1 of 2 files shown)


lightly_studio/tests/api/routes/api/test_dataset.py
```diff
@@ -137,8 +137,8 @@ def test_export_dataset(db_session: Session, test_client: TestClient) -> None:
 
     # Tag two samples.
```
mihnea-necsulescu, line 138:
```suggestion
    # Tag two images.
```
same thing for the rest of the comments - I think we should update those as part of this PR, too

michal-lightly: That would be too time consuming because it can't be searched for automatically. The current state is better than the one before, I'd like to merge as is.

mihnea-necsulescu (approved):

PR #52, opened 2025-10-28
Introduce dataset hierarchy

michal-lightly:
## What has changed and why?

Introduce dataset hierarchy.

* Add `parent_dataset_id` to the DatasetTable
* Add `parent` and `children` relationships
* Add a resolver function to fetch the hierarchy

## How has it been tested?

Added tests.

- [ ] Yes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Introduce button to reset viewport changes for embedding plot.
 - Improve UX for label picker when adding labels.
 - Changed the grid slider to define how many items will appear per row.
+- Updated the panel slider style.
 
 ### Removed
 
--- a/lightly_studio_view/src/routes/datasets/[dataset_id]/+layout.svelte
+++ b/lightly_studio_view/src/routes/datasets/[dataset_id]/+layout.svelte
@@ -41,6 +41,7 @@
     import PlotPanel from '$lib/components/PlotPanel/PlotPanel.svelte';
     import { Button } from '$lib/components/ui/index.js';
     import { PaneGroup, Pane, PaneResizer } from 'paneforge';
+    import { GripVertical } from '@lucide/svelte';
 
     const { data, children } = $props();
     const {
@@ -312,8 +313,12 @@
                     </Pane>
 
                     <PaneResizer
-                        class="mx-2 w-1 cursor-col-resize bg-border transition-colors hover:bg-accent"
-                    />
+                        class="relative mx-2 flex w-1 cursor-col-resize items-center justify-center"
+                    >
+                        <div class="bg-brand z-10 flex h-7 min-w-5 items-center justify-center">
+                            <GripVertical class="text-diffuse-foreground" />
+                        </div>
+                    </PaneResizer>
 
                     <Pane defaultSize={50} minSize={30} class="flex flex-col">
                         <PlotPanel />


lightly_studio/src/lightly_studio/models/dataset.py
```diff
@@ -1,14 +1,11 @@
 """This module contains the Dataset model and related enumerations."""
 
-from __future__ import annotations
```
michal-lightly, line 3: Relationships can't be defined properly with annotations enabed.

lightly_studio/src/lightly_studio/models/dataset.py
```diff
@@ -46,17 +44,24 @@ class DatasetTable(DatasetBase, table=True):
...
-        default_factory=lambda: datetime.now(timezone.utc),
+    updated_at: datetime = Field(default_factory=lambda: datetime.now(timezone.utc))
+
+    parent: Optional["DatasetTable"] = Relationship(
+        back_populates="children",
+        sa_relationship_kwargs={"remote_side": "DatasetTable.dataset_id"},
+    )
+    children: List["DatasetTable"] = Relationship(
+        back_populates="parent",
+        sa_relationship_kwargs={"cascade": "all, delete-orphan"},
```
mihnea-necsulescu: DuckDB does not currently support cascade deletes, no? I don't think this will work out of the box (maybe it would be worth adding a delete root dataset test)

we will probably have to manually deal with this in the resolver delete function, i.e., first delete all children before attempting to delete a dataset

lightly_studio/src/lightly_studio/resolvers/datasets_resolver/get_hierarchy.py
```diff
@@ -0,0 +1,26 @@
...
+    root_dataset = dataset_resolver.get_by_id(session=session, dataset_id=root_dataset_id)
+    if root_dataset is None:
+        raise ValueError(f"Dataset with id {root_dataset_id} not found.")
+
+    to_process = [root_dataset]
+    all_datasets = []
+    while to_process:
+        current_dataset = to_process.pop()
+        all_datasets.append(current_dataset)
+        to_process.extend(current_dataset.children)
```
mihnea-necsulescu, line 25: wouldn't recursion be cleaner and more readable for this? 

just a thought - I think it's fine either way

lightly_studio/src/lightly_studio/models/dataset.py
michal-lightly: I was thinking whether I should remove it. I will. Dataset deletion should be rare - so it is not something we have to optimise for.

lightly_studio/src/lightly_studio/resolvers/datasets_resolver/get_hierarchy.py
michal-lightly: I would need a helper function. Not sure if it is cleaner:

```py
def get_hierarchy(session: Session, root_dataset_id: UUID) -> list[DatasetTable]:
    """Retrieve all child datasets of the given root dataset, including the root itself."""
    root_dataset = dataset_resolver.get_by_id(session=session, dataset_id=root_dataset_id)
    if root_dataset is None:
        raise ValueError(f"Dataset with id {root_dataset_id} not found.")

    return _get_hierarchy_helper(session, root_dataset)

def _get_hierarchy_helper(session: Session, dataset: DatasetTable): -> list[DatasetTable]:
    hierarchy = [dataset]
    for child in dataset.children:
        hierarchy.extend(get_hierarchy(session, child))
    return children
```

michal-lightly: I would prefer to keep it as is. I'll add a comment though.

PR #60, opened 2025-10-29
Use pydantic mypy plugin for a stricter type check

michal-lightly:
## What has changed and why?

Use pydantic mypy plugin for a stricter type check.

Motivation: It discovered issues when adding new pydantic types for dataset table changes.

## How has it been tested?

Tests pass.

- [ ] Yes


lightly_studio/src/lightly_studio/api/routes/api/sample.py
```diff
@@ -72,7 +72,7 @@ def read_samples(
     # TODO(Michal, 10/2025): Add SampleView to ImageView and then use a response model
     # instead of manual conversion.
     return ImageViewsWithCount(
-        data=[
+        samples=[
```
michal-lightly, line 75: We define an alias in the pydantic model. Both keys work (I tested also the GUI), but mypy was complaining about one.

PR #65, opened 2025-10-29
Preparing captions section to support multiple captions

LeonardoRosaa:
## What has changed and why?

Preparing the captions section to support multiple captions. This PR adds scrolling when the captions don’t fit within the available space.

## How has it been tested?

https://github.com/user-attachments/assets/2df55d98-ebdf-4322-8d0e-0ade4f765868

- [ ] Yes


lightly_studio/src/lightly_studio/api/routes/api/caption.py
```diff
@@ -9,21 +9,23 @@
...
-@captions_router.get("/captions", response_model=CaptionsListView)
+@captions_router.get("/captions", response_model=SampleCaptionsListView)
 def read_captions(
     dataset_id: Annotated[UUID, Path(title="Dataset Id")],
     session: SessionDep,
     pagination: Annotated[PaginatedWithCursor, Depends()],
-) -> GetAllCaptionsResult:
+) -> GetAllCaptionsFromSampleResult:
     """Retrieve captions for a dataset."""
-    return caption_resolver.get_all(
```
michal-lightly, line 26: This is the only place we use `caption_resolver.get_all`. We can remove the function (or better said - modify it to fit the new needs).

lightly_studio/src/lightly_studio/models/caption.py
```diff
@@ -57,17 +57,25 @@ class CaptionView(SQLModel):
...
-class CaptionDetailsView(CaptionView):
+class SampleCaptionView(SQLModel):
     """Response model for caption."""
 
-    sample: CaptionSampleView
+    caption_id: UUID
+    text: str
+
+
+class SampleCaptionDetailsView(BaseModel):
```
michal-lightly: Let's sync on the model changes. I believe this can be greatly simplified - we can use `CaptionView` instead of both `CaptionSampleView` and `SampleCaptionView`, which should definitely not exist at the same time.

Also let's be more mindful about renames - what is the reason to rename `CaptionsListView`?

lightly_studio/src/lightly_studio/resolvers/caption_resolver.py
```diff
@@ -20,6 +21,14 @@ class GetAllCaptionsResult(BaseModel):
     next_cursor: int | None = None
 
 
+class GetAllCaptionsFromSampleResult(BaseModel):
+    """Result wrapper for caption listings."""
+
+    samples: Sequence[SampleTable]
+    total_count: int
+    next_cursor: int | None = None
```
michal-lightly: Again here, is there a reason we need a new class?

lightly_studio/src/lightly_studio/resolvers/caption_resolver.py
```diff
@@ -78,3 +87,55 @@ def get_all(
         total_count=total_count,
         next_cursor=next_cursor,
     )
+
+
+def get_all_from_samples(
```
michal-lightly: Please add a test for this function.

lightly_studio/tests/conftest.py
```diff
@@ -410,14 +410,14 @@ def captions_test_data(
...
     # Create 4 captions in the first dataset
-    for i in range(4):
-        caption = CaptionCreate(
-            dataset_id=datasets[0].dataset_id,
-            sample_id=samples[0].sample_id,
-            text=f"Caption number {i}",
-        )
-
-        captions_to_create.append(caption)
+    for sample in samples:
```
michal-lightly: Our guideline is to avoid for loop in tests. Let's discuss how we can change this.

michal-lightly (changes requested):

lightly_studio/src/lightly_studio/resolvers/caption_resolver.py
LeonardoRosaa: Done!

LeonardoRosaa:
This PR introduces a few additional changes:

- Cleaned up the caption view classes.
- Removed the `get_all `function from the Caption Resolver. 
- The `GET /captions `endpoint now uses `get_all_captions_by_samples` instead.

lightly_studio/src/lightly_studio/resolvers/caption_resolver.py
```diff
@@ -39,42 +40,55 @@ def create_many(session: Session, captions: Sequence[CaptionCreate]) -> list[Cap
...
+        List of captions matching the filters, total number of samples with captions, next
+        cursor (pagination)
     """
-    query = select(CaptionTable).order_by(
-        col(CaptionTable.created_at).asc(),
-        col(CaptionTable.caption_id).asc(),
+    query = (
+        select(SampleTable)
+        .join(CaptionTable)
+        .where(CaptionTable.dataset_id == dataset_id)
```
JonasWurst: In my understanding, the dataset_id will not be present in the CpationTable in the near future. Can we use SampleTable instead?

lightly_studio/src/lightly_studio/resolvers/caption_resolver.py
```diff
@@ -39,42 +40,55 @@ def create_many(session: Session, captions: Sequence[CaptionCreate]) -> list[Cap
...
+        .distinct()
     )
-    count_query = select(func.count()).select_from(CaptionTable)
 
-    query = query.where(CaptionTable.dataset_id == dataset_id)
-    count_query = count_query.where(CaptionTable.dataset_id == dataset_id)
+    count_subquery = (
+        select(SampleTable.sample_id)
+        .join(CaptionTable)
+        .where(CaptionTable.dataset_id == dataset_id)
```
JonasWurst: see comment above

lightly_studio/src/lightly_studio/resolvers/caption_resolver.py
```diff
@@ -39,42 +40,55 @@ def create_many(session: Session, captions: Sequence[CaptionCreate]) -> list[Cap
...
-        col(CaptionTable.caption_id).asc(),
+    query = (
+        select(SampleTable)
+        .join(CaptionTable)
+        .where(CaptionTable.dataset_id == dataset_id)
+        .order_by(
+            col(CaptionTable.created_at).asc(),
+            col(CaptionTable.caption_id).asc(),
+        )
+        .distinct()
```
JonasWurst: Is the distinct required here?

lightly_studio/src/lightly_studio/resolvers/caption_resolver.py
```diff
@@ -39,42 +40,55 @@ def create_many(session: Session, captions: Sequence[CaptionCreate]) -> list[Cap
     return db_captions
 
 
-def get_all(
+def get_all_captions_by_sample(
```
JonasWurst: I think it is bit misleading that this one is in the caption resolver since it is actually returning list of samples.

lightly_studio/src/lightly_studio/examples/example_coco_caption.py
```diff
@@ -15,10 +15,13 @@
...
-captions_result = caption_resolver.get_all(session=dataset.session, dataset_id=dataset.dataset_id)
-print(captions_result.total_count)
+samples_with_captions_result = caption_resolver.get_all_captions_by_sample(
+    session=dataset.session, dataset_id=dataset.dataset_id
+)
+print(samples_with_captions_result.total_count)
 
-for caption in captions_result.captions[:10]:
-    print(caption)
+for samples in samples_with_captions_result.samples[:10]:
```
JonasWurst:
```suggestion
for sample in samples_with_captions_result.samples[:10]:
```

lightly_studio/src/lightly_studio/examples/example_coco_caption.py
```diff
@@ -15,10 +15,13 @@
...
+    for caption in samples.captions:
```
JonasWurst:
```suggestion
    for caption in sample.captions:
```

lightly_studio/src/lightly_studio/resolvers/caption_resolver.py
JonasWurst: The view used is not a caption view but rather a samples view. This is not obvious from the description.

Can we instead use the sampleView as is? Or if required generate a caption focused SampleView.

The problem I see: like this, we won't reuse it.

lightly_studio/src/lightly_studio/resolvers/caption_resolver.py
```diff
@@ -39,42 +40,55 @@ def create_many(session: Session, captions: Sequence[CaptionCreate]) -> list[Cap
...
 
     Args:
         session: Database session
         dataset_id: dataset_id parameter to filter the query
         pagination: Optional pagination parameters
 
     Returns:
-        List of captions matching the filters, total number of captions, next cursor (pagination)
+        List of captions matching the filters, total number of samples with captions, next
+        cursor (pagination)
```
JonasWurst: This comment does not match the actual return type

lightly_studio/src/lightly_studio/resolvers/caption_resolver.py
JonasWurst: I would appreciate a comment, explaining the full query:

Getting all samples, that hold captions. Ordered by the caption creation date and their id

lightly_studio/tests/api/routes/api/test_caption.py
```diff
@@ -64,15 +71,17 @@ def test_read_captions__last_page(
...
 
     assert response.status_code == HTTP_STATUS_OK
     result = response.json()
-    assert result["total_count"] == 4
+    assert result["total_count"] == 10
     assert result["nextCursor"] is None
-    assert [f"Caption number {i}" for i in range(2, 4)] == [data["text"] for data in result["data"]]
+    assert [f"Caption number {i}" for i in range(4)] == [
+        caption["text"] for caption in result["data"][0]["captions"][:4]
+    ]
```
JonasWurst:
```suggestion
    assert result["data"][0]["captions"][0]["text"] == "Caption number 0"
    assert result["data"][0]["captions"][1]["text"] == "Caption number 1"
    assert result["data"][0]["captions"][2]["text"] == "Caption number 2"
    assert result["data"][0]["captions"][3]["text"] == "Caption number 3"
```

However, it is questionable if this test adds much clarity here

lightly_studio/tests/api/routes/api/test_caption.py
```diff
@@ -64,15 +71,17 @@ def test_read_captions__last_page(
...
+            "limit": 10,
         },
     )
 
     assert response.status_code == HTTP_STATUS_OK
     result = response.json()
-    assert result["total_count"] == 4
+    assert result["total_count"] == 10
     assert result["nextCursor"] is None
-    assert [f"Caption number {i}" for i in range(2, 4)] == [data["text"] for data in result["data"]]
```
JonasWurst, line 75: This test is not sufficient anymore, we also need to test the filename or sample_id. At the moment, we might receive captions from the wrong sample as they are all similar.

Hence, smth in this direction:
```
    assert result["data"][0].sample_id = # get. the sample_id here, this should be the 2-th sample id 
    assert result["data"][1].sample_id = # get. the sample_id here, this should be the 3-th sample id 
    ...
```

lightly_studio/tests/api/routes/api/test_caption.py
```diff
@@ -32,9 +32,13 @@ def test_read_captions__first_page(
     )
 
     assert response.status_code == HTTP_STATUS_OK
+
     result = response.json()
```
JonasWurst: Please see my comments further down. Since we moved from List[Captions] to List[Samples] as return, we need to completely rework the tests.

lightly_studio/tests/conftest.py
JonasWurst: We can omit the inner loop, as there is no need any more to keep 4 captions. We needed them before, to test the pagination, now pagination is not happening over captions --> 2 is fine.

The outer loop, i feel is fine, since it is just used for test case creation.

lightly_studio/tests/core/test_dataset__coco_caption.py
```diff
@@ -37,22 +37,21 @@ def test_add_samples_from_coco_caption__details_valid(
...
-            ("image1.jpg", "Caption 2 of image 1"),
-            ("image2.jpg", "Caption 1 of image 2"),
+        expected = {
+            (samples[0].sample_id, ("Caption 1 of image 1", "Caption 2 of image 1")),
+            (samples[1].sample_id, ("Caption 1 of image 2",)),
         }
+        assert {
+            (sample.sample_id, tuple([caption.text for caption in sample.captions]))
+            for sample in captions_result.samples
+        } == expected
```
JonasWurst: Minor: I would merge this, no need to assign it first.

lightly_studio/tests/resolvers/test_caption_resolver.py
JonasWurst, line 1: Major: This file needs rework, after the changes, we are not testing the captions. we only access the number of samples.

lightly_studio_view/src/lib/hooks/useCaptionsInfinite/useCaptionsInfinite.ts
```diff
@@ -14,12 +14,11 @@ export const useCaptionsInfinite = (...props: Parameters<typeof readCaptionsInfi
...
     };
 
-    const data = writable<CaptionDetailsView[]>([]);
+    const data = writable<CaptionViewsBySample[]>([]);
 
     captionsQuery.subscribe((query) => {
         if (query.isSuccess) {
-            const allCaptions = query.data.pages.flatMap((page) => page.data);
-            data.set(allCaptions);
+            data.set(query.data.pages.flatMap((e) => e.data));
```
JonasWurst: Question: Why got page changed to e?

lightly_studio/tests/api/routes/api/test_caption.py
JonasWurst, line 1: Major: This file needs rework, as the tests do not fit the new variable type. See comments for more details

lightly_studio/src/lightly_studio/models/caption.py
```diff
@@ -57,17 +49,18 @@ class CaptionView(SQLModel):
     text: str
 
 
-class CaptionDetailsView(CaptionView):
-    """Response model for caption."""
+class CaptionViewsBySample(BaseModel):
```
JonasWurst: This new name is not yet well suited. Let's discuss in more detail :)

JonasWurst (changes requested):
Thanks for the PR. This is the right direction, but we will need some changes.

In general, I agree that packing the captions by samples is a more elegant approach. To avoid confusion in the future, I would try to make this more clear. E.g. I think it is rather a sample view than a CaptionView.

We also need to rework the tests, after the current changes, some of them do not test anymore what they were intended for.

lightly_studio/src/lightly_studio/models/caption.py
michal-lightly: I discussed the name with Leonardo, I suggested `CaptionViewsBySample` but let us know if you have a better suggestion. It gives caption views that belong to a sample.

lightly_studio/src/lightly_studio/resolvers/caption_resolver.py
michal-lightly: It is okay to have a resolver function using/returning other tables. The resolvers are just logical groupings, there is no hard rule.

I think in this case both sample_resolver and caption_resolver are suitable locations.

Regarding SampleView, that's a good point. The disadvantage though is that we would always load (potentially many) captions even when we don't need them. I don't think there is a clear better solution. I would suggest to go with the approach Leonardo chose just to move this forward. We can adjust it later.

lightly_studio/src/lightly_studio/models/caption.py
```diff
@@ -57,17 +49,18 @@ class CaptionView(SQLModel):
...
 
 
-class CaptionsListView(BaseModel):
+class CaptionViewsBySampleWithCount(BaseModel):
     """Response model for counted captions."""
 
     model_config = ConfigDict(populate_by_name=True)
 
-    captions: List[CaptionDetailsView] = PydanticField(..., alias="data")
+    samples: List[CaptionViewsBySample] = PydanticField(..., alias="data")
```
michal-lightly: Where do we construct CaptionViewsBySample?

lightly_studio/src/lightly_studio/api/routes/api/caption.py
```diff
@@ -9,21 +9,23 @@
...
 
 
-@captions_router.get("/captions", response_model=CaptionsListView)
+@captions_router.get("/captions", response_model=CaptionViewsBySampleWithCount)
 def read_captions(
     dataset_id: Annotated[UUID, Path(title="Dataset Id")],
     session: SessionDep,
     pagination: Annotated[PaginatedWithCursor, Depends()],
-) -> GetAllCaptionsResult:
+) -> GetAllCaptionsFromSampleResult:
```
michal-lightly: There's a type mismatch - `GetAllCaptionsFromSampleResult` can't be "cast" to `CaptionViewsBySampleWithCount`. I.e. `caption_resolver.get_all_captions_by_sample` does not return the correct data.

I also tested the app - I don't see any captions.

lightly_studio/src/lightly_studio/resolvers/caption_resolver.py
michal-lightly: Looking at the rest of the PR - it needs some changes. It might be easier to indeed just extend the SampleView.

michal-lightly (changes requested):

lightly_studio/src/lightly_studio/models/caption.py
LeonardoRosaa: We've the class declaration [here](https://github.com/lightly-ai/lightly-studio/blob/1d87e4bcfad752028fc7cc789327b11045a33bf5/lightly_studio/src/lightly_studio/models/caption.py#L52).

lightly_studio/src/lightly_studio/api/routes/api/caption.py
LeonardoRosaa: This code should work.  Have you exported a dataset that includes captions?

lightly_studio/src/lightly_studio/resolvers/caption_resolver.py
LeonardoRosaa: We could keep the code in the `caption_resolver`, it's not a problem. 

I've updated the code to use the `SampleView`.

lightly_studio/src/lightly_studio/resolvers/caption_resolver.py
```diff
@@ -39,42 +40,57 @@ def create_many(session: Session, captions: Sequence[CaptionCreate]) -> list[Cap
...
 
     Returns:
-        List of captions matching the filters, total number of captions, next cursor (pagination)
+        List of samples matching the filters, total number of samples with captions, next
+        cursor (pagination)
     """
-    query = select(CaptionTable).order_by(
-        col(CaptionTable.created_at).asc(),
-        col(CaptionTable.caption_id).asc(),
+    """Selects distinct samples with captions, and orders by creation time and caption ID."""
```
michal-lightly: Minor: There is an issue with the comment formatting.

lightly_studio/src/lightly_studio/resolvers/caption_resolver.py
```diff
@@ -9,13 +9,14 @@
 from sqlmodel import Session, col, func, select
 
 from lightly_studio.api.routes.api.validators import Paginated
-from lightly_studio.models.caption import CaptionCreate, CaptionTable
+from lightly_studio.models.caption import CaptionCreate, CaptionTable, CaptionViewsBySampleWithCount
+from lightly_studio.models.sample import SampleTable
 
 
-class GetAllCaptionsResult(BaseModel):
+class GetAllCaptionsFromSampleResult(BaseModel):
```
michal-lightly: This class has nothing to do with captions. Let me see the rest of the changes.

michal-lightly: This class is unused.

lightly_studio/src/lightly_studio/resolvers/caption_resolver.py
```diff
@@ -39,42 +40,57 @@ def create_many(session: Session, captions: Sequence[CaptionCreate]) -> list[Cap
...
+    """Selects distinct samples with captions, and orders by creation time and caption ID."""
+    query = (
+        select(SampleTable)
+        .join(CaptionTable)
+        .where(SampleTable.dataset_id == dataset_id)
+        .order_by(
+            col(CaptionTable.created_at).asc(),
+            col(CaptionTable.caption_id).asc(),
+        )
+        .distinct()
```
michal-lightly: Why do we need the distinct clause?

lightly_studio/src/lightly_studio/resolvers/caption_resolver.py
```diff
@@ -39,42 +40,57 @@ def create_many(session: Session, captions: Sequence[CaptionCreate]) -> list[Cap
...
-    count_query = select(func.count()).select_from(CaptionTable)
 
-    query = query.where(CaptionTable.dataset_id == dataset_id)
-    count_query = count_query.where(CaptionTable.dataset_id == dataset_id)
+    """Selects distinct samples with captions for counting total number."""
+    count_subquery = (
+        select(SampleTable.sample_id)
+        .join(CaptionTable)
+        .where(SampleTable.dataset_id == dataset_id)
+        .distinct()
```
michal-lightly: Also here

lightly_studio/src/lightly_studio/models/caption.py
```diff
@@ -57,17 +51,11 @@ class CaptionView(SQLModel):
...
 
 
-class CaptionDetailsView(CaptionView):
-    """Response model for caption."""
-
-    sample: CaptionSampleView
-
-
-class CaptionsListView(BaseModel):
+class CaptionViewsBySampleWithCount(BaseModel):
```
michal-lightly: This class has nothing to do with captions. It is confusing it is placed here.

LeonardoRosaa:
- Moved the logic to sample_resolver;
- Removed the old captions endpoint;
- Simplified the tests;
- Kept the order_by since it works;
- Didn’t remove the .subquery() because I received a few warnings from SQLModel.

PR #66, opened 2025-10-30
Make Dataset.create() idempotent to allow script reusability

eternal-f1ame:
### Problem
Users cannot run the same dataset indexing script multiple times. When `Dataset.create()` is called with a name that already exists in the database, it raises a `ValueError`, forcing users to either:
- Manually delete the database between runs
- Use `db_manager.connect(cleanup_existing=True)` (losing all data)
- Remember to use `Dataset.load()` instead of `Dataset.create()`

This is unintuitive for new users and prevents simple script reusability.

### Solution
Modified `Dataset.create()` to check if a dataset with the given name already exists:
- If it exists: load and return it (with embeddings enabled if available)
- If it doesn't exist: create it as before

This makes `Dataset.create()` idempotent and aligns it with the existing `Dataset.load_or_create()` behavior.

### Changes
- Updated `Dataset.create()` to load existing datasets instead of raising an error
- Updated `test_create__duplicate_names` to verify the new idempotent behavior
- Maintained full backward compatibility with existing code

### Example
```python
import lightly_studio as ls

# First run: creates dataset and indexes images
dataset = ls.Dataset.create()
dataset.add_samples_from_path(path="images/")
ls.start_gui()

# Second run: loads existing dataset with preserved state
dataset = ls.Dataset.create()
ls.start_gui()  # ✓ Works without errors

### DIFF
--- a/lightly_studio/src/lightly_studio/core/dataset.py
+++ b/lightly_studio/src/lightly_studio/core/dataset.py
@@ -249,6 +249,7 @@ class Dataset:
         path: PathLike,
         allowed_extensions: Iterable[str] | None = None,
         embed: bool = True,
+        tag_depth: int = 0,
     ) -> None:
         """Adding samples from the specified path to the dataset.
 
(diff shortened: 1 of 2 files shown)


michal-lightly:
Hi @eternal-f1ame ! Thanks for making a PR.

We already support the functionality you describe with `Dataset.load_or_create(...)`. Does that fit your need?

PR #67, opened 2025-10-30
Split image resolver

michal-lightly:
## What has changed and why?

Split image resolver into multiple files.

No logic change, only splitting into multiple files.

Note that the previous rename image_resolver->image_resolver_legacy was not needed in the end since I'm doing the split in one go.

Follow-up: Split the test file.

## How has it been tested?

CI

- [ ] Yes

### DIFF
--- a/lightly_studio/tests/api/routes/api/test_sample.py
+++ b/lightly_studio/tests/api/routes/api/test_sample.py
@@ -15,7 +15,9 @@ from lightly_studio.resolvers import (
     image_resolver,
     tag_resolver,
 )
-from lightly_studio.resolvers.image_resolver.get_all_by_dataset_id import GetAllSamplesByDatasetIdResult
+from lightly_studio.resolvers.image_resolver.get_all_by_dataset_id import (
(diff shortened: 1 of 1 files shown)


lightly_studio/src/lightly_studio/resolvers/image_resolver/create.py
```diff
@@ -0,0 +1,69 @@
...
+            sample_id=db_sample.sample_id,
+        )
+    )
+    session.add(db_image)
+    session.commit()
+    session.refresh(db_image)
+    return db_image
+
+
+def create_many(session: Session, samples: list[ImageCreate]) -> list[ImageTable]:
```
michal-lightly, line 45: I have kept it in the same file since it reuses the same helper class.

JonasWurst (approved):
LGTM

Minor: In the mindset of boy scout rule, we could have used this oppurtinuty to add proper docstrings to the other resolver methods as well. Atm, only get_samples_excluding has them

PR #68, opened 2025-10-30
Create video and videoframe tables

JonasWurst:
## What has changed and why?

This PR adds the video and video frame table. This is required to get the video support going. 

## How has it been tested?

Will be tested through the resolvers in follow up PRs

Changelog will be updated as soon as the video support is up.

- [ ] Yes

### DIFF
--- a/lightly_studio_view/src/lib/schema.d.ts
+++ b/lightly_studio_view/src/lib/schema.d.ts
@@ -17,11 +17,7 @@ export interface paths {
          */
         get: operations["read_datasets"];
         put?: never;
-        /**
-         * Create Dataset
-         * @description Create a new dataset in the database.
-         */
-        post: operations["create_dataset"];
+        post?: never;
         delete?: never;
         options?: never;
         head?: never;
@@ -2278,39 +2274,6 @@ export interface operations {
                 };
                 content: {
                     "application/json": components["schemas"]["DatasetView"][];
-                };
-            };
-            /** @description Validation Error */
-            422: {
-                headers: {
-                    [name: string]: unknown;
-                };
-                content: {
-                    "application/json": components["schemas"]["HTTPValidationError"];
-                };
-            };
-        };
-    };
-    create_dataset: {
-        parameters: {
-            query?: never;
-            header?: never;
-            path?: never;
-            cookie?: never;
-        };
-        requestBody: {
-            content: {
-                "application/json": components["schemas"]["DatasetCreate"];
-            };
-        };
-        responses: {
-            /** @description Successful Response */
-            201: {
-                headers: {
-                    [name: string]: unknown;
-                };
-                content: {
-                    "application/json": components["schemas"]["DatasetView"];
                 };
             };
             /** @description Validation Error */


lightly_studio/src/lightly_studio/models/video.py
```diff
@@ -0,0 +1,150 @@
...
+        AnnotationBaseTable,
+    )
+    from lightly_studio.models.sample import SampleTable, SampleView
+
+else:
+    AnnotationBaseTable = object
+    SampleTable = object
+
+
+# VideoFrame
```
michal-lightly: Optional: The comment does not bring much value, I suggest to remove it.
```suggestion
```

lightly_studio/src/lightly_studio/models/video.py
```diff
@@ -0,0 +1,150 @@
...
+        fps: float
+
+    """The frame number of the video frame."""
+    frame_number: int
+    frame_timestamp: int
+    sample_id: UUID
+    video_sample_id: UUID
+
+    # Video metadata routed from parent video
+    video: VideoFrameViewVideo
```
michal-lightly: Can't we simply use VideoView?

lightly_studio/src/lightly_studio/models/video.py
```diff
@@ -0,0 +1,150 @@
...
+    from lightly_studio.models.sample import SampleTable, SampleView
+
+else:
+    AnnotationBaseTable = object
+    SampleTable = object
+
+
+# VideoFrame
+
+class VideoFrameBase(SQLModel):
```
michal-lightly: Minor: I'd list Video first since it does not depend on the VideoFrame.

lightly_studio/src/lightly_studio/models/video.py
```diff
@@ -0,0 +1,150 @@
...
+class VideoFrameCreate(VideoFrameBase):
+    """VideoFrame class when inserting."""
+
+
+class VideoFrameTable(VideoFrameBase, table=True):
+    """This class defines the VideoFrame model."""
+
+    __tablename__ = "video_frame"
+    sample_id: UUID = Field(foreign_key="sample.sample_id", primary_key=True)
+    annotations: Mapped[List["AnnotationBaseTable"]] = Relationship(
```
michal-lightly: I don't think this will work. AnnotationBase is currently linked to image.sample_id, we will need to change it to sample.sample_id first (and if that does not work then think of a solution).

I would suggest removing this for now.

lightly_studio/src/lightly_studio/models/video.py
```diff
@@ -0,0 +1,150 @@
...
+
+
+class VideoFrameViewsWithCount(BaseModel):
+    """Response model for counted video frames."""
+
+    model_config = ConfigDict(populate_by_name=True)
+
+    samples: List[VideoFrameView] = PydanticField(..., alias="data")
+    total_count: int
+    next_cursor: Optional[int] = PydanticField(None, alias="nextCursor")
```
michal-lightly: Can you please check with the person that wrote the original code? The code looks suspicious:
* What is model_config and why do we use it?
* Why do we use aliases?

We can also add this class in a later PR not to block this one.

lightly_studio/src/lightly_studio/models/video.py
```diff
@@ -0,0 +1,150 @@
...
+    fps: float
+
+    """The file name of the video."""
+    file_name: str
+
+    """The path of the video."""
+    file_path_abs: str = Field(default=None)
+
+    """The dataset ID to which the video belongs."""
+    dataset_id: UUID = Field(default=None, foreign_key="dataset.dataset_id")
```
michal-lightly: Let's not add dataset_id, it is determined by the sample. I plan to remove dataset_id from all the tables.

lightly_studio/src/lightly_studio/models/video.py
```diff
@@ -0,0 +1,150 @@
...
+    video: VideoFrameViewVideo
+    sample: SampleView
+
+
+class VideoFrameViewsWithCount(BaseModel):
+    """Response model for counted video frames."""
+
+    model_config = ConfigDict(populate_by_name=True)
+
+    samples: List[VideoFrameView] = PydanticField(..., alias="data")
```
michal-lightly: This should be renamed to video_frames. The name `samples` is legacy, I did not manage to rename everything to `images`.

lightly_studio/src/lightly_studio/models/video.py
```diff
@@ -0,0 +1,150 @@
...
+
+
+class VideoView(SQLModel):
+    """Video class when retrieving."""
+
+    file_name: str
+    file_path_abs: str
+    sample_id: UUID
+    sample: SampleView
+    frames: List["VideoFrameView"] = []
```
michal-lightly: This probably should not be included. When we fetch videos (e.g. for the grid view) we probably don't want to fetch all the frames. We can add it later if needed.

lightly_studio/src/lightly_studio/models/video.py
```diff
@@ -0,0 +1,150 @@
...
+    __tablename__ = "video"
+    sample_id: UUID = Field(foreign_key="sample.sample_id", primary_key=True)
+    frames: Mapped[List["VideoFrameTable"]] = Relationship(back_populates="video")
+    sample: Mapped["SampleTable"] = Relationship()
+
+
+class VideoView(SQLModel):
+    """Video class when retrieving."""
+
+    file_name: str
```
michal-lightly: Minor: Can we list the fields the same order as in the VideoTable class? It is then easier to maintain.

lightly_studio/src/lightly_studio/models/video.py
```diff
@@ -0,0 +1,150 @@
...
+    sample_id: UUID
+    sample: SampleView
+    frames: List["VideoFrameView"] = []
+    width: int
+    height: int
+    duration: float
+    fps: float
+
+
+class VideoViewsWithCount(BaseModel):
```
michal-lightly: Same comment as for VideoFrameViewsWithCount. My recommendation is to skip these classes for now.

michal-lightly (changes requested):
Thanks, left some comments.

Please also regenerate the schema.

lightly_studio/src/lightly_studio/models/video.py
```diff
@@ -0,0 +1,105 @@
...
+class VideoTable(VideoBase, table=True):
+    """This class defines the Video ORM table."""
+
+    __tablename__ = "video"
+    sample_id: UUID = Field(foreign_key="sample.sample_id", primary_key=True)
+    frames: Mapped[List["VideoFrameTable"]] = Relationship(back_populates="video")
+    sample: Mapped["SampleTable"] = Relationship()
+
+
+class VideoView(SQLModel):
```
michal-lightly, line 55: Optional: I'm not sure why we don't use VideoBase here (I know this has been copied from image.py). Can you try that? Similar for VideoFrameView.

lightly_studio/src/lightly_studio/models/video.py
```diff
@@ -0,0 +1,105 @@
...
+
+
+class VideoFrameBase(SQLModel):
+    """Base class for the VideoFrame model."""
+
+    """The frame number of the video frame."""
+    frame_number: int
+
+    """The timestamp of the video frame."""
+    frame_timestamp: int
```
michal-lightly, line 75: Minor: Let's add a unit, depending on it a float might be needed (maybe even for ms):
```
frame_timestamp_ms: int / float
# or
frame_timestamp_s: float
```

michal-lightly: I'm actually not sure what info we get when we index a video. We might need to adjust this when the add samples function is implemented.

lightly_studio/src/lightly_studio/models/video.py
JonasWurst: I will align on this with Kondrat and rather do it in a follow up, since the Count Views are missing anyway

lightly_studio/src/lightly_studio/models/video.py
JonasWurst: Let's adjust when we have an implementation of add samples

PR #69, opened 2025-10-30
feat: refactor Plot component to make it more controllable and scalable to extend

ikondrat:
## What has changed and why?
- Moved out logic to separate hooks
   1.  useArrowData - to handle arrow data
   2. usePlotData - to prepare data for plot and handle selection case
   3. introduce 1 more category to highlight intersection of "filtering" + "range selection", check video

## How has it been tested?

Manually with 2 cases 

Case 1
1. Run the app
2. Go to samples
4. Select label
5. Open Plot Panel
6. Create rectangle selection and select some of the points colored as visible
7. You should see subset of filtered by label samples in the samples list

Case 2
1. Run the app
2. Go to samples
4. Open Plot Panel
5. Create rectangle selection and select some of the points colored as visible
6. You should see subset of selected samples  in the sample list

- [ ] Yes
- [ x] Not needed (internal change)

## Visual changes

https://github.com/user-attachments/assets/f2ec51f4-a531-4a13-8692-1f92e3a44726


lightly_studio_view/src/lib/hooks/useEmbeddingSelection/isPointInPolygon.ts
```diff
@@ -0,0 +1,24 @@
+type Point = {
+    x: number;
+    y: number;
+};
+
+export function isPointInPolygon(px: number, py: number, polygon: Point[]): boolean {
```
mihnea-necsulescu: this is a bit hard to follow (at least for me) - could we add some comments?

(also, found this library https://www.npmjs.com/package/@turf/boolean-point-in-polygon - may be worth looking into at some point to reduce the code)

mihnea-necsulescu: also, this is just an utility function that is only used only in usePlotData and its tests; doesn't seem like it's a hook - we should probably move it out of the hooks directory and closer to PlotPanel, maybe in the same directory even, or in an utils one

lightly_studio_view/src/lib/components/PlotPanel/usePlotData.ts
```diff
@@ -0,0 +1,91 @@
+import { writable, type Writable } from 'svelte/store';
+import { EmbeddingView, type Point } from 'embedding-atlas/svelte';
+import type { ComponentProps } from 'svelte';
+import type { ArrowData } from './useArrowData';
+import { isPointInPolygon } from '$lib/hooks/useEmbeddingSelection/isPointInPolygon';
+
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+const plotColumns = ['x', 'y', 'category'] as const;
+type PlotColumn = (typeof plotColumns)[number];
```
mihnea-necsulescu: I am not sure I get the need for `plotColumns` - can't PlotColumn just be defined as a union type?

```suggestion
type PlotColumn = 'x' | 'y' | 'category';
```

lightly_studio_view/src/lib/hooks/useEmbeddingSelection/isPointInPolygon.ts
ikondrat: @mihnea-necsulescu I wouldn't use 18.4 kB library  to replace 10 lines function

lightly_studio_view/src/lib/components/PlotPanel/usePlotData.ts
```diff
@@ -0,0 +1,91 @@
...
+
+    if (!data) {
+        return { data: plotData, error, selectedSampleIds };
+    }
+
+    let category = data.fulfils_filter as Uint8Array;
+    const sampleIds = data.sample_id as string[];
+
+    if (rangeSelection) {
+        const _sampledSampleIds: string[] = [];
```
mihnea-necsulescu: is this needed? doesn't seem like it's used anywhere

lightly_studio_view/src/lib/components/PlotPanel/usePlotData.ts
```diff
@@ -0,0 +1,91 @@
...
+
+        const getCategoryBySelection =
```
mihnea-necsulescu: this could maybe be extracted to a utils file/directory together with isPointInPolygon

lightly_studio_view/src/lib/components/PlotPanel/PlotPanel.svelte
```diff
@@ -1,281 +1,65 @@
...
+            arrowData: $arrowData,
+            rangeSelection
+        })
+    );
 
-        if (selectedSampleIds.length > 0) {
-            nextFilters.sample_ids = selectedSampleIds;
-        } else {
-            delete nextFilters.sample_ids;
+    const areEqualArrays = (a: string[], b: string[]) => {
```
mihnea-necsulescu: is this needed? wasn't lodash working as expected?

lightly_studio_view/src/lib/components/PlotPanel/useArrowData/useArrowData.ts
```diff
@@ -0,0 +1,47 @@
...
+    error: Writable<string | undefined>;
+};
+
+/**
+ * Hook to read Arrow data from a Blob.
+ *
+ * @param dataBlob
+ * @returns
+ */
+export function useArrowData({ blobData }: { blobData: Blob }): UseArrowDataReturn {
```
mihnea-necsulescu, line 22: can we add some comments here? it's a bit hard to follow

lightly_studio_view/src/lib/hooks/useEmbeddingSelection/isPointInPolygon.ts
mihnea-necsulescu: we could also remove the tests, but you're right, I didn't check the size

lightly_studio_view/src/lib/components/PlotPanel/PlotPanel.svelte
ikondrat: not really 
I just forgot what we have it in lodash already, updated

lightly_studio_view/src/lib/components/PlotPanel/useArrowData/useArrowData.ts
ikondrat: yes sure and added test as well to explain what does it do in test

lightly_studio_view/src/lib/components/PlotPanel/usePlotData.ts
ikondrat: yes it is leftover

PR #74, opened 2025-10-31
Add video and video frame view with count

JonasWurst:
## What has changed and why?

This PR adds the views with count as response model for `video` and `video_frame` models. This will be required to pass the data to the frontend (follow up pr).

## How has it been tested?

api routes will be introduced and tested in follow up.

- [ ] Yes


horatiualmasan (approved):
LGTM

PR #78, opened 2025-11-03
Check dataset type when adding samples

michal-lightly:
## What has changed and why?

Check the dataset type when adding images, which is currently the only sample type.

## How has it been tested?

New unit tests.

- [ ] Yes

### DIFF
--- a/lightly_studio/src/lightly_studio/models/video.py
+++ b/lightly_studio/src/lightly_studio/models/video.py
@@ -1,8 +1,10 @@
 """This module defines the Video and VideoFrame model for the application."""
 
-from typing import TYPE_CHECKING, List
+from typing import TYPE_CHECKING, List, Optional
 from uuid import UUID
 
+from pydantic import BaseModel, ConfigDict
+from pydantic import Field as PydanticField
 from sqlalchemy.orm import Mapped
 from sqlmodel import Field, Relationship, SQLModel
 
@@ -65,6 +67,16 @@ class VideoView(SQLModel):
     sample: SampleView
 
 
+class VideoViewsWithCount(BaseModel):
+    """Response model for counted videos."""
+
+    model_config = ConfigDict(populate_by_name=True)
+
+    data: List[VideoView] = PydanticField(...)
+    total_count: int
+    next_cursor: Optional[int] = PydanticField(None, alias="nextCursor")
+
+
 class VideoFrameBase(SQLModel):
     """Base class for the VideoFrame model."""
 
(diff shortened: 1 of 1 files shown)


lightly_studio/tests/resolvers/image_resolver/test_create.py
```diff
@@ -1,12 +1,52 @@
...
-    image_resolver,
-)
-from tests.helpers_resolvers import (
-    create_dataset,
-)
+from lightly_studio.resolvers import image_resolver
+from tests.helpers_resolvers import create_dataset
+
+
+def test_create(test_db: Session) -> None:
```
michal-lightly, line 10: Added a missing plain test for `create()`

lightly_studio/tests/resolvers/image_resolver/test_create.py
```diff
@@ -1,12 +1,52 @@
...
+
+    # Fetch from db
+    retrieved_images = image_resolver.get_all_by_dataset_id(
+        session=test_db, dataset_id=dataset.dataset_id
+    )
+    assert len(retrieved_images.samples) == 1
+    assert retrieved_images.samples[0].sample_id == image.sample_id
+
+
+def test_create__sample_type_mismatch(test_db: Session) -> None:
```
JonasWurst, line 37: Major: I'm missing a test for the `create_many` for the mismatch case

lightly_studio/tests/resolvers/image_resolver/test_create.py
JonasWurst: Alternatively, we could revisit if we need both methods: `create` and `create_many`

lightly_studio/src/lightly_studio/models/dataset.py
```diff
@@ -84,6 +83,10 @@ def get_samples(
         Returns:
             A sequence of ImageTable objects.
         """
+        # TODO(Michal, 11/2025): Import moved here to avoid circular imports. Remove this function
+        # completely in the future in favor of iter(Dataset).
+        from lightly_studio.resolvers import image_resolver
```
JonasWurst, line 88: What change did make this required?

michal-lightly: dataset.py imports image_resolver here
image_resolver newly imports dataset.py to get the SampleType

JonasWurst: Can we move `SampleType` to `type_defintions.py` instead -> no circular import

michal-lightly: I tried something similar. It did not work.

I prefer this solution - it makes it more clear the get_samples function should be deprecated, and that dataset.py should not depend on image_resolver.

lightly_studio/tests/resolvers/image_resolver/test_create.py
michal-lightly: Added a test

PR #81, opened 2025-11-04
Return sample IDs from create_many

michal-lightly:
## What has changed and why?

Return sample IDs from create_many instead of the full pydantic objects.

This addresses the issue that the returned objects would not be completely refreshed.

## How has it been tested?

Added tests.

- [ ] Yes


lightly_studio/tests/helpers_resolvers.py
```diff
@@ -104,7 +104,7 @@ def create_image(
     )
     # TODO(Michal, 11/2025): Prepared for create_many to return only sample IDs.
```
horatiualmasan: You can remove the TODO

Issue #7, opened 2025-10-21
💡 Feature Requests and Ideas for LightlyStudio


IgorSusmelj:
We would love your feedback on what to build next for **LightlyStudio**.

🗳️ **How to participate:**
- Drop your feature ideas or requests as a comment below.
- Upvote the comments that match what you want to see in **LightlyStudio** first.

We will regularly review the most upvoted ideas and keep you updated as we roll out new features.

Your feedback helps shape the future of open-source data tooling. Thank you for being part of the journey!

faileon:
Dockerfile would be nice.

eternal-f1ame:
@masakljun lovely, I will check this out over the weekend

leiyu1980:
Hope to add annotations for key points as soon as possible, and the key points need to be displayed with serial numbers.

PR #84, opened 2025-11-04
add tag_depth=1 support to add_samples_from_path

masakljun:
## What has changed and why?

Added support for automatically creating sample tags when adding data from a folder structure.
When calling add_samples_from_path with tag_depth=1, the function now creates a tag for each parent directory and assigns new samples to those tags.
This enables users to organize data imported from directory-based datasets.
Related to https://github.com/lightly-ai/lightly-studio/issues/7#issuecomment-3465939257
## How has it been tested?
Added 3 tests to test_dataset.py

- [ ] Yes

### DIFF
--- a/lightly_studio/src/lightly_studio/api/routes/api/video.py
+++ b/lightly_studio/src/lightly_studio/api/routes/api/video.py
@@ -1,5 +1,6 @@
 """API routes for dataset videos."""
 
+from typing import Optional
 from uuid import UUID
 
 from fastapi import APIRouter, Depends, Path
@@ -7,9 +8,9 @@ from typing_extensions import Annotated
 
 from lightly_studio.api.routes.api.validators import Paginated, PaginatedWithCursor
 from lightly_studio.db_manager import SessionDep
-from lightly_studio.models.video import VideoViewsWithCount
-from lightly_studio.resolvers.video_resolver import get_all_by_dataset_id
-from lightly_studio.resolvers.video_resolver.get_all_by_dataset_id import VideosWithCount
+from lightly_studio.models.video import VideoTable, VideoView, VideoViewsWithCount
+from lightly_studio.resolvers import video_resolver
+from lightly_studio.resolvers.video_resolver import VideosWithCount
 
 video_router = APIRouter(prefix="/datasets/{dataset_id}/video", tags=["video"])
 
@@ -27,11 +28,28 @@ def get_all_videos(
         dataset_id: The ID of the dataset to retrieve videos for.
         pagination: Pagination parameters including offset and limit.
 
-    Return:
+    Returns:
         A list of videos along with the total count.
     """
-    return get_all_by_dataset_id(
+    return video_resolver.get_all_by_dataset_id(
         session=session,
         dataset_id=dataset_id,
         pagination=Paginated(offset=pagination.offset, limit=pagination.limit),
     )
+
+
+@video_router.get("/{sample_id}", response_model=VideoView)
+def get_video_by_id(
+    session: SessionDep,
+    sample_id: Annotated[UUID, Path(title="Sample ID")],
+) -> Optional[VideoTable]:
+    """Retrieve a video for a given dataset ID by its ID.
+
+    Args:
+        session: The database session.
+        sample_id: The ID of the video to retrieve.
+
+    Returns:
+        A video object.
+    """
+    return video_resolver.get_by_id(session=session, sample_id=sample_id)
--- a/lightly_studio/src/lightly_studio/resolvers/video_resolver/__init__.py
+++ b/lightly_studio/src/lightly_studio/resolvers/video_resolver/__init__.py
@@ -2,10 +2,16 @@
 
 from lightly_studio.resolvers.video_resolver.create_many import create_many
 from lightly_studio.resolvers.video_resolver.filter_new_paths import filter_new_paths
-from lightly_studio.resolvers.video_resolver.get_all_by_dataset_id import get_all_by_dataset_id
+from lightly_studio.resolvers.video_resolver.get_all_by_dataset_id import (
+    VideosWithCount,
+    get_all_by_dataset_id,
+)
+from lightly_studio.resolvers.video_resolver.get_by_id import get_by_id
 
 __all__ = [
+    "VideosWithCount",
     "create_many",
     "filter_new_paths",
     "get_all_by_dataset_id",
+    "get_by_id",
 ]
--- a/lightly_studio/src/lightly_studio/resolvers/video_resolver/get_by_id.py
+++ b/lightly_studio/src/lightly_studio/resolvers/video_resolver/get_by_id.py
@@ -1,0 +1,25 @@
+"""Find a video by its id."""
+
+from __future__ import annotations
+
+from uuid import UUID
+
+from sqlmodel import Session, select
+
+from lightly_studio.models.video import VideoTable
+
+
+def get_by_id(session: Session, sample_id: UUID) -> VideoTable | None:
+    """Retrieve a video for a given dataset ID by its ID.
+
+    Args:
+        session: The database session.
+        sample_id: The ID of the video to retrieve.
+
+    Returns:
+        A video object or none.
+    """
+    query = select(VideoTable).where(
+        VideoTable.sample_id == sample_id,
+    )
+    return session.exec(query).one()
--- a/lightly_studio/tests/api/routes/api/test_video.py
+++ b/lightly_studio/tests/api/routes/api/test_video.py
@@ -1,11 +1,12 @@
+from uuid import UUID
+
 from fastapi.testclient import TestClient
 from sqlmodel import Session
 
 from lightly_studio.api.routes.api.status import HTTP_STATUS_OK
 from lightly_studio.models.dataset import SampleType
-from tests.helpers_resolvers import (
-    create_dataset,
-)
+from lightly_studio.resolvers import video_resolver
+from tests.helpers_resolvers import create_dataset
 from tests.resolvers.video_resolver.helpers import VideoStub, create_videos
 
 
@@ -38,3 +39,32 @@ def test_get_all_videos(test_client: TestClient, db_session: Session) -> None:
     assert result["total_count"] == 2
     assert data[0]["file_path_abs"].endswith("sample1.mp4")
     assert data[1]["file_path_abs"].endswith("sample2.mp4")
+
+
+def test_get_video_by_id(test_client: TestClient, db_session: Session) -> None:
+    dataset = create_dataset(session=db_session, sample_type=SampleType.VIDEO)
+    dataset_id = dataset.dataset_id
+
+    create_videos(
+        session=db_session,
+        dataset_id=dataset_id,
+        videos=[
+            VideoStub(path="/path/to/sample1.mp4"),
+            VideoStub(path="/path/to/sample2.mp4"),
+        ],
+    )
+    videos = video_resolver.get_all_by_dataset_id(
+        session=db_session,
+        dataset_id=dataset_id,
+    ).samples
+
+    sample_id = videos[0].sample_id
+
+    response = test_client.get(
+        f"/api/datasets/{dataset_id}/video/{sample_id}",
+    )
+
+    assert response.status_code == HTTP_STATUS_OK
+    result = response.json()
+    assert UUID(result["sample_id"]) == sample_id
+    assert result["file_path_abs"].endswith("sample1.mp4")
--- a/lightly_studio/tests/resolvers/video_resolver/test_get_by_id.py
+++ b/lightly_studio/tests/resolvers/video_resolver/test_get_by_id.py
@@ -1,0 +1,32 @@
+from sqlmodel import Session
+
+from lightly_studio.models.dataset import SampleType
+from lightly_studio.resolvers import (
+    video_resolver,
+)
+from tests.helpers_resolvers import create_dataset
+from tests.resolvers.video_resolver.helpers import VideoStub, create_videos
+
+
+def test_get_by_id(test_db: Session) -> None:
+    dataset = create_dataset(session=test_db, sample_type=SampleType.VIDEO)
+    dataset_id = dataset.dataset_id
+
+    create_videos(
+        session=test_db,
+        dataset_id=dataset_id,
+        videos=[
+            VideoStub(path="/path/to/sample1.mp4"),
+            VideoStub(path="/path/to/sample2.mp4"),
+        ],
+    )
+
+    videos = video_resolver.get_all_by_dataset_id(
+        session=test_db,
+        dataset_id=dataset_id,
+    ).samples
+
+    result = video_resolver.get_by_id(session=test_db, sample_id=videos[0].sample_id)
+
+    assert result is not None
+    assert result.file_name == "sample1.mp4"
--- a/lightly_studio_view/src/lib/schema.d.ts
+++ b/lightly_studio_view/src/lib/schema.d.ts
@@ -1161,10 +1161,37 @@ export interface paths {
          *         dataset_id: The ID of the dataset to retrieve videos for.
          *         pagination: Pagination parameters including offset and limit.
          *
-         *     Return:
+         *     Returns:
          *         A list of videos along with the total count.
          */
         get: operations["get_all_videos"];
+        put?: never;
+        post?: never;
+        delete?: never;
+        options?: never;
+        head?: never;
+        patch?: never;
+        trace?: never;
+    };
+    "/api/datasets/{dataset_id}/video/{sample_id}": {
+        parameters: {
+            query?: never;
+            header?: never;
+            path?: never;
+            cookie?: never;
+        };
+        /**
+         * Get Video By Id
+         * @description Retrieve a video for a given dataset ID by its ID.
+         *
+         *     Args:
+         *         session: The database session.
+         *         sample_id: The ID of the video to retrieve.
+         *
+         *     Returns:
+         *         A video object.
+         */
+        get: operations["get_video_by_id"];
         put?: never;
         post?: never;
         delete?: never;
@@ -4514,6 +4541,37 @@ export interface operations {
                 };
                 content: {
                     "application/json": components["schemas"]["VideoViewsWithCount"];
+                };
+            };
+            /** @description Validation Error */
+            422: {
+                headers: {
+                    [name: string]: unknown;
+                };
+                content: {
+                    "application/json": components["schemas"]["HTTPValidationError"];
+                };
+            };
+        };
+    };
+    get_video_by_id: {
+        parameters: {
+            query?: never;
+            header?: never;
+            path: {
+                sample_id: string;
+            };
+            cookie?: never;
+        };
+        requestBody?: never;
+        responses: {
+            /** @description Successful Response */
+            200: {
+                headers: {
+                    [name: string]: unknown;
+                };
+                content: {
+                    "application/json": components["schemas"]["VideoView"];
                 };
             };
             /** @description Validation Error */


lightly_studio/src/lightly_studio/core/dataset.py
```diff
@@ -288,6 +303,36 @@ def add_samples_from_path(
...
+
+            # Fetch newly created samples to get their file paths
+            newly_created_samples = (
+                self.query()
+                .match(SampleField.id.is_in(created_sample_ids))
+                .to_list()
+            )
+
+            for sample in newly_created_samples:
+                parent_dir = Path(sample.file_path_abs).parent.name.strip()
```
MalteEbner: This will not get the depth=1, but the depth=last directory.

E.g. if `path = "/my/path"` and `sample.file_path_abs="/my/path/dir_depth_1/dir_depth_2/img.jpg`, then `parent_dir="dir_depth_2"`.
The solution is to rather use a logic similar to `Path(sample.file_path_abs).relative_to(path).split("/")[0]`.

You also need to handle the case of an image being in the path directly, e.g. `path = "/my/path"` and `sample.file_path_abs="/my/path/img.jpg` should not lead to an error.

lightly_studio/src/lightly_studio/core/dataset.py
```diff
@@ -288,6 +303,36 @@ def add_samples_from_path(
...
+                self.query()
+                .match(SampleField.id.is_in(created_sample_ids))
+                .to_list()
+            )
+
+            for sample in newly_created_samples:
+                parent_dir = Path(sample.file_path_abs).parent.name.strip() 
+                if not parent_dir:
+                    parent_dir = "root"  
+                parent_dir_to_sample_ids[parent_dir].append(sample.sample_id)
```
MalteEbner: The current implementation is contrary to the GitHub issue https://github.com/lightly-ai/lightly-studio/issues/7#issuecomment-3465939257, which tells to not set a tag when an image is in the root directory.

```suggestion
                if parent_dir:
                    parent_dir_to_sample_ids[parent_dir].append(sample.sample_id)
```

lightly_studio/src/lightly_studio/core/dataset.py
```diff
@@ -288,6 +303,36 @@ def add_samples_from_path(
             image_paths=image_paths,
         )
 
+        if tag_depth == 1 and created_sample_ids:
+            print(f"Adding directory tags to {len(created_sample_ids)} new samples.")
```
MalteEbner: Can you move this whole logic into a separate function? E.g.

`def tag_images_based_on_directories(image_root: PathLike, image_ids: Sequence[UUID])`

lightly_studio/src/lightly_studio/core/dataset.py
```diff
@@ -288,11 +304,68 @@ def add_samples_from_path(
...
+
+        try:
+            input_path_abs = Path(input_path).absolute()
+        except Exception as e:
+            print(f"Warning: Could not resolve absolute path for {input_path}. Skipping tagging. Error: {e}")
+            return
+
+        newly_created_samples = (
+            self.query()
+            .match(SampleField.id.is_in(sample_ids))
```
masakljun: Hey @MalteEbner , a quick question/thought:

In my function I need to get the ids of the new samples, so I can then get the correct file paths for the creation of the new tags.

Initially I thought I could just get the ids from the file lister at the start of `add_samples_from_path()`. However, the `image_paths` list and the `created_sample_ids` are not necessarily a 1 to 1 match (if some images don't successfully get turned into samples).

Therefore, I think one option is to re-query to get the correct `file_path_abs` for each sample id that was actually created. 

While the other option is to add `sample_id` to the `SampleField` class. This seems like the cleaner option, as it makes SampleField more complete. Then I could do  .match(SampleField.sample_id.is_in(sample_ids)).

What do you think?

lightly_studio/src/lightly_studio/core/dataset.py
```diff
@@ -288,11 +305,68 @@ def add_samples_from_path(
...
 
+    def _tag_samples_by_directory(
+        self,
+        input_path: PathLike,
+        sample_ids: list[UUID],
+    ) -> None:
+        """Tags samples based on their first-level subdirectory relative to input_path."""
+
+        print(f"Adding directory tags to {len(sample_ids)} new samples.")
+        parent_dir_to_sample_ids = defaultdict(list)
```
MalteEbner: Minor: Can you move the `parent_dir_to_sample_ids` var closer to where it is used? E.g. directly before the for-loop?

lightly_studio/src/lightly_studio/core/dataset.py
```diff
@@ -288,11 +305,68 @@ def add_samples_from_path(
...
+
+        try:
+            input_path_abs = Path(input_path).absolute()
+        except Exception as e:
```
MalteEbner: In our [python code style notion docs](https://www.notion.so/getlightly/Python-code-style-290e575bc2134c3eb4c67d28e350d954), it links to the google style guide, which covers exceptions. 

Some relevant part of it:

> Never use catch-all except: statements, or catch Exception or StandardError

In this case: I would remove the try-catch completely and just let users run into the error.

lightly_studio/src/lightly_studio/core/dataset.py
```diff
@@ -288,11 +305,68 @@ def add_samples_from_path(
...
+
+                if len(relative_path.parts) > 1:
+                    tag_name = relative_path.parts[0].strip()
+                    if tag_name:
+                        parent_dir_to_sample_ids[tag_name].append(sample.sample_id)
+
+            except ValueError:
+                print(f"Warning: Sample {sample.file_name} is not in the input path. Skipping tagging.")
+            except Exception as e:
+                print(f"Warning: Error processing tag for {sample.file_name}: {e}. Skipping.")
```
MalteEbner: Similarly to above, I would remove this exception catching completely.

lightly_studio/src/lightly_studio/core/dataset.py
```diff
@@ -288,11 +305,68 @@ def add_samples_from_path(
...
+            print(f"Warning: Could not resolve absolute path for {input_path}. Skipping tagging. Error: {e}")
+            return
+
+        # todo: this doesnt work ATM since sample_id is not attribute of SampleField
+        newly_created_samples = (
+            self.query()
+            .match(SampleField.sample_id.is_in(sample_ids)) 
+            .to_list()
+        )
+        for sample in newly_created_samples:
```
MalteEbner: The self.query is the public interface. Internally, we can use the resolvers directly. E.g. we can use
[sample_resolver.get_many_by_id](https://github.com/lightly-ai/lightly-studio/blob/e534887e58781421641b4d713a73b180317a0292/lightly_studio/src/lightly_studio/resolvers/sample_resolver.py#L40)

lightly_studio/tests/core/test_dataset.py
```diff
@@ -356,6 +358,78 @@ def test_compute_typicality_metadata(
...
+        img.save(tmp_path / "site_1" / "deep_dir" / "img2.png")
+        (tmp_path / " site_2 ").mkdir()
+        img.save(tmp_path / " site_2 " / "img3.png")
+
+        # Run with tag_depth=1
+        dataset = Dataset.create()
+        dataset.add_samples_from_path(tmp_path, tag_depth=1, embed=False)
+
+        # Order by filename to have a consistent order for checking
+        samples = dataset.query().order_by(SampleField.file_name).to_list()
```
MalteEbner: We don't need a consistent filename order, as the samples are only used for creating a dictionary.
```suggestion
        # Order by filename to have a consistent order for checking
        samples = dataset.query().to_list()
```

lightly_studio/tests/core/test_dataset.py
```diff
@@ -356,6 +358,78 @@ def test_compute_typicality_metadata(
...
+        assert len(samples) == 4
+
+        # Create a map of filename -> {set of tag names} for easy checking
+        tag_map = {s.file_name: {tag.name for tag in s.tags} for s in samples}
```
MalteEbner: We usually name dictionaries in the format of `f"{key_name}_to_{value_name}"`.

```suggestion
        # Create a map of filename -> {set of tag names} for easy checking
        sample_filename_to_tags = {s.file_name: {tag.name for tag in s.tags} for s in samples}
```

lightly_studio/tests/core/test_dataset.py
```diff
@@ -356,6 +358,78 @@ def test_compute_typicality_metadata(
...
         assert samples[2].metadata["typicality"] == pytest.approx(0.3853082)
 
+    def test_add_samples_from_path_tag_depth_invalid(
+        self,
+        patch_dataset: None,  # noqa: ARG002
+        tmp_path: Path,
+    ) -> None:
+        """Tests that tag_depth > 1 raises an error."""
+        dataset = Dataset.create(name="test_dataset")
+        with pytest.raises(NotImplementedError):
```
MalteEbner: When asserting errors, we usually assert parts of the error message using the `match` argument.

```suggestion
        with pytest.raises(NotImplementedError, match=r"tag_depth > 1 is not yet implemented for add_samples_from_path"):
```

Not sure if this works, you might need to escape some special chars of the match argument. Best try it out.

lightly_studio/tests/core/test_dataset.py
```diff
@@ -356,6 +358,78 @@ def test_compute_typicality_metadata(
...
+    ) -> None:
+        """Tests the default behavior (tag_depth=0) adds samples but no tags."""
+                
+        # Create a valid image
+        img = Image.new("RGB", (10, 10), color="red")
+        
+        img.save(tmp_path / "root_img.png")
+        
+        (tmp_path / "site_1").mkdir()
+        img.save(tmp_path / "site_1" / "img1.png")
```
MalteEbner: Please use the test helper function [create_images](https://github.com/lightly-ai/lightly-studio/blob/e534887e58781421641b4d713a73b180317a0292/lightly_studio/tests/helpers_resolvers.py#L127) instead.

lightly_studio/tests/core/test_dataset.py
```diff
@@ -356,6 +358,78 @@ def test_compute_typicality_metadata(
...
+        """Tests that tag_depth=1 correctly tags samples based on directory structure."""
+        # Setup file structure with all edge cases
+        img = Image.new("RGB", (10, 10), color="blue")
+        img.save(tmp_path / "root_img.png")
+        (tmp_path / "site_1").mkdir()
+        img.save(tmp_path / "site_1" / "img1.png")
+        (tmp_path / "site_1" / "deep_dir").mkdir(parents=True)
+        img.save(tmp_path / "site_1" / "deep_dir" / "img2.png")
+        (tmp_path / " site_2 ").mkdir()
+        img.save(tmp_path / " site_2 " / "img3.png")
```
MalteEbner: Please use the test helper function [create_images](https://github.com/lightly-ai/lightly-studio/blob/e534887e58781421641b4d713a73b180317a0292/lightly_studio/tests/helpers_resolvers.py#L127) instead.

lightly_studio/src/lightly_studio/core/dataset.py
MalteEbner: > In my function I need to get the ids of the new samples, so I can then get the correct file paths for the creation of the new tags. However, the image_paths list and the created_sample_ids are not necessarily a 1 to 1 match (if some images don't successfully get turned into samples).

Yes, you got that correctly. The goal is to add the tag to all samples that were actually created. Thus you can just use the `created_sample_ids` as you already do in your PR.

MalteEbner: > While the other option is to add sample_id to the SampleField class. This seems like the cleaner option, as it makes SampleField more complete. Then I could do .match(SampleField.sample_id.is_in(sample_ids)).

See https://github.com/lightly-ai/lightly-studio/pull/84#discussion_r2494580096

lightly_studio/src/lightly_studio/core/dataset.py
```diff
@@ -282,20 +291,69 @@ def add_samples_from_path(
                 path=str(path), allowed_extensions=allowed_extensions_set
             )
         )
+
+        if tag_depth > 1:
+            raise NotImplementedError(
+                "tag_depth > 1 is not yet implemented for add_samples_from_path."
+            )
```
MalteEbner: I thought about the location of this. I would move this into `_tag_samples_by_directory`.
Then we can call `_tag_samples_by_directory` from `add_samples_from_coco` and `add_samples_from_yolo` in the future. 

See also the comment one below.

lightly_studio/src/lightly_studio/core/dataset.py
```diff
@@ -282,20 +291,69 @@ def add_samples_from_path(
...
 
-        # Process images.
+        # Process images
         created_sample_ids = add_samples.load_into_dataset_from_paths(
             session=self.session,
             dataset_id=self.dataset_id,
             image_paths=image_paths,
         )
 
+        if tag_depth == 1 and created_sample_ids:
```
MalteEbner: Let's move the validation of `tag_depth` into `_tag_samples_by_directory`.
If tag_depth=0: early return
if tag_depth>1: raise NotImplementedError
else: normal path

```suggestion
        if created_sample_ids:
```

See also the comment one above.

lightly_studio/src/lightly_studio/core/dataset.py
```diff
@@ -282,20 +291,69 @@ def add_samples_from_path(
...
+        newly_created_samples = [Sample(inner=image) for image in newly_created_images]
+
+        print(f"Adding directory tags to {len(sample_ids)} new samples.")
+        parent_dir_to_sample_ids: defaultdict[str, list[UUID]] = defaultdict(list)
+        for sample in newly_created_samples:
+            sample_path_abs = Path(sample.file_path_abs)
+            relative_path = sample_path_abs.relative_to(input_path_abs)
+
+            if len(relative_path.parts) > 1:
+                tag_name = relative_path.parts[0].strip()
```
MalteEbner: I  would leave the stripping away. If users actually have a space in the path, e.g. `/my/ dir/` and `/my/dir /` and `/my/dir/`, they probably care about preserving that space.

This makes it also possible for users to get the directory name from the tag name, i.e. users may want to  do something like the following. When stripping spaces, it will fail instead.

```
for tag = dataset.get_tags() # get_tags() is a placeholder, it does not exist
     directory = Path(input_path) / tag.name
     os.copy(directory, new_path / tag.name)
```


```suggestion
                tag_name = relative_path.parts[0]
```

lightly_studio/src/lightly_studio/core/dataset.py
```diff
@@ -282,20 +291,69 @@ def add_samples_from_path(
...
         if embed:
             _generate_embeddings(
                 session=self.session, dataset_id=self.dataset_id, sample_ids=created_sample_ids
             )
 
+    def _tag_samples_by_directory(
+        self,
+        input_path: PathLike,
+        sample_ids: list[UUID],
+    ) -> None:
```
MalteEbner: Optional, idea:
Sorry for only thinking about this now, I just had a closer look at the overall directory structure.

What do you think about moving this function to [add_samples.py](https://github.com/lightly-ai/lightly-studio/blob/main/lightly_studio/src/lightly_studio/core/add_samples.py) instead?

Feel free to challenge that idea and give me reasons why not to do it ;)

lightly_studio/tests/core/test_dataset.py
```diff
@@ -356,6 +356,108 @@ def test_compute_typicality_metadata(
...
+        self,
+        patch_dataset: None,  # noqa: ARG002
+    ) -> None:
+        """Tests that tag_depth > 1 raises an error."""
+        dataset = Dataset.create(name="test_dataset")
+        with pytest.raises(
+            NotImplementedError,
+            match="tag_depth > 1 is not yet implemented for add_samples_from_path",
+        ):
+            dataset.add_samples_from_path(".", tag_depth=2)
```
MalteEbner: All 3 tests currently call `dataset.add_samples_from_path` and not `_tag_samples_by_directory`. Thus they are more integration tests than unittests.

I would change them to call `_tag_samples_by_directory` instead. 
From a testing point of view, this makes sure that only and exactly the relevant part is tested.
From a test implementation point of view, it allows to leave away all the the mocking of `lightly_studio.dataset.fsspec_lister.iter_files_from_path`, the files do not need to be on disk anymore. It is sufficient if the samples are in the DB, which happens with `create_images(`.

MalteEbner (commented):
The logic of `_tag_samples_by_directory` is now almost correctly, only some minor improvement ideas.

Regarding the tests: After moving the `tag_depth` logic into `_tag_samples_by_directory`, the tests should only test `_tag_samples_by_directory` and not the whole `add_samples_from_path` anymore. See https://github.com/lightly-ai/lightly-studio/pull/84/files#r2502176747

lightly_studio/src/lightly_studio/core/dataset.py
masakljun: Thanks for the suggestion. I'm not familiar with the codebase yet, but I had a look through `add_samples.py` and `dataset.py`, and what you said makes sense to me.

It feels cleaner to keep the Dataset class as the main interface for users and move the actual tagging logic next to the other functions that handle adding data to the database.

So yeah I think it makes sense to move as it will be more organized that way

lightly_studio/tests/core/test_dataset.py
masakljun: Good idea, thanks!

masakljun:
> The logic of `_tag_samples_by_directory` is now almost correctly, only some minor improvement ideas.
> 
> Regarding the tests: After moving the `tag_depth` logic into `_tag_samples_by_directory`, the tests should only test `_tag_samples_by_directory` and not the whole `add_samples_from_path` anymore. See https://github.com/lightly-ai/lightly-studio/pull/84/files#r2502176747

Thank you again for your review and great feedback. I now changed the logic but have some issues still with tests not going through, will try to resolve and then push the changes.

lightly_studio/tests/core/test_add_samples.py
```diff
@@ -228,6 +235,112 @@ def test_create_label_map(db_session: Session) -> None:
...
+
+    images_to_create = [
+        ImageStub(path=mock_image_paths[0]),
+        ImageStub(path=mock_image_paths[1]),
+    ]
+    created_images = create_images(
+        db_session=db_session,
+        dataset_id=dataset_table.dataset_id,
+        images=images_to_create,
+    )
```
MalteEbner, line 268: This can be shortened a lot, no need for all the temporary variables.

```suggestion
    dataset_table = create_dataset(db_session, "test_dataset")
    mock_root_path = "/mock/path"
    created_images = create_images(
        db_session=db_session,
        dataset_id=dataset_table.dataset_id,
        images=[
            ImageStub(path=f"{mock_root_path}/root_img.png"),
            ImageStub(path=f"{mock_root_path}/site_1/img1.png")
        ],
    )
```

lightly_studio/tests/core/test_add_samples.py
```diff
@@ -228,6 +235,112 @@ def test_create_label_map(db_session: Session) -> None:
...
+        ImageStub(path=mock_image_paths[0]),
+        ImageStub(path=mock_image_paths[1]),
+        ImageStub(path=mock_image_paths[2]),
+        ImageStub(path=mock_image_paths[3]),
+    ]
+    created_images = create_images(
+        db_session=db_session,
+        dataset_id=dataset_table.dataset_id,
+        images=images_to_create,
+    )
```
MalteEbner, line 304: This can be shortened a lot, see the comment above.

lightly_studio/tests/core/test_add_samples.py
```diff
@@ -14,11 +16,16 @@
...
 from lightly_studio.core import add_samples
 from lightly_studio.models.image import ImageCreate
 from lightly_studio.models.sample import SampleTable
 from lightly_studio.resolvers import caption_resolver, image_resolver
-from tests.helpers_resolvers import create_dataset
+from tests.helpers_resolvers import (
+    ImageStub,
+    create_dataset,
+    create_images,
+)
```
MalteEbner, line 27: Code style: We don't import functions directly, but rather their file. Then call `helpers_resolver.create_dataset` and `helpers_resolvers.create_image`. 

```suggestion
from tests.helpers_resolvers import (
    ImageStub,
)
from tests import helpers_resolvers
```

lightly_studio/tests/core/test_add_samples.py
```diff
@@ -341,6 +325,35 @@ def test_tag_samples_by_directory_tag_depth_1(
...
+    mocker.patch("lightly_studio.core.dataset._generate_embeddings")
+    spy_tagger = mocker.spy(add_samples, "tag_samples_by_directory")
+
+    dataset_table = helpers_resolvers.create_dataset(db_session, "test_dataset")
+    dataset = Dataset(dataset=dataset_table)
+    dataset.session = db_session
+
+    _create_sample_images([tmp_path / "image1.jpg"])
+
+    dataset.add_samples_from_path(path=str(tmp_path), tag_depth=0)
```
MalteEbner: I did not test it, but is setting `embed=False` sufficient, so that `mocker.patch("lightly_studio.core.dataset._generate_embeddings")` is not needed?

```suggestion
    spy_tagger = mocker.spy(add_samples, "tag_samples_by_directory")

    dataset_table = helpers_resolvers.create_dataset(db_session, "test_dataset")
    dataset = Dataset(dataset=dataset_table)
    dataset.session = db_session

    _create_sample_images([tmp_path / "image1.jpg"])

    dataset.add_samples_from_path(path=str(tmp_path), tag_depth=0, embed=False)
```

Feel free to resolve this comment yourself if it does not work.

lightly_studio/tests/core/test_add_samples.py
```diff
@@ -341,6 +325,35 @@ def test_tag_samples_by_directory_tag_depth_1(
     assert sample_filename_to_tags["root_img.png"] == set()
 
 
+def test_add_samples_from_path_calls_tag_samples_by_directory(
```
MalteEbner: Major: This belongs into `lightly_studio/tests/core/test_dataset__path.py`, as it is testing a function in `dataset.py`, not in `add_samples.py`.
As `test_dataset.py` would become to large if we put all tests into it, we have split it into different test files, including `test_dataset__path.py` for testing `Dataset.add_samples_from_path`.

You can move it easily in VSCode via right-clicking on the function name -> Refactor -> Move symbol to. This will also move all imports.

lightly_studio/tests/core/test_add_samples.py
```diff
@@ -341,6 +325,35 @@ def test_tag_samples_by_directory_tag_depth_1(
...
+    The add_samples.tag_samples_by_directory helper.
+    """
+    mocker.patch("lightly_studio.core.dataset._generate_embeddings")
+    spy_tagger = mocker.spy(add_samples, "tag_samples_by_directory")
+
+    dataset_table = helpers_resolvers.create_dataset(db_session, "test_dataset")
+    dataset = Dataset(dataset=dataset_table)
+    dataset.session = db_session
+
+    _create_sample_images([tmp_path / "image1.jpg"])
```
MalteEbner: Minor: This can be done more easily, see https://github.com/lightly-ai/lightly-studio/blob/e716cbaa06b9276389695cc8a6138597aacc059a/lightly_studio/tests/core/test_dataset__path.py#L177-L179

masakljun: Thanks for the suggestion!

However, I tried using `Dataset.create()`, but it looks like it causes an `IntegrityError` in this specific test. It seems `Dataset.create()` uses the app's persistent session, but this test requires everything to use the isolated `db_session` fixture.

So in the end I now changed back to my original, as this then causes the test to pass.

lightly_studio/tests/core/test_add_samples.py
masakljun: Ah, yes, you're right, using `embed=False` is much cleaner. Thanks!

lightly_studio/tests/core/test_dataset__path.py
```diff
@@ -189,3 +193,30 @@ def _create_sample_images(image_paths: list[Path]) -> None:
...
+
+    dataset.add_samples_from_path(path=str(tmp_path), tag_depth=0, embed=False)
+
+    spy_tagger.assert_called_once_with(
+        session=db_session,
+        dataset_id=dataset.dataset_id,
+        input_path=str(tmp_path),
+        sample_ids=mocker.ANY,
+        tag_depth=0,
+    )
```
MalteEbner: As this is a test for a method and not a class-less function, it should be a method of the `TestDataset` class earlier in this file. Thus just move it there. See `test_dataset_add_samples_from_path__dont_embed` in the same file a few lines earlier, it can go directly below it, while still being a method. Note that this means that the `self` argument needs to be added to `test_add_samples_from_path_calls_tag_samples_by_directory`

MalteEbner: Feel free to implement this change and then resolve the comment yourself.

PR #85, opened 2025-11-04
Adding Basic Operator Parameters

JonasWurst:
## What has changed and why?

This PR adds the parameter base implementation, as well as the string, bool, int, and float version. For those, more easy solutions would have been possible. However I decided for this design, since we want to keep it flexible for more custom types as well. E.g. ClassChoiseParameter, or PathParamter, etc. Having separate classes ensures highest flexibility here.

The actual operator will be part of a follow pr.

## How has it been tested?

unit tests are in place

- [ ] Yes

### DIFF
--- a/lightly_studio/src/lightly_studio/plugins/parameter.py
+++ b/lightly_studio/src/lightly_studio/plugins/parameter.py
@@ -43,7 +43,7 @@ class BuiltinParameter(BaseParameter, Generic[T]):
     def _validate(self, value: T) -> T:
         if isinstance(value, self._type):
             return cast(T, value)
-        raise TypeError(f"Expected value of type '{self._type.__name__}'")
+        raise TypeError(f"Expected value of type '{self._type.__name__}' but got {type(value)}'")
 
 
 class IntParameter(BuiltinParameter[int]):
--- a/lightly_studio/tests/plugins/test_paramter.py
+++ b/lightly_studio/tests/plugins/test_paramter.py
@@ -29,9 +29,9 @@ def test_builtin_parameters() -> None:
     _ = FloatParameter(name="test_float", default=42.0)
 
     # invalid default types
-    with pytest.raises(TypeError, match="Expected value of type 'int'"):
+    with pytest.raises(TypeError, match="Expected value of type 'int' but got <class 'str'>"):
         _ = IntParameter(name="test_int", default="42")
-    with pytest.raises(TypeError, match="Expected value of type 'int'"):
+    with pytest.raises(TypeError, match="Expected value of type 'int' but got <class 'float'>"):
         _ = IntParameter(name="test_int", default=42.0)
-    with pytest.raises(TypeError, match="Expected value of type 'float'"):
+    with pytest.raises(TypeError, match="Expected value of type 'float' but got <class 'int'>"):
         _ = FloatParameter(name="test_float", default=42)


lightly_studio/src/lightly_studio/plugins/parameter.py
```diff
@@ -0,0 +1,132 @@
...
+from __future__ import annotations
+
+from abc import ABC, abstractmethod
+from typing import Any
+
+
+class BaseParameter(ABC):
+    """Base parameter definition shared across operator parameters."""
+
+    __slots__ = ("default", "description", "name", "param_type", "required")
```
MalteEbner: Is there a reason this is not a dataclass? 
- The init could be deleted. The check if the default is of the correct type can be done in a postinit. 
- The to_dict could be made easier by using dataclasses.asdict: https://docs.python.org/3/library/dataclasses.html#dataclasses.asdict
- The tests would become shorter as we don't not need to test anymore that dataclasses indeed work as dataclasses. 


And why are the `__slots__` added? They are a memory optimisation we don't need at all, as there are max a few instances.

lightly_studio/src/lightly_studio/plugins/parameter.py
```diff
@@ -0,0 +1,132 @@
...
+            "required": self.required,
+        }
+
+    @staticmethod
+    @abstractmethod
+    def _validate(value: Any) -> Any:
+        """Validate the parameter value."""
+
+
+class IntParameter(BaseParameter):
```
MalteEbner: Did you consider having a generic for these types? This would save us a lot of code to review and tests.

It would change the logic such that a bool is then a valid value for a float or int parameter, but that is fine for me.

lightly_studio/src/lightly_studio/plugins/parameter.py
```diff
@@ -0,0 +1,70 @@
...
+        if not hasattr(self, "_parameter_type") or self._parameter_type is None:
+            raise NotImplementedError("Subclasses must define _parameter_type class attribute")
+        self._type = self._parameter_type
+        self.param_type = self._parameter_type.__name__
+        super().__post_init__()
+
+    def _validate(self, value: T) -> T:
+        if isinstance(value, self._type):
+            return cast(T, value)
+        raise TypeError(f"Expected value of type '{self._type.__name__}'")
```
MalteEbner: Minor: Would be nice to tell the type that it actually got.
```suggestion
        raise TypeError(f"Expected value of type '{self._type.__name__} but got {type(value)}'")
```

lightly_studio/tests/plugins/test_paramter.py
```diff
@@ -0,0 +1,37 @@
...
+    # Test dict conversion yield the expected format
+    param = IntParameter(name="test_int", description="abc", default=42, required=False)
+
+    assert asdict(param) == {
+        "name": "test_int",
+        "description": "abc",
+        "default": 42,
+        "required": False,
+        "param_type": "int",
+    }
```
MalteEbner, line 21: We can leave this test away, as it only tests that dataclasses and the asdict function by python work as expected. It does not test our code.

```suggestion
```

lightly_studio/src/lightly_studio/plugins/parameter.py
```diff
@@ -0,0 +1,70 @@
...
+            self.default = self._validate(self.default)
+
+    @abstractmethod
+    def _validate(self, value: Any) -> Any:
+        """Validate the parameter value."""
+
+
+class BuiltinParameter(BaseParameter, Generic[T]):
+    """Represents a built-in operator parameter."""
+
```
MalteEbner, line 34: Can we type the default here? Not sure if this works.

```suggestion
class BuiltinParameter(BaseParameter, Generic[T]):
    """Represents a built-in operator parameter."""

    default: T | None
```

lightly_studio/tests/plugins/test_paramter.py
JonasWurst: not only that. it also validates that we receive the schema that we expect in the frontend. Hence, I will keep it.

lightly_studio/src/lightly_studio/plugins/parameter.py
JonasWurst: Great Idea. I tried but it did not change. Required changes would be more complex. Let's rely on the validator.

PR #89, opened 2025-11-05
Remove dataset_id from ImageCreate

michal-lightly:
## What has changed and why?

Remove dataset_id from ImageCreate. A step towards removing dataset_id from ImageTable.

## How has it been tested?

Tests.

- [ ] Yes

### DIFF
--- a/lightly_studio/src/lightly_studio/resolvers/image_resolver/create_many.py
+++ b/lightly_studio/src/lightly_studio/resolvers/image_resolver/create_many.py
@@ -16,24 +16,22 @@ class ImageCreateHelper(ImageCreate):
     """Helper class to create ImageTable with sample_id."""
 
     sample_id: UUID
+    dataset_id: UUID
 
 
-def create_many(session: Session, samples: list[ImageCreate]) -> list[UUID]:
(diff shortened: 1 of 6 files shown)


lightly_studio/src/lightly_studio/resolvers/image_resolver/create_many.py
```diff
@@ -16,24 +16,22 @@ class ImageCreateHelper(ImageCreate):
     """Helper class to create ImageTable with sample_id."""
 
     sample_id: UUID
+    dataset_id: UUID
```
michal-lightly, line 19: Needed while dataset_id is in ImageTable.

horatiualmasan (approved):
LGTM

lightly_studio/src/lightly_studio/resolvers/image_resolver/create_many.py
horatiualmasan: We would need `dataset_id ` also after it is not in ImageTable for creating the samples and checking the sample type, no?

michal-lightly: Yes, however in this class it exists just for inserting in the ImageTable, so this particular line will go away.

PR #90, opened 2025-11-05
Add videos from path

horatiualmasan:
## What has changed and why?

Add the option to load video files from a folder.

## How has it been tested?

Created new unit tests.
You can manually test by updating example.py and add 
`dataset.add_videos_from_path(path="path_to_videos", allowed_extensions=[".mp4"],fps=10,)`

- [ ] Yes

### DIFF
--- a/lightly_studio/src/lightly_studio/core/add_samples.py
+++ b/lightly_studio/src/lightly_studio/core/add_samples.py
@@ -362,23 +362,6 @@ def tag_samples_by_directory(
     print(f"Created {len(parent_dir_to_sample_ids)} tags from directories.")
 
 
-def _log_loading_results(
-    session: Session, dataset_id: UUID, logging_context: _LoadingLoggingContext
-) -> None:
-    n_samples_end = sample_resolver.count_by_dataset_id(session=session, dataset_id=dataset_id)
-    n_samples_inserted = n_samples_end - logging_context.n_samples_before_loading
-    print(
-        f"Added {n_samples_inserted} out of {logging_context.n_samples_to_be_inserted}"
-        " new samples to the dataset."
-    )
-    if logging_context.example_paths_not_inserted:
-        # TODO(Jonas, 09/2025): Use logging instead of print
-        print(
-            f"Examples of paths that were not added: "
-            f" {', '.join(logging_context.example_paths_not_inserted)}"
-        )
-
-
 def _create_batch_samples(
     session: Session, dataset_id: UUID, samples: list[ImageCreate]
 ) -> tuple[dict[str, UUID], list[str]]:


lightly_studio/src/lightly_studio/core/add_samples.py
```diff
@@ -53,21 +54,6 @@ class _AnnotationProcessingContext:
     label_map: dict[int, UUID]
 
 
-@dataclass
```
horatiualmasan, line 59: Moved to separate file as it will be used also for logging the video loading info.

michal-lightly: Please drop the leading underscores. That's a python naming convention to indicate private functions/classes.

Sidenote: That could have been a separate PR :)

lightly_studio/src/lightly_studio/core/add_videos.py
```diff
@@ -0,0 +1,213 @@
+"""Functions to add samples and their annotations to a dataset in the database."""
```
michal-lightly:
```suggestion
"""Functions to add video samples and their annotations to a dataset in the database."""
```

lightly_studio/src/lightly_studio/core/add_videos.py
```diff
@@ -0,0 +1,213 @@
...
+from lightly_studio.core.logging import _LoadingLoggingContext, _log_loading_results
+from lightly_studio.models.video import VideoCreate, VideoFrameCreate
+from lightly_studio.resolvers import (
+    dataset_resolver,
+    sample_resolver,
+    video_frame_resolver,
+    video_resolver,
+)
+
+_DEFAULT_VIDEO_CHANNEL = 0
```
michal-lightly: Minor: For constants, even private, it is ok not to have a leading underscore.

lightly_studio/src/lightly_studio/core/add_videos.py
```diff
@@ -0,0 +1,213 @@
...
+    """Load video frames from file paths into the dataset using PyAV.
+
+    Args:
+        session: The database session.
+        dataset_id: The ID of the dataset to load video frames into.
+        video_paths: An iterable of file paths to the videos to load.
+        video_channel: The video channel from which frames are loaded.
+        fps: Optional FPS value to control frame extraction. If provided, only frames
+            at the specified FPS intervals will be extracted. If None, all frames
+            will be extracted.
```
michal-lightly: This needs more detail. What if video is encoded at 2 fps and we request:
* 1.1 fps?
* 0.99 fps?
* 4 fps?

What if the video is encoded at a non-constant fps?

I would not mind introducing this parameter in a separate PR. It requires careful tests.

lightly_studio/src/lightly_studio/core/add_videos.py
```diff
@@ -0,0 +1,213 @@
...
+
+    Returns:
+        A list of UUIDs of the created samples.
+    """
+    created_sample_ids: list[UUID] = []
+    file_paths_new, file_paths_exist = video_resolver.filter_new_paths(
+        session=session, file_paths_abs=list(video_paths)
+    )
+    video_logging_context = _LoadingLoggingContext(
+        n_samples_to_be_inserted=sum(1 for _ in video_paths),
```
michal-lightly: Minor: Is this possible?
```suggestion
        n_samples_to_be_inserted=len(video_paths),
```

michal-lightly: Also, shouldn't `n_samples_to_be_inserted` be `len(file_paths_new)`?

lightly_studio/src/lightly_studio/core/dataset.py
JonasWurst, line 1: I have a few questions here:

- we shoud rename add_samples to add_iamges --> i assume this will happen in a follow up?
- we should add the new way of loading also to class docstring where the other examples are provided: L67-71
-

lightly_studio/src/lightly_studio/core/add_videos.py
```diff
@@ -0,0 +1,213 @@
...
+                video_sample_id=video_sample_id,
+                video_container=video_container,
+                video_stream=video_stream,
+                fps=fps,
+            )
+            created_sample_ids.extend(frame_sample_ids)
+
+            video_container.close()
+
+        except Exception as e:
```
michal-lightly: Major: Do not catch all the exceptions. Instead, list the exceptions that should be caught.

lightly_studio/src/lightly_studio/core/add_videos.py
```diff
@@ -0,0 +1,213 @@
...
+    # Get the video frames dataset ID
+    video_frames_dataset_id = dataset_resolver.get_or_create_video_frame_child(
+        session=session, dataset_id=dataset_id
+    )
+    for video_path in tqdm(
+        file_paths_new,
+        desc="Loading videos",
+        unit=" videos",
+    ):
+        try:
```
michal-lightly, line 83: Minor: I suggest moving loading a single video to its own function.

lightly_studio/src/lightly_studio/core/add_videos.py
JonasWurst: please make it more specific for videos. also, we are not yet loading any annotations

lightly_studio/src/lightly_studio/core/add_videos.py
```diff
@@ -0,0 +1,213 @@
...
+            # Open video and extract metadata
+            fs, fs_path = fsspec.core.url_to_fs(video_path)
+            content = fs.cat_file(fs_path)
+            video_buffer = BytesIO(content)
```
michal-lightly: Does this load the whole video into memory, or does it stream the video?

lightly_studio/src/lightly_studio/core/add_videos.py
JonasWurst:
```suggestion
DEFAULT_VIDEO_CHANNEL = 0
```
Or is it intentionally "private"?

lightly_studio/src/lightly_studio/core/add_videos.py
```diff
@@ -0,0 +1,213 @@
...
+            # When opening a file for reading, container is InputContainer
```
michal-lightly: Minor: I do not understand what is this comment trying to say. Can you reformulate?

lightly_studio/src/lightly_studio/core/add_videos.py
```diff
@@ -0,0 +1,213 @@
...
+        session=session, dataset_id=dataset_id
+    )
+    for video_path in tqdm(
+        file_paths_new,
+        desc="Loading videos",
+        unit=" videos",
+    ):
+        try:
+            # Open video and extract metadata
+            fs, fs_path = fsspec.core.url_to_fs(video_path)
```
michal-lightly: Minor: Is it possible to use named parameters?

lightly_studio/src/lightly_studio/core/add_videos.py
```diff
@@ -0,0 +1,213 @@
...
+                video_buffer,
+            )  # type: ignore
+            video_stream = video_container.streams.video[video_channel]
+
+            # Get video metadata
+            framerate = float(video_stream.average_rate) if video_stream.average_rate else 0.0
+            video_width = video_stream.width if video_stream.width else 0
+            video_height = video_stream.height if video_stream.height else 0
+            if video_stream.duration and video_stream.time_base:
+                video_duration = float(video_stream.duration * video_stream.time_base)
```
michal-lightly: We have to change the fields we store in the VideoTable and store duration_pts and time_base. The issue is that without it we would be doing floating point arithmetic which might cause issues since it is not precise.

That's something for a follow-up PR.

lightly_studio/src/lightly_studio/core/add_videos.py
```diff
@@ -0,0 +1,213 @@
...
+                        duration_s=float(video_duration),
+                        fps=framerate,
+                        file_name=Path(video_path).name,
+                    )
+                ],
+            )
+
+            if not video_sample_ids:
+                video_container.close()
+                continue
```
michal-lightly: Let's instead check that the length is 1: `if len(video_sample_ids) != 1`

If that happens it is likely an internal error. What do you think about raising here instead?

lightly_studio/src/lightly_studio/core/add_videos.py
```diff
@@ -0,0 +1,213 @@
...
+
+            video_container.close()
+
+        except Exception as e:
+            print(f"Error processing video {video_path}: {e}")
+            continue
+
+    _log_loading_results(
+        session=session, dataset_id=dataset_id, logging_context=video_logging_context
+    )
```
michal-lightly: In a non-batched setting, this can be done before the loop.

lightly_studio/src/lightly_studio/core/add_videos.py
```diff
@@ -0,0 +1,213 @@
...
+            video_stream = video_container.streams.video[video_channel]
+
+            # Get video metadata
+            framerate = float(video_stream.average_rate) if video_stream.average_rate else 0.0
+            video_width = video_stream.width if video_stream.width else 0
+            video_height = video_stream.height if video_stream.height else 0
+            if video_stream.duration and video_stream.time_base:
+                video_duration = float(video_stream.duration * video_stream.time_base)
+            else:
+                video_duration = 0.0
```
JonasWurst: I don't know if 0.0 is a good default, as we could have valid videos if 0.0 secs well. (one frame)

Should we keep it as None?
Should we look into alternative ways, (e.g. get it with last and first frame?)

lightly_studio/src/lightly_studio/core/add_videos.py
```diff
@@ -0,0 +1,213 @@
...
+    )
+    return created_sample_ids
+
+
+def _create_video_frame_samples(  # noqa: PLR0913
+    session: Session,
+    dataset_id: UUID,
+    video_sample_id: UUID,
+    video_container: av.container.InputContainer,
+    video_stream: av.video.stream.VideoStream,
```
michal-lightly: Minor: Our convention is to import classes directly:
```
from av.container import InputContainer
from av.video.stream import VideoStream
```

lightly_studio/src/lightly_studio/core/add_videos.py
```diff
@@ -0,0 +1,213 @@
...
+                        height=video_height,
+                        duration_s=float(video_duration),
+                        fps=framerate,
+                        file_name=Path(video_path).name,
+                    )
+                ],
+            )
+
+            if not video_sample_ids:
+                video_container.close()
```
JonasWurst, line 132: when will this happen?

lightly_studio/src/lightly_studio/core/add_videos.py
```diff
@@ -0,0 +1,213 @@
...
+                continue
+
+            video_sample_id = video_sample_ids[0]
```
JonasWurst:
```suggestion
```

lightly_studio/src/lightly_studio/core/add_videos.py
```diff
@@ -0,0 +1,213 @@
...
+
+            # Create video frame samples by parsing all frames
+            frame_sample_ids = _create_video_frame_samples(
+                session=session,
+                dataset_id=video_frames_dataset_id,
+                video_sample_id=video_sample_id,
```
JonasWurst:
```suggestion
                video_sample_id=video_sample_ids[0],
```

lightly_studio/src/lightly_studio/core/add_videos.py
```diff
@@ -0,0 +1,213 @@
...
+                video_container=video_container,
+                video_stream=video_stream,
+                fps=fps,
+            )
+            created_sample_ids.extend(frame_sample_ids)
```
JonasWurst: should we also keep track of the videos created? not only the frames

lightly_studio/src/lightly_studio/core/add_videos.py
```diff
@@ -0,0 +1,213 @@
...
+
+    # Get time base for converting PTS to seconds
+    time_base = video_stream.time_base if video_stream.time_base else None
+
+    # Decode all frames
+    for decoded_index, frame in enumerate(video_container.decode(video_stream)):
+        # Get the presentation timestamp in seconds from the frame
+        # Convert frame.pts from time base units to seconds
+        if frame.pts is not None and time_base is not None:
+            frame_pts_seconds = float(frame.pts * time_base)
```
michal-lightly: Minor:
```suggestion
            frame_timestamp_s = float(frame.pts * time_base)
```

lightly_studio/src/lightly_studio/core/add_videos.py
```diff
@@ -0,0 +1,213 @@
...
+    fps: float | None = None,
+) -> list[UUID]:
+    """Create video frame samples for a video by parsing all frames.
+
+    Args:
+        session: The database session.
+        dataset_id: The ID of the dataset to load video frames into.
+        video_sample_id: The ID of the video sample to create frames for.
+        video_container: The PyAV container with the opened video.
+        video_stream: The video stream from the container.
```
JonasWurst: do we need both? isn't the video channel + container all we need?

lightly_studio/src/lightly_studio/core/dataset.py
```diff
@@ -256,6 +256,42 @@ def __getitem__(self, key: _SliceType) -> DatasetQuery:
...
+        path: PathLike,
+        allowed_extensions: Iterable[str] | None = None,
+        fps: float | None = None,
+    ) -> None:
+        """Adding video frames from the specified path to the dataset.
+
+        Args:
+            path: Path to the folder containing the videos to add.
+            allowed_extensions: An iterable container of allowed video file
+                extensions.
```
michal-lightly: There should be a default for allowed extensions

lightly_studio/src/lightly_studio/models/video.py
```diff
@@ -84,7 +84,8 @@ class VideoFrameBase(SQLModel):
     """The frame number of the video frame."""
     frame_number: int
 
-    """The timestamp of the video frame in seconds."""
+    """Presentation timestamp in the video's time_base and in seconds."""
+    frame_timestamp_pts: int
```
michal-lightly, line 88: Sidenote: "pts" means "presentation timestamp" so the variable name is "frame timestamp presentation timestamp". I don't mind though.

lightly_studio/tests/core/test_add_videos.py
```diff
@@ -0,0 +1,141 @@
...
+    Returns:
+        The path to the created video file.
+    """
+    output_path.parent.mkdir(parents=True, exist_ok=True)
+
+    # Open output container
+    output_container = av.open(str(output_path), mode="w")  # type: ignore[attr-defined]
+
+    # Add video stream
+    stream = output_container.add_stream("libx264", rate=fps)
```
michal-lightly, line 167: Optional: You can help mypy by defining the interface you expect:
```py
class PyAVStream(Protocol):
    width: int
    # etc...

    def encode(av_frame: <???>) -> <???>:
        ...
```

And then
```py
stream: PyAVStream = output_container.add_stream("libx264", rate=fps)
```

lightly_studio/src/lightly_studio/core/add_videos.py
```diff
@@ -0,0 +1,213 @@
...
+            frame_pts_seconds = float(frame.pts * time_base)
+        else:
+            # Fallback to frame.time if pts or time_base is not available
+            frame_pts_seconds = frame.time if frame.time is not None else -1.0
+
+        # Apply FPS filtering if specified
+        if fps is not None and fps > 0:
+            # Only include frames that are at least min_time_interval apart
+            if frame_pts_seconds - last_timestamp < min_time_interval:
+                continue
```
JonasWurst: I don't think that this is very robust. If even the slightest jitter, that you have often, this will already fail.

lightly_studio/src/lightly_studio/core/add_videos.py
```diff
@@ -0,0 +1,213 @@
...
+) -> list[UUID]:
+    """Create video frame samples for a video by parsing all frames.
+
+    Args:
+        session: The database session.
+        dataset_id: The ID of the dataset to load video frames into.
+        video_sample_id: The ID of the video sample to create frames for.
+        video_container: The PyAV container with the opened video.
+        video_stream: The video stream from the container.
+        fps: Optional FPS value to control frame extraction. If provided, only frames
```
JonasWurst: Major: lets drop the complete fps def logic and just extract the frames as they are. There are too many special cases we would need to consider --> separate topic and also not urgent

after seeing the logic, I'm not sure this is really a good idea. For now I would drop all the fps definition.

Ideally, we would look into some more sophisticated logic/handling (e.g. interpolation, resampling)

also it does not provide higher frame-rate.

lightly_studio/src/lightly_studio/core/add_videos.py
```diff
@@ -0,0 +1,213 @@
...
+            print(f"Error processing video {video_path}: {e}")
+            continue
+
+    _log_loading_results(
+        session=session, dataset_id=dataset_id, logging_context=video_logging_context
+    )
+    return created_sample_ids
+
+
+def _create_video_frame_samples(  # noqa: PLR0913
```
michal-lightly: Major: This function contains non-trivial logic and should have its own tests.

lightly_studio/src/lightly_studio/core/add_videos.py
```diff
@@ -0,0 +1,213 @@
...
+            if frame_pts_seconds - last_timestamp < min_time_interval:
+                continue
+            last_timestamp = frame_pts_seconds
+
+        sample = VideoFrameCreate(
+            frame_number=decoded_index,
+            frame_timestamp_s=frame_pts_seconds,
+            frame_timestamp_pts=frame.pts if frame.pts is not None else -1,
+            video_sample_id=video_sample_id,
+            parent_sample_id=video_sample_id,
```
JonasWurst: why do we need both?

JonasWurst (changes requested):
See the comments.

I did not review the tests

lightly_studio/src/lightly_studio/core/add_videos.py
horatiualmasan: n_samples_to_be_inserted is set to the total input number, and when logging is done it will show  how many of the provided inputs were added. It is the same as we do for images.

lightly_studio/src/lightly_studio/core/add_videos.py
horatiualmasan: Yes, updated to stream the content and not load all at once.

lightly_studio/src/lightly_studio/core/add_videos.py
horatiualmasan: av.container.open can return `InputContainer | OutputContainer`, I added the type so mypy won't complain (I see that it is not needed anymore so I will remove it). Updated the comment.

lightly_studio/src/lightly_studio/core/add_videos.py
horatiualmasan: Updated to pass channel  instead of video_stream

lightly_studio/src/lightly_studio/core/add_videos.py
horatiualmasan: Leftover from a merge, removed video_sample_id

lightly_studio/src/lightly_studio/core/add_videos.py
horatiualmasan: If there is an issue when adding to the database

lightly_studio/src/lightly_studio/core/add_videos.py
horatiualmasan: Updated to None if it is not available in the metadata. I would look into alternatives if this comes up at some point.

lightly_studio/src/lightly_studio/core/add_videos.py
horatiualmasan: I would leave it like this for now, can do it later.

lightly_studio/src/lightly_studio/core/add_samples.py
```diff
@@ -131,7 +117,7 @@ def load_into_dataset_from_paths(
         created_sample_ids.extend(created_path_to_id.values())
         logging_context.update_example_paths(paths_not_inserted)
 
-    _log_loading_results(session=session, dataset_id=dataset_id, logging_context=logging_context)
+    log_loading_results(session=session, dataset_id=dataset_id, logging_context=logging_context)
```
michal-lightly, line 123: Minor: Our convention is to explicitly call functions from the containing module (classes import directly, for function only the containing module):
```
logging.log_loading_results(...)
```
Also `logging` is not the best name as there exists a standard `logging` module we want to eventually use.

The change can be in a separate PR.

lightly_studio/src/lightly_studio/core/add_videos.py
```diff
@@ -0,0 +1,211 @@
...
+    session: Session,
+    dataset_id: UUID,
+    video_paths: Iterable[str],
+    video_channel: int = DEFAULT_VIDEO_CHANNEL,
+) -> tuple[list[UUID], list[UUID]]:
+    """Load video frames from file paths into the dataset using PyAV.
+
+    Args:
+        session: The database session.
+        dataset_id: The ID of the dataset to load video frames into.
```
michal-lightly: Minor: It is not clear to me from the description if we pass video dataset id or video frame dataset id here.

lightly_studio/src/lightly_studio/core/add_videos.py
```diff
@@ -0,0 +1,211 @@
...
+    ):
+        try:
+            # Open video and extract metadata
+            fs, fs_path = fsspec.core.url_to_fs(url=video_path)
+            video_file = fs.open(path=fs_path, mode="rb")
+            try:
+                # Open video container for reading (returns InputContainer)
+                video_container = container.open(
+                    file=video_file,
+                )
```
michal-lightly: Optional:
```suggestion
                video_container = container.open(file=video_file)
```

lightly_studio/src/lightly_studio/core/add_videos.py
```diff
@@ -0,0 +1,211 @@
...
+                video_stream = video_container.streams.video[video_channel]
+
+                # Get video metadata
+                framerate = float(video_stream.average_rate) if video_stream.average_rate else 0.0
+                video_width = video_stream.width if video_stream.width else 0
+                video_height = video_stream.height if video_stream.height else 0
```
michal-lightly: Very optional: Also this should work, up to your taste: `video_stream.height or 0`

lightly_studio/src/lightly_studio/core/add_videos.py
```diff
@@ -0,0 +1,211 @@
...
+                # Ensure file is closed even if container operations fail
+                video_file.close()
+
+        except (FileNotFoundError, OSError, IndexError, av.AVError) as e:
+            print(f"Error processing video {video_path}: {e}")
+            continue
+
+    log_loading_results(
+        session=session, dataset_id=dataset_id, logging_context=video_logging_context
+    )
```
michal-lightly, line 143: I still think we can move this before the for loop. Isn't that the case? Or is the message such that it would look strange before loading?

lightly_studio/src/lightly_studio/core/add_videos.py
```diff
@@ -0,0 +1,211 @@
...
+        video_sample_id: The ID of the video sample to create frames for.
+        video_container: The PyAV container with the opened video.
+        video_channel: The video channel from which frames are loaded.
+
+    Returns:
+        A list of UUIDs of the created video frame samples.
+    """
+    created_sample_ids: list[UUID] = []
+    samples_to_create: list[VideoFrameCreate] = []
+    video_stream = video_container.streams.video[video_channel]
```
michal-lightly, line 168: Here we are creating a second video stream from the `video_container`. Is there a performance penalty? Shouldn't we rather pass the stream we already have in the calling function? Also is this stream properly closed? (I see we have `video_file.close()` in the `finally` block.)

lightly_studio/src/lightly_studio/core/dataset.py
```diff
@@ -256,6 +258,37 @@ def __getitem__(self, key: _SliceType) -> DatasetQuery:
...
+        self,
+        path: PathLike,
+        allowed_extensions: Iterable[str] | None = None,
+    ) -> None:
+        """Adding video frames from the specified path to the dataset.
+
+        Args:
+            path: Path to the folder containing the videos to add.
+            allowed_extensions: An iterable container of allowed video file
+                extensions. If None, uses default VIDEO_EXTENSIONS.
```
```suggestion
                extensions in lowercase, including the leading dot. If None, uses default VIDEO_EXTENSIONS.
```

lightly_studio/src/lightly_studio/models/video.py
```diff
@@ -30,7 +30,7 @@ class VideoBase(SQLModel):
     height: int
 
     """The duration of the video in seconds."""
-    duration_s: float
+    duration_s: float = Field(default=None)
```
michal-lightly: How can the default be None if the field is float (i.e. non-optional)?

lightly_studio/tests/core/test_add_videos.py
horatiualmasan: I would not add it now

lightly_studio/src/lightly_studio/core/dataset.py
horatiualmasan: - add_samples to add_iamges it is a different pr

- updated the docstring

lightly_studio/src/lightly_studio/core/add_videos.py
horatiualmasan: I would have it as the final message after the videos were processed and all the frames are created.

lightly_studio/src/lightly_studio/core/add_videos.py
horatiualmasan: The streams are created when we open the container, accessing them should not be a big time penalty. I assume they are also properly closed when we do video_container.close()

lightly_studio/src/lightly_studio/models/video.py
horatiualmasan: made optional and added a video with None duration to the tests to the tests

lightly_studio/src/lightly_studio/core/add_videos.py
michal-lightly: What is the reason not to pass the video_stream from the caller though?

horatiualmasan: Jonas mentions in a comment we could use video channel + container as parameters.

PR #91, opened 2025-11-05
Prepare the header for video support

LeonardoRosaa:
## What has changed and why?

This small change adds video hierarchy support in the header. We’ll make it dynamic in the near future.

## How has it been tested?

Screenshot 2025-11-05 at 11 13 36

- [ ] Yes


lightly_studio_view/src/lib/components/NavigationMenu/NavigationMenu.svelte
```diff
@@ -4,16 +4,19 @@
...
+    import { Image, ComponentIcon, WholeWord, Video, Frame } from '@lucide/svelte';
+    import type { SampleType } from '$lib/api/lightly_studio_local';
 
     const {
-        datasetId
+        datasetId,
+        sampleType
     }: {
         datasetId: string;
+        sampleType?: SampleType;
```
michal-lightly, line 15: Why is sampleType optional?

lightly_studio_view/src/lib/hooks/useDataset/useDataset.ts
```diff
@@ -0,0 +1,19 @@
+import { readDataset, SampleType } from '$lib/api/lightly_studio_local';
+
+export const useDataset = (datasetId: string) => {
+    const get_details = async (): Promise<{ sampleType: SampleType } | null> => {
+        const { data } = await readDataset({
+            path: { dataset_id: datasetId }
+        });
+
+        if (!data) return null;
```
michal-lightly, line 9: @mihnea-necsulescu Is it possible to handle the error such that we don't propagate null?

lightly_studio_view/src/lib/hooks/useDataset/useDataset.ts
```diff
@@ -0,0 +1,19 @@
+import { readDataset, SampleType } from '$lib/api/lightly_studio_local';
+
+export const useDataset = (datasetId: string) => {
```
mihnea-necsulescu, line 3: I was wondering if we even need this, can't we simply call `readDataset` in the header? I don't think it makes sense to introduce this hook just to return the sample type - wdyt?

mihnea-necsulescu: we can also handle the error case there

PR #92, opened 2025-11-05
Prevent adding annotation label duplicates

JonasWurst:
## What has changed and why?

Users reported a bug that happened when adding samples with `add_samples_from_yolo`. This happened only when using multiple splits.

Reason: for each split, the annotation labels are created, even though they already exist. Later in the UI only the unique names only one occurence is used, hence the filtering was broken.

This PR fixes the underlying issue in two ways. On annotation label creation labels are reused instead of created if they exist. Furthermore, the label_name field is now unique, preventing adding duplicated labels in general.

## How has it been tested?

Unit test is implemented. Validated that unit test fails on the previous version.

- [x] Yes

### DIFF
--- a/lightly_studio/src/lightly_studio/plugins/parameter.py
+++ b/lightly_studio/src/lightly_studio/plugins/parameter.py
@@ -1,0 +1,70 @@
+"""Parameter for operators for LightlyStudio plugins."""
+
+from __future__ import annotations
+
+from abc import ABC, abstractmethod
+from dataclasses import dataclass
+from typing import Any, Generic, TypeVar, cast
+
+T = TypeVar("T")
+
+
+@dataclass
+class BaseParameter(ABC):
+    """Base parameter definition shared across operator parameters."""
+
+    name: str
+    description: str = ""
+    default: Any = None
+    required: bool = True
+    param_type: str | None = None
+
+    def __post_init__(self) -> None:
+        """Run value validation once the dataclass is initialized."""
+        if self.default is not None:
+            self.default = self._validate(self.default)
+
+    @abstractmethod
+    def _validate(self, value: Any) -> Any:
+        """Validate the parameter value."""
+
+
+class BuiltinParameter(BaseParameter, Generic[T]):
+    """Represents a built-in operator parameter."""
+
+    def __post_init__(self) -> None:
+        """Set up type information and validate default value."""
+        if not hasattr(self, "_parameter_type") or self._parameter_type is None:
+            raise NotImplementedError("Subclasses must define _parameter_type class attribute")
+        self._type = self._parameter_type
+        self.param_type = self._parameter_type.__name__
+        super().__post_init__()
+
+    def _validate(self, value: T) -> T:
+        if isinstance(value, self._type):
+            return cast(T, value)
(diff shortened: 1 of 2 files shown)


lightly_studio/src/lightly_studio/core/add_samples.py
```diff
@@ -385,6 +385,15 @@ def _create_label_map(
...
+            session=session, label_name=category.name
+        )
+        if existing_label is not None:
+            label_map[category.id] = existing_label.annotation_label_id
+            continue
+
+        # Create new label
         label = AnnotationLabelCreate(annotation_label_name=category.name)
         stored_label = annotation_label_resolver.create(session=session, label=label)
         label_map[category.id] = stored_label.annotation_label_id
```
michal-lightly: There's a cleaner way to write equivalent code, without the `continue` jump:
```py
        label = annotation_label_resolver.get_by_label_name(
            session=session, label_name=category.name
        )
        if label is None:
            # Create a new label
            label_create = AnnotationLabelCreate(annotation_label_name=category.name)
            label = annotation_label_resolver.create(session=session, label=label_create)

        label_map[category.id] = label.annotation_label_id
```

lightly_studio/tests/core/test_add_samples.py
```diff
@@ -199,23 +201,53 @@ def test_create_batch_samples(db_session: Session) -> None:
...
+    label_map_2 = add_samples._create_label_map(
+        session=db_session,
+        input_labels=label_input_2,
+    )
+
+    assert len(label_map) == 2  # dog and cat
+    assert len(label_map_2) == 3  # dog, cat and bird
+    assert label_map_2[0] == label_map[0]  # dog exists already
+    assert label_map_2[1] == label_map[1]  # cat exists already
+    assert label_map_2[2] != label_map[1]  # bird is new
```
michal-lightly: This check does not make sense, it tests that bird is not dog. We can e.g. do:
```py
assert label_map_2[2] not in label_map_1.values()
```

Also I suggest to add a comment explaining what we are comparing (annotation_label_id for a category id)

lightly_studio/tests/core/test_add_samples.py
```diff
@@ -199,23 +201,53 @@ def test_create_batch_samples(db_session: Session) -> None:
...
+
+
+def _get_labelformat_input(
+    filename: str = "image.jpg", category_names: list[str] | None = None
+) -> LabelformatObjectDetectionInput:
     """Creates a LabelformatObjectDetectionInput for testing.
 
     Args:
         filename: The name of the image file.
+        category_names: The names of the categories.
```
michal-lightly: Minor: Add info about the default

CHANGELOG.md
```diff
@@ -21,6 +21,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Preventing adding of duplicated annotation labels: Fixed an issue that occurred when adding samples from yolo using multiple splits.
```
michal-lightly: Minor:
```suggestion
- Prevent duplicated annotation labels: Fixed an issue that occurred when adding samples from yolo using multiple splits. 
```

PR #93, opened 2025-11-05
Remove dataset_id from ImageView

michal-lightly:
## What has changed and why?

* Remove dataset_id from ImageView
* Introduce ImageView.sample
* Fixed an issue with pydantic type propagation

Step towards removing dataset_id from ImageTable.

## How has it been tested?

Existing tests.

- [ ] Yes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Prevent duplicated annotation labels: Fixed an issue that occurred when adding samples from yolo using multiple splits. 
+
 ### Security
 
 ## \[0.4.1\] - 2025-10-27
--- a/lightly_studio/src/lightly_studio/core/add_samples.py
+++ b/lightly_studio/src/lightly_studio/core/add_samples.py
@@ -385,9 +385,16 @@ def _create_label_map(
         desc="Processing categories",
         unit=" categories",
     ):
-        label = AnnotationLabelCreate(annotation_label_name=category.name)
-        stored_label = annotation_label_resolver.create(session=session, label=label)
-        label_map[category.id] = stored_label.annotation_label_id
+        # Use label if already exists
+        label = annotation_label_resolver.get_by_label_name(
+            session=session, label_name=category.name
+        )
+        if label is None:
+            # Create new label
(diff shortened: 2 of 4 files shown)


lightly_studio/src/lightly_studio/api/routes/api/sample.py
```diff
@@ -75,13 +75,22 @@ def read_samples(
                 file_name=image.file_name,
                 file_path_abs=image.file_path_abs,
                 sample_id=image.sample_id,
-                dataset_id=image.dataset_id,
                 annotations=image.annotations,
                 captions=image.sample.captions,
-                tags=image.sample.tags,
```
michal-lightly, line 81: The change was needed here because type info is more properly propagated at runtime.

lightly_studio/src/lightly_studio/models/image.py
```diff
@@ -19,10 +20,8 @@
     from lightly_studio.models.metadata import (
         SampleMetadataView,
     )
-    from lightly_studio.models.sample import SampleTable
```
michal-lightly, line 22: The import was gated by TYPE_CHECKING. This prevented the type information to be propagated to schema.ts and the type was not checked at runtime.

lightly_studio/src/lightly_studio/models/image.py
```diff
@@ -71,10 +70,10 @@ class ImageTable(ImageBase, table=True):
 ]
 
 
-class ImageView(SQLModel):
```
michal-lightly, line 74: We do not need SQLModel here, the class is not used with the db.

lightly_studio_view/src/lib/schema.d.ts
```diff
@@ -1978,6 +1974,44 @@ export interface components {
...
+            /**
+             * Updated At
+             * Format: date-time
+             */
+            updated_at: string;
+            /**
+             * Tags
+             * @default []
+             */
+            tags: unknown[];
```
LeonardoRosaa, line 2006: Do you know why this has the `unknown` type? I'd suggest avoiding it.

horatiualmasan (approved):
LGTM

lightly_studio_view/src/lib/schema.d.ts
michal-lightly: This is a general issue in our codebase that is out of scope to solve in this PR. It occurs because we gate some imports behind `if TYPE_CHECKING`.

PR #94, opened 2025-11-06
Introduce BaseOperator

JonasWurst:
## What has changed and why?

This PR adds the abstract base operator class. The registry will be added in a follow up pr.

## How has it been tested?

For now, no tests as it only holds the abstract class. We can test with a concrete implementation at a later point.

- [ ] Yes

### DIFF
--- a/lightly_studio/src/lightly_studio/api/routes/api/sample.py
+++ b/lightly_studio/src/lightly_studio/api/routes/api/sample.py
@@ -75,13 +75,22 @@ def read_samples(
                 file_name=image.file_name,
                 file_path_abs=image.file_path_abs,
                 sample_id=image.sample_id,
-                dataset_id=image.dataset_id,
                 annotations=image.annotations,
                 captions=image.sample.captions,
-                tags=image.sample.tags,
+                tags=[
+                    ImageView.ImageViewTag(
+                        tag_id=tag.tag_id,
+                        name=tag.name,
+                        kind=tag.kind,
+                        created_at=tag.created_at,
+                        updated_at=tag.updated_at,
+                    )
+                    for tag in image.sample.tags
+                ],
                 metadata_dict=image.sample.metadata_dict,
                 width=image.width,
                 height=image.height,
+                sample=image.sample,
             )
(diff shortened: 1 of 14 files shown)


lightly_studio/src/lightly_studio/plugins/base_operator.py
```diff
@@ -0,0 +1,60 @@
...
+
+    @property
+    @abstractmethod
+    def description(self) -> str:
+        """Return the description of the operator."""
+
+    @property
+    @abstractmethod
+    def parameters(self) -> list[BaseParameter]:
+        """Return the list of parameters this operator expects."""
```
MalteEbner, line 39: Did you also consider using a dataclass instead of these abstract properties?

This enforces the children to have a class-level instead of instance-level name and description, which is not necessarily needed.
Given that the children of the BaseOperator need to be defined by external users, I am also wondering whether the approach of using the harder to understand concept of abstract properties is the best one.

If it was a dataclass, users could create a class instance without needing to write all the code on class level.

```python

class MyExportOperator(BaseOperator)

    def execute(...) -> ...
         ...

export_operator_instance = MyExportOperator(
    name=export,
    description="",
    parameters=[StringParameter(...)]
)
```

lightly_studio/src/lightly_studio/plugins/base_operator.py
```diff
@@ -0,0 +1,60 @@
...
+
+    @abstractmethod
+    def execute(
```
MalteEbner, line 42: Good choice to make this an abstract method

lightly_studio/src/lightly_studio/plugins/base_operator.py
JonasWurst: I think at least the parameters i would keep as they are, since it is direclty linked to the execute, and what can be used there. I feel splitting it between class definiton and instansation increases the chance that things break.

For the name and description, we could do that, but I see only little benefit since we already need to do it for execute and parameters anyway.

Hence I would stick to this solution, as It keeps things in one place.

PR #95, opened 2025-11-06
Use SampleTable.dataset_id instead of ImageTable.dataset_id in queries

michal-lightly:
## What has changed and why?

Preparation to remove dataset_id from ImageTable.

* Use SampleTable.dataset_id instead of ImageTable.dataset_id in queries
* Fetch dataset_id from SampleTable in core/sample.py

## How has it been tested?

CI

- [ ] Yes


lightly_studio/src/lightly_studio/models/sample.py
```diff
@@ -39,7 +39,7 @@ class SampleBase(SQLModel):
     """Base class for the Sample model."""
 
     """The dataset ID to which the sample belongs."""
-    dataset_id: Optional[UUID] = Field(default=None, foreign_key="dataset.dataset_id")
```
michal-lightly, line 42: Removed optional due to mypy issues. It should be always set.

lightly_studio/tests/core/test_sample.py
```diff
@@ -58,10 +57,6 @@ def test_basic_fields_set(self, mocker: MockerFixture, test_db: Session) -> None
         assert new_image_table.file_name == "sample1.png"
         assert new_image_table.width == 1000
 
-        # Can't set a foreign key to a non-existent id.
-        with pytest.raises(IntegrityError):
-            sample.dataset_id = UUID(int=123456)
```
michal-lightly, line 63: Made dataset_id a non-settable property. Attempts to set the field are caught by mypy.

PR #97, opened 2025-11-06
Operator System: registry

JonasWurst:
## What has changed and why?

This PR adds the registry for the operator system. In a next PR the api endpoints will be added.

## How has it been tested?

Unit test for registry and a test operator are part of the PR.

- [ ] Yes

### DIFF
--- a/lightly_studio/src/lightly_studio/core/dataset.py
+++ b/lightly_studio/src/lightly_studio/core/dataset.py
@@ -35,6 +35,7 @@ from lightly_studio.models.annotation.annotation_base import (
 )
 from lightly_studio.models.dataset import DatasetCreate, DatasetTable, SampleType
 from lightly_studio.models.image import ImageTable
+from lightly_studio.models.sample import SampleTable
 from lightly_studio.resolvers import (
     dataset_resolver,
     embedding_model_resolver,
@@ -162,7 +163,9 @@ class Dataset:
     def __iter__(self) -> Iterator[Sample]:
         """Iterate over samples in the dataset."""
         for sample in self.session.exec(
-            select(ImageTable).where(ImageTable.dataset_id == self.dataset_id)
+            select(ImageTable)
+            .join(ImageTable.sample)
+            .where(SampleTable.dataset_id == self.dataset_id)
         ):
             yield Sample(inner=sample)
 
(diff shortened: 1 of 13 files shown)


lightly_studio/src/lightly_studio/plugins/operator_registry.py
```diff
@@ -0,0 +1,50 @@
...
+        self._operators[operator_id] = operator
+
+    def get_all_metadata(self) -> list[RegisteredOperatorMetadata]:
+        """Get all registered operators with their names and parameters."""
+        return [
+            RegisteredOperatorMetadata(
+                operator_id=operator_id,
+                name=operator.name,
+                parameters=[asdict(param) for param in operator.parameters],
+            )
```
MalteEbner, line 37: Why do we need the `get_all_metadata` method? And why do we need to pass the `parameters`?

For the GUI components:
- The dropdown only needs the operator id and name (and perhaps description), but not the parameters
- The actual dialog needs the paramaters of a single operator and can use `get_by_id` for it.


Suggestion: Remove this method from this PR. Implement a similar method once we implement the endpoint in  [LIG-7909: Operator Plugins v1: Implement endpoints](https://linear.app/lightly/issue/LIG-7909/operator-plugins-v1-implement-endpoints) and know what we actually need in the frontend.

lightly_studio/tests/plugins/test_operator_registry.py
```diff
@@ -0,0 +1,106 @@
...
+
+from lightly_studio.plugins.base_operator import BaseOperator, OperatorResult
+from lightly_studio.plugins.operator_registry import OperatorRegistry
+from lightly_studio.plugins.parameter import BaseParameter, BoolParameter, StringParameter
+from tests.helpers_resolvers import create_dataset
+
+
+class TestOperator(BaseOperator):
+    @property
+    def name(self) -> str:
```
MalteEbner: Optional: This might be reused by other operator tests. We could already move it to `operator_helpers.py`

lightly_studio/tests/plugins/test_operator_registry.py
```diff
@@ -0,0 +1,106 @@
...
+        """Return the operator name."""
+        return "test operator"
+
+    @property
+    def description(self) -> str:
```
MalteEbner: Optional: make this a dataclass with default values for all properties instead. This is less code, easier to understand and allows testing with two operators with two different names easily.

lightly_studio/tests/plugins/test_operator_registry.py
JonasWurst: I would only move it once we reuse

lightly_studio/src/lightly_studio/plugins/operator_registry.py
JonasWurst: I dropped the parameters, but kept the method itself as we will need it for sure for the dropdopwn, also it comes in handy for the tests.

MalteEbner (approved):
LGTM^2

PR #100, opened 2025-11-07
add to docs about using cloud storage and current support limitations

masakljun:
## What has changed and why?

Added a new Using Cloud Storage section to the docs explaining how to enable loading images from S3, GCS, and other cloud providers.
It includes installation instructions for the lightly-studio[cloud-storage] extra and clarifies that annotations must still be local.

- [ ] Yes

### DIFF
--- a/lightly_studio/src/lightly_studio/api/routes/api/annotations/create_annotation.py
+++ b/lightly_studio/src/lightly_studio/api/routes/api/annotations/create_annotation.py
@@ -26,7 +26,7 @@ class AnnotationCreateInput(BaseModel):
 
     annotation_label_id: UUID
     annotation_type: AnnotationType
-    sample_id: UUID
+    parent_sample_id: UUID
     x: int | None = None
     y: int | None = None
     width: int | None = None
--- a/lightly_studio/src/lightly_studio/core/add_samples.py
+++ b/lightly_studio/src/lightly_studio/core/add_samples.py
@@ -411,7 +411,7 @@ def _process_object_detection_annotations(
         new_annotations.append(
             AnnotationCreate(
                 dataset_id=context.dataset_id,
-                sample_id=context.sample_id,
+                parent_sample_id=context.sample_id,
                 annotation_label_id=context.label_map[obj.category.id],
                 annotation_type="object_detection",
                 x=int(x),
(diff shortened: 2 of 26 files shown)


README.md
```diff
@@ -124,6 +124,17 @@ Run the script with `python example_coco_captions.py`. Now you can inspect sampl
 
 LightlyStudio has a powerful Python interface. You can not only index datasets but also query and manipulate them using code.
 
+### ☁️ Using Cloud Storage
+To load images directly from a cloud storage provider (like AWS S3, GCS, etc.), you must first install the required dependencies:
+
+```py
+pip install lightly-studio[cloud-storage]
+```
```
MalteEbner, line 132: This is not 100% true. 

In general: We are using fsspec which supports many different file systems. Best link the full list of supported file systems: 
- built-in: https://filesystem-spec.readthedocs.io/en/latest/api.html#built-in-implementations, which includes loading from URLs directly.
- external: https://filesystem-spec.readthedocs.io/en/latest/api.html#other-known-implementations All the external file systems require installing another pip package.

- Installing `pip install lightly-studio[cloud-storage]` will only install `s3fs` (for AWS S3), `gcfs` (google cloud storage) and `adlfs` (azure data lake). See https://github.com/lightly-ai/lightly-studio/blob/9b1ee84ef03017c3ffd650b8a5d5f0c66a4d86f0/lightly_studio/pyproject.toml#L60-L64
I would not mention it but rather tell users to head to https://filesystem-spec.readthedocs.io/en/latest/api.html#other-known-implementations, find the file system they need and install it via pip. E.g. `pip install s3fs`, `pip install gcfs` and `pip install adlfs` for AWS S3, gcloud and Azure respectively.

masakljun: Thanks for the feedback.

I agree that just mentioning [cloud-storage] doesn't tell the full story especially since `fsspec` already supports some file systems and has many more beyond the ones in our cloud-storage list.

My main goal with this section was to fix the issue from the original ticket, where a new user with an s3 bucket didn't realize they needed to install s3fs. If we only point to the fsspec docs, I'm slightly worried we will still leave some (less experienced, e.g., like me) users stuck, since they might not even know what fsspec is or which library applies to their case.

What do you think about keeping a combined approach , something like this? (see below)

Keep `pip install lightly-studio[cloud-storage]` as the quick, recommended command for the common providers (S3, GCS, Azure).

Add a short line right after explaining what that installs (s3fs, gcsfs, adlfs).

Then mention that we use fsspec and link to their docs for anyone who needs another file system.

So that way new users get a simple command while more advanced users still have the full technical context if they need it.

PR #108, opened 2025-11-10
Add operator endpoints

JonasWurst:
## What has changed and why?

Adding the api endpoints for the operator support. 3 methods are implemented:
- get_operators: That shall be used to construct the dropdown menu
- get_operator_parameters: that shall be used to constructe the operator dialog
- execute_operator: that is used to run an operator

## How has it been tested?

tests are implemented.

- [ ] Yes

### DIFF
--- a/lightly_studio_view/src/lib/schema.d.ts
+++ b/lightly_studio_view/src/lib/schema.d.ts
@@ -1049,7 +1049,6 @@ export interface paths {
         trace?: never;
     };
     "/api/operators": {
-    "/api/datasets/{dataset_id}/video/": {
         parameters: {
             query?: never;
             header?: never;
@@ -1112,6 +1111,48 @@ export interface paths {
          *         The execution result.
          */
         post: operations["execute_operator"];
+        delete?: never;
+        options?: never;
+        head?: never;
+        patch?: never;
+        trace?: never;
+    };
+    "/api/datasets/{video_frame_dataset_id}/frame/": {
+        parameters: {
+            query?: never;
+            header?: never;
+            path?: never;
+            cookie?: never;
+        };
+        /**
+         * Get All Frames
+         * @description Retrieve a list of all frames for a given dataset ID with pagination.
+         *
+         *     Args:
+         *         session: The database session.
+         *         video_frame_dataset_id: The ID of the dataset to retrieve frames for.
+         *         pagination: Pagination parameters including offset and limit.
+         *
+         *     Return:
+         *         A list of frames along with the total count.
+         */
+        get: operations["get_all_frames"];
+        put?: never;
+        post?: never;
+        delete?: never;
+        options?: never;
+        head?: never;
+        patch?: never;
+        trace?: never;
+    };
+    "/api/datasets/{dataset_id}/video/": {
+        parameters: {
+            query?: never;
+            header?: never;
+            path?: never;
+            cookie?: never;
+        };
+        /**
          * Get All Videos
          * @description Retrieve a list of all videos for a given dataset ID with pagination.
          *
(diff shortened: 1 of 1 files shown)


lightly_studio/src/lightly_studio/api/routes/api/operator.py
```diff
@@ -0,0 +1,75 @@
...
+class ExecuteOperatorRequest(BaseModel):
+    """Request model for executing an operator."""
+
+    parameters: dict[str, Any]
+
+
+@operator_router.get("")
+def get_operators() -> list[dict[str, Any]]:
+    """Get all registered operators (id, name)."""
+    return [asdict(operator) for operator in operator_registry.get_all_metadata()]
```
MalteEbner: I would rather keep returning the `RegisteredOperatorMetadata` instead of converting to dict too keep this typed. Then it is also typed in the frontend.

lightly_studio/src/lightly_studio/api/routes/api/operator.py
```diff
@@ -0,0 +1,75 @@
...
+
+
+@operator_router.get("/{operator_id}/parameters")
+def get_operator_parameters(operator_id: str) -> list[dict[str, Any]]:
```
MalteEbner: Similarly, I would return a `list[BaseParameter]` instead.

lightly_studio/src/lightly_studio/api/app.py
```diff
@@ -98,6 +99,8 @@ def use_route_names_as_operation_ids(app: FastAPI) -> None:
 api_router.include_router(features.features_router)
 api_router.include_router(metadata.metadata_router)
 api_router.include_router(selection.selection_router)
+api_router.include_router(selection.selection_router)
```
MalteEbner: It is added twice ;) 
```suggestion
```

lightly_studio/tests/api/routes/api/test_operator.py
```diff
@@ -0,0 +1,167 @@
...
+
+
+def test_get_operator_parameters__multiple_parameters(
+    test_client: TestClient,
+    isolated_operator_registry: OperatorRegistry,
+) -> None:
+    # Add an operator with multiple parameters:
+    # - BoolParameter(name="test flag", required=True)
+    # - StringParameter(name="test str", required=True)
+    isolated_operator_registry.register(TestOperator(name="multi"))
```
MalteEbner, line 94: This makes the test hard to verify, as it relies on values defined in another file. Furthermore, it will cause the test to break if any parameter in the `TestOperator` changes or new ones are added.

According to our testing code style tech talks / notion page, this test should rather define the parameters itself, e.g.
```python
operator = TestOperator(
    name="multi", 
    parameters=[
         BoolParameter(name="test flag", required=True),
         StringParameter(name="test str", required=True),
    ]
 )
 isolated_operator_registry.register(operator)
```

This would only require making the `TestOperator.parameters` a normal dataclass attribute.

MalteEbner (approved):
General design and functions look good to me. I would just make sure that the endpoint parameters are typed as much as possible.

lightly_studio/tests/api/routes/api/test_operator.py
JonasWurst: I decided to keep the TestOperator in each file now. This gives us duplicated code. However, if we make it a data-class, this is also confusing since the parameters will be defined without seeing the execute, which needs to be tightly coupled. 

Hence my decision is to introduce a bit of redundancy, which is ok for damp tests.

lightly_studio/tests/plugins/test_operator_registry.py
JonasWurst, line 1: moved the class to the bottom and added a missing assert

MalteEbner, line 63: Optional: I still believe that making the parameters a normal dataclasss attribute of the `TestOperator` would be better, it would make the tests more damp.

Optional: The execute function could also be a normal dataclass attribute of type `Callable[[Self, Session, UUID, dict[str, Any]], OperatorResult]`

By making the dataclass frozen, we could also solve the problem of having lists as default values being shared across instances.

MalteEbner: We discussed in a call that my suggestion leads to mypy errors. Thus we will merge it as it is now.

We will have a look at why it causes mypy errors separately.

JonasWurst: as discussed. Lets's keep it like this for now.

PR #109, opened 2025-11-10
Add a note about Linux and Firefox

lukas-lightly:
## What has changed and why?
We know that Chrome & NVIDIA on Linux don't work, we should recommend Firefox to these users.

## How has it been tested?
Rendered locally.

- [ ] Yes

### DIFF
--- a/lightly_studio/docs/docs/index.md
+++ b/lightly_studio/docs/docs/index.md
@@ -214,6 +214,9 @@ git clone https://github.com/lightly-ai/dataset_examples dataset_examples
 4.  This server reads from `lightly_studio.db` and serves data to the **UI Application** running in your browser (`http://localhost:8001`).
 5.  Images are streamed directly from your disk for display in the UI.
 
+!!! note "For Linux Users"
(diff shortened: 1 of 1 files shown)


lightly_studio/docs/docs/index.md
```diff
@@ -214,6 +214,9 @@ git clone https://github.com/lightly-ai/dataset_examples dataset_examples
 4.  This server reads from `lightly_studio.db` and serves data to the **UI Application** running in your browser (`http://localhost:8001`).
 5.  Images are streamed directly from your disk for display in the UI.
 
+!!! note "For Linux Users"
+    We recommend using Firefox for the best experience with embedding plots, as other browsers might not render them correctly.
+
```
michal-lightly, line 219: Optional: We can add info that this is relevant especially for nVidia gpus if we know it is specific.

lukas-lightly: Let's not narrow it too much now, so far we've tested 5 setups and only NVIDIA & AMD GPUs. I think we need more HW tests, e.g. with Intel cards, different driver versions, etc.

PR #111, opened 2025-11-10
Remove dataset_id from AnnotationCreate

horatiualmasan:
## What has changed and why?

First part of removing dataset_id from Annotation tables

## How has it been tested?

Existing tests work

- [ ] Yes

### DIFF
--- a/lightly_studio/src/lightly_studio/api/app.py
+++ b/lightly_studio/src/lightly_studio/api/app.py
@@ -23,11 +23,13 @@ from lightly_studio.api.routes.api import (
     embeddings2d,
     export,
     features,
+    frame,
     metadata,
     sample,
     selection,
     settings,
     text_embedding,
+    video,
 )
 from lightly_studio.api.routes.api.exceptions import (
     register_exception_handlers,
@@ -98,7 +100,8 @@ api_router.include_router(embeddings2d.embeddings2d_router)
 api_router.include_router(features.features_router)
 api_router.include_router(metadata.metadata_router)
 api_router.include_router(selection.selection_router)
-
+api_router.include_router(frame.frame_router)
+api_router.include_router(video.video_router)
 
 app.include_router(api_router)
 # images serving
--- a/lightly_studio/src/lightly_studio/api/routes/api/frame.py
+++ b/lightly_studio/src/lightly_studio/api/routes/api/frame.py
@@ -1,0 +1,41 @@
+"""API routes for dataset frames."""
+
+from __future__ import annotations
+
+from uuid import UUID
+
+from fastapi import APIRouter, Depends, Path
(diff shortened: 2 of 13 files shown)


lightly_studio/tests/api/routes/api/annotations/test_update_annotations_label.py
```diff
@@ -24,7 +24,11 @@ def test_update_annotations_labels(
     with patch(
         "lightly_studio.services.annotations_service.update_annotations"
     ) as mock_update_annotations:
-        mock_update_annotations.return_value = [annotation]
+        # Create a copy of the annotation with the updated label_id
+        updated_annotation = annotation.model_copy(
+            update={"annotation_label_id": label.annotation_label_id}
+        )
+        mock_update_annotations.return_value = [updated_annotation]
```
michal-lightly: Why is this needed?

This is by the way a legacy test that uses unittest.mock.patch. A test that manipulates a real annotation would be better (out of scope of this PR, unless you feel like fixing it).

lightly_studio/tests/resolvers/annotations/test_annotations_filters.py
```diff
@@ -48,23 +48,31 @@ def filter_test_data(
     tag1 = create_tag(session=test_db, dataset_id=dataset1.dataset_id, tag_name="tag1")
     tag2 = create_tag(session=test_db, dataset_id=dataset2.dataset_id, tag_name="tag2")
 
-    # Create annotations
-    annotation_ids = annotation_resolver.create_many(
+    # Create annotations for dataset1
+    annotation_ids = list(
```
michal-lightly: Please change the output type of `create_many` to `list[UUID]` and remove `list` here. The return values should be concrete (and the inputs general).

lightly_studio/tests/resolvers/annotations/test_annotations_filters.py
michal-lightly: This logic is now unnecessarily complicated - we just need to create two annotations and return them. We can do it this way:
```py
annotation1_id = annotation_resolver.create_many(...)[0]
annotation2_id = annotation_resolver.create_many(...)[0]
annotation1 = annotation_resolver.get_by_id(session=test_db, annotation_id=annotation1_id)
annotation2 = annotation_resolver.get_by_id(session=test_db, annotation_id=annotation2_id)
```

lightly_studio/tests/api/routes/api/annotations/test_update_annotations_label.py
horatiualmasan: The initial test was working by chance, it was mocking the result as the initial value. Removed it and just check the values of the result.

PR #118, opened 2025-11-10
Added video grid view

LeonardoRosaa:
## What has changed and why?

Added the video grid view and the endpoint for streaming video media.

## How has it been tested?

- Upload a video dataset;
- Open the GUI.

- [ ] Yes

### DIFF
--- a/lightly_studio/src/lightly_studio/api/routes/api/dataset.py
+++ b/lightly_studio/src/lightly_studio/api/routes/api/dataset.py
@@ -50,6 +50,14 @@ def read_datasets(
     return dataset_resolver.get_all(session=session, offset=paginated.offset, limit=paginated.limit)
 
 
+@dataset_router.get("/datasets/root_dataset", response_model=DatasetView)
+def read_root_dataset(
+    session: SessionDep,
+) -> DatasetTable:
+    """Retrieve the root dataset from the database."""
+    return dataset_resolver.get_root_dataset(session=session)
+
+
 @dataset_router.get("/datasets/{dataset_id}", response_model=DatasetViewWithCount)
 def read_dataset(
     session: SessionDep,
--- a/lightly_studio/tests/api/routes/api/test_dataset.py
+++ b/lightly_studio/tests/api/routes/api/test_dataset.py
@@ -68,6 +68,30 @@ def test_delete_dataset(test_client: TestClient, db_session: Session) -> None:
     assert response.status_code == HTTP_STATUS_NOT_FOUND
 
 
+def test_read_root_dataset(test_client: TestClient, db_session: Session) -> None:
+    client = test_client
+    dataset_id = create_dataset(session=db_session, dataset_name="example_dataset").dataset_id
+    create_dataset(session=db_session, dataset_name="child", parent_dataset_id=dataset_id)
+
+    response = client.get("/api/datasets/root_dataset")
+    assert response.status_code == HTTP_STATUS_OK
+
+    dataset = response.json()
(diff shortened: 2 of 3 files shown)


lightly_studio_view/src/lib/components/NavigationMenu/NavigationMenu.svelte
horatiualmasan: I would remove unsupported features from the menu and add them as we implement.

lightly_studio_view/.env
```diff
@@ -1,7 +1,7 @@
 PUBLIC_SAMPLES_URL=/images
 
 PUBLIC_LIGHTLY_STUDIO_API_URL='/'
-
+PUBLIC_VIDEOS_SAMPLES_URL=http://localhost:8001/videos/samples
```
horatiualmasan: Could you keep the same notation as we have for images? without localhost?

lightly_studio_view/src/routes/datasets/[dataset_id]/videos/+page.svelte
```diff
@@ -0,0 +1,81 @@
...
+            <ImageSizeControl />
+        </div>
+    </div>
+    <Separator class="mb-4 bg-border-hard" />
+
+    <div class="h-full w-full flex-1 overflow-hidden" bind:this={viewport} bind:clientWidth>
+        {#if $query.isPending && items.length === 0}
+            <div class="flex h-full w-full items-center justify-center">
+                <Spinner />
+                <div>Loading samples...</div>
```
Comment (horatiualmasan) [inline lightly_studio_view/src/routes/datasets/[dataset_id]/videos/+page.svelte]:
```suggestion
                <div>Loading  video samples...</div>
```

lightly_studio_view/.env
```diff
@@ -1,7 +1,7 @@
 PUBLIC_SAMPLES_URL=/images
 
 PUBLIC_LIGHTLY_STUDIO_API_URL='/'
-
+PUBLIC_VIDEOS_SAMPLES_URL=http://localhost:8001/video/media
```
horatiualmasan: Can you use similar url to images? without localhost

lightly_studio_view/src/routes/datasets/[dataset_id]/videos/+page.svelte
```diff
@@ -0,0 +1,81 @@
...
+            </div>
+        {:else if $query.isSuccess && items.length > 0}
+            <Grid
+                itemCount={items.length}
+                itemHeight={itemSize}
+                itemWidth={itemSize}
+                height={viewport?.clientHeight}
+                class="overflow-none overflow-y-auto dark:[color-scheme:dark]"
+                style="--sample-width: {videoSize}px; --sample-height: {videoSize}px;"
+                overScan={100}
```
Comment (horatiualmasan) [inline lightly_studio_view/src/routes/datasets/[dataset_id]/videos/+page.svelte]: Loading 100 videos seems a bit much, as it takes some time to load all.

horatiualmasan (commented):
I think for the streaming endpoint we need to implement HTTP Range requests. Now we always send the full content. With 100 videos loaded chrome consumes around 2.5 gb, it should not be that much.

Some functionality feedback:
When I reload the page it gets stuck, possibly we would need to stops streaming if the client disconnects.
Possibly related to this the next samples do not load if I try to scroll pass the 100 initial loaded samples.
The scroll bar does not resize correctly if I resize the window.

lightly_studio_view/src/routes/datasets/[dataset_id]/videos/+page.svelte
Comment (michal-lightly) [reply lightly_studio_view/src/routes/datasets/[dataset_id]/videos/+page.svelte]: Even better: `Loading videos...`

PR #120, opened 2025-11-11
Add caption update resolver

JonasWurst:
## What has changed and why?

To support editable captions, this pr adds the update resolver as well as the required get_by_ids resolver.

## How has it been tested?

tests are implemented

- [ ] Yes


lightly_studio/src/lightly_studio/resolvers/caption_resolver.py
```diff
@@ -78,3 +78,46 @@ def get_all(
         total_count=total_count,
         next_cursor=next_cursor,
     )
+
+
+def get_by_ids(session: Session, caption_ids: list[UUID]) -> list[CaptionTable]:
+    """Retrieve captions by IDs."""
+    results = session.exec(
+        select(CaptionTable).where(col(CaptionTable.caption_id).in_(caption_ids))
```
MalteEbner: Let's use a `Sequence[UUID]` as input type instead:
- more permissive on the input
- makes it clear by type that the `caption_ids` are only read, never written to.

See e.g. also https://github.com/lightly-ai/lightly-studio/blob/45ba5dfbc0b7aa0647f8025fed3f65ad211a8c08/lightly_studio/src/lightly_studio/resolvers/annotation_label_resolver/get_by_ids.py#L13-L22


```suggestion
def get_by_ids(session: Session, caption_ids: Sequence[UUID]) -> list[CaptionTable]:
    """Retrieve captions by IDs."""
    results = session.exec(
        select(CaptionTable).where(col(CaptionTable.caption_id).in_(set(caption_ids)))
```

lightly_studio/src/lightly_studio/resolvers/caption_resolver.py
```diff
@@ -78,3 +78,46 @@ def get_all(
...
+    ).all()
+    # Return samples in the same order as the input IDs
+    caption_map = {caption.caption_id: caption for caption in results}
+    return [caption_map[id_] for id_ in caption_ids if id_ in caption_map]
+
+
+def update_text(
```
MalteEbner, line 93: The implementation is very different from updating annotations: https://github.com/lightly-ai/lightly-studio/blob/45ba5dfbc0b7aa0647f8025fed3f65ad211a8c08/lightly_studio/src/lightly_studio/resolvers/annotation_label_resolver/update.py#L17

Can we align on one implementation style?

MalteEbner: Discussion in call: It is the same as in https://github.com/lightly-ai/lightly-studio/blob/45ba5dfbc0b7aa0647f8025fed3f65ad211a8c08/lightly_studio/src/lightly_studio/resolvers/annotation_resolver/update_bounding_box.py#L24 and thus fine.

PR #125, opened 2025-11-11
Add "input" class balancing

lukas-lightly:
## What has changed and why?
Input class balancing is the last missing class balancing type.

## How has it been tested?
Added test

- [x] Yes

### DIFF
--- a/lightly_studio/src/lightly_studio/resolvers/sample_resolver/sample_filter.py
+++ b/lightly_studio/src/lightly_studio/resolvers/sample_resolver/sample_filter.py
@@ -1,0 +1,62 @@
+"""SampleFilter class."""
+
+from typing import List, Optional
+from uuid import UUID
+
+from pydantic import BaseModel
+from sqlmodel import col, select
+
+from lightly_studio.models.annotation.annotation_base import AnnotationBaseTable
+from lightly_studio.models.annotation_label import AnnotationLabelTable
+from lightly_studio.models.metadata import SampleMetadataTable
+from lightly_studio.models.sample import SampleTable
+from lightly_studio.models.tag import TagTable
+from lightly_studio.resolvers.metadata_resolver import metadata_filter
+from lightly_studio.resolvers.metadata_resolver.metadata_filter import MetadataFilter
+from lightly_studio.type_definitions import QueryType
+
+
+class SampleFilter(BaseModel):
+    """Encapsulates filter parameters for querying samples."""
+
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Added uniform class balancing option to `AnnotationClassBalancingStrategy`.
+- Added input class balancing option to `AnnotationClassBalancingStrategy`.
```
MalteEbner:
```suggestion
- Added class balancing with a uniform or the input distribution as target. These options can be set for the `AnnotationClassBalancingStrategy`.
```

lightly_studio/src/lightly_studio/selection/select_via_db.py
```diff
@@ -79,6 +79,15 @@ def _get_class_balancing_data(
         target_keys_set = {a.annotation_label_id for a in annotations}
         target_keys = list(target_keys_set)
         target_values = [1.0 / len(target_keys)] * len(target_keys)
+    elif strat.distribution == "input":
+        # Count the number of times each label appears in the input
+        input_label_count: dict[UUID, int] = defaultdict(int)
+        for annotation in annotations:
+            input_label_count[annotation.annotation_label_id] += 1
```
MalteEbner: Optional: using `from collections import Counter` would be faster (uses C under the hood) and more pythonic:
```suggestion
        input_label_count = Counter(a.annotation_label_id for a in annotations)
```

lightly_studio/src/lightly_studio/selection/select_via_db.py
```diff
@@ -79,6 +79,15 @@ def _get_class_balancing_data(
...
+        target_keys, target_values = (
```
LeonardoRosaa, line 85: Minor: I would suggest changing it:

```
target_keys = list(input_label_count)
target_values = list(input_label_count.values())
```

We have the same result using `list(input_label_count)` or `list(input_label_count.keys())`.

lukas-lightly: I prefer the original, `.keys()` is explicit. Without it you need to know some details of the language and it's harder to read.

lightly_studio/src/lightly_studio/selection/select_via_db.py
lukas-lightly: Thanks I didn't know about it.

PR #130, opened 2025-11-12
add tagging info to readme and docs and changelog

masakljun:
## What has changed and why?

This PR completes the follow-up tasks for the `tag_depth=1` feature (from PR #[*84*](https://github.com/lightly-ai/lightly-studio/pull/84)).

It updates the documentation to make the new feature visible to users.

**Changes:**
* Updates `CHANGELOG.md` with an entry for the new feature.
* Adds a tip to the quickstart examples in `README.md` and `docs/index.md` showing how to use `tag_depth=1`.

### DIFF
--- a/lightly_studio/src/lightly_studio/api/routes/api/dataset.py
+++ b/lightly_studio/src/lightly_studio/api/routes/api/dataset.py
@@ -58,6 +58,15 @@ def read_root_dataset(
     return dataset_resolver.get_root_dataset(session=session)
 
 
+@dataset_router.get("/datasets/dataset_hierarchy", response_model=List[DatasetView])
+def read_dataset_hierarchy(
+    session: SessionDep,
+) -> list[DatasetTable]:
+    """Retrieve the dataset hierarchy from the database, starting with the root node."""
+    root_dataset_id = dataset_resolver.get_root_dataset(session=session).dataset_id
+    return dataset_resolver.get_hierarchy(session=session, root_dataset_id=root_dataset_id)
+
+
 @dataset_router.get("/datasets/{dataset_id}", response_model=DatasetViewWithCount)
 def read_dataset(
     session: SessionDep,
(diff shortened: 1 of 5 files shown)


lightly_studio/docs/docs/index.md
```diff
@@ -68,6 +68,8 @@ git clone https://github.com/lightly-ai/dataset_examples dataset_examples
     # Indexes the dataset, creates embeddings and stores everything in the database. Here we only load images.
     dataset = ls.Dataset.create()
     dataset.add_samples_from_path(path="dataset_examples/coco_subset_128_images/images")
+    # Tip: If your images are in subfolders like 'train/dog/img1.png', 
+    # use `tag_depth=1` to automatically tag them with 'dog'
```
MalteEbner: The tip is not super clear, given that that tag level depends on the path. 

One option to solve that:
We could use `images`, so that it is clear that the tag is relative to the `path` argument passed in `dataset.add_samples_from_path`
```suggestion
    # Tip: If your images are in subfolders like 'images/dog/img1.png' and `images/cat/img2.png`, 
    # use `tag_depth=1` to automatically tag them with 'dog' and 'cat' respectively. The tags are created relative to the `path` argument passed.
```

I am not super happy with this idea, it still sounds a bit too complicated and imprecise to me. Thus I am very open to other suggestions. 

Same for the readme.

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Added class balancing with a uniform or the input distribution as target. These options can be set for the `AnnotationClassBalancingStrategy`.
+- Added `tag_depth=1` parameter to `Dataset.add_samples_from_path` to automatically create tags from parent directory names.
```
MalteEbner: If we define the `path` argument as parent directory, then it creates them based on subdirectory names.
```suggestion
- Added `tag_depth=1` parameter to `Dataset.add_samples_from_path` to automatically create tags from subdirectory names.
```

lightly_studio/docs/docs/index.md
masakljun: Ok yes I agree, I tried out something else (new subsection), hopefully this makes it clearer?

PR #131, opened 2025-11-12
Rename samples backend route to images

michal-lightly:
## What has changed and why?

Rename samples backend route to images.

Follow-ups (for later):
* The function names should contain "image", not "sample"
* Frontend routes

## How has it been tested?

CI tests and a manual test.

- [ ] Yes


lightly_studio_view/src/lib/services/loadDimensionBounds.ts
```diff
@@ -23,7 +23,7 @@ export const loadDimensionBounds = async ({
     };
 
     try {
-        const response = await client.GET('/api/datasets/{dataset_id}/samples/dimensions', {
```
michal-lightly, line 26: Note: We have this single endpoint in our code (not autogenerated).

lightly_studio/src/lightly_studio/api/routes/api/image.py
JonasWurst, line 1: should we rename the file as well?

JonasWurst: also the return types etc. Do you have follow up prs for that?

michal-lightly: I forgot to rename the file, will do that.

Return types are already mostly renamed, but your comment applies for the input types. Will do that in a follow-up. The function names have to be still renamed in a follow-up.

PR #132, opened 2025-11-12
Sample Detail view: editable captions

JonasWurst:
## What has changed and why?

This pr adds the frontend changes to make captions editable.

## How has it been tested?

manually. 

- [x] Yes


lightly_studio/src/lightly_studio/api/routes/api/caption.py
```diff
@@ -41,3 +41,20 @@ def update_caption_text(
...
     return caption_resolver.update_text(session=session, caption_id=caption_id, text=text)
+
+
+@captions_router.get("/captions/{caption_id}", response_model=CaptionView)
+def get_caption(
+    session: SessionDep,
+    dataset_id: Annotated[  # noqa: ARG001
+        UUID,
+        Path(title="Dataset Id", description="The ID of the dataset"),
+    ],  # We need dataset_id because otherwise the path would not match
```
LeonardoRosaa: Minor: Have you tested without dataset_id? I'm not sure, but it could work.

lightly_studio/tests/api/routes/api/test_caption.py
LeonardoRosaa, line 95: A few notes about the unit tests:

- `test_update_caption_text` is testing two endpoints. We could use the `caption_resolver` to retrieve the captions instead of calling the endpoint.
- We're missing a unit test for the `GET /captions/{caption_id}` endpoint.
- Our testing policy is to avoid loops. In this case, we don’t need to retrieve all captions, only the updated one, so we could use the `caption_resolver.get_by_ids` function instead.

lightly_studio/src/lightly_studio/api/routes/api/caption.py
michal-lightly: +1, we don't need dataset_id here

lightly_studio/tests/api/routes/api/test_caption.py
michal-lightly: +1

Regarding for loops - it is not a hard rule, just a strong recommendation. In this case it is not really a for loop, but a filter of "first caption with given id". I agree that we should not use the endpoint but rather a resolver to check the outcome.

lightly_studio/src/lightly_studio/api/routes/api/caption.py
JonasWurst: done

lightly_studio/tests/api/routes/api/test_caption.py
JonasWurst: done

PR #136, opened 2025-11-13
Fix: Properly declare Python 3.13 compatibility. Bump version to 0.4.3.

michal-lightly:
## What has changed and why?

Bugfix: Properly declare Python 3.13 compatibility in pyproject.toml. Otherwise, the package installed from PyPI rejects python 3.13.

## How has it been tested?

Uploaded a new package wheel to TestPyPI and installed it in a python 3.13 virtual environment locally.

- [x] Yes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,8 +9,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- Added class balancing with a uniform or the input distribution as target. These options can be set for the `AnnotationClassBalancingStrategy`.
-
 ### Changed
 
 ### Deprecated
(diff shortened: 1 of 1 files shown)


lightly_studio/pyproject.toml
```diff
@@ -1,9 +1,9 @@
 [project]
 name = "lightly-studio"
-version = "0.4.2"
+version = "0.4.3"
 description = "LightlyStudio is a lightweight, fast, and easy-to-use data exploration tool for data scientists and engineers."
 readme = "USAGE.md"
-requires-python = ">=3.8,<3.13"
+requires-python = ">=3.8,<=3.13"
```
lukas-lightly: I'm afraid `<=3.13` disqualifies `3.13.1` and higher.
```suggestion
requires-python = ">=3.8,<3.14"
```

lukas-lightly: > As with version matching, the release segment is zero padded as necessary to ensure the release segments are compared with the same length.

From https://peps.python.org/pep-0440/#inclusive-ordered-comparison

PR #146, opened 2025-11-13
Show captions by sample in caption grid view

michal-lightly:
## What has changed and why?

Show multiple captions for a sample in Caption grid view. The frontend layout was taken from this PR: #65 

Note that iteration on the GUI might be needed, the main goal of this PR is to display the data at all.

Changes:
* Changed the `readSamples` endpoint to pass pagination in get params. This is needed for Tanstack to autogenerate functions for infinite scrolling.
* Changes the useCaptionsInfinite hook to useSamplesInfinite

## How has it been tested?

* Manually
* Modified endpoint tests in the backend

Screenshot 2025-11-14 at 11 26 48

- [x] Yes


lightly_studio_view/src/lib/hooks/useSamplesInfinite/useSamplesInfinite.ts
```diff
@@ -0,0 +1,38 @@
+import { createInfiniteQuery, useQueryClient } from '@tanstack/svelte-query';
```
michal-lightly, line 1: This file is moved `useCaptionsInfinite.ts` with change from Captions to Samples.

lightly_studio_view/src/lib/components/Captions/CaptionsItem/CaptionsItem.svelte
```diff
@@ -1,12 +1,12 @@
 <script lang="ts">
```
michal-lightly, line 1: Taken from this draft PR: #65

lightly_studio_view/src/lib/components/Captions/Captions.svelte
```diff
@@ -83,7 +83,7 @@
             >
                 {#snippet item({ index, style })}
                     <div {style} class={`pb-[${GridGap}]`}>
-                        <CaptionsItem item={items[index]} />
+                        <CaptionsItem --max-height={`${captionSize}px`} item={items[index]} />
```
JonasWurst, line 86: if we have very long captions, what will this do? I would rather limt the height and then people can use a scrollbar?

lightly_studio_view/src/lib/hooks/useSamplesInfinite/useSamplesInfinite.ts
```diff
@@ -0,0 +1,38 @@
...
+    };
+
+    const data = writable<SampleView[]>([]);
+
+    samplesQuery.subscribe((query) => {
+        if (query.isSuccess) {
+            const allCaptions = query.data.pages.flatMap((page) => page.data);
+            data.set(allCaptions);
+        }
+    });
```
JonasWurst, line 24: I would rather make this file a pure special case:

useSamplesCaptionsInfinte.ts

and remove the hascaption option but rather hard code it here. We are not using it anywhere else.

lightly_studio_view/src/lib/components/Captions/Captions.svelte
michal-lightly: That's what this does - introduces the scrollbar. `captionSize` I believe is the image height.

lightly_studio_view/src/lib/hooks/useSamplesInfinite/useSamplesInfinite.ts
michal-lightly: We are not using it elsewhere yet because it is a new function. I think it might get used for general filtering.

The main reason I'd like to keep it as is is the general design. My understanding is there should be a 1:1 correspondence between hooks and API endpoints. So the hook should expose the full endpoint options.

lightly_studio_view/src/lib/components/Captions/Captions.svelte
michal-lightly: Update: It is both the image width and image height.

PR #150, opened 2025-11-14
Initial implementation of dynamic dataset hierarchy

LeonardoRosaa:
## What has changed and why?

Implemented an initial version to support a dynamic dataset hierarchy in the GUI.

## How has it been tested?

Upload a video dataset and an image dataset, and the header should render the correct hierarchy.

- [ ] Yes

### DIFF
--- a/lightly_studio_view/src/lib/components/Header/Header.test.ts
+++ b/lightly_studio_view/src/lib/components/Header/Header.test.ts
@@ -7,7 +7,7 @@ import '@testing-library/jest-dom';
 import { Page } from '@sveltejs/kit';
 import type { ReversibleAction } from '$lib/hooks/useGlobalStorage';
 
-import { readDataset } from '$lib/api/lightly_studio_local';
+import { readRootDataset } from '$lib/api/lightly_studio_local';
 
 describe('Header', () => {
     const setup = (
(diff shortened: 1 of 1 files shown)


lightly_studio_view/src/lib/components/NavigationMenu/NavigationMenu.svelte
```diff
@@ -5,57 +5,86 @@
...
+        if (!menuItem) return [];
 
-    const videoMenu = () => [
-        {
-            title: 'Videos',
-            id: 'videos',
-            href: routeHelpers.toVideos(datasetId),
-            isSelected: pageId === APP_ROUTES.videos,
-            icon: Video
+        let children = dataset.children;
```
horatiualmasan, line 53: This does not lists all children recursively. You will only get the direct children. Have you thought about multiple levels ?

horatiualmasan (approved):
LGTM

lightly_studio_view/src/lib/components/NavigationMenu/NavigationMenu.svelte
LeonardoRosaa: We will implement it later. This is an initial version for a dynamic dataset hierarchy.

PR #152, opened 2025-11-14
Fix port availability check for dual-stack localhost binding

eternal-f1ame:
# Fix port availability check for dual-stack localhost binding

## Description

This PR fixes a bug in `_is_port_available()` where the server would fail to start with `RuntimeError: Could not find an available port` even when ports were actually available.

### Root Cause

The bug occurred when checking port availability on `localhost` on systems with dual-stack networking (IPv4/IPv6):

1. The function attempted to bind to both `AF_INET` (IPv4) and `AF_INET6` (IPv6) address families
2. On systems where IPv6 is not fully configured for localhost, the IPv6 bind would fail
3. The function used AND logic: it returned `False` if **any** address family failed to bind
4. This caused the function to report ports as unavailable even when IPv4 binding succeeded

### Solution

Changed the logic to use OR semantics:
- The port is now considered available if **at least one** address family can successfully bind
- Added early return on first successful bind attempt
- Port is only unavailable if **all** address families fail to bind

### Changes Made

**`lightly_studio/src/lightly_studio/api/server.py`:**
- Modified `_is_port_available()` function to return `True` on first successful bind
- Added comment explaining the OR logic
- Changed from "all must succeed" to "any succeeds" pattern

**`lightly_studio/tests/api/test_server.py`:**
- Added test case `test__get_available_port__localhost_with_ipv6_failure()`
- Validates that port availability check succeeds even when IPv6 binding might fail

### Manual Testing
```bash
import lightly_studio as ls

# Indexes the dataset, creates embeddings and stores everything in the database. Here we only load images.
dataset = ls.Dataset.create()
dataset.add_samples_from_path(path="dataset_examples/coco_subset_128_images/images")

# Start the UI server on localhost:8001.
# Use env variables LIGHTLY_STUDIO_HOST and LIGHTLY_STUDIO_PORT to customize it.
ls.start_gui()
```
**Before fix:** `RuntimeError: Could not find an available port`  
**After fix:** Server starts successfully on port 8001

### Unit Tests
```bash
cd lightly_studio
make test
```
All existing tests pass, plus new test case validates the fix.

## Impact

- **Low risk**: Changes only affect port availability checking logic
- **Backwards compatible**: No API or behavior changes for end users
- **Improves reliability**: Server now starts correctly on systems with partial IPv6 support

## Related Issues

Fixes the issue where users encountered:
```
RuntimeError: Could not find an available port.
```

## Checklist

- [x] Code follows project coding guidelines
- [x] Added unit test for the bug fix
- [x] All existing tests pass
- [x] Manual testing confirms fix works
- [x] Commit message follows conventional format

### DIFF
--- a/lightly_studio/src/lightly_studio/api/server.py
+++ b/lightly_studio/src/lightly_studio/api/server.py
@@ -73,10 +73,12 @@ def _is_port_available(host: str, port: int) -> bool:
             # Fallback for hostnames like 'localhost'
             families = [socket.AF_INET, socket.AF_INET6]
 
+    # The port is available if we can bind to at least one address family.
     for family in families:
         with socket.socket(family, socket.SOCK_STREAM) as s:
             try:
                 s.bind((host, port))
+                return True
             except OSError:
-                return False
-    return True
+                continue
+    return False
--- a/lightly_studio/tests/api/test_server.py
+++ b/lightly_studio/tests/api/test_server.py
@@ -81,3 +81,19 @@ def test__get_available_port__preferred_port_in_use() -> None:
     server = Server(host, port)
     s.close()
     assert server.port != port
+
+
+def test__get_available_port__localhost_with_ipv6_failure() -> None:
+    """Test that port is available even if IPv6 binding fails for localhost.
+    
+    This tests the fix for the bug where _is_port_available returned False
+    when checking 'localhost' because IPv6 binding failed, even though IPv4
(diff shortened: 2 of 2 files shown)


lukas-lightly:
@eternal-f1ame The fix is in https://github.com/lightly-ai/lightly-studio/pull/352
It will be in the next release, some time in the next two weeks.

PR #154, opened 2025-11-17
Use correct matrix.python version in unittests

MalteEbner:
## What has changed and why?

Use the correct python version in python unittests (the one from matrix.python). Previously, it always used python 3.8, see the prints of `Downloading cpython-3.8.20-linux-x86_64-gnu (download) (19.9MiB)
 Downloading cpython-3.8.20-linux-x86_64-gnu (download)
Using CPython 3.8.20` in https://github.com/lightly-ai/lightly-studio/actions/runs/19440391599/job/55622023406

Python 3.13 behaves slightly differently, thus some code and tests had to be upgraded as well.

The uv.lock file used to pin `numpy==1.26.4,` which is not compatible with python 3.13.The lock file was upgraded using `uv lock --upgrade-package numpy`, while keeping the `pyproject.toml` untouched.

## How has it been tested?

By CI. See `Using CPython 3.13.9 interpreter at: /opt/hostedtoolcache/Python/3.13.9/x64/bin/python3.13` in https://github.com/lightly-ai/lightly-studio/actions/runs/19441052654/job/55624150381?pr=154

By existing unittests.

- [ ] Yes

### DIFF
--- a/lightly_studio_view/src/lib/components/Captions/Captions.svelte
+++ b/lightly_studio_view/src/lib/components/Captions/Captions.svelte
@@ -12,7 +12,7 @@
         datasetId: string;
     } = $props();
 
-    const { data, query, loadMore } = $derived(
+    const { data, query, loadMore, refresh } = $derived(
         useSamplesInfinite({
             body: { filters: { dataset_id: datasetId, has_captions: true } }
         })
@@ -49,7 +49,8 @@
     });
 
     let items = $derived($data);
-    const GridGap = 16;
+    const GridGap = 8;
+    const innerCardMargin = 32;
 
     const height = $derived(viewportHeight + GridGap);
 </script>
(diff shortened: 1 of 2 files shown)


lightly_studio/src/lightly_studio/core/dataset.py
```diff
@@ -153,7 +153,7 @@ def load_or_create(
         if dataset.sample_type != sample_type:
             raise ValueError(
                 f"Dataset with name '{name}' already exists with sample type "
-                f"'{dataset.sample_type}', but '{sample_type}' was requested."
+                f"'{dataset.sample_type.value}', but '{sample_type.value}' was requested."
```
MalteEbner, line 156: Otherwise on python 3.13
```bash
E AssertionError: Regex pattern did not match.
E Regex: "is having sample type 'video', expected 'image'"
E Input: "[...] is having sample type 'SampleType.VIDEO', expected 'SampleType.IMAGE'."
```

lightly_studio/tests/core/test_add_videos.py
```diff
@@ -94,7 +94,7 @@ def test__create_video_frame_samples(db_session: Session, tmp_path: Path) -> Non
                 width=320,
                 height=240,
                 duration_s=2.0,  # 2 frames / 1 fps = 2 seconds
-                fps=1.0,
+                fps=1,
```
michal-lightly, line 97: This is strange, `VideoCreate.fps` is a float. Why was the change needed?

MalteEbner: > `VideoCreate.fps` is a float.

It can continue being a float. This change only affects that tests which create a virtual video, which requires the `rate`paratemeter to `output_container.add_stream("libx264", rate=fps)` to be an `int` or a `Fraction`.

> Why was the change needed?
See https://github.com/lightly-ai/lightly-studio/pull/154#discussion_r2537513164

PR #155, opened 2025-11-18
Caption editable in Caption GridView

JonasWurst:
## What has changed and why?

This pr adds the editable captions to the caption grid view. It is reusing the panel component for the captions

## How has it been tested?

locally

- [x] Yes


lightly_studio_view/src/lib/schema.d.ts
michal-lightly, line 27: Weird that there is a schema change - can you run `make build` again before merging?

PR #160, opened 2025-11-18
Delete Caption Frontend

JonasWurst:
## What has changed and why?

Adding the ability to delete captions to caption grid view and sample detail view.

## How has it been tested?

manually

- [ ] Yes

### DIFF
--- a/lightly_studio_view/src/lib/components/Captions/CaptionsItem/CaptionsItem.svelte
+++ b/lightly_studio_view/src/lib/components/Captions/CaptionsItem/CaptionsItem.svelte
@@ -1,11 +1,13 @@
 <script lang="ts">
     import { toast } from 'svelte-sonner';
+    import { page } from '$app/state';
     import { Card, CardContent } from '$lib/components';
     import type { SampleView } from '$lib/api/lightly_studio_local';
     import { SampleImage } from '$lib/components';
     import CaptionField from '$lib/components/CaptionField/CaptionField.svelte';
     import { useSettings } from '$lib/hooks/useSettings';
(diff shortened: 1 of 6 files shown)


lightly_studio_view/src/lib/components/Captions/CaptionsItem/CaptionsItem.svelte
```diff
@@ -19,6 +21,24 @@
 
     let objectFit = $derived($gridViewSampleRenderingStore); // Use store value directly
     $inspect(item);
+
+    const { deleteCaption } = useDeleteCaption();
+
+    const handleDeleteCaption = async (captionId: string) => {
+        if (!item) return;
+
+        const _delete = async () => {
```
LeonardoRosaa: It seems that we don't need to wrap the delete logic in the `_delete`  function. What do you think?

lightly_studio_view/src/lib/components/Captions/CaptionsItem/CaptionsItem.svelte
```diff
@@ -27,7 +47,11 @@
             <SampleImage sample={item} {objectFit} />
             <div class="flex h-full w-full flex-1 flex-col overflow-auto px-4 py-2">
                 {#each item.captions as caption}
-                    <CaptionField {caption} {onUpdate} />
+                    <CaptionField
+                        {caption}
+                        onDeleteCaption={() => handleDeleteCaption(caption.caption_id)}
```
LeonardoRosaa: Minor: I'd rename it to `onDeleteCaption` instead of `handleDeleteCaption`

PR #161, opened 2025-11-18
Remove AnnotationImageView and AnnotationDetailsView

horatiualmasan:
## What has changed and why?

Remove `AnnotationImageView` and `AnnotationDetailsView` in preparation of removing Image Table references from Annotation table.

## How has it been tested?
Current tests work
Tested locally

- [ ] Yes

### DIFF
--- a/README.md
+++ b/README.md
@@ -160,7 +160,7 @@ LightlyStudio has a powerful Python interface. You can not only index datasets b
 To load images directly from a cloud storage provider (like AWS S3, GCS, etc.), you must first install the required dependencies:
 
 ```py
-pip install lightly-studio[cloud-storage]
+pip install "lightly-studio[cloud-storage]"
 ```
 
 This installs the necessary libraries: s3fs (for S3), gcsfs (for GCS), and adlfs (for Azure).
--- a/lightly_studio/docs/docs/index.md
+++ b/lightly_studio/docs/docs/index.md
@@ -264,7 +264,7 @@ LightlyStudio has a powerful Python interface. You can not only index datasets b
 To load images directly from a cloud storage provider (like AWS S3, GCS, etc.), you must first install the required dependencies:
 
 ```py
-pip install lightly-studio[cloud-storage]
+pip install "lightly-studio[cloud-storage]"
 ```
 
 This installs the necessary libraries: s3fs (for S3), gcsfs (for GCS), and adlfs (for Azure).


lightly_studio_view/src/lib/components/AnnotationsGrid/AnnotationsGridItem/AnnotationsGridItem.svelte
```diff
@@ -28,20 +29,27 @@
     }: Props = $props();
 
     const padding = 20;
-    const sample = annotation.sample;
+    const { sample: sampleQuery } = useImage({ sampleId: annotation.parent_sample_id });
```
horatiualmasan: Maybe there is a better way to do it and not get the sample for each annotation.

lightly_studio_view/src/lib/components/AnnotationDetails/AnnotationDetailsPanel/AnnotationDetailsPanel.svelte
```diff
@@ -9,12 +9,15 @@
     import Button from '$lib/components/ui/button/button.svelte';
     import { useRemoveTagFromAnnotation } from '$lib/hooks/useRemoveTagFromAnnotation/useRemoveTagFromAnnotation';
     import SegmentTags from '../../SegmentTags/SegmentTags.svelte';
+    import type { ImageSample } from '$lib/services/types';
 
     const {
         annotationId,
+        sample: sampleProp,
```
michal-lightly: Frontend question: Where is the `sampleProp` symbol defined?

lightly_studio_view/src/lib/components/AnnotationDetails/AnnotationDetailsPanel/AnnotationDetailsPanel.svelte
```diff
@@ -9,12 +9,15 @@
...
         onUpdate
     }: {
         annotationId: string;
+        sample: ImageSample;
```
michal-lightly: Optional: `sample` is probably not the best name since we access `sample.sample` later. I was thinking `image` would be better, eventually it might be `image_or_frame` - not sure how this will evolve.

lightly_studio_view/src/lib/components/AnnotationDetails/AnnotationDetails.svelte
```diff
@@ -41,7 +41,7 @@
         sample
     }: {
         dataset: Dataset;
-        sample: Sample;
+        sample: ImageSample;
```
michal-lightly: I see we have more places with naming issues. I would suggest to rename to `image` for now (it matches the type).

lightly_studio_view/src/lib/components/AnnotationsGrid/AnnotationsGridItem/AnnotationsGridItem.svelte
michal-lightly: Can you add a todo and perhaps ask @mihnea-necsulescu what does he think?

lightly_studio_view/src/lib/components/AnnotationsGrid/AnnotationsGridItem/AnnotationsGridItem.svelte
```diff
@@ -98,15 +106,17 @@
...
     );
 </script>
 
-{#if isLoaded}
+{#if isLoaded && sample}
     <div
         class="crop rounded-lg bg-black"
         class:annotation-selected={selected}
-        style={`
+        style={`
```
michal-lightly: Phantom whitespace

lightly_studio_view/src/lib/components/AnnotationsGrid/AnnotationsGridItem/AnnotationsGridItem.svelte
```diff
@@ -98,15 +106,17 @@
...
     const uniqueImageUrl = $derived(
-        `${PUBLIC_SAMPLES_URL}/sample/${sample.sample_id}${datasetVersion ? `?v=${datasetVersion}` : ''}`
+        sample
+            ? `${PUBLIC_SAMPLES_URL}/sample/${annotation.parent_sample_id}${datasetVersion ? `?v=${datasetVersion}` : ''}`
+            : ''
     );
 </script>
 
-{#if isLoaded}
+{#if isLoaded && sample}
```
michal-lightly: Why do we need to assert the sample too? `isLoaded` is definition uses the sample, and it looks like we cast it to bool.

michal-lightly: Also, why here we do not need to cast to bool using `!!sample`?

lightly_studio_view/src/lib/components/AnnotationsGrid/AnnotationsGridItem/AnnotationsGridItem.svelte
LeonardoRosaa: This change introduces a new request for each image to fetch its details in the annotations grid view. 

I'm not sure this is the best approach, but I’d suggest that the back end return the appropriate sample as part of the response. For example, if we need a list of image annotations, the back end should also join the Image table and include it in the response. 

I tested it in [this draft](https://github.com/lightly-ai/lightly-studio/pull/148).

lightly_studio_view/src/lib/components/AnnotationDetails/AnnotationDetailsPanel/AnnotationDetailsPanel.svelte
horatiualmasan: it was just a renaming as I used sample below to get it reactive, but that was not really needed. removed.

lightly_studio_view/src/lib/components/AnnotationsGrid/AnnotationsGridItem/AnnotationsGridItem.svelte
horatiualmasan: It's needed to make typescript checks happy, we use sample below and it complains it could be undefined. Added bool conversion here also to be consistent.

lightly_studio_view/src/lib/components/AnnotationDetails/AnnotationDetails.svelte
horatiualmasan: renamed to image

lightly_studio_view/src/lib/components/AnnotationsGrid/AnnotationsGridItem/AnnotationsGridItem.svelte
horatiualmasan: Added a todo, will look more into this once we have also video frame annotations

PR #163, opened 2025-11-18
Use separate environment variables for test scripts

michal-lightly:
## What has changed and why?

Previously, running examples (and therefore manual testing) was clumsy:
* We have different dataset formats, but used one env variable to pass data to all example scripts
* Coco examples had hardcoded paths

Changes in this PR:
* Create separate env variables for different dataset types (folder, coco, yolo, videos)
* Added defaults to `.env.example`
    * Except videos, the will work out of the box if the user has `dataset_examples` cloned in the `lightly_studio` subdir (what we instruct in readme)

## How has it been tested?

Manually ran all the examples - no env var change was needed.

- [ ] Yes

### DIFF
--- a/lightly_studio/src/lightly_studio/api/routes/api/frame.py
+++ b/lightly_studio/src/lightly_studio/api/routes/api/frame.py
@@ -40,7 +40,7 @@ def get_all_frames(
 
 
 @frame_router.get("/{sample_id}", response_model=VideoFrameView)
-def get_by_id(
+def get_frame_by_id(
     session: SessionDep,
     sample_id: Annotated[UUID, Path(title="Sample Id")],
 ) -> VideoFrameTable:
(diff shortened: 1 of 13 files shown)


lightly_studio/src/lightly_studio/examples/example_yolo.py
```diff
@@ -1,6 +1,5 @@
 """Example of how to add samples in yolo format to a dataset."""
 
-from pathlib import Path
```
JonasWurst:
```suggestion
```

lightly_studio/src/lightly_studio/examples/example.py
```diff
@@ -1,6 +1,5 @@
 """Example of how to load samples from path with the dataset class."""
 
-from pathlib import Path
```
JonasWurst:
```suggestion
```

PR #164, opened 2025-11-18
Link CaptionTable to SampleTable

michal-lightly:
## What has changed and why?

Link CaptionTable to SampleTable.

* Makes CaptionTable.sample_id a foreign key in SampleTable, effectively making caption a sample
* Creates a child dataset for captions. This change is in `caption_resolver.create_many`, which is also changed to return sample_ids.

In a follow-up, CaptionTable.dataset_id should be removed in favour getting the dataset from the linked samples.

## How has it been tested?

Existing and new tests.

- [ ] Yes

### DIFF
--- a/lightly_studio/pyproject.toml
+++ b/lightly_studio/pyproject.toml
@@ -49,7 +49,7 @@ dependencies = [
     #   lightly-mundig==$version
     # Then lock the version. The uv.lock file is updated accordingly.
     # uv lock --upgrade-package lightly-mundig==$version
-    "lightly-mundig==0.1.7",
+    "lightly-mundig==0.1.9",
     "pyarrow>=17.0.0",
     "av>=10.0.0",
     "opencv-python>=4.11.0.86",
@@ -134,8 +134,6 @@ packages = ["src/lightly_studio"]
 
 [tool.uv.sources]
 lightly-edge-sdk = { index = "lightly-pypi" }
-# When you want to use public PyPi mundig versions, comment out the line below.
-lightly_mundig = { index = "lightly-pypi" }
 
 [tool.uv]
 keyring-provider = "subprocess"
--- a/lightly_studio/src/lightly_studio/metadata/compute_similarity.py
+++ b/lightly_studio/src/lightly_studio/metadata/compute_similarity.py
@@ -1,0 +1,68 @@
+"""Computes similarity from embeddings."""
+
+from uuid import UUID
+
+from lightly_mundig import Similarity  # type: ignore[import-untyped]
+from sqlmodel import Session
+
+from lightly_studio.dataset.env import LIGHTLY_STUDIO_LICENSE_KEY
+from lightly_studio.resolvers import metadata_resolver, sample_embedding_resolver
+from lightly_studio.resolvers.sample_resolver.sample_filter import SampleFilter
+
+
+def compute_similarity_metadata(
+    session: Session,
+    key_dataset_id: UUID,
+    embedding_model_id: UUID,
+    query_tag_id: UUID,
+    metadata_name: str,
+) -> None:
+    """Computes similarity for each sample in the dataset from embeddings.
+
+    Similarity is a measure of how similar a sample is to its nearest neighbor
+    in the embedding space. It can be used to find duplicates.
+
+    The computed similarity values are stored as metadata for each sample.
+
+    Args:
+        session:
+            The database session.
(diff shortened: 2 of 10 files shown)


lightly_studio/src/lightly_studio/api/routes/api/caption.py
```diff
@@ -60,22 +60,29 @@ def create_caption(
...
         captions=[
             CaptionCreate(
                 dataset_id=parent_sample.dataset_id,
                 parent_sample_id=create_caption_input.parent_sample_id,
                 text=create_caption_input.text,
             ),
         ],
-    )[0]
+    )
+    assert len(sample_ids) == 1, "Expected exactly one caption to be created."
```
JonasWurst, line 82: typically we use if -> error blocks

lightly_studio/src/lightly_studio/api/routes/api/caption.py
```diff
@@ -60,22 +60,29 @@ def create_caption(
     create_caption_input: Annotated[CaptionCreateInput, Body()],
 ) -> CaptionTable:
     """Create a new caption."""
+    # Get the parent dataset
```
JonasWurst:
```suggestion
    # Get the parent sample
```

lightly_studio/src/lightly_studio/resolvers/caption_resolver.py
```diff
@@ -20,23 +23,55 @@ class GetAllCaptionsResult(BaseModel):
...
-def create_many(session: Session, captions: Sequence[CaptionCreate]) -> list[CaptionTable]:
-    """Create many captions in bulk.
+class CaptionCreateHelper(CaptionCreate):
+    """Helper class to create CaptionTable with sample_id."""
+
+    sample_id: UUID
+
+
+def create_many(
+    session: Session, parent_dataset_id: UUID, captions: Sequence[CaptionCreate]
```
JonasWurst, line 33: instead of introducing the parent_dataset_id, i would rather get that info from the captions, since they also contain the parent_sample_id, which can lead to the parent_dataset_id. wdyt?

lightly_studio/tests/conftest.py
```diff
@@ -428,6 +428,7 @@ def captions_test_data(
 
     _ = caption_resolver.create_many(
         session=db_session,
+        parent_dataset_id=samples[0].sample.dataset_id,
```
JonasWurst: I think this fixture got removed.

lightly_studio/tests/resolvers/test_caption_resolver.py
```diff
@@ -68,6 +76,63 @@ def test_create_many(test_db: Session) -> None:
...
 
 
+def test_create_many__check_dataset_ids(test_db: Session) -> None:
+    dataset = create_dataset(session=test_db)
+    dataset_id = dataset.dataset_id
+    image = create_image(session=test_db, dataset_id=dataset_id)
+
+    created_ids = caption_resolver.create_many(
+        session=test_db,
+        parent_dataset_id=dataset.dataset_id,
```
JonasWurst:
```suggestion
        parent_dataset_id=dataset_id,
```

lightly_studio/tests/resolvers/test_caption_resolver.py
```diff
@@ -68,6 +76,63 @@ def test_create_many(test_db: Session) -> None:
...
+        captions=[
+            CaptionCreate(
+                dataset_id=dataset_id,
```
JonasWurst, line 89: For me it is confusing why we need this

JonasWurst: it is technically the paren_dataset_id, maybe we should rename

lightly_studio/tests/resolvers/test_caption_resolver.py
```diff
@@ -68,6 +76,63 @@ def test_create_many(test_db: Session) -> None:
...
+
+
+def test_create_many__relationships(test_db: Session) -> None:
+    dataset = create_dataset(session=test_db)
+    dataset_id = dataset.dataset_id
+    image = create_image(session=test_db, dataset_id=dataset_id)
+
+    created_ids = caption_resolver.create_many(
+        session=test_db,
+        parent_dataset_id=dataset.dataset_id,
```
JonasWurst: see above

PR #165, opened 2025-11-18
Add function to get or create an arbitrary child dataset

michal-lightly:
## What has changed and why?

Generalised the `get_or_create_video_frame_child` to `get_or_create_child_dataset` with a given sample type.

## How has it been tested?

Modified unit tests.

- [ ] Yes

### DIFF
--- a/lightly_studio/src/lightly_studio/resolvers/dataset_resolver/get_or_create_child_dataset.py
+++ b/lightly_studio/src/lightly_studio/resolvers/dataset_resolver/get_or_create_child_dataset.py
@@ -42,7 +42,7 @@ def get_or_create_child_dataset(
         return child_datasets[0].dataset_id
     if len(child_datasets) > 1:
         raise ValueError(
-            f"Multiple child datasets with sample type {sample_type} found "
+            f"Multiple child datasets with sample type {sample_type.value} found "
             f"for dataset id {dataset_id}."
         )
 


lightly_studio/src/lightly_studio/resolvers/dataset_resolver/get_or_create_child_dataset.py
```diff
@@ -0,0 +1,58 @@
+"""Returns a video frame child dataset for a given video dataset, creating it if necessary."""
+
+from __future__ import annotations
```
michal-lightly, line 3: This file is renamed `get_or_create_video_frame_child.py`.

lightly_studio/src/lightly_studio/resolvers/dataset_resolver/get_or_create_child_dataset.py
```diff
@@ -0,0 +1,58 @@
...
+        The uuid of the child dataset.
+
+    Raises:
+        ValueError: If multiple child datasets with the given sample type exist.
+    """
+    # Get filtered child datasets.
+    dataset = dataset_resolver.get_by_id(session=session, dataset_id=dataset_id)
+    if dataset is None:
+        raise ValueError(f"Dataset with id {dataset_id} not found.")
+    child_datasets = [ds for ds in dataset.children if ds.sample_type == sample_type]
```
michal-lightly, line 38: There's a logic change from the `get_or_create_video_frame_child` implementation:

1. We don't check the parent dataset type anymore
2. We fetch a direct child - using `dataset.children` instead of fetching the whole hierarchy. I don't see a use-case for requesting a non-direct child datasets (at least for now)

horatiualmasan: If we don't check the parent dataset type anymore we could then add video frame dataset as a child to a image dataset. Should we add a parent dataset type parameter or just leave the check up to the user outside this function?

PR #167, opened 2025-11-19
Simplify batching logic when loading samples

michal-lightly:
## What has changed and why?

Improve annotation and caption insertion batch logic.

There were several issues with the logic:
* Scattered across multiple functions
* The `(annotations|captions)_to_create` variable was created in a different function then where it was used and modified (used as an in-out parameter)
* Illogical "insert remaining annotations/captions" call in the caller, where it should have been handled inside a function
* Unnecessary extra batch size variable

## How has it been tested?

Existing tests.

- [ ] Yes

### DIFF
--- a/lightly_studio/src/lightly_studio/core/add_samples.py
+++ b/lightly_studio/src/lightly_studio/core/add_samples.py
@@ -150,7 +150,6 @@ def load_into_dataset_from_labelformat(
     # Create label mapping
     label_map = _create_label_map(session=session, input_labels=input_labels)
 
-    annotations_to_create: list[AnnotationCreate] = []
     samples_to_create: list[ImageCreate] = []
     created_sample_ids: list[UUID] = []
     path_to_anno_data: dict[str, ImageInstanceSegmentation | ImageObjectDetection] = {}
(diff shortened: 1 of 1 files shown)


lightly_studio/src/lightly_studio/core/add_samples.py
```diff
@@ -478,15 +460,19 @@ def _process_instance_segmentation_annotations(
...
     session: Session,
     created_path_to_id: dict[str, UUID],
     path_to_anno_data: dict[str, ImageInstanceSegmentation | ImageObjectDetection],
     dataset_id: UUID,
     label_map: dict[int, UUID],
-    annotations_to_create: list[AnnotationCreate],
 ) -> None:
     """Process annotations for a batch of samples."""
+    if len(created_path_to_id) == 0:
+        return
```
michal-lightly, line 472: Added this check for extra safety and parity with the captions function.

michal-lightly: Also made the check more explicit (instead of an implicit cast to bool).

PR #171, opened 2025-11-19
Link annotations to SampleTable

horatiualmasan:
## What has changed and why?

Link AnnotationBaseTable to SampleTable

## How has it been tested?

Existing tests

- [ ] Yes

### DIFF
--- a/lightly_studio/tests/resolvers/test_annotation_resolver.py
+++ b/lightly_studio/tests/resolvers/test_annotation_resolver.py
@@ -146,34 +146,29 @@ def test_create_and_get_annotation__for_video_frame_with_ordering(test_db: Sessi
     dataset_id = create_dataset(session=test_db, sample_type=SampleType.VIDEO).dataset_id
 
     # Create video.
-    sample_video_id_b = create_videos(
+    video_ids = create_videos(
         session=test_db,
         dataset_id=dataset_id,
         videos=[
             VideoStub(path="/path/to/b_video.mp4"),
-        ],
-    )[0]
-    sample_video_id_a = create_videos(
-        session=test_db,
-        dataset_id=dataset_id,
-        videos=[
             VideoStub(path="/path/to/a_video.mp4"),
         ],
-    )[0]
-
+    )
+    video_id_b = video_ids[0]
+    video_id_a = video_ids[1]
     # Create video frames.
     frames_to_create = [
         VideoFrameCreate(
             frame_number=1,
             frame_timestamp_s=0.1,
             frame_timestamp_pts=1,
-            parent_sample_id=sample_video_id_b,
+            parent_sample_id=video_id_b,
         ),
         VideoFrameCreate(
             frame_number=1,
             frame_timestamp_s=0.1,
             frame_timestamp_pts=1,
-            parent_sample_id=sample_video_id_a,
+            parent_sample_id=video_id_a,
         ),
     ]
 


lightly_studio/tests/resolvers/annotations/test_annotations_base_get_all_annotations_details.py
```diff
@@ -79,6 +79,19 @@ def test_default_ordering_by_file_path_abs(
 ) -> None:
     """Test that annotations are ordered by sample file path."""
     annotations = annotation_resolver.get_all(db_session).annotations
-    sample_paths: list[str] = [a.sample.file_path_abs for a in annotations if a.sample]
+
+    # Get all sample IDs from annotations
+    sample_ids = [a.parent_sample_id for a in annotations if a.parent_sample_id]
+
+    # Fetch all images using image_resolver
```
horatiualmasan: As annotation is not linked anymore to images file pats for the test need to be obtained using the image resolver.

lightly_studio/src/lightly_studio/resolvers/annotation_resolver/get_all.py
```diff
@@ -43,13 +44,21 @@ def get_all(
...
+        .outerjoin(
+            VideoFrameTable,
+            col(VideoFrameTable.sample_id) == col(AnnotationBaseTable.parent_sample_id),
+        )
+        .outerjoin(VideoTable, col(VideoTable.sample_id) == col(VideoFrameTable.parent_sample_id))
+        .order_by(
+            func.coalesce(ImageTable.file_path_abs, VideoTable.file_path_abs, "").asc(),
+            col(AnnotationBaseTable.created_at).asc(),
+            col(AnnotationBaseTable.annotation_id).asc(),
+        )
```
michal-lightly, line 60: Side-note: The ordering here is a little non-systematic, the query in theory needs to be adjusted for every sample type which can be linked to an annotation. A solution would be to order by `sample_id`, but then e2e tests which assume an order start to fail. I think it is fine to keep it as is.

A potential solution in a distant future could be to introduce an `order` field on `SampleTable`, then we could even let users to reorder samples arbitrarily.

lightly_studio/tests/resolvers/annotations/test_annotations_base_get_all_annotations_details.py
michal-lightly: Please rewrite the test with setting up the test data in this function. That solves the issue. It also solves the issue that now we don't even know how the samples were inserted - their filenames might have been created in order, in which case this test would be weak.

michal-lightly: Should we also add a test for video frame annotation fetching?

lightly_studio/tests/resolvers/test_annotation_resolver.py
```diff
@@ -135,6 +142,74 @@ def test_create_and_get_annotation(test_db: Session, test_data: _TestData) -> No
...
+            VideoStub(path="/path/to/b_video.mp4"),
+        ],
+    )[0]
+    sample_video_id_a = create_videos(
+        session=test_db,
+        dataset_id=dataset_id,
+        videos=[
+            VideoStub(path="/path/to/a_video.mp4"),
+        ],
+    )[0]
```
michal-lightly: Optional:
```py
video_ids = create_videos(...)
video_id_b = video_ids[0]
video_id_a = video_ids[1]
```

lightly_studio/tests/resolvers/test_annotation_resolver.py
```diff
@@ -135,6 +142,74 @@ def test_create_and_get_annotation(test_db: Session, test_data: _TestData) -> No
     assert retrieved_annotation == dog_annotation
 
 
+def test_create_and_get_annotation__for_video_frame(test_db: Session) -> None:
```
michal-lightly: Optional:
```suggestion
def test_create_and_get_annotation__video_frame_order(test_db: Session) -> None:
```

horatiualmasan: updated

lightly_studio/tests/resolvers/annotations/test_annotations_base_get_all_annotations_details.py
horatiualmasan: updated

PR #172, opened 2025-11-19
Add `compute_similarity_metadata` function

lukas-lightly:
## What has changed and why?
The query samples are defined using a tag.

Uses the new lightly-mundig. I've migrated to the public PyPI instead of internal-pypi.

## How has it been tested?
Added a test.

- [ ] Yes

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -7,7 +7,7 @@ We welcome contributions of all kinds, including:
 
 After you have your changes ready, and you create a new pull request, a maintainer will review your PR, may ask for changes, suggest improvements, or approve once ready.
 
-## Development
+## Development Quickstart
 
 
 ```bash
@@ -23,10 +23,17 @@ This will:
 
 For starting it again, you can skip the build step by just calling `make start-example`.
 
+Backend code lives in the `lightly_studio` subdirectory, frontend code in `lightly_studio_view`.
 To run static checks and unit tests use the following commands
 
 ```bash
-cd lightly_studio/lightly_studio
+# Backend
+cd lightly_studio
+make static-checks
+make test
+
+# Frontend
+cd lightly_studio_view
 make static-checks
 make test
 ```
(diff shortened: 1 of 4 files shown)


lightly_studio/src/lightly_studio/metadata/compute_similarity.py
```diff
@@ -0,0 +1,68 @@
...
+from lightly_studio.resolvers import metadata_resolver, sample_embedding_resolver
+from lightly_studio.resolvers.sample_resolver.sample_filter import SampleFilter
+
+
+def compute_similarity_metadata(
+    session: Session,
+    key_dataset_id: UUID,
+    embedding_model_id: UUID,
+    query_tag_id: UUID,
+    metadata_name: str,
```
MalteEbner, line 18: Please set a good default for the `metadata_name`.  See e.g. in the issue

> The name for the metadata should reflect that it is the similarity to a tag. E.g. use "similarity__{tag_name}__{current_date}" as the default name.

The doctoring should also be updated.

lightly_studio/tests/metadata/test_compute_similarity.py
```diff
@@ -0,0 +1,69 @@
...
+    ]
+    samples = []
+    for i, embedding in enumerate(embeddings):
+        image = create_image(session=test_db, dataset_id=dataset_id, file_path_abs=f"sample{i}.jpg")
+        create_sample_embedding(
+            session=test_db,
+            sample_id=image.sample_id,
+            embedding=embedding,
+            embedding_model_id=embedding_model_id,
+        )
```
MalteEbner: There is also the `helpers_resolvers.create_samples_with_embeddings` helper you created. Can it be used here instead?
https://github.com/lightly-ai/lightly-studio/blame/50c3011a7cb5aaf2a6fd929a74e819a30df37ffb/lightly_studio/tests/helpers_resolvers.py#L318

lightly_studio/src/lightly_studio/metadata/compute_similarity.py
lukas-lightly: This is a low-level function, I'd prefer to do this a level above. E.g. here you don't have the tag name, but only `query_tag_id`, so it's really hard to construct the tag.
Here's WIP code which does this one level above, inside `Dataset` and then calls this function. I'll post it as a PR next.
Image

lightly_studio/tests/metadata/test_compute_similarity.py
lukas-lightly: Done, also for the typicality calculation.

lightly_studio/src/lightly_studio/metadata/compute_similarity.py
MalteEbner: Great idea!

PR #174, opened 2025-11-19
Fix video frame orientation

LeonardoRosaa:
## What has changed and why?

(Delete this: Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context. List any dependencies that are required for this change.)

## How has it been tested?

- [ ] Yes

### DIFF
--- a/lightly_studio/src/lightly_studio/api/routes/video_frames_media.py
+++ b/lightly_studio/src/lightly_studio/api/routes/video_frames_media.py
@@ -33,7 +33,7 @@ async def stream_frame(sample_id: UUID, session: SessionDep) -> StreamingRespons
     # Open video with OpenCV
     cap = cv2.VideoCapture(video_path)
     if not cap.isOpened():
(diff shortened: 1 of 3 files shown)


lightly_studio/src/lightly_studio/api/routes/video_frames_media.py
```diff
@@ -15,28 +18,66 @@
...
 
 
+@lru_cache(maxsize=512)
+def _get_video_metadata(video_path: str) -> dict[str, Any]:
+    probe = ffmpeg.probe(video_path)
+    return next(s for s in probe["streams"] if s["codec_type"] == "video")
+
+
+@lru_cache(maxsize=512)
+def _get_rotation(video_path: str) -> int:
```
horatiualmasan: I would not do this here, as we need to open the file again to read it's content. I would do it when we first load the videos to add them to the database and add this information to the Video Table

PR #176, opened 2025-11-20
Remove CaptionTable.dataset_id

michal-lightly:
## What has changed and why?

Remove CaptionTable.dataset_id.

Also removed `caption_resolver.get_all()` which was the main place dataset_id was used. We used the function only in tests. Replaced it with asserts on `sample.captions` relationship.

## How has it been tested?

Updated tests.

- [ ] Yes


horatiualmasan (approved):
LGTM

PR #177, opened 2025-11-20
Class balancing for an incomplete target

lukas-lightly:
## What has changed and why?
E.g. for a distribution, that should be 50% 'cat', the rest of the categories can be anything else.

## How has it been tested?
Added a test.

- [ ] Yes

### DIFF
--- a/lightly_studio/src/lightly_studio/selection/select_via_db.py
+++ b/lightly_studio/src/lightly_studio/selection/select_via_db.py
@@ -113,11 +113,8 @@ def _process_explicit_target_distribution(
 
     all_label_ids = {a.annotation_label_id for a in annotations}
     unused_label_ids = all_label_ids - set(label_id_to_target.keys())
-    unused_count = len(unused_label_ids)
-    if unused_count == 0 and total_targets + EPSILON < 1.0:
-        raise ValueError(f"Targets sum to less than 1.0 and all classes are used: {total_targets}")
-
-    # `total_targets` can be more than 1.1, in that case we are not filling in any gaps.
+    # `total_targets` can be more or less than 1.0. Both can be ignored, selection will still
+    # try correctly to reach the target.
     remaining_ratio = max(1.0 - total_targets, 0.0)
     return label_id_to_target, unused_label_ids, remaining_ratio
 


lightly_studio/src/lightly_studio/selection/select_via_db.py
```diff
@@ -70,6 +72,56 @@ def _aggregate_class_distributions(
...
+        label_id_to_target[annotation_label.annotation_label_id] = target
+        total_targets += target
+
+    if 1.0 - total_targets > EPSILON:
+        all_label_ids = {a.annotation_label_id for a in annotations}
+        unused_label_ids = all_label_ids - set(label_id_to_target.keys())
+
+        unused_count = len(unused_label_ids)
+        for label_id in unused_label_ids:
+            label_id_to_target[label_id] = (1.0 - total_targets) / unused_count
```
lukas-lightly: We use a uniform distribution for the remaining targets. "Input distribution" might be a better choice. I could do this in a followup PR, WDYT?

lightly_studio/src/lightly_studio/selection/select_via_db.py
```diff
@@ -62,20 +64,71 @@ def _aggregate_class_distributions(
...
+    Args:
+        session: The SQLAlchemy session.
+        target_distribution:
+            A dictionary mapping annotation label names to their target proportions.
+        annotations:
+            A sequence of all annotations to consider for class balancing.
+
+    Returns:
+        A dictionary mapping annotation label IDs to their effective target proportions, the set of
+        unused label IDs and the target value remaining to 1.0.
```
MalteEbner: Minor: The following style is nicer in my eyes
```suggestion
    Returns:
        Tuple of:
            A dictionary mapping annotation label IDs to their effective target proportions.
            The set of unused label IDs.
            The target value remaining to 1.0.
```

lightly_studio/src/lightly_studio/selection/select_via_db.py
```diff
@@ -62,20 +64,71 @@ def _aggregate_class_distributions(
...
+    if 1.0 - total_targets > EPSILON:
+        all_label_ids = {a.annotation_label_id for a in annotations}
+        unused_label_ids = all_label_ids - set(label_id_to_target.keys())
+
+        unused_count = len(unused_label_ids)
+        if unused_count == 0:
+            raise ValueError(
+                f"Targets sum to less than 1.0 and all classes are used: {total_targets}"
+            )
+        return label_id_to_target, unused_label_ids, 1.0 - total_targets
```
MalteEbner: Major: We must use the other class even if total_targets >= 1. In that case set the remainder to 0.0. This is necessary to enforce that nothing of the other classes is selected instead of not caring whether they are selected or not. 


I tried it out live on your branch and the coco dataset. With `target={"person": 0.9, "airplane": 0.1}`, a lot of non-person and non-airplane samples are selected. With `target={"person": 0.9, "airplane": 0.0999}` none of them are selected. However, the target passed to mundig should include `"other": 0.0` in both cases, not only the second one.

```suggestion
    all_label_ids = {a.annotation_label_id for a in annotations}
    unused_label_ids = all_label_ids - set(label_id_to_target.keys())

    unused_count = len(unused_label_ids)
    if unused_count == 0:
        raise ValueError(
            f"Targets sum to less than 1.0 and all classes are used: {total_targets}"
        )
    return label_id_to_target, unused_label_ids, max(1.0 - total_targets, 0.0)
```

lightly_studio/src/lightly_studio/selection/select_via_db.py
```diff
@@ -90,17 +143,23 @@ def _get_class_balancing_data(
...
+        label_id_to_target, unused_label_ids, remaining_ratio = (
+            _process_explicit_target_distribution(
+                session=session,
+                target_distribution=strat.target_distribution,
+                annotations=annotations,
+            )
+        )
+        if len(unused_label_ids) >= 1:
+            other_uuid = uuid4()
+            # Handle the case when the target values do not sum up to 1.0.
```
MalteEbner:
```suggestion
            # Handle the case when not all classes have a target.
```

lightly_studio/tests/selection/test_select_via_db.py
```diff
@@ -403,6 +403,84 @@ def test_select_via_database_with_annotation_class_balancing_missing_class(
...
+                    "cat": 0.5,
+                },
+            )
+        ],
+    )
+    select_via_database(
+        session=test_db,
+        config=config,
+        input_sample_ids=sample_ids,
+    )
```
MalteEbner, line 471: Can you add a `mocker.spy` on `mundig.add_class_balancing`, ensuring that the passed `target` and `class_distributions` are correct?

lightly_studio/tests/selection/test_select_via_db.py
```diff
@@ -403,6 +403,84 @@ def test_select_via_database_with_annotation_class_balancing_missing_class(
...
+        selection_result_tag_name="selection-tag",
+        strategies=[
+            AnnotationClassBalancingStrategy(
+                # This distribution gets translated to `cat: 0.5, dog: 0.25, bird: 0.25`
+                target_distribution={
+                    "cat": 0.5,
+                },
+            )
+        ],
+    )
```
MalteEbner, line 464: Can you add as subtest or another test using `target={"cat": 1.1 }`and ensure that the target passed to mundig is [1.1, 0.0], with 0.0 being the other classes?

lightly_studio/src/lightly_studio/selection/select_via_db.py
MalteEbner: Best also add a test for this, see https://github.com/lightly-ai/lightly-studio/pull/177/files#r2545975128

lightly_studio/tests/selection/test_select_via_db.py
lukas-lightly: Done

lightly_studio/tests/selection/test_select_via_db.py
lukas-lightly: Done, however, the numpy array was problematic, so I had to do a quite elaborate assert. Please check it.

lightly_studio/src/lightly_studio/selection/select_via_db.py
lukas-lightly: Done, also added a test.

lightly_studio/src/lightly_studio/selection/select_via_db.py
```diff
@@ -62,20 +64,70 @@ def _aggregate_class_distributions(
...
+        label_id_to_target[annotation_label.annotation_label_id] = target
+        total_targets += target
+
+    all_label_ids = {a.annotation_label_id for a in annotations}
+    unused_label_ids = all_label_ids - set(label_id_to_target.keys())
+    unused_count = len(unused_label_ids)
+    if unused_count == 0 and total_targets + EPSILON < 1.0:
+        raise ValueError(f"Targets sum to less than 1.0 and all classes are used: {total_targets}")
+
+    # `total_targets` can be more than 1.1, in that case we are not filling in any gaps.
```
MalteEbner:
```suggestion
    # `total_targets` can be more or less than 1.0. Both can be ignored, selection will still try correctly to reach the target.
```

MalteEbner (approved):
LGTM, thanks for implementing this so quickly!

lightly_studio/tests/selection/test_select_via_db.py
MalteEbner: Using np.testing is the correct way :)

PR #178, opened 2025-11-20
Add tests for input and uniform distributions

masakljun:
This PR adds direct unit tests for the private helper functions [_get_class_balancing_data()](https://github.com/lightly-ai/lightly-studio/blob/0a06dd43df834db2c5a131a60e5fbae05a306384/lightly_studio/src/lightly_studio/selection/select_via_db.py#L73) and [_aggregate_class_distributions ](https://github.com/lightly-ai/lightly-studio/blob/0a06dd43df834db2c5a131a60e5fbae05a306384/lightly_studio/src/lightly_studio/selection/select_via_db.py#L35) located in `select_via_db.py`. Previously, these functions were only covered implicitly by larger integration tests. This change ensures the logic for calculating target values and aggregating sample distributions is verified in isolation.

Key changes:
- Added `test_aggregate_class_distributions`: Verifies that annotations are correctly counted and mapped to samples to create the distribution matrix.

- Added `test_get_class_balancing_data` variants:

1. `_input`: Ensures the target distribution correctly mirrors the input sample counts.

2. `_uniform`: Verifies that target values are calculated as a perfectly even split.

3. `_target`: Confirms that custom dictionary distributions are preserved and ordered correctly.

I performed a self-review using Codex and I have a note on imports: Codex suggested refactoring the direct imports of private functions (_get...). However, to maintain consistency with the existing testing style in test_select_via_db.py, I have kept the direct imports as is.

### DIFF
--- a/lightly_studio/tests/selection/test_select_via_db.py
+++ b/lightly_studio/tests/selection/test_select_via_db.py
@@ -690,7 +690,7 @@ def test_aggregate_class_distributions() -> None:
     """Tests that annotation counting works correctly."""
     label_id_a = uuid4()
     label_id_b = uuid4()
-    label_id_c = uuid4()
+    label_id_c = uuid4()  # This label will be ignored in targets
 
     sample_id_1 = uuid4()
     sample_id_2 = uuid4()
@@ -701,8 +701,8 @@ def test_aggregate_class_distributions() -> None:
 
     sample_id_to_annotation_label_ids = {
         sample_id_1: [label_id_a, label_id_b],
-        sample_id_2: [label_id_a, label_id_c],
-        sample_id_3: [label_id_b, label_id_b],
+        sample_id_2: [label_id_a, label_id_c],  # C should be ignored
+        sample_id_3: [label_id_b, label_id_c],  # C should be ignored
     }
 
     class_distributions = _aggregate_class_distributions(
@@ -711,11 +711,15 @@ def test_aggregate_class_distributions() -> None:
         target_annotation_ids=target_annotation_ids,
     )
 
+    # Columns correspond to [label_id_a, label_id_b]
+    # Row 0 (Sample 1): 1 A, 1 B
+    # Row 1 (Sample 2): 1 A, 0 B (C is ignored)
+    # Row 2 (Sample 3): 0 A, 1 B (C is ignored)
     expected_distributions = np.array(
         [
             [1.0, 1.0],
             [1.0, 0.0],
-            [0.0, 2.0],
+            [0.0, 1.0],
         ],
         dtype=np.float32,
     )
@@ -749,6 +753,9 @@ def test_get_class_balancing_data_input(test_db: Session) -> None:
         annotation_type=AnnotationType.CLASSIFICATION,
     )
 
+    # The order of target keys depends on the insertion order in this list.
+    # 'cat' appears first, 'dog' appears second.
+    # Target Keys: [cat, dog]
     all_annotations = [ann_cat_1, ann_cat_2, ann_dog_1]
     input_sample_ids = [sample_id_1, sample_id_2]
 
@@ -770,6 +777,9 @@ def test_get_class_balancing_data_input(test_db: Session) -> None:
     expected_vals = [2, 1]
     assert target_vals == expected_vals
 
+    # Columns: [Cat, Dog]
+    # Row 0 (Sample 1): 1 Cat, 0 Dog
+    # Row 1 (Sample 2): 1 Cat, 1 Dog
     expected_dist = np.array([[1.0, 0.0], [1.0, 1.0]])
     np.testing.assert_array_equal(class_dist, expected_dist)
 


lightly_studio/tests/selection/test_select_via_db.py
```diff
@@ -537,6 +543,258 @@ def test_select_via_database_with_annotation_class_balancing_input(
...
+    assert len(target_vals) == 2
+    assert target_vals == pytest.approx([0.5, 0.5], abs=1e-9)
+
+    expected_col_cat = np.array([1.0, 1.0])
+    expected_col_dog = np.array([0.0, 1.0])
+
+    col_1 = class_dist[:, 0]
+    col_2 = class_dist[:, 1]
+
+    is_order_cat_dog = np.array_equal(col_1, expected_col_cat) and np.array_equal(
```
lukas-lightly: Did you experience unstable order of columns? In such a case we might have a bug.

lightly_studio/tests/selection/test_select_via_db.py
```diff
@@ -537,6 +543,258 @@ def test_select_via_database_with_annotation_class_balancing_input(
...
+    sample_id_2 = uuid4()
+    sample_id_3 = uuid4()
+
+    dataset_id = uuid4()
+
+    target_annotation_ids = [label_id_a, label_id_b]
+    input_sample_ids = [sample_id_1, sample_id_2, sample_id_3]
+
+    sample_id_to_annotations = {
+        sample_id_1: [
```
lukas-lightly: There's a PR in review which simplifies this greatly, this parameter becomes `Mapping[UUID, list[UUID]]`  (from `Mapping[UUID, Sequence[AnnotationBaseTable]]`):
https://github.com/lightly-ai/lightly-studio/pull/177/files#diff-ead969226381bb9d6986bafd70abd2d560d59a3bbe97a8a2b59a18b01eb7b5cf

It will greatly simplify the test setup here, you won't need to create the `AnnotationBaseTable()` object which takes up a lot of lines.

Let's merge PR #177 first and then these tests become much shorter.

lightly_studio/tests/selection/test_select_via_db.py
masakljun: No, I added that extra check because Codex said a possible issue during my self review. It mentioned that `select_via_db.py` turns a set into a list without sorting (which could in theory produce a different order each run). After running the test a few times on my machine, the order seems to be the same every time. Since it looks stable in practice, I now removed the extra logic and switched to checking the fixed order now.

lightly_studio/tests/selection/test_select_via_db.py
```diff
@@ -681,6 +686,210 @@ def test_select_via_database_with_annotation_class_balancing_input(
...
+    sample_id_2 = uuid4()
+    sample_id_3 = uuid4()
+
+    target_annotation_ids = [label_id_a, label_id_b]
+    input_sample_ids = [sample_id_1, sample_id_2, sample_id_3]
+
+    sample_id_to_annotation_label_ids = {
+        sample_id_1: [label_id_a, label_id_b],
+        sample_id_2: [label_id_a, label_id_c],
+        sample_id_3: [label_id_b, label_id_b],
```
lukas-lightly: Label B twice for one sample seems wrong. I think you meant C.

lightly_studio/tests/selection/test_select_via_db.py
```diff
@@ -681,6 +686,210 @@ def test_select_via_database_with_annotation_class_balancing_input(
...
+
+    expected_distributions = np.array(
+        [
+            [1.0, 1.0],
+            [1.0, 0.0],
+            [0.0, 2.0],
+        ],
+        dtype=np.float32,
+    )
+    np.testing.assert_array_equal(class_distributions, expected_distributions)
```
ikondrat, line 726: Can you pIease add explanation why expected_distributions is correct in this case?

lightly_studio/tests/selection/test_select_via_db.py
```diff
@@ -681,6 +686,210 @@ def test_select_via_database_with_annotation_class_balancing_input(
...
+        strat=strat,
+        annotations=all_annotations,
+        input_sample_ids=input_sample_ids,
+        sample_id_to_annotation_label_ids=sample_id_to_annotation_label_ids,
+    )
+
+    expected_vals = [2, 1]
+    assert target_vals == expected_vals
+
+    expected_dist = np.array([[1.0, 0.0], [1.0, 1.0]])
```
ikondrat, line 783: I would here also add few notes why we expecting 
`[1.0, 0.0], [1.0, 1.0]`

lightly_studio/tests/selection/test_select_via_db.py
masakljun: Added

lightly_studio/tests/selection/test_select_via_db.py
masakljun: Added

lightly_studio/tests/selection/test_select_via_db.py
masakljun: Yes, correct, thanks for catching. I've updated now

PR #185, opened 2025-11-21
Add example script to load video annotations

michal-lightly:
## What has changed and why?

Add example script to load video annotations. The logic is manual and slightly hacky, until video annotation loading is supported in labelformat.

## How has it been tested?

Tested with a draft branch to display the annotations.

- [ ] Yes

### DIFF
--- a/lightly_studio/.vscode/launch.json
+++ b/lightly_studio/.vscode/launch.json
@@ -49,6 +49,15 @@
             "console": "integratedTerminal"
         },
         {
+            "name": "Start: 'Video' example",
+            "type": "debugpy",
+            "request": "launch",
+            "program": "${workspaceFolder}/e2e-tests/index_with_video.py",
+            "python": "${workspaceFolder}/.venv/bin/python",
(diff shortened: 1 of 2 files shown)


lightly_studio/src/lightly_studio/core/add_samples.py
```diff
@@ -462,8 +463,8 @@ def _process_instance_segmentation_annotations(
 
 def _process_batch_annotations(
     session: Session,
-    created_path_to_id: dict[str, UUID],
-    path_to_anno_data: dict[str, ImageInstanceSegmentation | ImageObjectDetection],
+    created_path_to_id: Mapping[str, UUID],
+    path_to_anno_data: Mapping[str, ImageInstanceSegmentation | ImageObjectDetection],
```
michal-lightly, line 467: This makes mypy happy, it is possible to pass in `dict[str, ImageObjectDetection]` instead of `dict[str, ImageInstanceSegmentation | ImageObjectDetection]`. That's because `dict` is invariant, but `Mapping` is covariant.

PR #190, opened 2025-11-24
Add similarity API endpoint

lukas-lightly:
## What has changed and why?
I've refactored `compute_similarity_metadata()` to get query tag by name and not by ID, so that `compute_similarity_metadata()` can create the metadata name from the query name if it's not provided.

I've also introduce TagNotFoundError exception.

## How has it been tested?
Added a test.

- [ ] Yes

### DIFF
--- a/lightly_studio/tests/metadata/test_compute_similarity.py
+++ b/lightly_studio/tests/metadata/test_compute_similarity.py
@@ -37,9 +37,10 @@ def test_compute_similarity_metadata(test_db: Session) -> None:
     )
 
     query_tag = create_tag(session=test_db, dataset_id=dataset_id, tag_name="query_tag")
+    query_tag_id = query_tag.tag_id
     tag_resolver.add_sample_ids_to_tag_id(
         session=test_db,
-        tag_id=query_tag.tag_id,
+        tag_id=query_tag_id,
         sample_ids=[samples[0].sample_id, samples[2].sample_id],
     )
 
@@ -47,7 +48,7 @@ def test_compute_similarity_metadata(test_db: Session) -> None:
         session=test_db,
         key_dataset_id=dataset_id,
         embedding_model_id=embedding_model_id,
-        query_tag_id=query_tag.tag_id,
+        query_tag_id=query_tag_id,
         metadata_name="similarity",
     )
 


lightly_studio/src/lightly_studio/metadata/compute_similarity.py
```diff
@@ -31,26 +34,38 @@ def compute_similarity_metadata(
             The ID of the dataset the similarity is computed on.
         embedding_model_id:
             The ID of the embedding model to use for the computation.
-        query_tag_id:
-            The ID of the tag describing the query.
+        query_tag_name:
```
JonasWurst: I would stick to query_tag_id here.

see proposal below

lightly_studio/src/lightly_studio/metadata/compute_similarity.py
```diff
@@ -31,26 +34,38 @@ def compute_similarity_metadata(
...
             "LIGHTLY_STUDIO_LICENSE_KEY environment variable is not set. "
             "Please set it to your LightlyStudio license key."
         )
+    query_tag = tag_resolver.get_by_name(
+        session=session,
+        tag_name=query_tag_name,
+        dataset_id=key_dataset_id,
+    )
+    if query_tag is None:
+        raise TagNotFoundError(f"Query tag {query_tag_name} not found")
```
JonasWurst: If we switch back to id, we can get the tag via `tag_resolver.get_by_id`, this should also carry the name.

We should still introduce the error, but using the is ofc.

lightly_studio/src/lightly_studio/metadata/compute_similarity.py
```diff
@@ -59,10 +74,14 @@ def compute_similarity_metadata(
     )
     query_embeddings = [sample.embedding for sample in query_samples]
     similarity_values = similarity.calculate_similarity(query_embeddings=query_embeddings)
+    if metadata_name is None:
```
JonasWurst, line 69: Here we can use the retrieved tag_name

lightly_studio/tests/metadata/test_compute_similarity.py
```diff
@@ -37,18 +37,17 @@ def test_compute_similarity_metadata(test_db: Session) -> None:
...
     )
 
     compute_similarity.compute_similarity_metadata(
         session=test_db,
         key_dataset_id=dataset_id,
         embedding_model_id=embedding_model_id,
-        query_tag_id=query_tag_id,
+        query_tag_name="query_tag",
         metadata_name="similarity",
     )
```
JonasWurst, line 53: we should add a test for the meta data name

lightly_studio/src/lightly_studio/api/routes/api/metadata.py
```diff
@@ -89,3 +91,73 @@ def compute_typicality_metadata(
...
+
+    Raises:
+        HTTPException: 404 if invalid embedding model or query tag is given.
+    """
+    try:
+        embedding_model = embedding_model_resolver.get_by_name(
+            session=session,
+            dataset_id=dataset.dataset_id,
+            embedding_model_name=request.embedding_model_name,
+        )
```
JonasWurst, line 142: I think this could also be done inside the compute_similarity_metadata.

At least it is not obvious why we get this id here, while the tag id inside.

lightly_studio/tests/api/routes/api/test_metadata.py
```diff
@@ -90,3 +90,40 @@ def test_compute_typicality_metadata(test_client: TestClient, db_session: Sessio
...
+        session=db_session, dataset_id=dataset_id
+    ).samples
+    tag_resolver.add_sample_ids_to_tag_id(
+        session=db_session,
+        tag_id=query_tag.tag_id,
+        sample_ids=[samples[0].sample_id, samples[2].sample_id],
+    )
+
+    response = test_client.post(
+        f"/api/datasets/{dataset_id}/metadata/similarity", json={"query_tag_name": "query_tag"}
```
JonasWurst: What is the embedding_model_name used for, if not required. What is its default?

lightly_studio/tests/api/routes/api/test_metadata.py
```diff
@@ -90,3 +90,40 @@ def test_compute_typicality_metadata(test_client: TestClient, db_session: Sessio
...
+    )
+
+    assert response.status_code == 200
+    metadata_name = response.text[1:-1]  # We strip the double-quotes
+    assert metadata_name.startswith("similarity_query_tag_20")
+
+    samples = image_resolver.get_all_by_dataset_id(
+        session=db_session, dataset_id=dataset_id
+    ).samples
+    assert len(samples) == 10
```
JonasWurst: why do we need this?

lightly_studio/tests/metadata/test_compute_similarity.py
JonasWurst: also for throwing the exception.

lightly_studio/tests/api/routes/api/test_metadata.py
```diff
@@ -90,3 +90,40 @@ def test_compute_typicality_metadata(test_client: TestClient, db_session: Sessio
         )
         assert typicality_value is not None
         assert isinstance(typicality_value, float)
+
+
+def test_compute_similarity_metadata(test_client: TestClient, db_session: Session) -> None:
```
JonasWurst, line 93: I miss a test for the two exceptions that can raise.

JonasWurst (changes requested):
Some tests are missing. Also I would rather get all ids in the route, or get all of them in the similarity method.

lightly_studio/src/lightly_studio/api/routes/api/metadata.py
lukas-lightly: It's because of the `raise HTTPException(...)`. You'd have to translate the internal exception, just like we do below for `TagNotFoundError`.

lightly_studio/src/lightly_studio/metadata/compute_similarity.py
lukas-lightly: Done

lightly_studio/src/lightly_studio/metadata/compute_similarity.py
lukas-lightly: Done

lightly_studio/src/lightly_studio/metadata/compute_similarity.py
lukas-lightly: Done

lightly_studio/tests/api/routes/api/test_metadata.py
lukas-lightly: It's present in other tests too, so it was copied, but you are right, it's not needed.

lightly_studio/tests/api/routes/api/test_metadata.py
lukas-lightly: You need an embedding to calculate similarity, typicality, etc. If you don't specify it, the only one model is selected. See the docs:

```
def get_by_name(
    session: Session, dataset_id: UUID, embedding_model_name: str | None
) -> EmbeddingModelTable:
    """Helper function to resolve the embedding model name to its ID.

    Args:
        session: The database session.
        dataset_id: The ID of the dataset.
        embedding_model_name: The name of the embedding model.
            If None, expects the dataset to have exactly one embedding model and
            returns it. Otherwise raises a ValueError.
            If set, expects the dataset to have an embedding model with the given name.
            Otherwise raises a ValueError.
```

lightly_studio/tests/metadata/test_compute_similarity.py
JonasWurst, line 1: This change can be complete removed

lightly_studio/src/lightly_studio/api/routes/api/metadata.py
JonasWurst: Yes, I meant to have another custom Exception for the embedding model not found. We can keep it like this for now.

lightly_studio/tests/api/routes/api/test_metadata.py
lukas-lightly: I'll do this in a followup

lightly_studio/tests/metadata/test_compute_similarity.py
lukas-lightly: Done

lightly_studio/tests/metadata/test_compute_similarity.py
lukas-lightly: I'll do this in a followup

PR #192, opened 2025-11-24
Fix video frame orientation

michal-lightly:
## What has changed and why?

Fix frame orientation issue.

We use two libraries to load videos:
- PyAV when indexing the video
- OpenCV when serving a single frame (due to its speed)

However, OpenCV does not read video metadata properly and can return frames in wrong orientation

**Solution**

During indexing, use PyAV to get frame orientation. Note that PyAV only makes this metadata available on the frame level, not on the video level.

**Considered options**

`ffmpeg-python` exposes `ffmpeg.probe()` function that returns the metadata. However, it calls `ffmpeg` in a subprocess and thus is not suitable for use with cloud storage.

## How has it been tested?

Loaded a video dataset that was previously displayed wrong.

- [ ] Yes


lightly_studio_view/src/lib/components/VideoFrameItem/VideoFrameItem.svelte
```diff
@@ -25,6 +25,9 @@
 
 <style>
     img {
+        object-fit: cover;
+        background-color: black;
```
michal-lightly: @LeonardoRosaa This change is from your PR, should I keep it?

PR #200, opened 2025-11-25
Fixed label disappearing after a few clicks

LeonardoRosaa:
## What has changed and why?

Fixed the issue with the label disappearing after a few clicks. The main change is switching to use the correct dataset (often the root dataset). 

## How has it been tested?

Add tags for images and annotations, and then refresh the page within the annotations view. Confirm that the annotation tags are displayed. Then, navigate to the images view and confirm that the tags are also displayed.

- [ ] Yes

### DIFF
--- a/lightly_studio/src/lightly_studio/api/routes/api/metadata.py
+++ b/lightly_studio/src/lightly_studio/api/routes/api/metadata.py
@@ -5,13 +5,15 @@ from __future__ import annotations
 from typing import List
 from uuid import UUID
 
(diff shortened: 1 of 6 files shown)


ikondrat:
I also noticed perf degradation comparing to previous release
Screenshot 2025-11-26 at 08 50 05
Screenshot 2025-11-26 at 08 45 39


I think it is fine to merge it with fix and release but please create follow up tasks for mentioned issues

PR #204, opened 2025-11-26
Navigate to ImageDetail from CaptionGrid

michal-lightly:
## What has changed and why?

Navigate to ImageDetail from CaptionGrid.

This is **not** possible to implement currently. Navigation to ImageDetail requires to know `sampleIndex` to make the prev/next buttons work properly. That's not possible - we need to change the prev/next logic not to rely on `sampleIndex`.

## How has it been tested?

- [ ] Yes

### DIFF
--- a/lightly_studio_view/src/lib/components/Captions/Captions.svelte
+++ b/lightly_studio_view/src/lib/components/Captions/Captions.svelte
@@ -88,6 +88,8 @@
                         <CaptionsItem
                             maxHeight={`${captionSize}px`}
                             item={items[index]}
+                            datasetId={datasetId}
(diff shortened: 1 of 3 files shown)


lightly_studio_view/src/lib/components/Captions/Captions.svelte
```diff
@@ -88,6 +88,8 @@
                         <CaptionsItem
                             maxHeight={`${captionSize}px`}
                             item={items[index]}
+                            datasetId={datasetId}
+                            sampleIndex={index}
```
michal-lightly, line 92: This is not the correct index.

PR #206, opened 2025-11-26
Add a function to convert captions dataset to coco dict

michal-lightly:
## What has changed and why?

Add a function to convert captions dataset to coco dict.

Follow-up: Use the function to create an export file.

## How has it been tested?

New test.

- [ ] Yes

### DIFF
--- a/lightly_studio/src/lightly_studio/api/routes/api/frame.py
+++ b/lightly_studio/src/lightly_studio/api/routes/api/frame.py
@@ -93,7 +93,7 @@ def _build_annotation_view(a: AnnotationBaseTable) -> AnnotationView:
     return AnnotationView(
         parent_sample_id=a.parent_sample_id,
         dataset_id=a.dataset_id,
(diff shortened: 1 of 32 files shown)


lightly_studio/src/lightly_studio/export/coco_captions.py
```diff
@@ -0,0 +1,69 @@
...
+
+    Args:
+        samples: The samples to export.
+
+    Returns:
+        A dictionary in COCO captions format.
+    """
+    coco_images: list[CocoCaptionImage] = []
+    coco_annotations: list[CocoCaptionAnnotation] = []
+    annotation_id = 1
```
horatiualmasan: Annotations start from 1 and images from 0. Should we be consistent or is it this way in the coco captions format?

michal-lightly: I'll change it to 0, good catch

PR #208, opened 2025-11-27
Use dialog for export. SidePanel removal

JonasWurst:
## What has changed and why?

To prepare the introduction of the menu, the export shall be handled in a dialog instead of a sidebar.

## How has it been tested?

locally

- [ ] Yes

### DIFF
--- a/.github/workflows/end2end_test.yml
+++ b/.github/workflows/end2end_test.yml
@@ -18,10 +18,13 @@ env:
   E2E_BASE_URL: http://localhost:8001
   LIGHTLY_STUDIO_LICENSE_KEY: ${{ secrets.MUNDIG_LICENSE_KEY }}
 
+
 jobs:
-  end-to-end-test:
-    name: End2End Test
+  prepare-caches:
+    name: Build Cache
     runs-on: ubuntu-latest
+    outputs:
+      cache-hit: ${{ steps.cache-build.outputs.cache-hit }}
     steps:
       - name: Checkout Code
         uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


lightly_studio_view/src/lib/components/ExportSamples/ExportSamples.svelte
mihnea-necsulescu, line 80: it would be a good idea to also handle the error case in this PR

lightly_studio_view/src/lib/components/ExportSamples/ExportSamples.svelte
```diff
@@ -100,93 +103,151 @@
...
+                                >
+                                    <Select.Trigger class="w-full">
+                                        {triggerContent}
+                                    </Select.Trigger>
+                                    <Select.Content>
+                                        <Select.Group>
+                                            <Select.GroupHeading
+                                                >Annotation tags</Select.GroupHeading
+                                            >
+                                            {#if $tags.filter((tag) => tag.kind == 'annotation').length === 0}
```
mihnea-necsulescu:
```suggestion
                                            {#if $tags.filter((tag) => tag.kind === 'annotation').length === 0}
```

lightly_studio_view/src/lib/components/ExportSamples/ExportSamples.svelte
```diff
@@ -100,93 +103,151 @@
...
+                                                <div
+                                                    class="py-1.5 pl-8 pr-2 text-sm italic text-muted-foreground"
+                                                >
+                                                    no tags available
+                                                </div>
+                                            {/if}
+                                            {#each $tags.filter((tag) => tag.kind == 'annotation') as annotationTag}
```
mihnea-necsulescu:
```suggestion
                                            {#each $tags.filter((tag) => tag.kind === 'annotation') as annotationTag}
```

lightly_studio_view/src/lib/components/ExportSamples/ExportSamples.svelte
```diff
@@ -100,93 +103,151 @@
...
+                                                <Select.Item
+                                                    value={annotationTag.tag_id}
+                                                    label={annotationTag.name}
+                                                    >{annotationTag.name}</Select.Item
+                                                >
+                                            {/each}
+                                            <Select.GroupHeading>Sample tags</Select.GroupHeading>
+                                            {#if $tags.filter((tag) => tag.kind == 'sample').length === 0}
```
mihnea-necsulescu:
```suggestion
                                            {#if $tags.filter((tag) => tag.kind === 'sample').length === 0}
```

lightly_studio_view/src/lib/components/ExportSamples/ExportSamples.svelte
```diff
@@ -100,93 +103,151 @@
...
+                                                <div
+                                                    class="py-1.5 pl-8 pr-2 text-sm italic text-muted-foreground"
+                                                >
+                                                    no tags available
+                                                </div>
+                                            {/if}
+                                            {#each $tags.filter((tag) => tag.kind == 'sample') as sampleTag}
```
mihnea-necsulescu:
```suggestion
                                            {#each $tags.filter((tag) => tag.kind === 'sample') as sampleTag}
```

lightly_studio_view/src/lib/components/ExportSamples/ExportSamples.svelte
mihnea-necsulescu: let's also fix this typo 

```suggestion
            "Select a tag to export its samples (required)"
```

lightly_studio_view/src/lib/components/ExportSamples/ExportSamples.svelte
```diff
@@ -100,93 +103,151 @@
...
+                        class="border text-foreground"
+                        data-testid={errorMessage ? 'alert-destructive' : 'alert-success'}
+                    >
+                        <div class="flex items-center gap-2">
+                            <span class="text-destructive-foreground">{errorMessage}</span>
+                        </div>
+                    </Alert.Root>
+                </div>
+            {/if}
+            <Dialog.Footer>
```
mihnea-necsulescu, line 225: should we also add a cancel button here for consistency (e.g., the selection dialog has one)?

lightly_studio_view/src/lib/components/ExportSamples/ExportSamples.svelte
```diff
@@ -100,93 +103,151 @@
...
+                    {/if}
                 </div>
             </div>
-        {/if}
-    </div>
-</SidePanel>
+            {#if errorMessage}
+                <div class="my-4" transition:fade>
+                    <Alert.Root
+                        variant={errorMessage ? 'destructive' : 'default'}
```
mihnea-necsulescu: we're already inside `{#if errorMessage}`, the ternary operator is not needed

```suggestion
                        variant="destructive"
```

lightly_studio_view/src/lib/components/ExportSamples/ExportSamples.svelte
mihnea-necsulescu: we could extract a derived `annotationTags` and use that here instead of running the same filter twice

we could declare 
```ts
const annotationTags = $derived($tags.filter((tag) => tag.kind === 'annotation'));
```
in the <script> section and just use that here


```suggestion
                                            {#if annotationTags.length === 0}
                                                <div
                                                    class="py-1.5 pl-8 pr-2 text-sm italic text-muted-foreground"
                                                >
                                                    no tags available
                                                </div>
                                            {/if}
                                            {#each annotationTags as annotationTag}
```

lightly_studio_view/src/lib/components/ExportSamples/ExportSamples.svelte
mihnea-necsulescu: same thing here

lightly_studio_view/src/lib/components/SidePanel/SidePanel.svelte
```diff
@@ -1,112 +0,0 @@
...
-        cancelLabel = 'Cancel',
-        errorMessage = '',
-        successMessage = '',
-        isOpen = false,
-        isLoading = false,
-        isDisabled = false,
-        onSubmit,
-        onOpenChange,
-        children,
-        exportURL
```
mihnea-necsulescu, line 21: I suggest we only remove the export coupling (all exportURL-related occurrences) from this component, but still keep it - we may need it later, who knows

wdyt?

ikondrat: let's remove if it is not used, we can always get it back from repo

lightly_studio_view/src/lib/components/ExportSamples/ExportSamples.svelte
JonasWurst: I think, the close button of a dialog is enough.

lightly_studio_view/src/lib/components/ExportSamples/ExportSamples.svelte
JonasWurst: Discussed with kondrat. Not blocking, will keep it as it was before

lightly_studio_view/src/lib/components/ExportSamples/ExportSamples.svelte
JonasWurst: see comment above

PR #212, opened 2025-11-27
example to reproduce issue with hitting limit of pool

ikondrat:
## What has changed and why?

- Fetch coco-10k dataset (you can get it from [PR](https://github.com/lightly-ai/dataset_examples/pull/10))
- Run `make start-coco-10k` and set the limit of images per row at least 7
- Scroll
- It would loading stop on some step  (check my video)

In the logs you can see
```
INFO:     ::1:57872 - "GET /api/datasets/6b2ec36a-2498-48b4-91c2-9fbf4eff6857/metadata/info HTTP/1.1" 200 OK
ERROR:asyncio:socket.accept() out of system resource
socket: <asyncio.TransportSocket fd=19, family=AddressFamily.AF_INET6, type=SocketKind.SOCK_STREAM, proto=6, laddr=('::1', 8001, 0, 0)>
Traceback (most recent call last):
  File "/Users/kondrat/.local/share/uv/python/cpython-3.8.20-macos-aarch64-none/lib/python3.8/asyncio/selector_events.py", line 164, in _accept_connection
  File "/Users/kondrat/.local/share/uv/python/cpython-3.8.20-macos-aarch64-none/lib/python3.8/socket.py", line 292, in accept
OSError: [Errno 24] Too many open files

```
or 
```
sqlalchemy.exc.TimeoutError: QueuePool limit of size 5 overflow 10 reached,
```

I think it depends on where it hits the limit

UI wise it looks like:
- hitting the limit of pool connections of sql alchemy - will look like non-listing all of the items in the Grid
- Hitting the TCP connections limit (256 default for macos) looks like freesing interface because fastapi starts to timeout requests

### DIFF
--- a/lightly_studio/Makefile
+++ b/lightly_studio/Makefile
@@ -40,6 +40,11 @@ start-e2e: build
 	@echo "Starting server for e2e tests..."
 	uv run e2e-tests/index_dataset_for_end2end_ui_tests.py
 
+.PHONY: start-coco-10k
+start-coco-10k: build
+	@echo "Starting server for e2e tests with COCO 10k dataset..."
+	uv run e2e-tests/index_coco_10k.py
+
 .PHONY: start-e2e-with-captions
 start-e2e-with-captions: build
 	@echo "Starting server for e2e tests with captions..."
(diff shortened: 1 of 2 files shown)


IgorSusmelj:
Can we close this issue?

PR #213, opened 2025-11-27
Add getter/setter methods for captions

lukas-lightly:
## What has changed and why?
We want the user to manipulate sample captions easily.

## How has it been tested?
Added tests.

- [x] Yes

### DIFF
--- a/lightly_studio/tests/core/test_sample.py
+++ b/lightly_studio/tests/core/test_sample.py
@@ -203,7 +203,6 @@ class TestSample:
         sample = Sample(inner=image_table)
 
         # Test adding a caption.
-        assert sample.captions == []
         sample.add_caption("caption3")
         assert sample.captions == ["caption3"]
         sample.add_caption("caption2")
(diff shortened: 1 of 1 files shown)


lightly_studio/tests/core/test_sample.py
```diff
@@ -193,3 +193,35 @@ def test_metadata__schema_must_match(self, test_db: Session) -> None:
...
+        # Test adding a caption.
+        assert sample.captions == []
+        sample.add_caption("caption3")
+        assert sample.captions == ["caption3"]
+        sample.add_caption("caption2")
+        assert sample.captions == ["caption3", "caption2"]
+        sample.add_caption("caption1")
+        assert sample.captions == ["caption3", "caption2", "caption1"]
+
+    def test_captions_setter(self, test_db: Session) -> None:
```
michal-lightly, line 213: The test names are swapped - the other tests the setter, this one tests add_caption

lukas-lightly: No, setter was tested in the right test, but the test also used `add_caption`. I've changed it to only use the setter.

In both cases we need to use the getter, hence there's no special test for the getter, both tests use it.

PR #214, opened 2025-11-27
feat: fix for issues related with hitting limit of sqlalchemy pool + TCP connections for 

ikondrat:
## What has changed and why?

This is PR fixes 2 issues because they are both about hitting limits. Depending on system state and it is one of them:
- Hitting the limit or parallel TCP connections 
`OSError: [Errno 24] Too many open files`
- Hitting the limit of sqlalchemy pool connections
`sqlalchemy.exc.TimeoutError: QueuePool limit of size 5 overflow 10 reached, `

Fix for this is:
- increase pool size for sqlalchemy to 60
- define concurrency parameters for uvicorn server, with defining 

## How has it been tested?

Check https://github.com/lightly-ai/lightly-studio/pull/212 to reproduce the issue 

and follow the same steps on this PR, the issue should not be reproduced anymore

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Added reading and updating of captions to the `Sample` class.
+- Added export functionality for image datasets with captions to the python `Dataset` class interface.
 
 ### Changed
 
--- a/lightly_studio/src/lightly_studio/export/export_dataset.py
+++ b/lightly_studio/src/lightly_studio/export/export_dataset.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import json
 from pathlib import Path
 from typing import Iterable
 
@@ -9,6 +10,7 @@ from labelformat.formats import COCOObjectDetectionOutput
 from sqlmodel import Session
 
 from lightly_studio.core.sample import Sample
+from lightly_studio.export import coco_captions
 from lightly_studio.export.lightly_studio_label_input import LightlyStudioObjectDetectionInput
 from lightly_studio.type_definitions import PathLike
 
@@ -50,6 +52,17 @@ class DatasetExport:
             output_json=Path(output_json),
         )
 
+    def to_coco_captions(self, output_json: PathLike | None = None) -> None:
+        """Exports captions to a COCO format JSON file.
+
+        Args:
+            output_json: The path to the output COCO JSON file. If not provided,
+                defaults to "coco_export.json" in the current working directory.
+        """
+        if output_json is None:
+            output_json = DEFAULT_EXPORT_FILENAME
+        to_coco_captions(samples=self.samples, output_json=Path(output_json))
+
 
 def to_coco_object_detections(
     session: Session,
@@ -71,3 +84,21 @@ def to_coco_object_detections(
         samples=samples,
     )
     COCOObjectDetectionOutput(output_file=output_json).save(label_input=export_input)
+
+
+def to_coco_captions(
+    samples: Iterable[Sample],
+    output_json: Path,
+) -> None:
+    """Exports captions to a COCO format JSON file.
+
+    This function is for internal use. Use `Dataset.query().export().to_coco_captions()`
+    instead.
+
+    Args:
+        samples: The samples to export.
+        output_json: The path to save the output JSON file.
+    """
+    coco_captions_dict = coco_captions.to_coco_captions_dict(samples=samples)
+    with output_json.open("w") as f:
+        json.dump(coco_captions_dict, f, indent=2)
--- a/lightly_studio/tests/export/test_coco_captions.py
+++ b/lightly_studio/tests/export/test_coco_captions.py
@@ -54,3 +54,16 @@ def test_to_coco_caption_dict(
             {"id": 1, "image_id": 1, "caption": "caption two"},
         ],
     }
+
+
+def test_to_coco_caption_dict__empty(
+    db_session: Session,
+) -> None:
+    """Tests conversion to COCO captions format."""
+    dataset = create_dataset(session=db_session)
+
+    # Call the function under test
+    samples = DatasetQuery(dataset=dataset, session=db_session)
+    coco_dict = coco_captions.to_coco_captions_dict(samples=samples)
+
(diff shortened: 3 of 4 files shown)


lightly_studio/e2e-tests/index_coco_10k.py
```diff
@@ -0,0 +1,25 @@
+"""End-to-end demonstration of the lightly_studio dataset loading and UI.
```
michal-lightly: Preferably let's not add this file. It is not documented how to get the coco10k dataset, and users or other developers won't have it locally - this looks like a helper script for one user.

You can achieve the same by setting a coco dataset in your `.env` file and calling:
```
uv run src/lightly_studio/examples/example_coco.py
```

lightly_studio/src/lightly_studio/db_manager.py
```diff
@@ -41,7 +41,14 @@ def __init__(
         self._engine_url = engine_url if engine_url else "duckdb:///lightly_studio.db"
         if cleanup_existing:
             _cleanup_database_file(engine_url=self._engine_url)
-        self._engine = create_engine(url=self._engine_url, poolclass=poolclass)
+        # Total available connections = pool_size + max_overflow = 20 + 40 = 60
+        # (default is (pool_size=5, max_overflow=10))
+        self._engine = create_engine(
+            url=self._engine_url,
+            poolclass=poolclass,
+            pool_size=20,
```
IgorSusmelj: Let's do pool size of 10. (default of duckdb is 10. And 5 was default in SQLAlchemy)

lightly_studio/src/lightly_studio/api/server.py
```diff
@@ -30,8 +30,19 @@ def __init__(self, host: str, port: int) -> None:
...
-        # start the app
-        uvicorn.run(app, host=self.host, port=self.port, http="h11")
+        # start the app with connection limits and timeouts
+        uvicorn.run(
+            app,
+            host=self.host,
+            port=self.port,
+            http="h11",
+            # https://uvicorn.dev/settings/#resource-limits
+            limit_concurrency=1000,  # Max concurrent connections
```
IgorSusmelj: Let's do 100 if default before was 40

ikondrat: updated

lightly_studio/src/lightly_studio/db_manager.py
ikondrat: updated with 10

lightly_studio_view/src/lib/components/LazyTrigger/LazyTrigger.svelte
```diff
@@ -36,4 +36,4 @@
     });
 </script>
 
-<div bind:this={triggerRef}></div>
+<div bind:this={triggerRef} style="height: 1px; width: 1px;"></div>
```
michal-lightly, line 39: What is this change?

lightly_studio/Makefile
```diff
@@ -40,6 +40,11 @@ start-e2e: build
 	@echo "Starting server for e2e tests..."
 	uv run e2e-tests/index_dataset_for_end2end_ui_tests.py
 
+.PHONY: start-coco-10k
+start-coco-10k: build
+	@echo "Starting server for e2e tests with COCO 10k dataset..."
+	uv run e2e-tests/index_coco_10k.py
```
michal-lightly: Please remove

lightly_studio_view/src/lib/components/LazyTrigger/LazyTrigger.svelte
ikondrat: it is to make div always visible, to trigger intersection event

lightly_studio/e2e-tests/index_coco_10k.py
ikondrat: removed

lightly_studio/src/lightly_studio/db_manager.py
```diff
@@ -41,7 +42,22 @@ def __init__(
...
+
+        # Configuration for testing
+        self._engine = create_engine(
+            url=self._engine_url,
+            poolclass=poolclass,
+        )
+
+        # Total available connections = pool_size + max_overflow = 10 + 40 = 50
+        # (default is (pool_size=5, max_overflow=10))
+        if "PYTEST_CURRENT_TEST" not in os.environ:
```
michal-lightly: We are setting `poolclass=StaticPool` in our code in `tests/conftest.py:53`.

A naive solution would be to add `pool_size` and `max_overflow` also as the `__ini__()` function arguments.

A better solution is to bundle the pool settings instead together - that's what I would recommend:
```
@dataclass
class PoolSettings:
    poolclass: type[Pool]
    pool_size: int
    max_overflow: int
```
And make `__init__()` accept `PoolSettings | None`. If None, we set it to prod values, and non-None we pass in conftest.

lightly_studio/src/lightly_studio/db_manager.py
```diff
@@ -27,21 +27,33 @@ def __init__(
...
-                in-memory databases for testing, otherwise different DB connections
-                connect to different in-memory databases.
+            single_threaded: If True, creates a single-threaded engine suitable for testing.
         """
         self._engine_url = engine_url if engine_url else "duckdb:///lightly_studio.db"
         if cleanup_existing:
             _cleanup_database_file(engine_url=self._engine_url)
-        self._engine = create_engine(url=self._engine_url, poolclass=poolclass)
+
+        self._engine = None
```
michal-lightly: Minor: This line can be deleted.

lightly_studio/tests/core/conftest.py
```diff
@@ -23,7 +23,10 @@ def patch_dataset(
     mocker.patch.object(
         db_manager,
         "get_engine",
-        return_value=db_manager.DatabaseEngine("duckdb:///:memory:"),
+        return_value=db_manager.DatabaseEngine(
+            engine_url="duckdb:///:memory:",
+            single_threaded=True,
```
michal-lightly, line 28: Are you sure we need it here? We were using default poolclass before here. Is it set automatically to StaticPool for in-memory db?

ikondrat: yes unfortunatelly, it switches automatically

PR #218, opened 2025-11-28
Split caching steps in e2e tests yaml

michal-lightly:
## What has changed and why?

Split cache building from e2e test execution. Preparation for test with captions and videos.

* Use YAML anchors to avoid duplication
* Reduced caches from 5 to 3
    * Merged python build cache (which was not used properly) with frontend build cache and the overall "e2e" cache
* Improved server error detection

## How has it been tested?

Ran the CI multiple times. Also tested:
* Error on server startup
* When cache is hit, and when not hit

- [ ] Yes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Added reading and updating of captions to the `Sample` class.
 - Added export functionality for image datasets with captions to the python `Dataset` class interface.
+- Added keyboard shortcut support for toggling annotation edit mode.
+- Added sliders to adjust brightness and contrast for more accurate labeling.
+- Reduce the minimum size of bounding box creation from 10px side length to 4px.
 
 ### Changed
 
--- a/README.md
+++ b/README.md
@@ -101,7 +101,7 @@ from lightly_studio.utils import download_example_dataset
 dataset_path = download_example_dataset(download_dir="dataset_examples")
 
 # Create a dataset and populate it with videos.
-dataset = ls.Dataset.create()
+dataset = ls.Dataset.create(sample_type=ls.SampleType.VIDEO)
 dataset.add_videos_from_path(path=f"{dataset_path}/youtube_vis_50_videos/train/videos")
 
 # Start the UI server.
@@ -270,7 +270,7 @@ s = samples[0]
 s.sample_id        # Sample ID (UUID)
 s.file_name        # Image file name (str), e.g. "img1.png"
 s.file_path_abs    # Full image file path (str), e.g. "full/path/img1.png"
-s.tags             # The list of sample tags (list[str]), e.g. ["tag1", "tag2"]
+s.tags             # The set of sample tags (set[str]), e.g. {"tag1", "tag2"}
 s.metadata["key"]  # dict-like access for metadata (any)
 
 # Set sample attributes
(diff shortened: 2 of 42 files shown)


.github/workflows/end2end_test.yml
```diff
@@ -35,46 +38,42 @@ jobs:
...
+        id: cache-build
         uses: actions/cache@v4
         with:
-          path: |
+          lookup-only: true
+          path: &cache-build-path |
             lightly_studio_view/build
             lightly_studio/src/lightly_studio/dist_lightly_studio_view_app
-          key: ui-build-${{ hashFiles('lightly_studio_view/src/**', 'lightly_studio_view/public/**', 'lightly_studio_view/package.json', 'lightly_studio_view/package-lock.json') }}
+          key: &cache-build-key python-build-${{ hashFiles('.github/workflows/end2end_test.yml', 'lightly_studio/e2e-tests/index_dataset_for_end2end_ui_tests.py', 'lightly_studio/src/**', 'lightly_studio/pyproject.toml', 'lightly_studio_view/e2e/**', 'lightly_studio_view/src/**', 'lightly_studio_view/public/**', 'lightly_studio_view/package.json', 'lightly_studio_view/package-lock.json') }}
```
michal-lightly: Added `.github/workflows/end2end_test.yml`

.github/workflows/end2end_test.yml
```diff
@@ -87,61 +86,121 @@ jobs:
...
         run: |
           rm -rf lightly_studio/src/lightly_studio/dist_lightly_studio_view_app
           cp -r lightly_studio_view/build/ lightly_studio/src/lightly_studio/dist_lightly_studio_view_app
+
+  end-to-end-test:
+    name: End2End Test
+    runs-on: ubuntu-latest
+    needs: prepare-caches
+    # Skip the job if there's no change that affects end-to-end tests.
+    if: needs.prepare-caches.outputs.cache-hit != 'true'
```
michal-lightly, line 153: Note: We don't need to check for every step anymore.

.github/workflows/end2end_test.yml
```diff
@@ -35,46 +38,59 @@ jobs:
...
-          path: |
+          lookup-only: true
+          path: &cache-build-path |
             lightly_studio_view/build
             lightly_studio/src/lightly_studio/dist_lightly_studio_view_app
-          key: ui-build-${{ hashFiles('lightly_studio_view/src/**', 'lightly_studio_view/public/**', 'lightly_studio_view/package.json', 'lightly_studio_view/package-lock.json') }}
+          key: &cache-build-key >-
+            python-build-${{
+              hashFiles(
+                '.github/workflows/end2end_test.yml',
```
michal-lightly, line 52: Added `end2end_test.yml`

.github/workflows/end2end_test.yml
```diff
@@ -163,6 +238,11 @@ jobs:
               echo "✅ Server is up and running after ${elapsed} seconds!"
               break
             fi
+            if ! ps -p ${{ env.SERVER_PID }} > /dev/null; then
+              echo "❌ Server process has exited unexpectedly."
+              exit 1
+            fi
```
michal-lightly, line 248: Exit early if server stating failed.

.github/workflows/end2end_test.yml
lukas-lightly, line 97: Seeing `enable-cache: true`  - doesn't the `uv` action have its own caching?

michal-lightly: It does have some caching, I do not understand details though. I kept the logic as it was.

Note that we do not attempt to cache python dependencies in this job, we just need to build the frontend that we cache (which needs backend for the schema).

PR #228, opened 2025-12-01
Add captions export endpoint

michal-lightly:
## What has changed and why?

Add captions export endpoint. Modelled after the existing annotations export endpoint.

## How has it been tested?

Added a unit test.

- [ ] Yes


lightly_studio/src/lightly_studio/api/routes/api/caption.py
```diff
@@ -73,7 +73,6 @@ def create_caption(
         parent_dataset_id=parent_sample.dataset_id,
         captions=[
             CaptionCreate(
-                dataset_id=parent_sample.dataset_id,
```
michal-lightly, line 76: `CaptionCreate` does not have the `dataset_id` field. We don't get static checks because SQLModel does some magic in the background.

PR #232, opened 2025-12-01
Implement a new endpoint to count annotations for videos

LeonardoRosaa:
## What has changed and why?

Implemented a new endpoint to count annotations for videos. This endpoint considers annotations on video frames, as we are not using video annotations directly.

## How has it been tested?

Unit tests.

- [ ] Yes

### DIFF
--- a/lightly_studio/src/lightly_studio/resolvers/video_resolver/count_video_frame_annotations_by_video_dataset.py
+++ b/lightly_studio/src/lightly_studio/resolvers/video_resolver/count_video_frame_annotations_by_video_dataset.py
@@ -4,7 +4,6 @@ from typing import Any, List, Optional, Tuple
 from uuid import UUID
 
 from pydantic import BaseModel
-
 from sqlmodel import Session, asc, col, func, select
 from sqlmodel.sql.expression import Select
 
--- a/lightly_studio/tests/api/routes/api/test_video.py
+++ b/lightly_studio/tests/api/routes/api/test_video.py
@@ -170,3 +170,91 @@ def test_get_fields_bounds(test_client: TestClient, db_session: Session) -> None
     assert result["width"]["max"] == 300
     assert result["height"]["min"] == 250
     assert result["height"]["max"] == 450
+
+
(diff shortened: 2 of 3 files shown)


lightly_studio/src/lightly_studio/resolvers/video_resolver/count_video_frame_annotations_by_video_dataset.py
```diff
@@ -0,0 +1,88 @@
+"""Count video frame annotations by video dataset."""
+
+from typing import Any, List, Optional, Tuple
+from uuid import UUID
+
+from pydantic import BaseModel
+
+# from sqlalchemy import Select, select
```
horatiualmasan: Remove commented code

lightly_studio/src/lightly_studio/resolvers/video_resolver/video_count_annotations_filter.py
```diff
@@ -0,0 +1,50 @@
...
+from sqlmodel import col, select
+
+from lightly_studio.models.annotation.annotation_base import AnnotationBaseTable
+from lightly_studio.models.annotation_label import AnnotationLabelTable
+from lightly_studio.models.video import VideoFrameTable, VideoTable
+from lightly_studio.resolvers.video_resolver.video_filter import VideoFilter
+from lightly_studio.type_definitions import QueryType
+
+
+class VideoCountAnnotationsFilter(BaseModel):
```
horatiualmasan, line 15: Can't we just use VideoFilters? without adding a new class? We already have the annotation_frames_label_ids but as ids and not as string.

LeonardoRosaa: We can't use the same because the count annotation filters use the annotation label name instead of the annotation label ID

horatiualmasan: Ok, we can leave it as it is for now. Maybe add a TODO to check if we can't use annotation label IDs everywhere so we don't need the extra class.

PR #243, opened 2025-12-02
Add endpoint to retrieve video frame fields bounds

LeonardoRosaa:
## What has changed and why?

Introduced an endpoint that returns the bounds (min and max) of the `frame_number` field for video frames.

## How has it been tested?

Unit tests.

- [ ] Yes


lightly_studio/src/lightly_studio/api/routes/api/frame.py
```diff
@@ -72,6 +74,27 @@ def get_all_frames(
     )
 
 
+@frame_router.get("/bounds", response_model=Optional[VideoFrameFieldsBoundsView])
```
michal-lightly: Minor: For stronger type checks, it is better not to specify the response model if it matches the return type.
```suggestion
@frame_router.get("/bounds")
```

PR #249, opened 2025-12-03
Operator Dialog

JonasWurst:
## What has changed and why?

This adds the operator dialog. The different input types are intentionally decoupled from the dialog itself. Ideally we onyl need to new types and components for new input types there

## How has it been tested?

Locally tested. For e2e test, we would need to define and register operators first.

- [x] Yes

### DIFF
--- a/lightly_studio_view/src/lib/components/Header/Header.svelte
+++ b/lightly_studio_view/src/lib/components/Header/Header.svelte
@@ -8,7 +8,7 @@
     import { page } from '$app/state';
     import NavigationMenu from '../NavigationMenu/NavigationMenu.svelte';
     import { isSamplesRoute } from '$lib/routes';
-    import { useRootDataset } from '$lib/hooks/useRootDataset/useRootDataset';
+    import { useRootDatasetOptions } from '$lib/hooks/useRootDataset/useRootDataset';
     import { get } from 'svelte/store';
     import Menu from '$lib/components/Header/Menu.svelte';
 
@@ -22,6 +22,8 @@
     const isFSCEnabled = $derived.by(() => {
         return $featureFlags.some((flag) => flag === 'fewShotClassifierEnabled');
     });
+
+    const { rootDataset } = useRootDatasetOptions();
 
     const { setIsEditingMode, isEditingMode, reversibleActions, executeReversibleAction } =
         page.data.globalStorage;
@@ -46,9 +48,9 @@
                 <a href="/"><Logo /></a>
             </div>
             <div class="flex flex-1 justify-start">
-                {#await useRootDataset() then dataset}
-                    <NavigationMenu {dataset} />
-                {/await}
+                {#if $rootDataset.data}
+                    <NavigationMenu dataset={$rootDataset.data} />
+                {/if}
             </div>
             <div class="flex flex-auto justify-end gap-2">
                 <Menu {isSamples} {hasEmbeddingSearch} {isFSCEnabled} />
--- a/lightly_studio_view/src/lib/components/Footer/Footer.svelte
+++ b/lightly_studio_view/src/lib/components/Footer/Footer.svelte
@@ -2,7 +2,7 @@
     import { BookOpen, Mail } from '@lucide/svelte';
     import { page } from '$app/state';
     import { version, git_sha, is_tagged_commit } from '$lib/version.json';
-    import { isAnnotationsRoute } from '$lib/routes';
+    import { isAnnotationsRoute, isVideoFramesRoute, isVideosRoute } from '$lib/routes';
 
     type FooterProps = {
         totalSamples?: number;
@@ -18,11 +18,23 @@
         filteredAnnotations = 0
     }: FooterProps = $props();
 
+    function getItemType(): string {
+        if (isAnnotationsRoute(page.route.id)) {
+            return 'annotations';
+        } else if (isVideoFramesRoute(page.route.id)) {
+            return 'video frames';
+        } else if (isVideosRoute(page.route.id)) {
+            return 'videos';
+        } else {
+            return 'images';
+        }
+    }
+
     const statsText = $derived.by(() => {
         const isAnnotationView = isAnnotationsRoute(page.route.id);
         const total = isAnnotationView ? totalAnnotations : totalSamples;
         const filtered = isAnnotationView ? filteredAnnotations : filteredSamples;
-        const itemType = isAnnotationView ? 'annotations' : 'images';
+        const itemType = getItemType();
 
         if (!total) return '';
 
--- a/lightly_studio_view/src/lib/components/Header/Header.test.ts
+++ b/lightly_studio_view/src/lib/components/Header/Header.test.ts
@@ -1,13 +1,13 @@
 import { describe, it, expect, vi, beforeEach } from 'vitest';
 import { fireEvent, render, screen } from '@testing-library/svelte';
-import { writable } from 'svelte/store';
+import { readable, writable } from 'svelte/store';
 import Header from './Header.svelte';
 import * as appState from '$app/state';
 import '@testing-library/jest-dom';
 import { Page } from '@sveltejs/kit';
 import type { ReversibleAction } from '$lib/hooks/useGlobalStorage';
 
-import { readRootDataset } from '$lib/api/lightly_studio_local';
+import { useRootDatasetOptions } from '$lib/hooks/useRootDataset/useRootDataset';
 
 describe('Header', () => {
     const setup = (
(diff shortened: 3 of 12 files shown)


lightly_studio_view/src/lib/components/Header/Header.svelte
```diff
@@ -51,7 +57,7 @@
                 {/await}
             </div>
             <div class="flex flex-auto justify-end gap-2">
-                <Menu {isSamples} {hasEmbeddingSearch} {isFSCEnabled} />
+                <Menu {isSamples} {hasEmbeddingSearch} {isFSCEnabled} {datasetId} />
```
mihnea-necsulescu: I don't think the dataset id should be passed as a prop to components; when we will introduce multiple datasets, this will most likely cause a bug: the header/menu will have a dataset id set on the first render -> user switches the dataset -> header/menu will still have the old dataset it -> operators will execute on a different dataset than the selected one

the dataset id is usually retrieved from the page params `page.params.dataset_id`

mihnea-necsulescu: https://github.com/lightly-ai/lightly-studio/blob/main/lightly_studio_view/src/routes/datasets/%5Bdataset_id%5D/%2Blayout.svelte#L272

this should also be fixed

lightly_studio_view/src/lib/hooks/useOperatorsDialog/useOperatorsDialog.ts
```diff
@@ -1,18 +1,22 @@
 import { writable } from 'svelte/store';
 
 const isOperatorsDialogOpen = writable(false);
+const operatorDatasetId = writable<string | undefined>(undefined);
```
mihnea-necsulescu: I don't think this would be needed either if we get the dataset id from the page params

lightly_studio_view/src/lib/components/Operator/OperatorsMenu.svelte
```diff
@@ -36,18 +44,23 @@
         loadOperators();
     });
 
+    const { isOperatorsDialogOpen, operatorDatasetId, openOperatorsDialog, closeOperatorsDialog } =
+        useOperatorsDialog();
+    const datasetId = $derived.by(
+        () => datasetIdProp ?? $operatorDatasetId ?? page.params.dataset_id
+    );
```
mihnea-necsulescu: can be just

```suggestion
    const datasetId = $derived(page.params.dataset_id);
```

lightly_studio_view/src/lib/components/Operator/OperatorDialog.svelte
```diff
@@ -0,0 +1,222 @@
...
+
+    let { operatorMetadata, datasetId, isOpen, onOpenChange }: Props = $props();
+    let operator: Operator | null = $state(null);
+    let isLoadingParameters = $state(false);
+    let loadError = $state<string | null>(null);
+    let parameters = $state<ParameterValues>({});
+    let isExecuting = $state(false);
+    let executionError = $state<string | undefined>(undefined);
+    let executionSuccess = $state<string | undefined>(undefined);
+    let fetchToken = 0;
```
mihnea-necsulescu: I'm not sure I understand the need for this `fetchToken` - is it used as a protection against race conditions when loading the operator parameters? if yes, I'm not sure that would even be possible; the time needed to load the params would probably be much smaller than the time it would take a user to switch to another operator in the meantime

I suggest we remove it for now and add a todo/simple comment to keep this in mind; if we decide that we actually need it, there are better alternatives to handle this (abort controllers for example)

mihnea-necsulescu: we could also add a loading overlay while the operator executes? this would block the dialog -> users can't close it while executing (this may be bad UX though)

lightly_studio_view/src/lib/components/Operator/OperatorDialog.svelte
```diff
@@ -0,0 +1,222 @@
...
+        previousIsOpen = isOpen;
+    });
+
+    async function handleExecute() {
+        if (!operator || !datasetId || !isFormValid) {
+            executionError = !datasetId
+                ? 'Dataset not available. Please open a dataset first.'
+                : 'Please fill in all required parameters.';
+            return;
+        }
```
mihnea-necsulescu, line 93: this check seems to be redundant: the execute button would be disabled anyway if the form is invalid (which also checks if the operator is set)

the dataset id _should_ always be present, but we can still keep that check just to be sure

lightly_studio_view/src/lib/components/Operator/OperatorDialog.svelte
```diff
@@ -0,0 +1,222 @@
...
+            fetchToken += 1;
+            operator = null;
+            parameters = {};
+            loadError = null;
+            isLoadingParameters = false;
+            resetExecutionState();
+            return;
+        }
+        loadOperatorDefinition(operatorMetadata);
+    });
```
mihnea-necsulescu, line 76: I'm not sure I understand this - the operator metadata can't change while an operator has already been selected, no? you'd have to close the dialog, reopen it, and pick a new operator

or am I missing something?

mihnea-necsulescu: I'm saying this can be done once when the component mounts, I think, or load them directly in the menu

but this should be fine for now

lightly_studio_view/src/lib/components/Operator/parameterTypeConfig.ts
```diff
@@ -0,0 +1,96 @@
...
+
+const parseIntegerValue = (value: string) => (value === '' ? '' : Number.parseInt(value, 10));
+const parseFloatValue = (value: string) => (value === '' ? '' : Number.parseFloat(value));
+const identity = (value: string) => value;
+
+export function isValueFilled(
+    value: ParameterValue,
+    type: OperatorParameterType | 'default'
+): boolean {
+    if (value === undefined || value === null) return false;
```
mihnea-necsulescu, line 36: image

this is a bit confusing for the checkbox case: if I check it, then uncheck it, the "this field is required" message disappears, because it's only checking for null and undefined - how should we handle this case?

lightly_studio_view/src/lib/components/Header/Header.svelte
JonasWurst: first one is fixed
for the second point I will create a follow up pr.

lightly_studio_view/src/lib/components/Operator/OperatorDialog.svelte
JonasWurst: Kept it, as it does not hurt. But in general you are right

lightly_studio_view/src/lib/components/Operator/OperatorDialog.svelte
JonasWurst: removed the token and aded a todo

lightly_studio_view/src/lib/components/Operator/parameterTypeConfig.ts
JonasWurst: fixed

lightly_studio_view/src/lib/components/Operator/OperatorDialog.svelte
JonasWurst: I removed the check for empty meta data. Now it loads the operator definition.

JonasWurst:
> two more things:
> 
> 1. after selecting an operator and closing the dialog, then reopening the operator list, the previously selected operator is still highlighted
> 2. the background lightens after selecting an operator:
> 
> image image

Resolved the different background, the highlighting was not reproducible for me.

lightly_studio_view/src/lib/components/Operator/OperatorDialog.svelte
mihnea-necsulescu: let's do it like 

```suggestion
    // Load operator definition and initialize parameters when metadata changes
    $effect(() => {
        if (isOpen && operatorMetadata) {
            loadOperatorDefinition(operatorMetadata);
        }
    });
```
since loadOperatorDefinition expects the metadata to not be null (and also check if the dialog is open before actually loading)

lightly_studio_view/src/lib/components/Operator/OperatorsMenu.svelte
```diff
@@ -1,18 +1,20 @@
...
     import { getOperators, type RegisteredOperatorMetadata } from '$lib/api/lightly_studio_local';
     import Loader2 from '@lucide/svelte/icons/loader-2';
     import AlertCircle from '@lucide/svelte/icons/alert-circle';
     import ChevronRight from '@lucide/svelte/icons/chevron-right';
     import * as Dialog from '$lib/components/ui/dialog';
     import { useOperatorsDialog } from '$lib/hooks/useOperatorsDialog/useOperatorsDialog';
+    import OperatorDialog from '$lib/components/Operator/OperatorDialog.svelte';
 
     let operators: RegisteredOperatorMetadata[] = $state([]);
     let selectedOperatorId: string | undefined = $state(undefined);
```
mihnea-necsulescu, line 12: I think that the previously selected operator remains highlighted because of this - this is never reset

I think we should set it back to undefined when the menu closes?

lightly_studio_view/src/lib/components/Operator/OperatorsMenu.svelte
```diff
@@ -107,20 +112,9 @@
...
-            </Dialog.Header>
-            {#if activeOperator}
-                <p class="text-sm text-foreground">Operator ID: {activeOperator.operator_id}</p>
-            {/if}
-        </Dialog.Content>
-    </Dialog.Portal>
-</Dialog.Root>
+<OperatorDialog
+    operatorMetadata={activeOperatorMetadata}
+    {datasetId}
```
mihnea-necsulescu: let's not pass the dataset id to the dialog either, and get it from the page params there too

PR #250, opened 2025-12-03
Use tags for video samples in GUI

horatiualmasan:
## What has changed and why?

Add tag functionality for videos. 
Now you can:

From video grid page: 
- view existing tags
- filter videos using tags
- create tag for videos
- add videos to tag 
- 
From video details page:
- view existing tags from the video sample
- remove tag from sample
-
## How has it been tested?
Tested locally

https://github.com/user-attachments/assets/48b517f2-2bee-4506-a8a7-6455051cd43e

- [x] Yes

### DIFF
--- a/lightly_studio_view/src/routes/datasets/[dataset_id]/+layout.svelte
+++ b/lightly_studio_view/src/routes/datasets/[dataset_id]/+layout.svelte
@@ -102,6 +102,10 @@
             gridType = 'samples';
         } else if (isCaptions) {
             gridType = 'captions';
+        } else if (isVideoFrames) {
+            gridType = 'video_frames';
+        } else if (isVideos) {
+            gridType = 'videos';
         }
 
         // Temporary hack to remember where the user was when navigating
@@ -187,13 +191,21 @@
     const annotationsLabels = $derived(
         selectedAnnotationFilter.length > 0 ? selectedAnnotationFilter : undefined
     );
+    const metadataFilters = $derived(
+        metadataValues ? createMetadataFilters($metadataValues) : undefined
+    );
     const rootDatasetId = $derived(dataset.parent_dataset_id ?? datasetId);
     const annotationCounts = $derived.by(() => {
         if (isVideoFrames) {
             return useVideoFrameAnnotationCounts({
                 datasetId: rootDatasetId,
                 filter: {
-                    annotations_labels: annotationsLabels
+                    annotations_labels: annotationsLabels,
+                    video_filter: {
+                        sample_filter: {
+                            metadata_filters: metadataFilters
+                        }
+                    }
                 }
             });
         } else if (isVideos) {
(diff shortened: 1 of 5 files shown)


lightly_studio_view/src/routes/datasets/[dataset_id]/videos/[sample_id]/+page.svelte
```diff
@@ -12,10 +13,36 @@
...
+            })
+        ) ?? []
+    );
+
+    const handleRemoveTag = async (tagId: string) => {
+        if (!sample?.sample_id) return;
+        try {
+            await removeTagFromSample(sample.sample_id, tagId);
+            // Refresh the page data to get updated tags
+            await invalidateAll();
```
Comment (LeonardoRosaa) [inline lightly_studio_view/src/routes/datasets/[dataset_id]/videos/[sample_id]/+page.svelte:40]: We should avoid refreshing the page data, as it results in a lot of unnecessary calls. Although this isn’t being addressed in this PR, I suggest opening an issue to track it.

lightly_studio_view/src/routes/datasets/[dataset_id]/+layout.svelte
```diff
@@ -102,6 +102,8 @@
             gridType = 'samples';
         } else if (isCaptions) {
             gridType = 'captions';
+        } else if (isVideos) {
+            gridType = 'samples';
```
Comment (LeonardoRosaa) [inline lightly_studio_view/src/routes/datasets/[dataset_id]/+layout.svelte]: Should be `videos`?

lightly_studio_view/src/routes/datasets/[dataset_id]/videos/+page.svelte
```diff
@@ -75,7 +100,30 @@
                 {#snippet item({ index, style })}
                     {#if items[index]}
                         {#key items[index].sample_id}
-                            <div {style}>
+                            <div
```
Comment (LeonardoRosaa) [inline lightly_studio_view/src/routes/datasets/[dataset_id]/videos/+page.svelte:109]: We’re currently building similar components for the samples, annotations, and videos grid views. To reduce code duplication and improve maintainability, let’s consider creating a reusable component that can serve all three views.

lightly_studio_view/src/routes/datasets/[dataset_id]/+layout.svelte
Comment (horatiualmasan) [reply lightly_studio_view/src/routes/datasets/[dataset_id]/+layout.svelte]: It should be `sample`, as it will be used in the create tag dialog to create either a sample tag or an annotation tag. It was sample by default so I removed the unnecessary code.

lightly_studio_view/src/routes/datasets/[dataset_id]/videos/[sample_id]/+page.svelte
Comment (horatiualmasan) [reply lightly_studio_view/src/routes/datasets/[dataset_id]/videos/[sample_id]/+page.svelte:40]: Yes, we should find a better way to handle this.  For the current use case it is probably fine as it will not happen that often to cause performance issues.

lightly_studio_view/src/routes/datasets/[dataset_id]/videos/+page.svelte
Comment (horatiualmasan) [reply lightly_studio_view/src/routes/datasets/[dataset_id]/videos/+page.svelte:109]: Yes, it will also be used for video frames. I would do it in a follow-up once we have all the functionality ready.

Comment (LeonardoRosaa) [reply lightly_studio_view/src/routes/datasets/[dataset_id]/videos/+page.svelte:109]: Could you create a follow-up issue?

PR #252, opened 2025-12-03
Allow filter by video frame number

LeonardoRosaa:
## What has changed and why?

Added a filter by video frame number to the GUI.

## How has it been tested?

https://github.com/user-attachments/assets/1a41c75c-4e60-4850-a298-aa63a726ba9f

- [ ] Yes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added metadata section to video and video frame details.
 - Added tag support for videos in the GUI.
 - Added plugins: This is the initial version for plugins. It supports the execution of operators.
+- `lightly_studio.core.sample.Sample` renamed to `ImageSample`.
 
 ### Changed
 
--- a/lightly_studio/src/lightly_studio/core/add_samples.py
+++ b/lightly_studio/src/lightly_studio/core/add_samples.py
@@ -29,7 +29,7 @@ from sqlmodel import Session
 from tqdm import tqdm
 
 from lightly_studio.core.loading_log import LoadingLoggingContext, log_loading_results
-from lightly_studio.core.sample import Sample
+from lightly_studio.core.sample import ImageSample
 from lightly_studio.models.annotation.annotation_base import AnnotationCreate
 from lightly_studio.models.annotation_label import AnnotationLabelCreate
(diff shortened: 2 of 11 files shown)


lightly_studio_view/src/lib/components/VideoFrameBoundsFilter/VideoFrameBoundsFilter.svelte
```diff
@@ -0,0 +1,43 @@
...
+        <div class="flex justify-between text-sm text-diffuse-foreground">
+            <span>{formatInteger($videoFramesBoundsValues.frame_number.min)}</span>
+            <span>{formatInteger($videoFramesBoundsValues.frame_number.max)}</span>
+        </div>
+        <div class="relative p-2">
+            <Slider
+                type="multiple"
+                class="filter-width"
+                min={$videoFramesBounds?.frame_number.min}
+                max={$videoFramesBounds?.frame_number.max}
```
mihnea-necsulescu: ? not really needed since we're already inside #if $videoFramesBounds

lightly_studio_view/src/lib/hooks/useVideoFramesBounds/useVideoFramesBounds.ts
```diff
@@ -0,0 +1,57 @@
...
+const videoFramesBoundsValues = useSessionStorage<VideoFrameFieldsBoundsView | null>(
+    'lightlyStudio_video_frames_bounds_values',
+    null
+);
+const videoFramesBounds = useSessionStorage<VideoFrameFieldsBoundsView | null>(
+    'lightlyStudio_video_frames_bounds',
+    null
+);
+
+const isInitialized = writable(false as boolean);
```
mihnea-necsulescu:
```suggestion
const isInitialized = writable<boolean>(false);
```

mihnea-necsulescu: also this may cause bugs when we add support for multiple datasets (stale bounds after switching to another video dataset); it's fine for now, but we should probably add a comment just so we know where to check

mihnea-necsulescu:
image
shouldn't the bounds start at 1 instead of 0 (1 - 5, in this case)?

mihnea-necsulescu (approved):

PR #253, opened 2025-12-03
Add metadata filters for videos frames.

LeonardoRosaa:
## What has changed and why?

Added metadata filters for video frames in the GUI, and introduced a `force` field to refresh metadata across pages.

## How has it been tested?

- Upload a video dataset and add metadata
- Open the GUI, and the filter options should be there.

- [ ] Yes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Added plugins: This is the initial version for plugins. It supports the execution of operators.
+
 ### Changed
 
 ### Deprecated
--- a/lightly_studio/src/lightly_studio/examples/example_operators.py
+++ b/lightly_studio/src/lightly_studio/examples/example_operators.py
@@ -13,7 +13,13 @@ import lightly_studio as ls
 from lightly_studio import db_manager
 from lightly_studio.plugins.base_operator import BaseOperator, OperatorResult
 from lightly_studio.plugins.operator_registry import operator_registry
-from lightly_studio.plugins.parameter import BaseParameter, BoolParameter, StringParameter
+from lightly_studio.plugins.parameter import (
+    BaseParameter,
+    BoolParameter,
+    FloatParameter,
+    IntParameter,
+    StringParameter,
+)
 
 
 @dataclass
@@ -28,7 +34,25 @@ class TestOperator(BaseOperator):
         """Return the list of parameters this operator expects."""
         return [
             BoolParameter(name="test flag", required=True),
-            StringParameter(name="test str", required=True),
+            StringParameter(name="test str", required=True, default="Good Morning"),
+            StringParameter(
+                name="test str 2", required=True, default="Another", description="Test Description"
+            ),
+            BoolParameter(
+                name="test flag 2", required=True, description="Test Description", default=True
+            ),
+            FloatParameter(name="test float", required=False, default=2.0, description="abc"),
+            IntParameter(name="test int", required=False, default=2),
+            IntParameter(name="test int 2", required=True, description="Yet another int"),
+            IntParameter(
+                name="test int 3",
+                required=True,
+                description=(
+                    "Yet another int with a very long description. this will show how we will "
+                    "overflow or what soever is happening to this text in the GUI. maybe it "
+                    "exceeds maybe not. only time will tell"
+                ),
+            ),
         ]
 
     def execute(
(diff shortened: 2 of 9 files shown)


lightly_studio_view/src/lib/hooks/useMetadataFilters/useMetadataFilters.ts
```diff
@@ -10,8 +10,8 @@ type MetadataValues = Record<string, { min: number; max: number }>;
 
 const isInitialized = writable(false as boolean);
 
-const loadInitialMetadataInfo = async (dataset_id: string) => {
-    if (get(isInitialized)) {
+const loadInitialMetadataInfo = async (dataset_id: string, force = false) => {
```
mihnea-necsulescu: the `force` bool seems like a quick fix to a bigger issue we should address in a follow-up: in the global storage, we are currently only storing the metadata for a single dataset - this became outdated since we introduced videos and had multiple datasets

I think that now, the global metadata will get overwritten every time we switch grids

we should probably start storing metadata per dataset, but that's outside of the scope of this PR, and can be addressed later (we should probably add a triage issue for it, though)

what we could do, in order to not modify the signature here and keep the changes minimal, is to replace `isInitialized` at line 11 with a `lastInitializedDatasetId` or something like that, and check that against the `dataset_id` we receive here

lightly_studio_view/src/lib/hooks/useMetadataFilters/useMetadataFilters.ts
```diff
@@ -100,13 +100,20 @@ export const useMetadataFilters = (dataset_id?: string) => {
         updateMetadataInfo
     } = useGlobalStorage();
 
+    function cleanMetadata() {
```
mihnea-necsulescu: do we need this? it seems like it's not used anywhere

lightly_studio_view/src/lib/components/CombinedMetadataDimensionsFilters/index.svelte
```diff
@@ -46,7 +51,7 @@
         metadataBounds: metadataBounds,
         metadataValues: metadataValues,
         updateMetadataValues: updateMetadataValues
-    } = useMetadataFilters(datasetId);
+    } = $derived(useMetadataFilters(datasetId, true));
```
mihnea-necsulescu: should this hook be inside $derived?

lightly_studio_view/src/lib/components/CombinedMetadataDimensionsFilters/index.svelte
```diff
@@ -5,15 +5,20 @@
     import Segment from '$lib/components/Segment/Segment.svelte';
     import { useDimensions } from '$lib/hooks/useDimensions/useDimensions';
     import { useMetadataFilters } from '$lib/hooks/useMetadataFilters/useMetadataFilters';
-    import { page } from '$app/state';
+    import { page } from '$app/stores';
```
mihnea-necsulescu: this is deprecated - we should import from `$app/state` instead

lightly_studio_view/src/lib/components/CombinedMetadataDimensionsFilters/index.svelte
```diff
@@ -5,15 +5,20 @@
...
+    import { derived } from 'svelte/store';
     import VideoFieldBoundsFilters from '../VideoFieldBoundsFilters/VideoFieldBoundsFilters.svelte';
 
-    const datasetId = page.params.dataset_id;
+    const params = derived(page, ($page) => $page.params);
+
+    const datasetId = $derived($params.dataset_id);
```
mihnea-necsulescu: is this change needed? couldn't we just wrap `page.params.dataset_id` inside $derived?

lightly_studio_view/src/routes/datasets/[dataset_id]/+layout.svelte
```diff
@@ -296,8 +306,13 @@
                                         onToggleAnnotationFilter={toggleAnnotationFilterSelection}
                                     />
 
-                                    {#if isSamples || isVideos}
-                                        <CombinedMetadataDimensionsFilters {isVideos} />
+                                    {#if isSamples || isVideos || isVideoFrames}
+                                        {#key gridType}
```
Comment (mihnea-necsulescu) [inline lightly_studio_view/src/routes/datasets/[dataset_id]/+layout.svelte]: shouldn't the key be the dataset id instead?

lightly_studio_view/src/lib/hooks/useMetadataFilters/useMetadataFilters.ts
LeonardoRosaa: You’re right, it’s definitely a workaround. I’ll go ahead and implement your suggestion. It looks much better.

lightly_studio_view/src/routes/datasets/[dataset_id]/+layout.svelte
Comment (LeonardoRosaa) [reply lightly_studio_view/src/routes/datasets/[dataset_id]/+layout.svelte]: We should use `page.params.dataset_id` because the `datasetId` from `data` does not update correctly at the `+layout` level.

Comment (LeonardoRosaa) [reply lightly_studio_view/src/routes/datasets/[dataset_id]/+layout.svelte]: This issue occurs because `data` is not reactive. I’ll address this in a separate PR.

mihnea-necsulescu (approved):

PR #256, opened 2025-12-04
Video Embedding: Download from lightly hosted s3

JonasWurst:
## What has changed and why?

Downloading the model weights from our s3 bucket. 

## How has it been tested?

Test script.

- [ ] Yes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Added tag support for videos in the GUI.
 - Added plugins: This is the initial version for plugins. It supports the execution of operators.
 
 ### Changed
--- a/lightly_studio_view/src/routes/datasets/[dataset_id]/+layout.svelte
+++ b/lightly_studio_view/src/routes/datasets/[dataset_id]/+layout.svelte
@@ -293,7 +293,7 @@
                         <div
                             class="min-h-0 flex-1 space-y-2 overflow-y-auto px-4 pb-2 dark:[color-scheme:dark]"
                         >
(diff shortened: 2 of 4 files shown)


lightly_studio/src/lightly_studio/vendor/perception_encoder/vision_encoder/config.py
```diff
@@ -7,28 +10,57 @@
...
+    Supports checkpoints from:
+    - Predefined downloadable model names from our repository
 
-        return hf_hub_download(repo_id=repo, filename=file)
-    else:
-        return path
+    Returns:
+        Path to the local checkpoint file.
+    """
+    if model_name in DOWNLOADABLE_MODEL_URL.keys():
```
michal-lightly: Minor: Early exit is more readable:
```py
if model_name not in DOWNLOADABLE_MODEL_URL.keys():
    raise ValueError(f"Unknown model name: '{model_name}'")
```

lightly_studio/src/lightly_studio/vendor/perception_encoder/vision_encoder/config.py
```diff
@@ -7,28 +10,57 @@
...
 
+        download_dir = Path(tempfile.gettempdir())
+        model_name_with_suffix = os.path.basename(urllib.parse.urlparse(model_url).path)
+        local_ckpt_path = download_dir / model_name_with_suffix
 
+        file_utils.download_file_if_does_not_exist(
+            url=model_url,
+            local_filename=local_ckpt_path,
+        )
+        return str(local_ckpt_path)
```
michal-lightly: The type annotations says Path. I think it is fine to return str, we don't need the cast in fetch_pe_checkpoint then.

Note that vendored code is not static checked (on purpose).

PR #258, opened 2025-12-04
Add tags to video frames

horatiualmasan:
## What has changed and why?

Add tag functionality for video frames.
Now you can:

From video frame grid page:

view existing tags
filter videos using tags
create tag for video frames
add video frames to tag

From video frame details page:

view existing tags from the video frame
remove tag from the video frame

## How has it been tested?
manual tests
https://github.com/user-attachments/assets/13762d10-caf6-4653-97e8-4637205b716b

- [ ] Yes

### DIFF
--- a/lightly_studio/src/lightly_studio/api/routes/video_frames_media.py
+++ b/lightly_studio/src/lightly_studio/api/routes/video_frames_media.py
@@ -2,11 +2,13 @@
 
 from __future__ import annotations
 
+import io
 from collections.abc import Generator
-from typing import Any
+from typing import Any, cast
 from uuid import UUID
 
 import cv2
+import fsspec
 from fastapi import APIRouter, HTTPException
 from fastapi.responses import StreamingResponse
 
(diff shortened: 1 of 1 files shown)


mihnea-necsulescu:
https://github.com/user-attachments/assets/bef4b8f9-ca0d-4ca9-88b5-034d93d38d46



seems like the selection state is not bound to a specific grid, like it is for images/annotations: I can create tags for videos from the video frame grid, and the created tag is seen only on the frame grid, but it filters the videos

lightly_studio_view/src/routes/datasets/[dataset_id]/videos/+page.svelte
```diff
@@ -64,6 +69,7 @@
         }
     }
     $effect(() => {
+        clearSelectedSamples();
         setfilteredSampleCount($totalCount);
```
Comment (mihnea-necsulescu) [inline lightly_studio_view/src/routes/datasets/[dataset_id]/videos/+page.svelte]: I think these shouldn't be done in the same effect, the selections will clear every time the total count changes (for example when filters change):

https://github.com/user-attachments/assets/060e2c29-f408-4afd-aa0c-e68e4b17b2e8

lightly_studio_view/src/routes/datasets/[dataset_id]/videos/+page.svelte
```diff
@@ -64,6 +69,7 @@
         }
     }
     $effect(() => {
+        clearSelectedSamples();
```
Comment (mihnea-necsulescu) [inline lightly_studio_view/src/routes/datasets/[dataset_id]/videos/+page.svelte]: also, I'm not sure this should be called here at all - there's currently a discrepancy between images/annotations and videos/video frames: for the former case, `clearSelectedSamples()` only seems to be called after creating the tag; if you select some images, go to the annotations view, then go back to the images view, the selection will still be there - this does not happen with videos/video frames

Comment (horatiualmasan) [reply lightly_studio_view/src/routes/datasets/[dataset_id]/videos/+page.svelte]: Yes, because the selection is stored separated for samples(images) and annotations, now videos and video frames are both samples and share the same global storage entry. I will group them by the dataset_id so they are independent and would not need clearing when you switch pages

lightly_studio_view/src/routes/datasets/[dataset_id]/videos/+page.svelte
Comment (horatiualmasan) [reply lightly_studio_view/src/routes/datasets/[dataset_id]/videos/+page.svelte]: Removed now as the selected samples for each page are stored separately. No need to clear here.

lightly_studio_view/src/lib/components/TagCreateDialog/TagCreateDialog.svelte
```diff
@@ -27,28 +27,34 @@
...
 
     // setup global selection state
     const {
-        selectedSampleIds,
+        getSelectedSampleIds,
         selectedSampleAnnotationCropIds,
         clearSelectedSampleAnnotationCrops,
         clearSelectedSamples
     } = useGlobalStorage();
+    const selectedSampleIds = getSelectedSampleIds(datasetId);
```
mihnea-necsulescu, line 42:
```suggestion
    const selectedSampleIds = $derived(getSelectedSampleIds(datasetId));
```

otherwise this is only called once when the component initializes, and the video dataset id would always be used (for example, the create tag button would be disabled for frames until you also select some videos in the video grid)

horatiualmasan: done

PR #259, opened 2025-12-04
VideoEmbeddingGenerator Protocol + EmbeddingManager refactor

JonasWurst:
## What has changed and why?

This PR introduces:
- a new VideoEmbeddingGenerator Protocol + renames the existing to ImageEmbeddingGenerator
- preparation refactoring in the embedding manger

## How has it been tested?

Test for the embedding manger will be implemented in a follow up PR

- [ ] Yes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Added tag support for videos in the GUI.
+- Added plugins: This is the initial version for plugins. It supports the execution of operators.
+
 ### Changed
 
 ### Deprecated
--- a/lightly_studio/src/lightly_studio/examples/example_operators.py
+++ b/lightly_studio/src/lightly_studio/examples/example_operators.py
@@ -13,7 +13,13 @@ import lightly_studio as ls
 from lightly_studio import db_manager
 from lightly_studio.plugins.base_operator import BaseOperator, OperatorResult
 from lightly_studio.plugins.operator_registry import operator_registry
-from lightly_studio.plugins.parameter import BaseParameter, BoolParameter, StringParameter
+from lightly_studio.plugins.parameter import (
+    BaseParameter,
+    BoolParameter,
+    FloatParameter,
+    IntParameter,
+    StringParameter,
+)
 
 
 @dataclass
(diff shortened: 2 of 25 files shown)


lightly_studio/src/lightly_studio/dataset/embedding_manager.py
lukas-lightly, line 171: We might consider expanding the definition of the default model. So far we only used images, so the default model was the only model available, the image model.

However, now we can have a default image and a default video model. This function can thus take a parameter representing the model type (image or video). You can note this as a TODO and we can improve it later.

lightly_studio/src/lightly_studio/dataset/embedding_manager.py
```diff
@@ -200,6 +202,15 @@ def load_or_get_default_model(
 
         return embedding_model.embedding_model_id
 
+    def _get_valid_model_id(self, embedding_model_id: UUID | None) -> UUID:
```
lukas-lightly, line 207: Same comment about separate default image and video models applies here too, see comment above.

lightly_studio/src/lightly_studio/dataset/embedding_manager.py
```diff
@@ -119,23 +122,22 @@ def embed_images(
...
-        model_id = embedding_model_id or self._default_model_id
-        if not model_id:
-            raise ValueError("No default embedding model registered.")
+        model_id = self._get_valid_model_id(embedding_model_id)
 
-        if model_id not in self._models:
-            raise ValueError(f"No embedding model found with ID {model_id}")
+        model = self._models[model_id]
+        if not isinstance(model, ImageEmbeddingGenerator):
+            raise ValueError("Embedding model not compatible with images.")
```
lukas-lightly, line 140: Could you add a test for this?

JonasWurst: is part of the next pr

lightly_studio/src/lightly_studio/dataset/embedding_manager.py
JonasWurst: yes, I was also thinking about this. However this might need some more through design. I will add a TODO

lightly_studio/src/lightly_studio/dataset/embedding_manager.py
JonasWurst: will add a TODO as well

PR #261, opened 2025-12-04
Add e2e tests for the captions page

michal-lightly:
## What has changed and why?

Added an e2e test for the captions page.

The test continues from the sample details page which creates a caption that we assert.

Changes:
* Move caption helpers out of SampleDetailsPage so they can be reused in CaptionsPage
* Move the test out of `sample-details` directory

## How has it been tested?

Added a new test.

- [ ] Yes

### DIFF
--- a/lightly_studio_view/e2e/utils.ts
+++ b/lightly_studio_view/e2e/utils.ts
@@ -6,7 +6,6 @@ import {
     SampleDetailsPage,
     CaptionsPage
 } from './pages';
-import { CaptionUtils } from './caption-utils';
 
 export async function gotoFirstPage(page: Page): Promise<void> {
     await page.goto('/');


lightly_studio_view/e2e/general/caption.e2e-test.ts
```diff
@@ -0,0 +1,98 @@
+import { test, expect } from '../utils';
+
+test.describe('caption-flow1', () => {
+    test('Add and edit captions in sample details', async ({
```
michal-lightly: The first two tests are copied from the old test file, with one change - use `captionUtils` instead of `sampleDetailsPage` for caption manipulation.

lightly_studio_view/e2e/general/caption.e2e-test.ts
```diff
@@ -0,0 +1,98 @@
...
+        await captionUtils.deleteNthCaption(0);
+
+        // Check the remaining caption
+        expect(await captionUtils.getCaptionCount()).toEqual(1);
+        expect(await captionUtils.getNthCaption(0).getByTestId('caption-input')).toHaveValue(
+            'another caption'
+        );
+    });
+
+    test('Add and edit captions in caption page', async ({ captionsPage, captionUtils }) => {
```
michal-lightly: New tests here and below.

lightly_studio_view/e2e/pages/sample-details-page.ts
```diff
@@ -84,46 +84,4 @@ export class SampleDetailsPage {
         await this.getLabelSelects().first().click();
         await this.setLabel(label);
     }
-
-    // Captions
```
michal-lightly, line 88: Moved to `caption-utils.e2e-test.ts`

lightly_studio_view/e2e/caption-utils.ts
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    // Get input value of the nth caption. Useful for checking the caption in edit mode.
+    getNthCaptionInput(index: number) {
+        return this.getNthCaption(index).getByTestId('caption-input').inputValue();
+    }
+
+    // TODO(Michal, 12/2025): Change the function signature once it is not possible to
+    // add a caption without text.
+    async addCaption(addButtonIndex: number = 0) {
```
michal-lightly, line 28: Changed: Added button index.

.github/workflows/end2end_test.yml
```diff
@@ -50,12 +50,12 @@ jobs:
...
-                'lightly_studio/src/**',
+                'lightly_studio/e2e-tests/**/*',
+                'lightly_studio/src/**/*',
                 'lightly_studio/pyproject.toml',
-                'lightly_studio_view/e2e/**',
-                'lightly_studio_view/src/**',
-                'lightly_studio_view/public/**',
+                'lightly_studio_view/e2e/**/*',
+                'lightly_studio_view/src/**/*',
+                'lightly_studio_view/public/**/*',
```
michal-lightly, line 58: I had an issue with the cache being hit even though there were changes. Added a star to be sure.

lightly_studio_view/e2e/general/caption.e2e-test.ts
```diff
@@ -0,0 +1,85 @@
...
+        // Start edit mode and delete the first caption
+        await sampleDetailsPage.clickEditButton();
+        await captionUtils.deleteNthCaption(0);
+
+        // Check the remaining caption
+        expect(await captionUtils.getCaptionCount()).toEqual(1);
+        expect(await captionUtils.getNthCaptionInput(0)).toEqual('another caption');
+    });
+
+    test('Add and edit captions in caption page', async ({ captionsPage, captionUtils }) => {
```
LeonardoRosaa: I’m not sure if it should stay here. It could also be considered a different flow:

- First flow: adding and deleting captions from sample details
- Second flow: adding and deleting captions from the captions grid view

We could move it to another file, then Playwright could run the tests in parallel.

lightly_studio_view/e2e/pages/captions-page.ts
```diff
@@ -15,12 +15,20 @@ export class CaptionsPage {
         await this.page.getByTestId('navigation-menu-captions').click();
 
         // Wait for the captions grid to be visible
-        await expect(this.getCaptions().first()).toBeVisible({
+        await expect(this.getNthGridItem(0)).toBeVisible({
             timeout: 10000
```
LeonardoRosaa: I’d suggest reducing this timeout. We should be aware if things are slow, and 10 seconds is too long.

lightly_studio_view/e2e/caption-utils.ts
```diff
@@ -0,0 +1,59 @@
...
+        const captionCountBefore = await this.getCaptionCount();
+        await this.page.getByTestId('add-caption-button').nth(addButtonIndex).click();
+        // Wait until the new caption field appears
+        await expect(this.page.getByTestId('caption-field')).toHaveCount(captionCountBefore + 1);
```
LeonardoRosaa, line 32: Should the assertion be placed here? Keeping assertions out of `.test` files might make tests harder to understand. I would expect assertions to be in the `.test` file rather than in a util file.

lightly_studio_view/e2e/general/caption.e2e-test.ts
michal-lightly: * The problem is that it is not possible to add captions from scratch in the captions page. We have to go to sample details first.
* We don't run e2e tests in parallel for one indexed dataset so that wouldn't be an option.

lightly_studio_view/e2e/caption-utils.ts
michal-lightly: It has to be here.

The purpose is not to assert the number, but to wait until the new caption appear (see the comment). Due to how playwright works this can be done with an assert.

lightly_studio_view/e2e/pages/captions-page.ts
michal-lightly: I copied this from annotations page. I'm not sure if we should reduce that though, I'll wait for an opinion from @mihnea-necsulescu. If to reduce, we could do it for all pages, and maybe set to 5 sec or so.

lightly_studio_view/e2e/pages/sample-details-page.ts
mihnea-necsulescu: I wouldn't completely remove those, just change the implementation to use caption utils (same thing for the captions page, too)

tests should ideally directly interact with page objects, it also makes the intent clearer (i.e., somePage.doSomething() vs utils.doSomething() - you don't really need additional context to understand what should happen with the former)

I suggest composing CaptionUtils both in the sample details page and the captions page 

something like
```ts
export class SampleDetailsPage {
    private captionUtils: CaptionUtils;

    constructor(public readonly page: Page) {
        this.page = page;
        this.captionUtils = new CaptionUtils(page);
    }

    getCaptionCount() {
        return this.captionUtils.getCaptionCount();
    }

...
```

this would add slightly more code, but I think it's worth it, and it would make the page definitions self-documenting (what can actually happen on a specific page)

lightly_studio_view/e2e/utils.ts
```diff
@@ -57,6 +60,11 @@ export const test = base.extend<Pages>({
 
         // Use the fixture value in the test.
         await use(captionsPage);
+    },
+
+    captionUtils: async ({ page }, use) => {
+        const captionUtils = new CaptionUtils(page);
+        await use(captionUtils);
```
mihnea-necsulescu: this is not really a page, but a collection of helper functions; a page fixture should reflect an actual page in the browser

this should be removed, and also line 22

lightly_studio_view/e2e/caption-utils.ts
```diff
@@ -0,0 +1,59 @@
+import { expect, type Page } from '@playwright/test';
```
mihnea-necsulescu, line 1: this file should be renamed - should be just caption-utils.ts

lightly_studio_view/e2e/general/caption.e2e-test.ts
mihnea-necsulescu: I would also keep them separate - either keep the previous caption test file in the sample-details directory and add a new caption-grid test file outside of it, or, if you want to keep them in the same file, they could be split into two `test.describe` groups (e.g., caption-flow-sample-details, caption-flow-captions-grid)

lightly_studio_view/e2e/pages/captions-page.ts
mihnea-necsulescu: I think we did this as a quick workaround for another problem which I don't really remember, but hope to have been fixed 

the default timeout for expect is 5 seconds AFAIK, so I think we should be good - we can remove it

lightly_studio_view/e2e/caption-utils.ts
michal-lightly: I was wondering when I created it - when do we use `.e2e-test.ts` extension?

mihnea-necsulescu (changes requested):
check comments

lightly_studio_view/e2e/caption-utils.ts
mihnea-necsulescu: that should be used for test files only

lightly_studio_view/e2e/utils.ts
```diff
@@ -6,6 +6,7 @@ import {
     SampleDetailsPage,
     CaptionsPage
 } from './pages';
+import { CaptionUtils } from './caption-utils';
```
mihnea-necsulescu: remove this import to please the linter

PR #262, opened 2025-12-04
Embedding Manager: Add Video Embed + tests

JonasWurst:
## What has changed and why?

Adding the video embed method to the embedding manger. adding missing test also for embed images.

## How has it been tested?

unit test

- [ ] Yes

### DIFF
--- a/.github/workflows/end2end_test.yml
+++ b/.github/workflows/end2end_test.yml
@@ -50,7 +50,7 @@ jobs:
             python-build-${{
               hashFiles(
                 '.github/workflows/end2end_test.yml',
-                'lightly_studio/e2e-tests/index_dataset_for_end2end_ui_tests.py',
+                'lightly_studio/e2e-tests/**',
                 'lightly_studio/src/**',
                 'lightly_studio/pyproject.toml',
                 'lightly_studio_view/e2e/**',
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,8 +9,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Added metadata section to video and video frame details.
 - Added tag support for videos in the GUI.
 - Added plugins: This is the initial version for plugins. It supports the execution of operators.
+- `lightly_studio.core.sample.Sample` renamed to `ImageSample`.
 
 ### Changed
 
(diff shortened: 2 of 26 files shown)


lightly_studio/src/lightly_studio/dataset/embedding_manager.py
```diff
@@ -167,6 +169,54 @@ def embed_images(
...
+            raise ValueError("Embedding model not compatible with videos.")
+
+        # Get the samples
+        filepaths = []
+        for sample_id in sample_ids:
+            sample = video_resolver.get_by_id(session=session, sample_id=sample_id)
+            if sample is not None:
+                filepaths.append(sample.file_path_abs)
+
+        assert len(filepaths) == len(sample_ids)
```
lukas-lightly: I think we should rather throw an exception than assert here. If a user provides a wrong UUID, it's a user error, deserving an exception.

lightly_studio/tests/dataset/test_embedding_manager.py
```diff
@@ -211,17 +212,71 @@ def test_embed_images(
...
+    ).embedding_model_id
+    assert manager._get_valid_model_id(embedding_model_id=None) == default_model_id
+
+    other_model_id = manager.register_embedding_model(
+        session=db_session,
+        embedding_generator=RandomEmbeddingGenerator(),
+        dataset_id=dataset.dataset_id,
+        set_as_default=False,
+    ).embedding_model_id
+    assert manager._get_valid_model_id(embedding_model_id=other_model_id) == other_model_id
```
lukas-lightly: The name `_get_valid_model_id` is confusing, now that I saw the test I understand better its sue. It also does validation, checks if the model is present in `self._models`. A better name would be `get_default_or_validate()`. It also doesn't make much sense to check whether `self. _default_model_id` is in `self._models` IMO, we should have performed this check when we assigned these variables, so such a check should always yield `true`.

And another thought, it might be better to split this function into two, as I wrote above, the validation doesn't make sense in the default case. I know the code is really short for it to be split, but it's currently hard to name the function and describe it.

JonasWurst: I renamed the function as suggested. to _get_default_or_validate

I think the name is well suited now.

Furthermore I changed the function slightly. Now if we use the default, it is not checked again. Also the error message is bit more clear now.

PR #263, opened 2025-12-04
Rename Sample to ImageSample

lukas-lightly:
## What has changed and why?
This is an effort to make Sample either a protocol or an abstract class, i.e. to have `class ImageSample(Sample)`. The definition of refactoring of `ImageSample` and addition of `Sample` will come in a later commit.

## How has it been tested?
Existing tests

- [x] Yes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Added metadata section to video and video frame details.
 - Added tag support for videos in the GUI.
 - Added plugins: This is the initial version for plugins. It supports the execution of operators.
 - `lightly_studio.core.sample.Sample` renamed to `ImageSample`.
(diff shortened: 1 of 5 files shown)


lightly_studio/src/lightly_studio/core/sample.py
```diff
@@ -102,7 +102,7 @@ def __init__(self, inner: ImageTable) -> None:
             inner: The ImageTable SQLAlchemy model instance.
         """
         self.inner = inner
-        self._metadata = SampleMetadata(self)
+        self._metadata = ImageSampleMetadata(self)
```
michal-lightly: Metadata are linked to Sample, not Image (also videos should have metadata with the same interface). Should stay SampleMetadata.

lukas-lightly: We don't have a `Sample` protocol or abstract class, so it takes a `ImageSample`, but OK, let's keep the old name and we will fix that interface later. I'll add a TODO.

PR #271, opened 2025-12-08
Export for image captions from GUI

michal-lightly:
## What has changed and why?

Export for image captions from GUI.

Added a new tab to the export dialogue. Copied the functionality for annotations.

Renamed the export tabs to save some space.

## How has it been tested?

- Manually
- Added an e2e test

Screenshot 2025-12-08 at 17 40 10

- [x] Yes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added tag support for videos in the GUI.
 - Added plugins: This is the initial version for plugins. It supports the execution of operators.
 - `lightly_studio.core.sample.Sample` renamed to `ImageSample`.
+- Added cloud storage support for video frames.
 - Added export for image captions.
 
 ### Changed
--- a/lightly_studio/src/lightly_studio/api/routes/api/dataset.py
+++ b/lightly_studio/src/lightly_studio/api/routes/api/dataset.py
@@ -17,6 +17,7 @@ from lightly_studio.api.routes.api.validators import Paginated
 from lightly_studio.db_manager import SessionDep
 from lightly_studio.models.dataset import (
     DatasetCreate,
+    DatasetOverviewView,
     DatasetTable,
     DatasetView,
     DatasetViewWithCount,
@@ -58,6 +59,7 @@ def read_root_dataset(
     return dataset_resolver.get_root_dataset(session=session)
 
 
+# TODO (Mihnea, 12/2025): Update this endpoint to receive a dataset ID.
 @dataset_router.get("/datasets/dataset_hierarchy", response_model=List[DatasetView])
 def read_dataset_hierarchy(
     session: SessionDep,
@@ -65,6 +67,12 @@ def read_dataset_hierarchy(
     """Retrieve the dataset hierarchy from the database, starting with the root node."""
     root_dataset_id = dataset_resolver.get_root_dataset(session=session).dataset_id
     return dataset_resolver.get_hierarchy(session=session, root_dataset_id=root_dataset_id)
+
+
+@dataset_router.get("/datasets/overview", response_model=List[DatasetOverviewView])
+def read_datasets_overview(session: SessionDep) -> list[DatasetOverviewView]:
+    """Retrieve root datasets with metadata for dashboard display."""
+    return dataset_resolver.get_root_datasets_overview(session=session)
 
 
 @dataset_router.get("/datasets/{dataset_id}", response_model=DatasetViewWithCount)
--- a/lightly_studio/src/lightly_studio/models/dataset.py
+++ b/lightly_studio/src/lightly_studio/models/dataset.py
@@ -45,6 +45,16 @@ class DatasetViewWithCount(DatasetView):
     total_sample_count: int
 
 
+class DatasetOverviewView(SQLModel):
+    """Dataset view for dashboard display."""
+
+    dataset_id: UUID
+    name: str
+    sample_type: SampleType
+    created_at: datetime
+    total_sample_count: int
+
+
 class DatasetTable(DatasetBase, table=True):
     """This class defines the Dataset model."""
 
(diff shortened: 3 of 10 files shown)


lightly_studio_view/src/lib/components/ExportSamples/ExportSamples.svelte
```diff
@@ -240,13 +237,31 @@
                         <Button
```
JonasWurst, line 236: Optional: we can remove " along with the corresponding samples" from the annotations tab. Let's keep it as similar as possible

lightly_studio_view/src/lib/components/ExportSamples/ExportSamples.svelte
```diff
@@ -119,9 +115,10 @@
 
             <div class="grid flex-1 gap-4 overflow-y-auto px-1">
                 <Tabs.Root bind:value={exportType} class="w-full">
-                    <Tabs.List class="grid w-full grid-cols-2">
-                        <Tabs.Trigger value="samples">Samples</Tabs.Trigger>
-                        <Tabs.Trigger value="annotations">Samples & Annotations</Tabs.Trigger>
+                    <Tabs.List class="grid w-full grid-cols-3">
+                        <Tabs.Trigger value="samples">Image Filenames</Tabs.Trigger>
```
JonasWurst, line 119: I like this new names. More intuitive!

lightly_studio_view/src/lib/components/ExportSamples/ExportSamples.svelte
```diff
@@ -240,13 +237,31 @@
...
+                    <!-- Captions tab -->
+
+                    <Tabs.Content value="captions" class="pt-2">
+                        <p class="text-sm text-muted-foreground">
+                            The captions will be exported in COCO format.
+                        </p>
+
+                        <Button
+                            class="relative my-4 w-full"
+                            disabled={isSubmitDisabled || $isLoading}
```
JonasWurst: `isSubmitDisabled` will be None anyway, as is can only be true for `exportType="samples"`

same for annotations tab

lightly_studio_view/e2e/general/export-captions.e2e-test.ts
```diff
@@ -0,0 +1,89 @@
+import { expect, test, gotoFirstPage } from '../utils';
+import { cocoDataset } from './fixtures';
+import fs from 'node:fs/promises';
+
+test.describe('Export Captions', () => {
+    test('Add a caption in sample details', async ({ samplesPage, sampleDetailsPage }) => {
```
JonasWurst: Minor: Why should we split this in two? adding a caption should be part of a different test. So we could do this all in one. We can keep the expects ofc

JonasWurst: Note: maybe my understanding of a test is wrong here.

lightly_studio_view/e2e/general/export-annotations.e2e-test.ts
```diff
@@ -11,8 +11,8 @@ test.describe('Export Annotations', () => {
         await page.getByTestId('menu-export').click();
         await expect(page.getByRole('heading', { name: 'Export' })).toBeVisible();
 
-        // Switch to the "Samples & Annotations" tab and wait until the anchor has the href
-        await page.getByRole('tab', { name: 'Samples & Annotations' }).click();
+        // Switch to the right tab and wait until the anchor has the href
```
JonasWurst:
```suggestion
        // Switch to the "Image Annotations" tab and wait until the anchor has the href
```

lightly_studio_view/e2e/general/export-captions.e2e-test.ts
michal-lightly: This is setup for the other test. I need to add a caption to test it is exported

lightly_studio_view/e2e/general/export-captions.e2e-test.ts
```diff
@@ -0,0 +1,89 @@
...
+
+        // Parse and assert COCO-like structure
+        const data = JSON.parse(jsonText);
+        expect(Array.isArray(data.images)).toBeTruthy();
+        expect(Array.isArray(data.annotations)).toBeTruthy();
+
+        // Inspect images
+        expect(data.images.length).toBeGreaterThan(0);
+        const img = data.images[0];
+        expect(img).toHaveProperty('id');
```
ikondrat, line 64: I think it is not best approach considering e2e test, for me e2e test is rather "like user" behaviour

Where is no user checking what img has id property

Now it is confusing why we need to check it? 
What shall we do if id dissappears?

E.g. for me it is unclear what this check is about
`expect(ann.id).toEqual(0);`

ikondrat (changes requested):
I would try to write tests what can help describing the issue, e.g. if it fails it should help to understand the issue

some of the check doing rather opposite for me

E.g.
expect(ann.id).toEqual(0);

if it fails what I should check?

lightly_studio_view/e2e/general/export-captions.e2e-test.ts
michal-lightly: We want to test that the can export captions. The test creates a caption and then downloads the export.

The question is what we should assert in the export. I wanted to assert its whole contents but that's not possible. I can remove the id check though.

lightly_studio_view/e2e/general/export-annotations.e2e-test.ts
michal-lightly: I'd like to avoid repeating the tab name. It is "right" as opposite of "wrong". I'll rephrase to "correct".

lightly_studio_view/e2e/general/export-captions.e2e-test.ts
michal-lightly: I changed the description slightly

ikondrat (approved):

PR #274, opened 2025-12-09
Add embedding to video dataset

JonasWurst:
## What has changed and why?

This adds the embedding execution to the video dataset creation.

## How has it been tested?

unit tests. local testing.

- [ ] Yes

### DIFF
--- a/lightly_studio/src/lightly_studio/core/sample.py
+++ b/lightly_studio/src/lightly_studio/core/sample.py
@@ -11,6 +11,7 @@ from sqlmodel import Session, col
 
 from lightly_studio.models.caption import CaptionCreate
 from lightly_studio.models.image import ImageTable
+from lightly_studio.models.sample import SampleTable
 from lightly_studio.resolvers import caption_resolver, metadata_resolver, tag_resolver
 
 T = TypeVar("T")
@@ -102,7 +103,15 @@ class ImageSample:
             inner: The ImageTable SQLAlchemy model instance.
         """
         self.inner = inner
+        self._sample_table = inner.sample
         self._metadata = SampleMetadata(self)
+
+    @property
+    def sample_table(self) -> SampleTable:
+        """Returns the SampleTable associated with this ImageSample."""
+        # TODO(lukas 12/2025): This should be later removed, as it exposes private implementation.
+        # Remove this once we add a `annotations` property and `embeddings` property.
+        return self._sample_table
 
     def get_object_session(self) -> Session:
         """Get the database session for this sample.
@@ -135,9 +144,9 @@ class ImageSample:
         )
 
         # Add the tag to the sample if not already associated.
-        if tag not in self.inner.sample.tags:
+        if tag not in self.sample_table.tags:
             tag_resolver.add_tag_to_sample(
-                session=session, tag_id=tag.tag_id, sample=self.inner.sample
+                session=session, tag_id=tag.tag_id, sample=self.sample_table
             )
 
     def remove_tag(self, name: str) -> None:
@@ -154,9 +163,9 @@ class ImageSample:
         )
 
         # Remove the tag from the sample if it exists and is associated
-        if existing_tag is not None and existing_tag in self.inner.sample.tags:
+        if existing_tag is not None and existing_tag in self.sample_table.tags:
             tag_resolver.remove_tag_from_sample(
-                session=session, tag_id=existing_tag.tag_id, sample=self.inner.sample
+                session=session, tag_id=existing_tag.tag_id, sample=self.sample_table
             )
 
     @property
@@ -166,7 +175,7 @@ class ImageSample:
         Returns:
             A set of tag names as strings.
         """
-        return {tag.name for tag in self.inner.sample.tags}
+        return {tag.name for tag in self.sample_table.tags}
 
     @tags.setter
     def tags(self, tags: Iterable[str]) -> None:
@@ -205,7 +214,7 @@ class ImageSample:
         Returns:
             The UUID of the dataset.
         """
-        return self.inner.sample.dataset_id
+        return self.sample_table.dataset_id
 
     def add_caption(self, text: str) -> None:
         """Add a caption to this sample.
@@ -228,7 +237,7 @@ class ImageSample:
     @property
     def captions(self) -> list[str]:
         """Returns the text of all captions."""
-        return [caption.text for caption in self.inner.sample.captions]
+        return [caption.text for caption in self.sample_table.captions]
 
     @captions.setter
     def captions(self, captions: Iterable[str]) -> None:
@@ -240,7 +249,7 @@ class ImageSample:
         session = self.get_object_session()
 
         # Delete all existing captions for this sample
-        caption_sample_ids = [c.sample_id for c in self.inner.sample.captions]
+        caption_sample_ids = [c.sample_id for c in self.sample_table.captions]
         for caption_sample_id in caption_sample_ids:
             caption_resolver.delete_caption(session=session, sample_id=caption_sample_id)
 
@@ -276,9 +285,9 @@ class SampleMetadata:
         Returns:
             The metadata value for the given key, or None if the key doesn't exist.
         """
-        if self._sample.inner.sample.metadata_dict is None:
+        if self._sample.sample_table.metadata_dict is None:
             return None
-        return self._sample.inner.sample.metadata_dict.get_value(key)
+        return self._sample.sample_table.metadata_dict.get_value(key)
 
     def __setitem__(self, key: str, value: Any) -> None:
         """Set a metadata key-value pair.
--- a/lightly_studio/src/lightly_studio/export/coco_captions.py
+++ b/lightly_studio/src/lightly_studio/export/coco_captions.py
@@ -53,12 +53,12 @@ def to_coco_captions_dict(samples: Iterable[ImageSample]) -> CocoCaptionsJson:
                 "height": image.height,
             }
         )
-        for caption in image.inner.sample.captions:
+        for caption in image.captions:
             coco_annotations.append(
                 {
                     "id": annotation_id,
                     "image_id": image_id,
-                    "caption": caption.text,
+                    "caption": caption,
                 }
             )
             annotation_id += 1
--- a/lightly_studio/src/lightly_studio/export/lightly_studio_label_input.py
+++ b/lightly_studio/src/lightly_studio/export/lightly_studio_label_input.py
@@ -93,7 +93,7 @@ def _sample_to_image_obj_det(
             annotation=annotation,
             label_id_to_category=label_id_to_category,
         )
-        for annotation in sample.inner.sample.annotations
+        for annotation in sample.sample_table.annotations
         if annotation.annotation_type == AnnotationType.OBJECT_DETECTION
     ]
     return ImageObjectDetection(
--- a/lightly_studio/tests/core/test_dataset__coco.py
+++ b/lightly_studio/tests/core/test_dataset__coco.py
@@ -68,12 +68,12 @@ class TestDataset:
         assert len(samples) == 2
         assert {s.file_name for s in samples} == {"image1.jpg", "image2.jpg"}
         assert all(
-            len(s.inner.sample.embeddings) == 1 for s in samples
+            len(s.sample_table.embeddings) == 1 for s in samples
         )  # Embeddings should be generated
 
         # Verify the first sample and annotation
-        bbox = samples[0].inner.sample.annotations[0].object_detection_details
-        annotation = samples[0].inner.sample.annotations[0].annotation_label
+        bbox = samples[0].sample_table.annotations[0].object_detection_details
+        annotation = samples[0].sample_table.annotations[0].annotation_label
         assert isinstance(bbox, ObjectDetectionAnnotationTable)
         assert bbox.height == 200.0
         assert bbox.width == 200.0
@@ -82,8 +82,8 @@ class TestDataset:
         assert annotation.annotation_label_name == "cat"
 
         # Verify the second sample and annotation
-        bbox = samples[1].inner.sample.annotations[0].object_detection_details
-        annotation = samples[1].inner.sample.annotations[0].annotation_label
+        bbox = samples[1].sample_table.annotations[0].object_detection_details
+        annotation = samples[1].sample_table.annotations[0].annotation_label
         assert isinstance(bbox, ObjectDetectionAnnotationTable)
         assert bbox.height == 250.0
         assert bbox.width == 250.0
@@ -405,7 +405,7 @@ class TestDataset:
 
         # Check that an embedding was not created
         samples = list(dataset)
-        assert all(len(sample.inner.sample.embeddings) == 0 for sample in samples)
+        assert all(len(sample.sample_table.embeddings) == 0 for sample in samples)
 
     def test_add_samples_from_coco__tags_created_for_split(
         self,
--- a/lightly_studio/tests/core/test_dataset__coco_caption.py
+++ b/lightly_studio/tests/core/test_dataset__coco_caption.py
@@ -34,13 +34,11 @@ class TestDataset:
         assert len(samples) == 2
         assert {s.file_name for s in samples} == {"image1.jpg", "image2.jpg"}
         assert all(
-            len(s.inner.sample.embeddings) == 1 for s in samples
+            len(s.sample_table.embeddings) == 1 for s in samples
         )  # Embeddings should be generated
 
         # Assert captions
-        captions_map = {
-            s.file_name: sorted(c.text for c in s.inner.sample.captions) for s in samples
-        }
+        captions_map = {s.file_name: sorted(s.captions) for s in samples}
 
         assert captions_map == {
             "image1.jpg": ["Caption 1 of image 1", "Caption 2 of image 1"],
@@ -118,7 +116,7 @@ class TestDataset:
 
         # Check that an embedding was not created
         samples = list(dataset)
-        assert all(len(sample.inner.sample.embeddings) == 0 for sample in samples)
+        assert all(len(sample.sample_table.embeddings) == 0 for sample in samples)
 
 
 def _create_sample_images(image_paths: list[Path]) -> None:
--- a/lightly_studio/tests/core/test_dataset__labelformat.py
+++ b/lightly_studio/tests/core/test_dataset__labelformat.py
@@ -153,12 +153,12 @@ class TestDataset:
         samples = sorted(samples, key=lambda sample: sample.file_path_abs)
 
         # Verify first image and annotation
-        annotation = samples[0].inner.sample.annotations[0].annotation_label
+        annotation = samples[0].sample_table.annotations[0].annotation_label
         assert samples[0].file_name == "001.jpg"
         assert annotation.annotation_label_name == "dog"
 
         # Verify first image and annotation
-        annotation = samples[1].inner.sample.annotations[0].annotation_label
+        annotation = samples[1].sample_table.annotations[0].annotation_label
         assert samples[1].file_name == "020.jpg"
         assert annotation.annotation_label_name == "cat"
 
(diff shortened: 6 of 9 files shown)


lightly_studio/src/lightly_studio/dataset/embedding_manager.py
```diff
@@ -250,51 +257,88 @@ def load_or_get_default_model(
...
 
 
 # TODO(Michal, 09/2025): Write tests for this function.
-def _load_embedding_generator_from_env() -> EmbeddingGenerator | None:
+def _load_embedding_generator_from_env(sample_type: SampleType) -> EmbeddingGenerator | None:
     """Load the embedding generator based on environment variable configuration."""
+    if sample_type == SampleType.IMAGE:
+        return _load_image_embedding_generator_from_env()
+    if sample_type == SampleType.VIDEO:
+        return _load_video_embedding_generator_from_env()
```
michal-lightly: The function should have a different function. There is only one embedding model for videos, we do not need to load anything from env

lightly_studio/src/lightly_studio/dataset/embedding_manager.py
```diff
@@ -60,13 +61,14 @@ class EmbeddingManager:
...
         self._models: dict[UUID, EmbeddingGenerator] = {}
-        self._default_model_id: UUID | None = None
+        self._default_model_id: dict[SampleType, UUID | None] = {}
 
     def register_embedding_model(
         self,
         session: Session,
         dataset_id: UUID,
         embedding_generator: EmbeddingGenerator,
+        sample_type: SampleType = SampleType.IMAGE,
```
michal-lightly: I would consider adding this to the db model

lightly_studio/src/lightly_studio/api/routes/api/text_embedding.py
```diff
@@ -35,9 +38,15 @@ def embed_text(
     ] = None,
 ) -> list[float]:
     """Retrieve embeddings for the input text."""
+    # TODO(Jonas, 12/2025): Remove this hack after model_id is provided from frontend
```
michal-lightly:
```suggestion
    # TODO(Jonas, 12/2025): Remove this hack after dataset_id is provided from frontend
```

lightly_studio/src/lightly_studio/api/routes/api/text_embedding.py
```diff
@@ -35,9 +38,15 @@ def embed_text(
...
+    # This is a hack, since at the moment, no valid embedding_model_id is passed from the frontend.
+    # so we fetch the root_dataset_id, which will be used inside embed_text to get the default model
+    # for this dataset.
+    root_dataset = dataset_resolver.get_root_dataset(session=session)
+    dataset_id = root_dataset.dataset_id
     try:
         text_embeddings = embedding_manager.embed_text(
-            TextEmbedQuery(query_text, embedding_model_id)
+            dataset_id=dataset_id, text_query=TextEmbedQuery(query_text, embedding_model_id)
```
michal-lightly: Minor: It would be nice to add the kwargs if you are touching the line:
```suggestion
            dataset_id=dataset_id, text_query=TextEmbedQuery(text=query_text, embedding_model_id=embedding_model_id)
```

lightly_studio/src/lightly_studio/core/dataset.py
```diff
@@ -647,16 +660,51 @@ def _generate_embeddings(session: Session, dataset_id: UUID, sample_ids: list[UU
...
+        sample_ids=sample_ids,
+        embedding_model_id=model_id,
     )
+
+    # Mark the embedding search feature as enabled.
+    if "embeddingSearchEnabled" not in features.lightly_studio_active_features:
+        features.lightly_studio_active_features.append("embeddingSearchEnabled")
+    # Mark the FSC feature as enabled.
+    if "fewShotClassifierEnabled" not in features.lightly_studio_active_features:
+        features.lightly_studio_active_features.append("fewShotClassifierEnabled")
```
michal-lightly: Let's move this into a helper function

lightly_studio/src/lightly_studio/core/dataset.py
```diff
@@ -647,16 +660,51 @@ def _generate_embeddings(session: Session, dataset_id: UUID, sample_ids: list[UU
         return
 
     embedding_manager = EmbeddingManagerProvider.get_embedding_manager()
-    model_id = embedding_manager.load_or_get_default_model(
+    model_id = embedding_manager.load_or_get_default_model(session=session, dataset_id=dataset_id)
+    if model_id is None:
+        logger.warning("No embedding model loaded. Skipping embedding generation.")
+        return
+
+    embedding_manager.embed_videos(
```
michal-lightly, line 668: It is still not clear to me whether this is the easiest solution.

An alternative would be:
* Add `embed_samples` to embedding_manager. It chooses the right embed function based on dataset sample type
* Keep only `_generate_embeddings` in this file

What is your opinion? I don't feel strongly about it.

lightly_studio/src/lightly_studio/dataset/embedding_manager.py
```diff
@@ -60,7 +62,7 @@ class EmbeddingManager:
     def __init__(self) -> None:
         """Initialize the embedding manager."""
         self._models: dict[UUID, EmbeddingGenerator] = {}
-        self._default_model_id: UUID | None = None
+        self._default_model_id: dict[UUID, UUID | None] = {}
```
michal-lightly: The meaning of keys and values is not clear. Let's rename the variable to `_dataset_id_to_default_model_id`.

lightly_studio/src/lightly_studio/dataset/embedding_manager.py
```diff
@@ -94,48 +96,55 @@ def register_embedding_model(
...
-    def embed_text(self, text_query: TextEmbedQuery) -> list[float]:
+    def embed_text(
+        self,
+        dataset_id: UUID,
+        text_query: TextEmbedQuery,
+    ) -> list[float]:
         """Generate an embedding for a text sample.
 
         Args:
+            dataset_id: The ID of the dataset to associate with the model.
```
michal-lightly: The comment should be expanded (multiple sentences are ok). I though this is true but it is not accurate either.
```suggestion
            dataset_id: The ID of the dataset associated with the model.
```

lightly_studio/src/lightly_studio/dataset/embedding_manager.py
```diff
@@ -234,17 +246,22 @@ def load_or_get_default_model(
...
         """
         # Return the existing default model ID if available.
-        # TODO(Michal, 09/2025): We do not check if the model belongs to the dataset.
-        # The design of EmbeddingManager needs to change to support multiple datasets.
-        if self._default_model_id is not None:
-            return self._default_model_id
 
-        # Load the embedding generator based on configuration.
-        embedding_generator = _load_embedding_generator_from_env()
+        if self._default_model_id.get(dataset_id, None) is not None:
```
michal-lightly: Style:
```suggestion
        if dataset_id in self._default_model_id:
```

lightly_studio/src/lightly_studio/dataset/embedding_manager.py
```diff
@@ -254,47 +271,81 @@ def load_or_get_default_model(
 
         return embedding_model.embedding_model_id
 
-    # TODO (Jonas 12/2025): We need to introduce default models per type
-    def _get_default_or_validate(self, embedding_model_id: UUID | None) -> UUID:
-        if embedding_model_id is None and self._default_model_id is None:
+    def _get_default_or_validate(self, dataset_id: UUID, embedding_model_id: UUID | None) -> UUID:
```
michal-lightly, line 272: Please add a docstring.

lightly_studio/src/lightly_studio/dataset/embedding_manager.py
```diff
@@ -254,47 +271,81 @@ def load_or_get_default_model(
...
-            return self._default_model_id
+        if embedding_model_id is None and default_model_id is not None:
+            return default_model_id
 
         if embedding_model_id not in self._models:
             raise ValueError(f"No embedding model found with ID {embedding_model_id}")
         return embedding_model_id
 
 
 # TODO(Michal, 09/2025): Write tests for this function.
```
michal-lightly: Please move the todo to `_load_image_embedding_generator_from_env`.

lightly_studio/src/lightly_studio/dataset/embedding_manager.py
```diff
@@ -254,47 +271,81 @@ def load_or_get_default_model(
...
+        except ImportError:
+            logger.warning("Embedding functionality is disabled.")
+    else:
+        logger.warning(f"Unsupported model type: '{env.LIGHTLY_STUDIO_EMBEDDINGS_MODEL_TYPE}'")
+        logger.warning("Embedding functionality is disabled.")
+    return None
+
+
+def _load_video_embedding_generator_from_env() -> EmbeddingGenerator | None:
+    if env.LIGHTLY_STUDIO_VIDEO_EMBEDDINGS_MODEL_TYPE == "PE":
```
michal-lightly: I would suggest to remove the env variable. We can introduce complexity once we need it, but it is not anytime soon.

Also let's rename the function to `_load_video_embedding_generator`.

lightly_studio/src/lightly_studio/dataset/embedding_manager.py
```diff
@@ -254,47 +271,81 @@ def load_or_get_default_model(
...
+            return PerceptionEncoderEmbeddingGenerator()
+        except ImportError:
+            logger.warning("Embedding functionality is disabled.")
+    else:
+        logger.warning(f"Unsupported model type: '{env.LIGHTLY_STUDIO_EMBEDDINGS_MODEL_TYPE}'")
+        logger.warning("Embedding functionality is disabled.")
+    return None
+
+
+def _load_video_embedding_generator_from_env() -> EmbeddingGenerator | None:
```
michal-lightly: The retun type should be specific:

```suggestion
def _load_video_embedding_generator_from_env() -> VideoEmbeddingGenerator | None:
```

Same for image.

lightly_studio/src/lightly_studio/dataset/env.py
```diff
@@ -9,6 +9,9 @@
 LIGHTLY_STUDIO_EMBEDDINGS_MODEL_TYPE: str = env.str(
     "LIGHTLY_STUDIO_EMBEDDINGS_MODEL_TYPE", "MOBILE_CLIP"
 )
+LIGHTLY_STUDIO_VIDEO_EMBEDDINGS_MODEL_TYPE: str = env.str(
```
michal-lightly: I suggest to remove the env variable

lightly_studio/src/lightly_studio/resolvers/embedding_model_resolver.py
```diff
@@ -34,8 +34,6 @@ def get_or_create(session: Session, embedding_model: EmbeddingModelCreate) -> Em
         db_model.name != embedding_model.name
         or db_model.parameter_count_in_mb != embedding_model.parameter_count_in_mb
         or db_model.embedding_dimension != embedding_model.embedding_dimension
-        # TODO(Michal, 09/2025): Allow same model for different datasets.
-        or db_model.dataset_id != embedding_model.dataset_id
```
michal-lightly, line 38: This is a plot twist: EmbeddingModel has dataset_id.

It is not clear what is the intended use. Currently it is legacy because we were removing dataset_ids from tables.

There are two possible interpretations:
* It is the "root dataset id" - with multiple (root) datasets the model belongs only to a single one.
* It is a "collection id" - the model belongs to a collection

I think the first interpretation makes more sense at the moment. WDYT? What do you think we should do with the field?

lightly_studio/tests/api/routes/api/test_text_embedding.py
```diff
@@ -11,11 +12,16 @@
...
 
 
-def test_embed_text(mocker: MockerFixture, test_client: TestClient) -> None:
+def test_embed_text(db_session: Session, mocker: MockerFixture, test_client: TestClient) -> None:
     # Initialize the embedding_manager with a mock variant so it does not update
     # the singleton.
+    # Create a db and fill with some samples, as the text_embeddings defaults to root_dataset
+    helpers_resolvers.fill_db_with_samples_and_embeddings(
+        test_db=db_session, n_samples=10, embedding_model_names=["test_embedding_model"]
+    )
```
michal-lightly: Wouldn't creating a dataset suffice? We should avoid setting up unnecessary data in tests.

lightly_studio/tests/api/routes/api/test_text_embedding.py
```diff
@@ -40,10 +46,16 @@ def test_embed_text(mocker: MockerFixture, test_client: TestClient) -> None:
...
 def test_embed_text_embedding_invalid_model_id(
+    db_session: Session,
     mocker: MockerFixture,
     test_client: TestClient,
 ) -> None:
     # Make the request to the `/samples` endpoint
+    # Create a db and fill with some samples, as the text_embeddings defaults to root_dataset
+    helpers_resolvers.fill_db_with_samples_and_embeddings(
+        test_db=db_session, n_samples=10, embedding_model_names=["test_embedding_model"]
+    )
```
michal-lightly: The same comment.

lightly_studio/src/lightly_studio/core/dataset.py
```diff
@@ -635,7 +644,11 @@ def compute_similarity_metadata(
         )
 
 
-def _generate_embeddings(session: Session, dataset_id: UUID, sample_ids: list[UUID]) -> None:
+def _generate_embeddings_video(
```
michal-lightly, line 647: Optional:`_generate_video_embeddings`

lightly_studio/src/lightly_studio/resolvers/embedding_model_resolver.py
JonasWurst: I think it makes sense to have this mapping (as discussed previously). However, I think the most scalable and future proof way would be to use the second idea but allow multiple collections. Example: for groups, you could assign the embedding model to all image collections

At the moment it is not used at all.

I recommend we look into the model per collection registry as soon as the groups have matured?

lightly_studio/src/lightly_studio/core/dataset.py
JonasWurst: I would keep it like this as the other one is called similar,

lightly_studio/src/lightly_studio/core/dataset.py
JonasWurst: If we push it to emebd_samples this would introduce additonal logic there. This way we have very little duplication.

I'd prefer the slightly duplicated version instead of adding more logic.

lightly_studio/src/lightly_studio/resolvers/embedding_model_resolver.py
michal-lightly: I think it makes sense to look into this a bit later.

lightly_studio/src/lightly_studio/core/dataset.py
michal-lightly: The other one should be renamed too :) The name was chosen in this PR.

lightly_studio/src/lightly_studio/dataset/embedding_manager.py
```diff
@@ -60,7 +62,7 @@ class EmbeddingManager:
     def __init__(self) -> None:
         """Initialize the embedding manager."""
         self._models: dict[UUID, EmbeddingGenerator] = {}
-        self._default_model_id: UUID | None = None
+        self._dataset_id_to_default_model_id: dict[UUID, UUID | None] = {}
```
michal-lightly: Minor: I noticed now that this should be:

```suggestion
        self._dataset_id_to_default_model_id: dict[UUID, UUID] = {}
```
If there is no deault then simply the key is missing.

lightly_studio/src/lightly_studio/dataset/embedding_manager.py
```diff
@@ -94,48 +97,53 @@ def register_embedding_model(
         self._models[model_id] = embedding_generator
 
         # Set as default if requested or if it's the first model
-        if set_as_default or self._default_model_id is None:
-            self._default_model_id = model_id
+        if set_as_default or self._dataset_id_to_default_model_id.get(dataset_id, None) is None:
```
michal-lightly:
```suggestion
        if set_as_default or dataset_id not in self._dataset_id_to_default_model_id:
```

lightly_studio/src/lightly_studio/dataset/embedding_manager.py
```diff
@@ -234,17 +246,21 @@ def load_or_get_default_model(
...
+        dataset = dataset_resolver.get_by_id(session=session, dataset_id=dataset_id)
+        if dataset is None:
+            raise ValueError("Provided dataset_id could not be found.")
+
+        embedding_generator = _load_embedding_generator_from_env(sample_type=dataset.sample_type)
         if embedding_generator is None:
             return None
 
         # Register the embedding model and set it as default.
+        # same config already is registered
```
michal-lightly: Minor: Seems like part of the comment got lost.

lightly_studio/src/lightly_studio/dataset/embedding_manager.py
```diff
@@ -254,47 +270,82 @@ def load_or_get_default_model(
...
 
-    # TODO (Jonas 12/2025): We need to introduce default models per type
-    def _get_default_or_validate(self, embedding_model_id: UUID | None) -> UUID:
-        if embedding_model_id is None and self._default_model_id is None:
+    def _get_default_or_validate(self, dataset_id: UUID, embedding_model_id: UUID | None) -> UUID:
+        """Get a valid model_id or raise error of non available.
+
+        Return the default embedding model that is registered for the dataset_id,
+        if the embedding_model_id is not provided.
+        Return the embedding_model_id if it is associated to a registered model.
```
michal-lightly: Optional:
```suggestion
        If embedding_model_id is not provided, returns the default model for dataset_id.
        If embedding_model_id is provided, validates that the model has been loaded and returns it.
```

Actually this function as very strange responsibilities, it is difficult to explain in the coment.

lightly_studio/tests/api/routes/api/test_text_embedding.py
```diff
@@ -11,11 +12,14 @@
     EmbeddingManager,
     EmbeddingManagerProvider,
 )
+from tests import helpers_resolvers
 
 
-def test_embed_text(mocker: MockerFixture, test_client: TestClient) -> None:
+def test_embed_text(db_session: Session, mocker: MockerFixture, test_client: TestClient) -> None:
     # Initialize the embedding_manager with a mock variant so it does not update
     # the singleton.
```
michal-lightly, line 23: Minor: This comment is relevant to the `mocker.patch.object` line, should be a bit lower.

lightly_studio/tests/api/routes/api/test_text_embedding.py
```diff
@@ -11,11 +12,14 @@
...
+    # Create a db and fill with some samples, as the text_embeddings defaults to root_dataset
```
michal-lightly: We dont fill it with samples, the comment can be simplified.

lightly_studio/tests/api/routes/api/test_text_embedding.py
```diff
@@ -40,10 +44,14 @@ def test_embed_text(mocker: MockerFixture, test_client: TestClient) -> None:
 
 
 def test_embed_text_embedding_invalid_model_id(
+    db_session: Session,
     mocker: MockerFixture,
     test_client: TestClient,
 ) -> None:
     # Make the request to the `/samples` endpoint
+    # Create a db and fill with some samples, as the text_embeddings defaults to root_dataset
```
michal-lightly: Same

lightly_studio/src/lightly_studio/dataset/embedding_manager.py
JonasWurst: This comment explains it better.

PR #277, opened 2025-12-09
Add our own log formatter

lukas-lightly:
## What has changed and why?
Our own logs were not visible, at least when running the examples. Also we've used the default formatting, which is not the best.

- [ ] Yes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added plugins: This is the initial version for plugins. It supports the execution of operators.
 - `lightly_studio.core.sample.Sample` renamed to `ImageSample`.
 - Added cloud storage support for video frames.
+- Added export for image captions.
 
 ### Changed
 
--- a/lightly_studio/src/lightly_studio/models/annotation/annotation_base.py
+++ b/lightly_studio/src/lightly_studio/models/annotation/annotation_base.py
@@ -161,6 +161,14 @@ class AnnotationViewsWithCount(BaseModel):
     next_cursor: Optional[int] = PydanticField(..., alias="nextCursor")
 
 
+class SampleAnnotationView(BaseModel):
+    """Response model for sample annotation view."""
+
+    model_config = ConfigDict(populate_by_name=True)
+
+    dataset_id: UUID
+
+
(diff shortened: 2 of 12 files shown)


lightly_studio/src/lightly_studio/setup_logging.py
```diff
@@ -9,10 +9,48 @@
...
+        formatter = logging.Formatter(log_fmt)
+        return formatter.format(record)
+
 
 # Set logging level to ERROR for labelformat.
 logging.getLogger("labelformat").setLevel(logging.ERROR)
 
+# Configure the root logger with ConsoleFormatter
+root_logger = logging.getLogger()
+if not root_logger.handlers:
```
lukas-lightly, line 49: Note: this `if not ...` was added by AI and I don't fully understand it. I'm trying to figure it out now.

lukas-lightly: This prevents log messages appearing twice in the output, as was the case in my previous attempts to solve this.

If someone sets up a root logger handler, we would not add our own. It will be redundant.

michal-lightly: Let's add this info to a comment?

lightly_studio/src/lightly_studio/setup_logging.py
```diff
@@ -9,10 +9,48 @@
...
+
+class ConsoleFormatter(logging.Formatter):
+    """Customer console formatter for Lightly Studio."""
+
+    green = "\x1b[33;32m"
+    yellow = "\x1b[33;20m"
+    red = "\x1b[31;20m"
+    bold_red = "\x1b[31;1m"
+    reset = "\x1b[0m"
+    time_str = "%(asctime)s "
```
michal-lightly, line 23: Optional: It would be nice if the timestamp was muted (e.g. gray). Or just another color.

lightly_studio/src/lightly_studio/setup_logging.py
lukas-lightly: Done

lightly_studio/src/lightly_studio/setup_logging.py
lukas-lightly: I can try different options. Will merge this now and I'll experiment with it tomorrow.

PR #281, opened 2025-12-10
Do not print uvicorn access log by default

lukas-lightly:
## What has changed and why?
Only if LIGHTLY_STUDIO_DEBUG is set to true.

## How has it been tested?
Locally

- [ ] Yes

### DIFF
--- a/lightly_studio/src/lightly_studio/api/app.py
+++ b/lightly_studio/src/lightly_studio/api/app.py
@@ -63,8 +63,8 @@ async def lifespan(_: FastAPI) -> AsyncGenerator[None, None]:
 if LIGHTLY_STUDIO_DEBUG:
     import logging
 
-    # TODO: move this into setup_logging.py, drop `basicConfig()`. Also everything seems to be on
-    # the INFO level. `logging.DEBUG` maybe doesn't make a difference.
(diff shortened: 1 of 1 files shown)


lukas-lightly:
The output is not great when using `LIGHTLY_STUDIO_DEBUG`, but at least the logs are silented otherwise:
image

Main problems:
* Different log formatting, our own logs have timestamp, logs from `uvicorn` don't.
* Every log message is INFO-level, which is wrong, these are clearly debug messages. We probably can't do much about uvicorn, unless the upstream is patched.

lightly_studio/src/lightly_studio/api/app.py
```diff
@@ -63,6 +63,8 @@ async def lifespan(_: FastAPI) -> AsyncGenerator[None, None]:
 if LIGHTLY_STUDIO_DEBUG:
     import logging
 
+    # TODO: move this into setup_logging.py, drop `basicConfig()`. Also everything seems to be on
```
michal-lightly:
```suggestion
    # TODO(Lukas, 12/2025): move this into setup_logging.py, drop `basicConfig()`. Also everything seems to be on
```

michal-lightly (approved):

PR #283, opened 2025-12-10
Introduce ImageSample.sample_table

lukas-lightly:
## What has changed and why?
This is an important refactoring, so that we do not access `.inner.sample` directly at multiple places in the code.

* It will later allow a generalization, having a common `Sample` abstract class, e.g. shared with `VideoSample`.
* The `.inner.sample` access would not be allowed in other languages, as it goes directly to the private data.

## How has it been tested?
Existing tests

- [ ] Yes

### DIFF
--- a/lightly_studio/src/lightly_studio/core/sample.py
+++ b/lightly_studio/src/lightly_studio/core/sample.py
@@ -11,6 +11,7 @@ from sqlmodel import Session, col
 
 from lightly_studio.models.caption import CaptionCreate
 from lightly_studio.models.image import ImageTable
+from lightly_studio.models.sample import SampleTable
 from lightly_studio.resolvers import caption_resolver, metadata_resolver, tag_resolver
 
 T = TypeVar("T")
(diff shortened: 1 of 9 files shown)


lightly_studio/src/lightly_studio/export/lightly_studio_label_input.py
```diff
@@ -93,7 +93,7 @@ def _sample_to_image_obj_det(
             annotation=annotation,
             label_id_to_category=label_id_to_category,
         )
-        for annotation in sample.inner.sample.annotations
+        for annotation in sample.sample_table.annotations
```
lukas-lightly, line 96: Notet that the need to access `sample.inner.sample` (or `sample.sample_table`), which is internal data showcases the shortcomings of our API.
I'll try to add a `annotations` property soon, so that this goes away completely.

lightly_studio/src/lightly_studio/core/sample.py
```diff
@@ -102,8 +103,16 @@ def __init__(self, inner: ImageTable) -> None:
             inner: The ImageTable SQLAlchemy model instance.
         """
         self.inner = inner
```
michal-lightly, line 105: Originally `inner` was planned to be `_inner`. It caused some issues though because there where some places where sample_table access was necessary. It would be nice if it could be private in the future.

PR #284, opened 2025-12-10
Create an abstract `Sample` class

lukas-lightly:
## What has changed and why?
The only implementation is `ImageSample`, but `VideoSample` will be soon added.

This is an important step towards different `*Sample` classes.

## How has it been tested?
Existing tests.

- [ ] Yes

### DIFF
--- a/lightly_studio/src/lightly_studio/core/dataset.py
+++ b/lightly_studio/src/lightly_studio/core/dataset.py
@@ -272,6 +272,7 @@ class Dataset(Generic[T]):
         path: PathLike,
         allowed_extensions: Iterable[str] | None = None,
         num_decode_threads: int | None = None,
+        embed: bool = True,
     ) -> None:
         """Adding video frames from the specified path to the dataset.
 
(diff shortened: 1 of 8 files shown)


lightly_studio/src/lightly_studio/core/dataset.py
```diff
@@ -57,10 +57,10 @@
 _SliceType = slice  # to avoid shadowing built-in slice in type annotations
 
 
-T = TypeVar("T", default=ImageSample)
+S = TypeVar("S", default=ImageSample, bound=Sample)
```
michal-lightly: `T` is customary name in python for a type variable. It is called differently only if more typevars are needed in the same scope.

lightly_studio/src/lightly_studio/core/sample.py
```diff
@@ -264,12 +252,41 @@ def captions(self, captions: Iterable[str]) -> None:
             )
 
 
+class ImageSample(Sample):
```
michal-lightly: Food for thought: Should we split this into a separate file already now / later / never? In my experience, things are reluctant to move unless created in a different location in the beginning.

JonasWurst: +1 for move

lightly_studio/src/lightly_studio/core/dataset.py
lukas-lightly: Done

lightly_studio/src/lightly_studio/core/sample.py
lukas-lightly: Moved under `image_sample.py`

PR #288, opened 2025-12-11
Pass dataset_id to embedText API endpoint

michal-lightly:
## What has changed and why?

Pass dataset_id to embedText API endpoint.

Changed the embedding logic to use the tanstack-generated function.

## How has it been tested?

Existing tests and manually.

- [ ] Yes

### DIFF
--- a/lightly_studio_view/src/routes/datasets/[dataset_id]/videos/+page.svelte
+++ b/lightly_studio_view/src/routes/datasets/[dataset_id]/videos/+page.svelte
@@ -1,6 +1,5 @@
 <script lang="ts">
-    import { ImageSizeControl, LazyTrigger, Spinner, SelectableBox } from '$lib/components';
-    import Separator from '$lib/components/ui/separator/separator.svelte';
+    import { LazyTrigger, Spinner, SelectableBox } from '$lib/components';
     import { useVideos } from '$lib/hooks/useVideos/useVideos';
     import { page } from '$app/stores';
     import { Grid } from 'svelte-virtual';
@@ -71,17 +70,6 @@
 </script>
 
 <div class="flex flex-1 flex-col space-y-4">
-    <div class="my-2 flex items-center space-x-4">
-        <div class="flex-1">
-            <div class="text-2xl font-semibold">Videos</div>
-        </div>
-
-        <div class="w-4/12">
-            <ImageSizeControl />
-        </div>
-    </div>
-    <Separator class="mb-4 bg-border-hard" />
-
     <div class="h-full w-full flex-1 overflow-hidden" bind:this={viewport} bind:clientWidth>
         {#if $query.isPending && items.length === 0}
             <div class="flex h-full w-full items-center justify-center">


lightly_studio_view/src/routes/datasets/[dataset_id]/+layout.svelte
```diff
@@ -116,32 +116,44 @@
...
+            const trimmedQuery = query_text.trim();
+            if (!trimmedQuery) return;
 
-            return response.data;
-        } catch (error) {
-            setError((error as unknown as Error).message);
-            console.error('Error during API call:', error);
+            submittedQueryText = trimmedQuery;
+            await tick();
+            get(embedTextQuery)?.refetch?.();
```
Comment (mihnea-necsulescu) [inline lightly_studio_view/src/routes/datasets/[dataset_id]/+layout.svelte]: those shouldn't be needed; $derived should reevaluate embedTextQuery with the new query text when it changes

lightly_studio_view/src/routes/datasets/[dataset_id]/+layout.svelte
```diff
@@ -116,32 +116,44 @@
...
+        if (!submittedQueryText) return;
 
+        if ($embedTextQuery.isError && $embedTextQuery.error) {
+            const queryError = $embedTextQuery.error as { error?: unknown } | Error;
+            const message = 'error' in queryError ? queryError.error : queryError.message;
+            setError(String(message));
+            return;
+        }
+
+        if ($embedTextQuery.data) {
```
Comment (mihnea-necsulescu) [inline lightly_studio_view/src/routes/datasets/[dataset_id]/+layout.svelte]: I think this should also check if $embedTextQuery.isSuccess; the data may be stale otherwise

lightly_studio_view/src/routes/datasets/[dataset_id]/+layout.svelte
```diff
@@ -116,32 +116,44 @@
...
+            submittedQueryText = trimmedQuery;
+            await tick();
+            get(embedTextQuery)?.refetch?.();
         }
     }
 
-    async function onKeyDown(event: KeyboardEvent) {
-        if (event.key === 'Enter') {
-            const textEmbedding = await handleTextEmbeddingSearch();
+    $effect(() => {
```
Comment (mihnea-necsulescu) [inline lightly_studio_view/src/routes/datasets/[dataset_id]/+layout.svelte:136]: these error and success cases should ideally be handled in separate effects

lightly_studio_view/src/routes/datasets/[dataset_id]/+layout.svelte
```diff
@@ -116,32 +116,44 @@
...
+        if (event.key === 'Enter') {
+            const trimmedQuery = query_text.trim();
+            if (!trimmedQuery) return;
 
-            return response.data;
-        } catch (error) {
-            setError((error as unknown as Error).message);
-            console.error('Error during API call:', error);
+            submittedQueryText = trimmedQuery;
+            await tick();
```
Comment (ikondrat) [inline lightly_studio_view/src/routes/datasets/[dataset_id]/+layout.svelte]: can you explain why you need tick here?

lightly_studio_view/src/routes/datasets/[dataset_id]/+layout.svelte
Comment (ikondrat) [reply lightly_studio_view/src/routes/datasets/[dataset_id]/+layout.svelte]: I support MIhnea here 

defining hook with $derived means useEmbedText will be recreated on state change

there is also another way to trigger refetch, 
It can be useful if you don't want to reinitialise the hook every time
https://github.com/lightly-ai/lightly-studio/blob/main/lightly_studio_view/src/lib/hooks/useSamplesInfinite/useSamplesInfinite.ts#L13

lightly_studio_view/e2e/general/bussines-flow-1.e2e-test.ts
```diff
@@ -91,52 +92,21 @@ test.describe('bussines-flow1', () => {
...
         await samplesPage.createTag(catsTagName);
-        await samplesPage.goto();
-
         expect(samplesPage.getTagsMenuItem(catsTagName)).toBeDefined();
 
+        // Filter by the newly created tag
         await samplesPage.pressTag(catsTagName);
         expect(await samplesPage.getSamples().count()).toBe(1);
-    });
-    test('Remove cat text filter and check UI', async ({ page, samplesPage }) => {
```
michal-lightly, line 101: This test split was removed.

The reason is that it uses the `samplesPage` fixture, which in the setup calls `samplesPage.goto`. Therefore many of the actions in the test did not make sense. E.g. clearing the search - it was already cleared by the goto.

PR #291, opened 2025-12-12
Update the header component in preparation for multiple datasets support

mihnea-necsulescu:
## What has changed and why?

After indexing multiple datasets, and switching between them (via URL), the header was always referencing the first root dataset, causing inconsistencies (e.g., view videos with image navigation menu). Changes made:
- updated the get_root_dataset to get the root dataset for any given dataset id (optional for now for backwards compatibility, there are still places where the old version is used) (e.g., get the video dataset id after passing in the video frame dataset it) - needed for the navigation menu to properly display the tabs
- updated the read_root_dataset endpoint to receive a dataset id as a path param
- updated the useRootDataset hook to account for the endpoint change (pass in the dataset it) + update all its uses
- updated the menu-related components to receive the current dataset
- brought back the old get most recent dataset for the / page (this would break with the new change)

## How has it been tested?

Manually + updated tests

To test: index two datasets (coco + videos preferably), note the dataset ids, open the GUI and switch between datasets by changing the URL -> the header should be consistent with the selected dataset 

- [ ] Yes


lightly_studio/src/lightly_studio/resolvers/dataset_resolver/get_root_dataset.py
```diff
@@ -2,27 +2,53 @@
...
+        ValueError: If no root dataset is found or dataset_id doesn't exist.
     """
+    if dataset_id is not None:
+        # Find the dataset.
+        dataset = session.get(DatasetTable, dataset_id)
+        if dataset is None:
+            raise ValueError(f"Dataset with ID {dataset_id} not found.")
+
+        # Traverse up the hierarchy until we find the root.
+        while dataset.parent_dataset_id is not None:
```
LeonardoRosaa, line 42: Consider replacing this `while` loop with a Common Table Expression (CTE)

mihnea-necsulescu: thought about that, but I don't think that's worth it; we'd be trading readability for a micro-optimization - hierarchies are shallow enough (a couple of levels) that I don't think a recursive CTE would provide much value in this case

mihnea-necsulescu: added a comment for this

ikondrat: It looks like we did smth wrong in architecture if we need to make recursive call to derive root dataset

Can we have something like "root dataset type" cc @michal-lightly (maybe not related with this PR) ?

PR #292, opened 2025-12-12
Enable video frame annotations filtering

LeonardoRosaa:
## What has changed and why?

Enabled filtering of video frame annotations using video frame filters. The video frame annotations filter utilizes the existing structure for video frame filters.

Key changes:

- Added a HashMap to store and retrieve the parent dataset (as a workaround), which is required to identify the parent sample type.

## How has it been tested?

https://github.com/user-attachments/assets/498eae2e-2d32-48c6-ab86-9275a97f75fe

- [ ] Yes

### DIFF
--- a/lightly_studio_view/src/routes/datasets/[dataset_id]/+layout.svelte
+++ b/lightly_studio_view/src/routes/datasets/[dataset_id]/+layout.svelte
@@ -209,7 +209,9 @@
     const { videoFramesBoundsValues } = useVideoFramesBounds();
     const { videoBoundsValues } = useVideoBounds();
 
-    const { rootDataset } = useRootDatasetOptions();
+    const { rootDataset } = useRootDatasetOptions({
+        datasetId: page.params.dataset_id
+    });
 
     const annotationCounts = $derived.by(() => {
         if (


lightly_studio_view/src/lib/hooks/useGlobalStorage.ts
```diff
@@ -69,6 +81,33 @@ export const useGlobalStorage = () => {
...
+            ...prev,
+            [dataset.dataset_id]: {
+                sampleType: dataset.sample_type,
+                parentDatasetId: dataset.parent_dataset_id,
+                datasetId: dataset.dataset_id
+            }
+        }));
+    };
+
+    const retrieveParentDataset = (
```
ikondrat: Can you please explain why it is required for filtering functionality?

lightly_studio_view/src/routes/datasets/[dataset_id]/+layout.svelte
```diff
@@ -200,10 +206,15 @@
     const { videoFramesBoundsValues } = useVideoFramesBounds();
     const { videoBoundsValues } = useVideoBounds();
 
+    const { rootDataset } = useRootDatasetOptions();
+
     const annotationCounts = $derived.by(() => {
-        if (isVideoFrames) {
+        if (
+            isVideoFrames ||
```
Comment (ikondrat) [inline lightly_studio_view/src/routes/datasets/[dataset_id]/+layout.svelte:218]: I am a bit confused, we do have double check here
isVideoFrames and
SampleType.VIDEO_FRAME

how is it possible what it is not enough to check only isVideoFrames?

Comment (LeonardoRosaa) [reply lightly_studio_view/src/routes/datasets/[dataset_id]/+layout.svelte:218]: It is not enough. We should check the parent dataset’s sample type, not the current dataset’s sample type. Otherwise, the filtering will never be enabled.

PR #294, opened 2025-12-15
Add VideoSample class

lukas-lightly:
## What has changed and why?
I've renamed `video_resolver.get_by_id()` to `.get_view_by_id()`, because it returns `VideoView` and the corresponding `image_resolver.get_by_id()` returns `ImageTable`, so `video_resolver.get_by_id()` now returns `VideoTable`. It's more consistent this way.

## How has it been tested?
Added a test

- [x] Yes


lightly_studio/tests/resolvers/video/helpers.py
```diff
@@ -36,6 +36,26 @@ class VideoStub:
     fps: float = 30.0
 
 
+def create_video(session: Session, dataset_id: UUID, video: VideoStub) -> VideoTable:
+    sample_ids = video_resolver.create_many(
```
michal-lightly, line 40: Optional: Could call `create_videos`.

lightly_studio/src/lightly_studio/resolvers/video_resolver/get_by_id.py
```diff
@@ -13,15 +13,28 @@
...
         sample_id: The ID of the video to retrieve.
 
     Returns:
-        A video object or none.
+        A VideoTable object or none.
+    """
+    return session.exec(select(VideoTable).where(VideoTable.sample_id == sample_id)).one()
+
+
+def get_view_by_id(session: Session, sample_id: UUID) -> VideoView | None:
```
michal-lightly: Please add a test for this function.

michal-lightly: Also, please split the function into a separate file.

lightly_studio/tests/resolvers/video/helpers.py
lukas-lightly: `create_videos` returns a `list[UUID]`, not `list[VideoTable]`. I could fix this one function (and I might have to in the future) or add `create_video` returning `VideoTable` which is currently useful and don't have to refactor other uses.

lightly_studio/src/lightly_studio/resolvers/video_resolver/get_by_id.py
lukas-lightly: Done and done

PR #296, opened 2025-12-15
Consider the dataset ID in annotation crop selection

LeonardoRosaa:
## What has changed and why?

Store the `dataset_id` field along with the annotations to identify the correct dataset and avoid carrying over the wrong annotation filters between views.

## How has it been tested?

The tests were done manually + unit tests.

- [ ] Yes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,15 +9,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- Added metadata section to video and video frame details.
-- Added tag support for videos in the GUI.
-- Added plugins: This is the initial version for plugins. It supports the execution of operators.
-- `lightly_studio.core.sample.Sample` renamed to `ImageSample`.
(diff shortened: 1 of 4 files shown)


lightly_studio_view/src/lib/hooks/useGlobalStorage.ts
```diff
@@ -153,19 +153,28 @@ export const useGlobalStorage = () => {
...
 
         // Individual sample annotation crop selection methods
-        toggleSampleAnnotationCropSelection: (annotationId: string) => {
+        toggleSampleAnnotationCropSelection: (datasetId: string, annotationId: string) => {
             selectedSampleAnnotationCropIds.update((state) => {
-                if (state.has(annotationId)) {
-                    state.delete(annotationId);
+                const annotations = state[datasetId] ?? new Set();
+
+                if (annotations?.has(annotationId)) {
```
mihnea-necsulescu: annotations will always be defined
```suggestion
                if (annotations.has(annotationId)) {
```

PR #298, opened 2025-12-15
Pass datasetId to list samples endpoint

michal-lightly:
## What has changed and why?

Dataset id is now required in the list samples endpoint.

Also made the request body non-optional.

## How has it been tested?

* Unit tests
* Manually clicked through GUI

- [ ] Yes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added cloud storage support for video frames.
 - Added export for image captions.
 - Added semantic search by adding perception encoder core as embeddings model.
+- Added `VideoSample` class.
 
 ### Changed
 
--- a/lightly_studio/docs/docs/api/sample.md
+++ b/lightly_studio/docs/docs/api/sample.md
@@ -1,5 +1,13 @@
 # Sample
 
-::: lightly_studio.core.sample
+## ImageSample
+
+::: lightly_studio.core.image_sample
     options:
-        members: [Sample]
+        members: [ImageSample]
+
+## VideoSample
+
+::: lightly_studio.core.video_sample
+    options:
+        members: [VideoSample]
--- a/lightly_studio/src/lightly_studio/api/routes/api/video.py
+++ b/lightly_studio/src/lightly_studio/api/routes/api/video.py
@@ -107,7 +107,7 @@ def get_video_by_id(
     Returns:
         A video object.
     """
-    return video_resolver.get_by_id(session=session, sample_id=sample_id)
+    return video_resolver.get_view_by_id(session=session, sample_id=sample_id)
 
 
 @video_router.post("/bounds", response_model=Optional[VideoFieldsBoundsView])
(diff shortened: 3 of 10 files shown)


lightly_studio/src/lightly_studio/api/routes/api/embeddings2d.py
```diff
@@ -22,22 +22,17 @@
 class GetEmbeddings2DRequest(BaseModel):
     """Request body for retrieving 2D embeddings."""
 
-    filters: ImageFilter | None = Field(
-        None,
-        description="Filter parameters identifying matching samples",
-    )
+    filters: ImageFilter = Field(description="Filter parameters identifying matching samples")
```
michal-lightly, line 25: Also changed the `get_2d_embeddings` endpoint from the previous PR to have a non-optional parameter,

lightly_studio_view/src/lib/hooks/useEmbeddings/useEmbeddings.ts
```diff
@@ -5,14 +5,10 @@ import { createQuery } from '@tanstack/svelte-query';
...
-                          filters
-                      }
-                  }
-                : undefined
-        )
+        get2dEmbeddingsOptions({
+            body: {
+                filters: filters ?? {}
+            }
+        })
```
michal-lightly, line 12: This change is necessary, `filters` is now a mandatory field.

lightly_studio/src/lightly_studio/api/routes/api/embeddings2d.py
```diff
@@ -22,22 +22,17 @@
...
     session: SessionDep,
-    body: GetEmbeddings2DRequest | None = None,
+    body: GetEmbeddings2DRequest,
 ) -> Response:
     """Return 2D embeddings serialized as an Arrow stream."""
     dataset_id = (
-        body.filters.sample_filter.dataset_id
-        if body and body.filters and body.filters.sample_filter
-        else None
+        body.filters.sample_filter.dataset_id if body.filters.sample_filter is not None else None
```
LeonardoRosaa, line 35: Minor: I believe this is not related to this PR, but should we also always consider the `dataset_id` field for this function? In other words, should it ever be `None`?.

lightly_studio/src/lightly_studio/api/routes/api/sample.py
```diff
@@ -47,6 +47,8 @@ def read_samples(
     Returns:
         A list of filtered samples.
     """
+    if body.filters.dataset_id is None:
```
LeonardoRosaa, line 50: Minor: Is `dataset_id` optional? If so, why is it defined as optional but required in the end?

lightly_studio/src/lightly_studio/api/routes/api/embeddings2d.py
michal-lightly: It can be undefined, in which case we store `None` and then error out on the next line.

lightly_studio/src/lightly_studio/api/routes/api/sample.py
michal-lightly: There are multiple places where we use SamplesFilter - not only for this one endpoint. Making it non-optional would require examining all the places of use to see if we need optional anywhere.

LeonardoRosaa (approved):

PR #300, opened 2025-12-15
handle recursive datasets in the header

LeonardoRosaa:
## What has changed and why?

Improved the header to handle recursive datasets and display the dataset hierarchy properly. This should make it possible to see child datasets.

## How has it been tested?

Testing was done manually. I mocked a dataset hierarchy for the screenshot below.

header

- [ ] Yes


LeonardoRosaa:
> image
> this seems to not be aligned

That’s expected. It should create a smooth misalignment since the items aren’t on the same line (hierarchy level).

PR #302, opened 2025-12-16
Update the breadcrumb components in preparation for multiple datasets

mihnea-necsulescu:
## What has changed and why?

Breadcrumbs were always referencing the first root dataset. They were updated to be bound to their current root dataset.

## How has it been tested?

Manually. To test: index two datasets (coco + videos preferably), note the dataset ids, open the GUI and switch between datasets by changing the URL -> all breadcrumbs should be consistent with the selected dataset (image, annotation, video, video frame): click on "home", note the dataset name

- [ ] Yes


ikondrat (approved):
lgtm approved

PR #303, opened 2025-12-16
Bug fix: Embedding Plot selection resettable

JonasWurst:
## What has changed and why?

Fix a bug: when embedding plot selection is active -> closing embedding plot panel, open again -> selection still active but can not reset anymore, since local selection is empty.
made the selection a globalStorage entity.

## How has it been tested?

local

- [x] Yes

### DIFF
--- a/lightly_studio/src/lightly_studio/api/routes/api/annotation_label.py
+++ b/lightly_studio/src/lightly_studio/api/routes/api/annotation_label.py
@@ -16,7 +16,7 @@ from lightly_studio.models.annotation_label import (
     AnnotationLabelCreate,
     AnnotationLabelTable,
 )
-from lightly_studio.resolvers import annotation_label_resolver
+from lightly_studio.resolvers import annotation_label_resolver, dataset_resolver
 
 annotations_label_router = APIRouter()
 
@@ -29,7 +29,6 @@ def create_annotation_label(
     input_label: AnnotationLabelCreate,
     session: SessionDep,
     # TODO(Michal, 12/2025): Pass the root dataset directly.
-    # TODO(Michal, 12/2025): Use the dataset id.
     dataset_id: Annotated[
         UUID,
         Path(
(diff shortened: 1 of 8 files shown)


lightly_studio_view/src/lib/hooks/useGlobalStorage.ts
```diff
@@ -46,23 +46,13 @@ const setIsEditingMode = (isEditing: boolean) => {
...
-    >
->({});
-
 export type TextEmbedding = {
     embedding: number[];
     queryText: string;
 };
 
 const showPlot = writable<boolean>(false);
+const rangeSelection = writable<Point[] | null>(null);
```
mihnea-necsulescu: I think we should probably group these by dataset id (for multiple datasets), otherwise they may get overwritten when switching datasets

(probably not a big deal, but we started doing this in almost all other places)

lightly_studio_view/src/lib/components/PlotPanel/PlotPanel.test.ts
```diff
@@ -16,11 +22,34 @@ vi.mock('$lib/hooks/useImageFilters/useImageFilters', () => ({
...
+    const getRangeSelection = vi.fn(() => rangeSelection);
+    const setRangeSelectionForDataset = vi.fn();
+    return {
+        useGlobalStorage: () => ({
+            setShowPlot: vi.fn(),
+            getRangeSelection,
+            setRangeSelectionForDataset
+        })
+    };
+});
```
mihnea-necsulescu, line 34: I think this can be just

```suggestion
vi.mock('$lib/hooks/useGlobalStorage', () => {
    return {
        useGlobalStorage: () => ({
            setShowPlot: vi.fn(),
            getRangeSelection: vi.fn(() => writable(null)),
            setRangeSelectionForDataset: vi.fn()
        })
    };
});
```

lightly_studio_view/src/lib/components/PlotPanel/PlotPanel.test.ts
```diff
@@ -16,11 +22,34 @@ vi.mock('$lib/hooks/useImageFilters/useImageFilters', () => ({
...
+            return () => {};
+        },
+        set: vi.fn((newValue) => {
+            value = newValue;
+        }),
+        update: vi.fn((updater) => {
+            value = updater(value);
+        })
+    };
+}
```
mihnea-necsulescu: and this can be removed

PR #311, opened 2025-12-18
Move export() to Dataset from DatasetQuery

lukas-lightly:
## What has changed and why?
Having export() attached to DatasetQuery makes a lot of sense, however due to the limitations of Python typing system, it is better suited for Dataset. The export formats will differ between different dataset types (image, video, etc.), so in the end we'll have different DatasetExport types: ImageDatasetExport, VideoDatasetExport, etc., all having different methods. The corresponding ImageDataset, VideoDataset, etc., classes with have a `export()` method returning these.

## How has it been tested?
Converted tests, added an important test with a `DatasetQuery.match()` filter.

- [x] Yes

### DIFF
--- a/lightly_studio/src/lightly_studio/api/routes/api/annotation_label.py
+++ b/lightly_studio/src/lightly_studio/api/routes/api/annotation_label.py
@@ -4,7 +4,8 @@ from __future__ import annotations
 
 from uuid import UUID
 
-from fastapi import APIRouter, HTTPException
+from fastapi import APIRouter, HTTPException, Path
+from typing_extensions import Annotated
 
 from lightly_studio.api.routes.api.status import (
     HTTP_STATUS_CREATED,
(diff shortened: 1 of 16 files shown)


lightly_studio/tests/export/test_export_dataset.py
```diff
@@ -1,11 +1,18 @@
 from __future__ import annotations
 
 import json
+from collections.abc import Generator
 from pathlib import Path
+from unittest.mock import ANY
```
michal-lightly: No need to import this, you can use `mocker.ANY`.

lightly_studio/tests/export/test_export_dataset.py
```diff
@@ -20,26 +27,47 @@
 )
 
 
+@pytest.fixture
+def patch_dataset(
```
michal-lightly: You should not need to write this fixture, we already have it elsewhere. You might need to move it "up" in the conftest.py hierarchy to make it available here.

lightly_studio/tests/export/test_export_dataset.py
```diff
@@ -20,26 +27,47 @@
...
-        dataset = create_dataset(session=db_session)
+        dataset = Dataset.create(name="test_dataset")
         images_to_create = [
             ImageStub(path="image0.jpg", width=100, height=100),
             ImageStub(path="image1.jpg", width=200, height=200),
             ImageStub(path="image2.jpg", width=300, height=300),
         ]
         images = create_images(
-            db_session=db_session, dataset_id=dataset.dataset_id, images=images_to_create
+            db_session=dataset.session, dataset_id=dataset.dataset_id, images=images_to_create
```
michal-lightly, line 39: Session should be also available as a fixture, I think `db_session`. The `patch_dataset` afaik depends on it.

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 ### Changed
+- Moved `DatasetQuery.export()` method under `Dataset`.
```
michal-lightly: Optional: Make it more clear how users should change their code:
```suggestion
- Migrated `DatasetQuery.export()` to `Dataset.export()`.
```

michal-lightly (changes requested):
Minor comments

lightly_studio/tests/export/test_export_dataset.py
lukas-lightly: Done

lightly_studio/tests/export/test_export_dataset.py
lukas-lightly: I've moved it up.

lightly_studio/tests/export/test_export_dataset.py
lukas-lightly: Unfortunately that doesn't work, but `db_session` also uses `duckdb:///:memory:`, however they don't know about each other. I think we crate two separate in-memory instances.

PR #315, opened 2025-12-18
 Add prototype UI for instance segmentation “add” action

LeonardoRosaa:
## What has changed and why?

This PR prototypes the instance segmentation "add" action in the GUI.

Implemented behavior:

- Allow users to define a segmentation mask.
- After the segmentation mask is created, a bounding box is generated around it.

## How has it been tested?

https://github.com/user-attachments/assets/65970b3b-f539-4aa8-91bf-fb71f3aca4e0

- [ ] Yes


lightly_studio_view/src/lib/components/SampleAnnotation/SampleAnnotation.svelte
```diff
@@ -48,7 +48,10 @@
         $customLabelColorsStore[label]?.color ?? getColorByLabel(label, 0.4).color
     );
 
-    const opacity = $derived($customLabelColorsStore[label]?.alpha * 0.4);
+    const opacity = $derived(segmentationMask ? 0.65 : $customLabelColorsStore[label]?.alpha * 0.4);
+    const boundingBoxOpacity = $derived(
+        segmentationMask ? 0 : $customLabelColorsStore[label]?.alpha * 0.4
```
mihnea-necsulescu, line 57: this basically means only keep the border (no fill) for segmentation masks, and keep both otherwise, no? can you add a comment?

lightly_studio_view/src/lib/components/SampleAnnotation/SampleAnnotation.svelte
```diff
@@ -48,7 +48,10 @@
         $customLabelColorsStore[label]?.color ?? getColorByLabel(label, 0.4).color
     );
 
-    const opacity = $derived($customLabelColorsStore[label]?.alpha * 0.4);
+    const opacity = $derived(segmentationMask ? 0.65 : $customLabelColorsStore[label]?.alpha * 0.4);
```
mihnea-necsulescu: since we have two opacity variables now, could you also rename this one to be more specific?

lightly_studio_view/src/lib/components/SampleAnnotation/SampleAnnotation.svelte
mihnea-necsulescu: also, shouldn't this be used at line 111, too? for instance segmentation, the box should have 0 opacity both in edit and non-edit modes, no? it's currently clear only in edit mode

lightly_studio_view/src/lib/components/SampleDetails/SampleDetails.svelte
```diff
@@ -456,6 +461,190 @@
...
+    let isDrawingSegmentation = $state(false);
+    let segmentationPath = $state<{ x: number; y: number }[]>([]);
+    let annotationType = $state<string | null>(
+        $lastAnnotationType[collectionId] ?? AnnotationType.OBJECT_DETECTION
+    );
+    let isSegmentationMask = $derived(annotationType == AnnotationType.INSTANCE_SEGMENTATION);
+
+    const canDrawSegmentation = $derived(isSegmentationMask && addAnnotationEnabled);
+
+    const computeBoundingBoxFromMask = (
```
mihnea-necsulescu, line 475: could you leave some comments for the helper functions? there are a lot of them, and it's a bit hard to follow

also this component is getting quite big, we should probably extract some logic at some point

lightly_studio_view/src/lib/hooks/useGlobalStorage.ts
```diff
@@ -34,6 +34,12 @@ const metadataBounds = useSessionStorage<MetadataBounds>('lightlyStudio_metadata
 const metadataValues = useSessionStorage<MetadataValues>('lightlyStudio_metadata_values', {});
 const metadataInfo = useSessionStorage<MetadataInfo[]>('lightlyStudio_metadata_info', []);
 
+// Store the most recently selected annotation type.
+const lastAnnotationType = useSessionStorage<Record<string, AnnotationType>>(
```
mihnea-necsulescu, line 38: I don't think we should store this here - I think the dropdowns should be cleared on navigation (at least that's what's happening with the annotation label dropdown); if I add some annotations on a sample, then navigate to the next one, the annotation type dropdown is preselected, while the label one is not, which seems off

Image

LeonardoRosaa: We should also handle the annotation label being preselected in a follow-up issue.

lightly_studio_view/src/lib/components/SampleDetails/SampleDetails.svelte
LeonardoRosaa: I totally agree, this component is quite big. We have a refactor issue for this component that we should address.

mihnea-necsulescu:
found another bug: if I select object detection, then instance segmentation in the annotation type dropdown, I can then create both types of annotations; but I honestly kind of like it.. it feels smoother: drag for object detection boxes, click and draw for instance segmentation, without having to go back and forth with the dropdown - maybe we could even make this the default behavior, remove the annotation type dropdown, and just inform the users how to draw annotations, wdyt?


https://github.com/user-attachments/assets/59d9dde2-4fe9-4a01-bdb3-15d7424d26d9

mihnea-necsulescu:
>4.  I’ve disabled operations like resizing and moving the bounding box in the instance segmentation annotation. We still need to investigate how to handle this properly.
>5. I need more information about this.

I think disabling the resizing and moving also fixed the latter, I wasn't able to reproduce it whatever I tried

LeonardoRosaa:
> found another bug: if I select object detection, then instance segmentation in the annotation type dropdown, I can then create both types of annotations; but I honestly kind of like it.. it feels smoother: drag for object detection boxes, click and draw for instance segmentation, without having to go back and forth with the dropdown - maybe we could even make this the default behavior, remove the annotation type dropdown, and just inform the users how to draw annotations, wdyt?
> 
>  Screen.Recording.2025-12-23.at.11.38.01.mov

I was thinking of doing it in the near future. I'm not sure if we should fix it now or in a follow-up issue, as this PR is getting quite large. WDYT?

mihnea-necsulescu:
> > found another bug: if I select object detection, then instance segmentation in the annotation type dropdown, I can then create both types of annotations; but I honestly kind of like it.. it feels smoother: drag for object detection boxes, click and draw for instance segmentation, without having to go back and forth with the dropdown - maybe we could even make this the default behavior, remove the annotation type dropdown, and just inform the users how to draw annotations, wdyt?
> > Screen.Recording.2025-12-23.at.11.38.01.mov
> 
> I was thinking of doing it in the near future. I'm not sure if we should fix it now or in a follow-up issue, as this PR is getting quite large. WDYT?

yep, a follow-up should be good

PR #316, opened 2025-12-18
Make DatasetQuery generic

lukas-lightly:
## What has changed and why?
DatasetQuery now takes sample_class: type[T] argument, in order to construct ImageSample/VideoSample inside __iter__().

We will later most likely also introduce
`ImageDatasetQuery(DatasetQuery[ImageSample])` which will be returned by `ImageDataset.query()`.

## How has it been tested?
Existing tests.

- [ ] Yes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,12 +10,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 ### Changed
+
 - Renamed `Dataset` to `Collection` in the internal code.
-
 - Migrated `DatasetQuery.export()` to `Dataset.export()`.
-
 - Reduced the package size by using opencv-python-headless.
-
+- AnnotationLabelTable is now linked to a dataset.
 
 ### Deprecated
 
--- a/lightly_studio/docs/docs/index.md
+++ b/lightly_studio/docs/docs/index.md
@@ -422,7 +422,10 @@ dataset = ls.Dataset.create(name="predictions_dataset")
 # Create label for your class (if it does not exist for the dataset)
 label = annotation_label_resolver.create(
     session=dataset.session,
-    label=AnnotationLabelCreate(annotation_label_name="person"),
+    label=AnnotationLabelCreate(
+        root_collection_id=dataset.dataset_id,
+        annotation_label_name="person"
+    ),
 )
 
 # Your model predictions (e.g., from a detector)
@@ -896,7 +899,7 @@ from lightly_studio.plugins.parameter import FloatParameter, StringParameter
 from lightly_studio.resolvers import annotation_label_resolver, annotation_resolver, image_resolver
 
 
-def _preload_label_map(session, class_names):
+def _preload_label_map(session, root_collection_id, class_names):
     """Pre-creates all necessary labels in the DB and returns a lookup map.
 
     Args:
@@ -914,7 +917,7 @@ def _preload_label_map(session, class_names):
 
         # Create if missing
         if label is None:
-            label_create = AnnotationLabelCreate(annotation_label_name=name)
+            label_create = AnnotationLabelCreate(root_collection_id=root_collection_id, annotation_label_name=name)
             label = annotation_label_resolver.create(session=session, label=label_create)
 
         label_map[name] = label.annotation_label_id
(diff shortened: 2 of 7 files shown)


lightly_studio/src/lightly_studio/core/dataset.py
```diff
@@ -171,15 +175,9 @@ def load_or_create(
         )
         return Dataset(dataset=dataset)
 
-    # TODO(lukas 12/2025): return `Iterator[T]` instead
-    def __iter__(self) -> Iterator[ImageSample]:
+    def __iter__(self) -> Iterator[T]:
```
lukas-lightly: Note that we reached the ultimate goal of making `Dataset.__iter__()` generic.

lightly_studio/src/lightly_studio/core/dataset_query/dataset_query.py
```diff
@@ -164,13 +184,15 @@ def order_by(self, *order_by: OrderByExpression) -> DatasetQuery:
         Raises:
             ValueError: If order_by() has already been called on this instance.
         """
+        if self.dataset.sample_type == SampleType.VIDEO:
```
michal-lightly: I think this is more true. Also the error message should be adjusted.
```suggestion
        if self.dataset.sample_type != SampleType.IMAGE:
```

lightly_studio/src/lightly_studio/core/dataset.py
```diff
@@ -102,9 +100,15 @@ class Dataset(Generic[T]):
     ```
     """
 
-    def __init__(self, dataset: DatasetTable) -> None:
+    def __init__(self, dataset: DatasetTable, sample_class: type[T] | None = None) -> None:
```
michal-lightly: It is not clear to me how would a user now create a Dataset. Shouldn't `Dataset.create` set the new parameter?

lightly_studio/src/lightly_studio/core/dataset_query/dataset_query.py
lukas-lightly: Done

lightly_studio/src/lightly_studio/core/dataset.py
lukas-lightly: I've reverted this, `Dataset` doesn't return `T` in any of it's methods now, it's `ImageSample` everywhere (as it was before).

I'll handle this in a future PR, I already have a draft branch.

lightly_studio/src/lightly_studio/core/dataset.py
lukas-lightly: ... we haven't after I reverted some changes.

lightly_studio/src/lightly_studio/core/dataset_query/dataset_query.py
```diff
@@ -225,19 +247,36 @@ def __getitem__(self, key: _SliceType) -> DatasetQuery:
...
+            video_query: SelectOfScalar[VideoTable] = (
+                select(VideoTable)
+                .join(VideoTable.sample)
+                .where(SampleTable.collection_id == self.dataset.collection_id)
+            )
+            for video_table in self.session.exec(video_query):
+                # Calling the constructor of `VideoSample`
+                yield self._sample_class(video_table)  # type: ignore[arg-type]
+        else:
+            return
```
michal-lightly: Optional: Maybe like this?
```
raise NotImplementedError("Iter is not implemented for sample type {}")
```

lightly_studio/src/lightly_studio/core/dataset_query/dataset_query.py
```diff
@@ -225,19 +247,36 @@ def __getitem__(self, key: _SliceType) -> DatasetQuery:
...
 
+    # TODO(lukas 12/2025): this only works for images currently.
+    def _process(self, query: SelectOfScalar[ImageTable]) -> SelectOfScalar[ImageTable]:
```
michal-lightly: Optional: A comment would be nice, or a different function name (`_compose_query`?)

lukas-lightly: Renamed and added a docstring

lightly_studio/src/lightly_studio/core/dataset_query/dataset_query.py
lukas-lightly: Done

PR #325, opened 2025-12-22
Update press tag function

horatiualmasan:
## What has changed and why?

Update `pressTag` function to check the UI element instead of waiting for `/images/list` response.
There are cases when the response is available before we trigger the wait and it will fail.

## How has it been tested?

Triggered ci e2e tests several times with no failure. 
Manual tests locally. With the previous implementation it would always fail at a fresh run of the test (empty db). With the update I could not get it to fail anymore.

- [ ] Yes

### DIFF
--- a/lightly_studio_view/e2e/pages/samples-page.ts
+++ b/lightly_studio_view/e2e/pages/samples-page.ts
@@ -145,9 +145,7 @@ export class SamplesPage {
     }
 
     async pressTag(tagName: string): Promise<void> {
-        await expect(this.page.getByTestId('sample-grid-item').first()).toBeVisible({
-            timeout: 10000
-        });
+        await expect(this.page.getByTestId('sample-grid-item').first()).toBeVisible();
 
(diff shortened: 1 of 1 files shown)


lightly_studio_view/e2e/pages/samples-page.ts
```diff
@@ -159,8 +159,6 @@ export class SamplesPage {
             }
         }
 
-        await responsePromise;
```
michal-lightly, line 162: You can move the expect afterwards - I think it is more logical, and how I did it in another helper function. Also, the default timeout (5s) I'd assume is sufficient, especially if the expect is afterwards. WDYT?

horatiualmasan: I would leave it at the beginning. I think as the press tag is used in tests mostly after we create the tag, the goal of the delay here is to go back from the tag creation dialog to the main grid view with the elements loaded and have everything in place before we try to select the tag. Also 5s would be enough, I will update.

PR #326, opened 2025-12-22
Rename root_collection_id to dataset_id

michal-lightly:
## What has changed and why?

Rename root_collection_id to dataset_id and rootCollectionId to datasetId using search -and-replace.

## How has it been tested?

Exsting tests.

- [ ] Yes

### DIFF
--- a/lightly_studio_view/e2e/pages/samples-page.ts
+++ b/lightly_studio_view/e2e/pages/samples-page.ts
@@ -145,9 +145,7 @@ export class SamplesPage {
     }
 
     async pressTag(tagName: string): Promise<void> {
-        const responsePromise = this.page.waitForResponse(
-            (response) => response.url().includes('/images/list') && response.status() === 200
-        );
+        await expect(this.page.getByTestId('sample-grid-item').first()).toBeVisible();
 
         const tagLabels = this.page.getByTestId('tags-menu-label');
         const labelCount = await tagLabels.count();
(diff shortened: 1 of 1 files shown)


lightly_studio/src/lightly_studio/resolvers/collection_resolver/get_hierarchy.py
```diff
@@ -10,17 +10,15 @@
...
+def get_hierarchy(session: Session, dataset_id: UUID) -> list[CollectionTable]:
     """Retrieve all child collections of the given root collection, including the root itself.
 
     The collections are returned in the depth-first order, starting with the root collection.
     The relative order of children of any given node is the order in CollectionTable.children.
     """
-    root_collection = collection_resolver.get_by_id(
-        session=session, collection_id=root_collection_id
-    )
+    root_collection = collection_resolver.get_by_id(session=session, collection_id=dataset_id)
```
ikondrat, line 19: hmm

I am a bit confused what would be usecase for this function?

I thought we decided to use dedicated endpoints per collection
e.g.

/datasets/{dataset_id}/images/{collection_id}

My assumption we always need only 1 CollectionLabel smth like

get_collection(
    ...
    collection_id=collection_id,
)

no?

michal-lightly: The question is unrelated to this PR.

We can remove the function if it is not needed. The idea was to use it for the navigation menu.

PR #327, opened 2025-12-22
Create ImageDataset and VideoDataset

lukas-lightly:
## What has changed and why?
In the core of the package we have an alias `import ImageDataset as Dataset`, meaning that `Dataset` is the same as `ImageDataset`.

We will remove this alias in a followup, it's there now to make this commit/PR shorter.

## How has it been tested?
Modified existing test

- [x] Yes

### DIFF
--- a/lightly_studio/docs/docs/index.md
+++ b/lightly_studio/docs/docs/index.md
@@ -423,7 +423,7 @@ dataset = ls.Dataset.create(name="predictions_dataset")
 label = annotation_label_resolver.create(
     session=dataset.session,
     label=AnnotationLabelCreate(
-        root_collection_id=dataset.dataset_id,
+        dataset_id=dataset.dataset_id,
         annotation_label_name="person"
     ),
 )
@@ -899,7 +899,7 @@ from lightly_studio.plugins.parameter import FloatParameter, StringParameter
 from lightly_studio.resolvers import annotation_label_resolver, annotation_resolver, image_resolver
 
 
-def _preload_label_map(session, root_collection_id, class_names):
+def _preload_label_map(session, dataset_id, class_names):
     """Pre-creates all necessary labels in the DB and returns a lookup map.
 
     Args:
@@ -917,7 +917,7 @@ def _preload_label_map(session, root_collection_id, class_names):
 
         # Create if missing
         if label is None:
-            label_create = AnnotationLabelCreate(root_collection_id=root_collection_id, annotation_label_name=name)
+            label_create = AnnotationLabelCreate(dataset_id=dataset_id, annotation_label_name=name)
             label = annotation_label_resolver.create(session=session, label=label_create)
 
         label_map[name] = label.annotation_label_id
--- a/lightly_studio/src/lightly_studio/api/routes/api/annotation_label.py
+++ b/lightly_studio/src/lightly_studio/api/routes/api/annotation_label.py
@@ -46,13 +46,13 @@ def create_annotation_label(
 ) -> AnnotationLabelTable:
     """Create a new annotation label in the database."""
     # TODO(Michal, 12/2025): Use a different model for label creation from the frontend.
-    root_collection_id = collection_resolver.get_dataset(
+    dataset_id = collection_resolver.get_dataset(
         session=session, collection_id=collection_id
     ).collection_id
     return annotation_label_resolver.create(
         session=session,
         label=AnnotationLabelCreate(
-            root_collection_id=root_collection_id,
+            dataset_id=dataset_id,
             annotation_label_name=input_label.annotation_label_name,
         ),
     )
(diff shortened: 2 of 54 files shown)


lightly_studio/src/lightly_studio/core/dataset.py
```diff
@@ -100,78 +99,13 @@ class Dataset(Generic[T]):
     ```
     """
 
-    def __init__(self, dataset: CollectionTable) -> None:
+    def __init__(self, collection: CollectionTable, sample_class: type[T]) -> None:
```
michal-lightly, line 102: `sample_class` is not used in this PR. Is that for a follow-up?

lightly_studio/src/lightly_studio/core/dataset.py
```diff
@@ -643,17 +577,38 @@ def compute_similarity_metadata(
...
-        """Return a DatasetExport instance which can export the dataset in various formats.
 
-        Args:
-            query:
-                The dataset query to export. If None, the default query `self.query()` is used.
-        """
-        if query is None:
-            query = self.query()
-        return DatasetExport(session=self.session, root_dataset_id=self.dataset_id, samples=query)
+def _load_collection(
```
michal-lightly: This is used from outside modeules:

```suggestion
def load_collection(
```

lightly_studio/src/lightly_studio/core/dataset.py
```diff
@@ -643,17 +577,38 @@ def compute_similarity_metadata(
...
+    name: str | None = None, sample_type: SampleType = SampleType.IMAGE
```
michal-lightly: Minor: If this is an internal API I'd suggest to make sample_type non-optional (i.e. no default).

lightly_studio/src/lightly_studio/core/image_dataset.py
```diff
@@ -0,0 +1,74 @@
...
+    SampleType,
+)
+from lightly_studio.resolvers import collection_resolver
+
+
+class ImageDataset(Dataset[ImageSample]):
+    """Image dataset."""
+
+    def __init__(self, collection: CollectionTable):
+        """Create ImageDataset from collection table.
```
michal-lightly: Optional: English

```suggestion
        """Create ImageDataset from a collection table.
```

lightly_studio/src/lightly_studio/core/dataset.py
```diff
@@ -643,17 +577,38 @@ def compute_similarity_metadata(
...
+
+    Args:
+        name: The name of the dataset. If None, a default name is used.
+        sample_type: The type of samples in the dataset. Defaults to SampleType.IMAGE.
+
+    Return:
+        A collection if it exists, or None if it doesn't.
+    """
+    if name is None:
+        name = "default_dataset"
```
```suggestion
        name = DEFAULT_DATASET_NAME
```

lightly_studio/src/lightly_studio/core/image_dataset.py
```diff
@@ -0,0 +1,74 @@
...
+
+        Args:
+            collection: collection table.
+        """
+        super().__init__(collection=collection, sample_class=ImageSample)
+
+    @staticmethod
```
michal-lightly: Methods that return an instance are usually declared as "class methods":
```py
@classmethod
def create(cls, name: str | None = None) -> ImageDataset:
```
The first argument is the class type. This is maybe answer to some of the topics we discussed in person? You could e.g. call `return cls(collection=collection)`.

I suggest to make it a class method even if the `cls` param is not used, for two reasons:
1) It is done like that usually
2) It might become handy in the future

There might be deeper python reasons why it is done this way.

lightly_studio/src/lightly_studio/core/dataset.py
lukas-lightly: Yes, it's used in the followup: https://github.com/lightly-ai/lightly-studio/pull/328

lightly_studio/src/lightly_studio/core/dataset.py
lukas-lightly: Done

lightly_studio/src/lightly_studio/core/dataset.py
lukas-lightly: Done

lightly_studio/src/lightly_studio/core/dataset.py
lukas-lightly: Done

lightly_studio/src/lightly_studio/core/image_dataset.py
lukas-lightly: Done

lightly_studio/src/lightly_studio/core/image_dataset.py
lukas-lightly: I'll investigate this. I propose to merge this as is and fix later, WDYT? There's a series of PRs coming which needs this functionality and none of the rest would benefit from the class methods, so I suggest to postpone the proper fix.

lukas-lightly: I've changed `staticmethod` to `classmethod` and added `cls` parameter. Will fix this properly later.

PR #330, opened 2025-12-22
Use classmethod in Dataset

lukas-lightly:
## What has changed and why?
Make the code more DRY

## How has it been tested?
Existing tests

- [ ] Yes

### DIFF
--- a/lightly_studio/src/lightly_studio/core/dataset.py
+++ b/lightly_studio/src/lightly_studio/core/dataset.py
@@ -98,13 +98,12 @@ class Dataset(Generic[T], ABC):
     ```
     """
 
-    def __init__(self, collection: CollectionTable, sample_class: type[T]) -> None:
+    def __init__(self, collection: CollectionTable) -> None:
         """Initialize a LightlyStudio Dataset."""
         self._inner = collection
         # TODO(Michal, 09/2025): Do not store the session. Instead, use the
         # dataset object session.
         self.session = db_manager.persistent_session()
-        self._sample_class = sample_class
 
     @staticmethod
     @abstractmethod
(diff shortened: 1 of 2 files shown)


lightly_studio/src/lightly_studio/core/dataset.py
```diff
@@ -106,6 +106,52 @@ def __init__(self, collection: CollectionTable, sample_class: type[T]) -> None:
...
+        """Returns the sample type."""
+
+    @staticmethod
+    @abstractmethod
+    def sample_class() -> type[T]:
+        """Returns the sample class type."""
+
+    @classmethod
+    def create(cls, name: str | None = None) -> Self:
+        """Create a new image dataset.
```
```suggestion
        """Create a new dataset.
```

lightly_studio/tests/core/test_dataset__path.py
```diff
@@ -208,7 +210,7 @@ def test_add_images_from_path_calls_tag_samples_by_directory(
 
         _create_sample_images([tmp_path / "image1.jpg"])
         dataset_table = helpers_resolvers.create_collection(db_session, "test_dataset")
-        dataset = Dataset(collection=dataset_table)
+        dataset = ImageDataset(collection=dataset_table, sample_class=ImageSample)
```
michal-lightly: Side note: We might be able to also get rid of the `sample_class` param now.

lukas-lightly: Done

lightly_studio/src/lightly_studio/core/dataset.py
```diff
@@ -131,7 +176,7 @@ def query(self) -> DatasetQuery[T]:
             A DatasetQuery instance for querying samples in this dataset.
         """
         return DatasetQuery(
-            dataset=self._inner, session=self.session, sample_class=self._sample_class
+            dataset=self._inner, session=self.session, sample_class=Dataset.sample_class()
```
michal-lightly: I'm not sure if this is right, I'd expect `self.sample_class()`. Can you double-check?

lukas-lightly: I think both work but mine looks more correct (it's also how other languages access static methods). But  maybe your solution is more Pythonic?

PR #331, opened 2025-12-23
Factor out get_embedding_count

michal-lightly:
## What has changed and why?

Preparation for an endpoint that returns whether a collection has embeddings.

* Move `_are_embeddings_available()` from `dataset.py` to a renamed function `collection_has_embeddings()` in a new file `embedding_utils.py`
* A new resolver function `get_embedding_count()` is used to avoid fetching all embeddings from the db

## How has it been tested?

Added tests.

- [ ] Yes

### DIFF
--- a/lightly_studio_view/src/lib/components/SampleAnnotation/SampleAnnotation.svelte
+++ b/lightly_studio_view/src/lib/components/SampleAnnotation/SampleAnnotation.svelte
@@ -48,7 +48,14 @@
         $customLabelColorsStore[label]?.color ?? getColorByLabel(label, 0.4).color
     );
 
-    const opacity = $derived($customLabelColorsStore[label]?.alpha * 0.4);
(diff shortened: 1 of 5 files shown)


lightly_studio/src/lightly_studio/dataset/embedding_utils.py
```diff
@@ -0,0 +1,37 @@
...
+
+from sqlmodel import Session
+
+from lightly_studio.dataset.embedding_manager import EmbeddingManagerProvider
+from lightly_studio.resolvers import (
+    sample_embedding_resolver,
+)
+
+
+def collection_has_embeddings(session: Session, collection_id: UUID) -> bool:
```
michal-lightly, line 13: Moved from `dataset.py` with a minor change - a different resolver function for `return`.

PR #332, opened 2025-12-23
Move export api functions

michal-lightly:
## What has changed and why?

Addressed an old TODO. Move export api functions from collection.py to export.py.

Renamed test_dataset.py to test_collection.py.

## How has it been tested?

Existing tests.

- [ ] Yes

### DIFF
--- a/lightly_studio/src/lightly_studio/core/dataset.py
+++ b/lightly_studio/src/lightly_studio/core/dataset.py
@@ -25,8 +25,7 @@ from typing_extensions import Self, TypeVar
 
 from lightly_studio import db_manager
 from lightly_studio.api import features
-from lightly_studio.core import add_samples, add_videos
-from lightly_studio.core.add_videos import VIDEO_EXTENSIONS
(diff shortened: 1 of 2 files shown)


lightly_studio/tests/api/routes/api/test_export.py
```diff
@@ -103,3 +106,34 @@ def test_export_collection_captions(
     assert (
         response.headers["Content-Disposition"] == "attachment; filename=coco_captions_export.json"
     )
+
```
LeonardoRosaa, line 109: Minor: We don't have tests for the `/export/stats` endpoint, right? We could do that in a follow-up issue.

LeonardoRosaa (approved):
LGTM :)

lightly_studio/tests/api/routes/api/test_export.py
michal-lightly: Yes. It is low priority though, it won't get picked, so there's no need to create the issue.

PR #336, opened 2025-12-23
Move image methods to ImageDataset

lukas-lightly:
## What has changed and why?
Image-related methods should be in `ImageDataset`, not in `Dataset`.

## How has it been tested?
Existing tests

- [ ] Yes

### DIFF
--- a/lightly_studio/src/lightly_studio/core/dataset.py
+++ b/lightly_studio/src/lightly_studio/core/dataset.py
@@ -16,8 +16,7 @@ from lightly_studio.core.dataset_query.dataset_query import DatasetQuery
 from lightly_studio.core.dataset_query.match_expression import MatchExpression
 from lightly_studio.core.dataset_query.order_by import OrderByExpression
 from lightly_studio.core.sample import Sample
-from lightly_studio.dataset import embedding_utils, fsspec_lister
-from lightly_studio.dataset.embedding_manager import EmbeddingManagerProvider
+from lightly_studio.dataset import embedding_utils
 from lightly_studio.metadata import compute_similarity, compute_typicality
 from lightly_studio.models.collection import CollectionCreate, CollectionTable, SampleType
 from lightly_studio.resolvers import (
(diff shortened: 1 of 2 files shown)


lightly_studio/src/lightly_studio/resolvers/collection_resolver/get_parent_collection_id.py
```diff
@@ -9,12 +9,13 @@
 
 from lightly_studio.models.collection import CollectionTable
 
-ParentCollection = aliased(CollectionTable)
-ChildCollection = aliased(CollectionTable)
-
 
 def get_parent_collection_id(session: Session, collection_id: UUID) -> CollectionTable | None:
     """Retrieve the parent collection for a given collection ID."""
+    # Note: the aliasing is done inside the function rather than at the module level, because
```
lukas-lightly, line 15: Full explanation of this by Gemini:

> ✦ The change in src/lightly_studio/resolvers/collection_resolver/get_parent_collection_id.py moved the aliased(CollectionTable) calls from the module
>  level into the get_parent_collection_id function. This fixed a crash during test discovery and execution for the following reasons:
>   1. Early ORM Configuration: In SQLAlchemy and SQLModel, calling aliased() on a model class triggers the internal "mapper configuration" process. This process validates all relationships for all models currently registered in the metadata.
>   2. Circular Import Chain: When uv run pytest starts, it imports lightly_studio. The lightly_studio/__init__.py file imports ImageDataset, which initiates a long chain of imports (ImageDataset -> add_samples -> image_sample -> sample -> caption_resolver -> collection_resolver ->      get_parent_collection_id).
>   3. Incomplete Model Registration: In the last commit, ImageDataset gained several new dependencies, causing this import chain to be triggered very
>      early. When the chain reaches get_parent_collection_id, SampleTable has already been imported and registered, but its related model, SampleEmbeddingTable, has not yet been imported.
>   4. The Crash: Because aliased() was at the module level, it fired as soon as get_parent_collection_id was imported. SQLAlchemy attempted to  configure the SampleTable mapper, saw a relationship to the string "SampleEmbeddingTable", but couldn't find that class in its registry yet, resulting in an sqlalchemy.exc.InvalidRequestError.
>  The Fix: Moving the aliased() calls inside the function defers their execution until the function is actually called. By that time, the application has finished its import phase, and all models (including SampleEmbeddingTable) are fully registered, allowing the ORM configuration to complete successfully.

lightly_studio/src/lightly_studio/core/dataset.py
```diff
@@ -229,340 +206,6 @@ def __getitem__(self, key: _SliceType) -> DatasetQuery[T]:
...
-                tag_id=tag.tag_id,
-                sample_ids=created_sample_ids,
-            )
-
-        if embed:
-            _generate_embeddings_image(
-                session=self.session, collection_id=self.dataset_id, sample_ids=created_sample_ids
-            )
-
-    def compute_typicality_metadata(
```
michal-lightly, line 499: Typicality and similarity computation should work for any collection with embeddings; isn't that the case?

lukas-lightly: You are right, I'll keep it in `Dataset`

michal-lightly (approved):

PR #349, opened 2025-12-29
Implement image similarity search drag and drop

IgorSusmelj:
## What has changed and why?

- Adds support for image similarity search
- User can search for similar images given a reference image directly in the frontend using 3 ways
  - Via drag and drop of an image into the text search input field
  - Via the small image icon on the right end of the text search input field (after clicking a file browser opens)
  - Via paste through the clipboard (e.g. copy any image from the desktop, screenshot or the web and then paste it to the text search input field)
- Minor modifications to the existing codebase to support single image embeddings and without writing to the database
- The input text field is blocked from further text input when in image search mode
- We show a small preview image in the input text field (it's just large enough to double check whether the right image was used)

Showcase video of the feature:
https://github.com/user-attachments/assets/793b4d52-15d0-4175-9811-ba6fb045be87

## How has it been tested?

- New unit tests in the backend
- Extensive manual testing in the frontend (MacOS, Chrome browser)

- [X] Yes

### DIFF
--- a/lightly_studio/src/lightly_studio/dataset/edge_embedding_generator.py
+++ b/lightly_studio/src/lightly_studio/dataset/edge_embedding_generator.py
@@ -100,9 +100,7 @@ class EdgeSDKEmbeddingGenerator(ImageEmbeddingGenerator):
             return embeddings[0]
         return []
 
-    def embed_images(
-        self, filepaths: list[str], show_progress: bool = True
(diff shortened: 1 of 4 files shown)


ikondrat:
I think we need to show confidence score, otherwise it will be not clear what we have other low-confidence image on the page

wdyt?

ikondrat (approved):
lgtm

PR #350, opened 2025-12-30
Graceful db handling during shutdown of app

IgorSusmelj:
## What has changed and why?

- In some situations we can end up having .db.wal files (write ahead log) from duckdb because we did not save everything to disk before the execution of the software stopped (e.g. when running ctrl-c)
- These changes ensure the persistent session/engine are closed on shutdown so DuckDB can checkpoint and clean up WAL files. We handle two cases
  - case 1: the python code was running the GUI (so uvicorn was running and needs to initiate the db flush)
  - case 2: the python code was used without the GUI (e.g. scripting) and then we can handle the shutdown directly in the db manager
- When cleanup_existing=True, we also remove any leftover .wal/.tmp files from previous runs.
- We also address another issue that can happen when restarting the webserver quickly. We had a graceful shutdown time and it can be that after termination of the server (e.g. through ctrl-c) the ports can appear busy briefly due to OS TIME_WAIT, and our availability pre-check treated those as unavailable.
  - to allow starting another server within 2-3s we change the availability check to SO_REUSEADDR to avoid false ‘port busy’ results during TIME_WAIT, reducing unnecessary port hopping.

## How has it been tested?

- new unit test for handling the shutdown and making sure the files are removed properly
- extensive manual testing

- [ ] Yes

### DIFF
--- a/lightly_studio/src/lightly_studio/api/server.py
+++ b/lightly_studio/src/lightly_studio/api/server.py
@@ -42,7 +42,7 @@ class Server:
             limit_max_requests=10000,  # Max requests before worker restart
             # https://uvicorn.dev/settings/#timeouts
             timeout_keep_alive=5,  # Keep-alive timeout in seconds
-            timeout_graceful_shutdown=30,  # Graceful shutdown timeout (might block port for up to 30 seconds)
+            timeout_graceful_shutdown=30,  # Graceful shutdown timeout
             access_log=env.LIGHTLY_STUDIO_DEBUG,
         )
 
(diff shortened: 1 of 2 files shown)


lightly_studio/src/lightly_studio/api/server.py
```diff
@@ -101,6 +102,9 @@ def _is_port_available(host: str, port: int) -> bool:
 
     for family in families:
         with socket.socket(family, socket.SOCK_STREAM) as s:
+            # Allow port binding during TIME_WAIT to avoid false "port busy" checks.
```
IgorSusmelj, line 105: In case anyone is interested in reading up. I found this link helpful: https://serverframework.com/asynchronousevents/2011/01/time-wait-and-its-design-implications-for-protocols-and-scalable-servers.html

PR #351, opened 2025-12-31
Fix docs after introducing VideoDataset

lukas-lightly:
## What has changed and why?
After the introduction of VideoDataset, we've fixed examples but didn't fix the docs.

## How has it been tested?
Using `make` in the `docs` directory.

- [ ] Yes

### DIFF
--- a/README.md
+++ b/README.md
@@ -134,13 +134,14 @@ Create a file named `example_video.py` with the following contents:
 
 ```python title="example_video.py"
 import lightly_studio as ls
+from lightly_studio.core.video_dataset import VideoDataset
 from lightly_studio.utils import download_example_dataset
 
 # Download the example dataset (will be skipped if it already exists)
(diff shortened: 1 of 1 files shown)


lightly_studio/docs/docs/api/dataset.md
```diff
@@ -1,5 +1,5 @@
 # Dataset
 
-::: lightly_studio.core.dataset
+::: lightly_studio
```
lukas-lightly, line 3: The original `lightly_studio.core.dataset` is now a generic class, having it in the public API is confusing.

`lightly_studio.Dataset` is now an alias for `ImageDataset`, it's what people are actually using.

The final naming is TBD, we might end up with `lightly_studio.ImageDataset` instead of `lightly_studio.Dataset`, but it will also break a lot of existing code.

PR #352, opened 2025-12-31
Add IPv6 detection

lukas-lightly:
## What has changed and why?
If IPv6 is not available, we need to translate "localhost" to a IPv4 address, otherwise uvicorn fails to start.

## How has it been tested?
Added a test. Also used `sudo sysctl -w net.ipv6.conf.all.disable_ipv6=1` in Linux to disable IPv6 temporarily.

- [x] Yes


ikondrat (approved):
lgtm

PR #356, opened 2026-01-05
Rename StringField to ComparableField

lukas-lightly:
## What has changed and why?
The class (and neighboring classes) generalize to all comparable types, those having the `__eq__()` method.

This allows match expressions using `VideoSampleField.duration_s`, and possibly others in the future.

## How has it been tested?
Added a test.

- [ ] Yes

### DIFF
--- a/lightly_studio_view/src/lib/components/TagCreateDialog/TagCreateDialog.svelte
+++ b/lightly_studio_view/src/lib/components/TagCreateDialog/TagCreateDialog.svelte
@@ -36,6 +36,7 @@
     import { useVideoBounds } from '$lib/hooks/useVideosBounds/useVideosBounds';
     import Spinner from '../Spinner/Spinner.svelte';
     import type { Writable } from 'svelte/store';
+    import { useImageFilters } from '$lib/hooks/useImageFilters/useImageFilters';
 
     export type UseTagsCreateDialog = {
         collectionId: string;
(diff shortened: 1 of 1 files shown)


lightly_studio/src/lightly_studio/core/dataset_query/video_sample_field.py
```diff
@@ -28,13 +28,12 @@ class VideoSampleField:
...
-    file_name = StringField(col(VideoTable.file_name))
+    file_name = ComparableField(col(VideoTable.file_name))
     width = NumericalField(col(VideoTable.width))
     height = NumericalField(col(VideoTable.height))
-    file_path_abs = StringField(col(VideoTable.file_path_abs))
+    file_path_abs = ComparableField(col(VideoTable.file_path_abs))
 
-    # TODO(lukas 01/2026): make `duration_s` work too. Option[float] is not ordinal.
-    # duration_s = NumericalField(col(VideoTable.duration_s))
+    duration_s = ComparableField(col(VideoTable.duration_s))
```
lukas-lightly, line 38: Note: this only allows `==` and `!=` operators on `duration_s`. It would be great to also support `<` comparison, but that's not possible out of the box with `Optional[float]`. I'll skip this effort for now, if anyone has good ideas how to implement this in a general manner, please let me know.

ikondrat (approved):
lgtm

lightly_studio/src/lightly_studio/core/dataset_query/video_sample_field.py
michal-lightly: Probably with `OptionalOrdinalField` and expressions involving None (e.g. `x < None`) evaluate as false.

Another option: Why is duration an Optional? Can we make it non-optional?

lukas-lightly: I had the same question as you, each video has duration, so I'm not sure why this field is optional. Maybe we can just fix that and the issue will resolve itself :)

PR #362, opened 2026-01-07
Allow interactive jupyter notebook with python and gui in parallel

MalteEbner:
## What has changed and why?

Adds functionality to run the GUI from python in the background. This is especially useful in notebooks where you want to
keep exploring or updating the dataset after the GUI starts. You can explore the dataset interactively in the GUI, then decided to add some metadata or do selections in Python, and then continue exploring the updated dataset in the GUI without having to restart it.

Added 3 new functions for it:
- `start_gui_background()`: Starts the GUI server in a background thread.
- `check_gui_background()`: Checks if the background GUI server is running.
- `stop_gui_background()`: Stops the background GUI server.

## How has it been tested? - Code

- Added unitests
- Manually using an example notebook as shown in the docs. I created tags both via python and the GUI switching between them and ensuring that both python and the GUI find the same tags.

## How has it been tested? - Docs
image

- [x] Yes, and the docs

### DIFF
--- a/lightly_studio/src/lightly_studio/models/collection.py
+++ b/lightly_studio/src/lightly_studio/models/collection.py
@@ -16,6 +16,7 @@ class SampleType(str, Enum):
     IMAGE = "image"
     ANNOTATION = "annotation"
     CAPTION = "caption"
+    GROUP = "group"
 
 
 class CollectionBase(SQLModel):
@@ -26,6 +27,10 @@ class CollectionBase(SQLModel):
         default=None, foreign_key="collection.collection_id"
     )
     sample_type: SampleType
+
+    # Group-specific fields
+    group_component_name: Optional[str] = None
+    group_component_index: Optional[int] = None
 
 
 class CollectionCreate(CollectionBase):
--- a/lightly_studio/src/lightly_studio/models/group.py
+++ b/lightly_studio/src/lightly_studio/models/group.py
@@ -1,0 +1,29 @@
+"""Group table definition."""
+
+from uuid import UUID
+
+from sqlalchemy.orm import Mapped
+from sqlmodel import Field, Relationship, SQLModel
+
+from lightly_studio.models.sample import SampleTable
+
+
+class GroupTable(SQLModel, table=True):
+    """This class defines the Group model."""
+
+    __tablename__ = "group"
+    sample_id: UUID = Field(foreign_key="sample.sample_id", primary_key=True)
+
+    sample: Mapped["SampleTable"] = Relationship(
+        sa_relationship_kwargs={
+            "lazy": "select",
+            "foreign_keys": "[GroupTable.sample_id]",
+        },
+    )
+
+
+class SampleGroupLinkTable(SQLModel, table=True):
+    """Model to define links between Group and Sample One-to-Many."""
+
+    sample_id: UUID = Field(foreign_key="sample.sample_id", primary_key=True)
+    parent_sample_id: UUID = Field(foreign_key="group.sample_id")
--- a/lightly_studio/src/lightly_studio/models/video.py
+++ b/lightly_studio/src/lightly_studio/models/video.py
@@ -43,7 +43,9 @@ class VideoTable(VideoBase, table=True):
 
     __tablename__ = "video"
     sample_id: UUID = Field(foreign_key="sample.sample_id", primary_key=True)
-    frames: Mapped[List["VideoFrameTable"]] = Relationship(back_populates="video")
+    frames: Mapped[List["VideoFrameTable"]] = Relationship(
+        sa_relationship_kwargs={"lazy": "select"}, back_populates="video"
+    )
     sample: Mapped["SampleTable"] = Relationship()
 
 
--- a/lightly_studio_view/src/lib/routes.ts
+++ b/lightly_studio_view/src/lib/routes.ts
@@ -95,7 +95,8 @@ export const routes = {
             const path = `/collections/${collectionId}/frames/${sampleId}`;
 
             return index === null ? path : path + `?index=${index}`;
-        }
+        },
+        groups: (collectionId: string) => `/collections/${collectionId}/groups`
     }
 };
 
@@ -128,5 +129,8 @@ export const routeHelpers = {
     },
     toFramesDetails: (collectionId: string, sampleId: string, index: number | null = null) => {
         return routes.collection.framesDetails(collectionId, sampleId, index);
+    },
+    toGroups: (collectionId: string) => {
+        return routes.collection.groups(collectionId);
     }
 };
--- a/lightly_studio_view/src/lib/schema.d.ts
+++ b/lightly_studio_view/src/lib/schema.d.ts
@@ -1998,6 +1998,10 @@ export interface components {
             /** Parent Collection Id */
             parent_collection_id?: string | null;
             sample_type: components["schemas"]["SampleType"];
+            /** Group Component Name */
+            group_component_name?: string | null;
+            /** Group Component Index */
+            group_component_index?: number | null;
         };
         /**
          * CollectionOverviewView
@@ -2030,6 +2034,10 @@ export interface components {
             /** Parent Collection Id */
             parent_collection_id?: string | null;
             sample_type: components["schemas"]["SampleType"];
+            /** Group Component Name */
+            group_component_name?: string | null;
+            /** Group Component Index */
+            group_component_index?: number | null;
             /**
              * Collection Id
              * Format: uuid
@@ -2056,6 +2064,10 @@ export interface components {
             /** Parent Collection Id */
             parent_collection_id?: string | null;
             sample_type: components["schemas"]["SampleType"];
+            /** Group Component Name */
+            group_component_name?: string | null;
+            /** Group Component Index */
+            group_component_index?: number | null;
             /**
              * Collection Id
              * Format: uuid
@@ -2087,6 +2099,10 @@ export interface components {
             /** Parent Collection Id */
             parent_collection_id?: string | null;
             sample_type: components["schemas"]["SampleType"];
+            /** Group Component Name */
+            group_component_name?: string | null;
+            /** Group Component Index */
+            group_component_index?: number | null;
             /**
              * Collection Id
              * Format: uuid
@@ -2739,7 +2755,7 @@ export interface components {
          * @description The type of samples in the collection.
          * @enum {string}
          */
-        SampleType: "video" | "video_frame" | "image" | "annotation" | "caption";
+        SampleType: "video" | "video_frame" | "image" | "annotation" | "caption" | "group";
         /**
          * SampleView
          * @description This class defines the Sample view model.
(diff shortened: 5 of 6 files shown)


lightly_studio/src/lightly_studio/api/server.py
```diff
@@ -50,19 +50,28 @@ def start(self) -> None:
             access_log=env.LIGHTLY_STUDIO_DEBUG,
         )
 
+    def create_uvicorn_server(self) -> uvicorn.Server:
+        """Create a Uvicorn server instance."""
+        return uvicorn.Server(self._build_config())
```
MalteEbner: This allows reusing the `create_uvicorn_server` method by `start_gui_background`

lightly_studio/src/lightly_studio/db_manager.py
```diff
@@ -74,7 +78,10 @@ def session(self) -> Generator[Session, None, None]:
             # Commit the persistent session to ensure it sees the latest data changes.
             # This prevents the persistent session from having stale data when it's used
             # after operations in short-lived sessions have modified the database.
-            self.get_persistent_session().commit()
+            # Only commit from the owning thread to avoid cross-thread session access.
+            persistent_session = self._get_persistent_session_if_owned()
+            if persistent_session is not None:
+                persistent_session.commit()
```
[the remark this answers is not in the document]
MalteEbner: I tried it, it works.

lightly_studio/src/lightly_studio/api/server.py
```diff
@@ -34,10 +34,10 @@ def __init__(self, host: str, port: int) -> None:
             env.LIGHTLY_STUDIO_PORT = self.port
             env.APP_URL = f"{env.LIGHTLY_STUDIO_PROTOCOL}://{env.LIGHTLY_STUDIO_HOST}:{env.LIGHTLY_STUDIO_PORT}"
 
-    def start(self) -> None:
-        """Start the API server using Uvicorn."""
-        config = uvicorn.Config(
-            app,
+    def _build_config(self) -> uvicorn.Config:
```
michal-lightly: Minor: Move this function to the bottom.

lightly_studio/src/lightly_studio/core/start_gui.py
```diff
@@ -11,6 +16,9 @@
 
 logger = logging.getLogger(__name__)
 
+if TYPE_CHECKING:
```
michal-lightly: Why is `if TYPE_CHECKING` needed? We should really avoid it. I was able to remove it from many files in models.

lightly_studio/src/lightly_studio/__init__.py
```diff
@@ -7,11 +7,26 @@
...
 
 
 # TODO (Jonas 08/25): This will be removed as soon as the new interface is used in the examples
 from lightly_studio.models.annotation.annotation_base import AnnotationType
 
-__all__ = ["AnnotationType", "ImageDataset", "SampleType", "VideoDataset", "start_gui"]
+start_gui = start_gui_module.start_gui
+start_gui_background = start_gui_module.start_gui_background
+check_gui_background = start_gui_module.check_gui_background
+stop_gui_background = start_gui_module.stop_gui_background
```
michal-lightly: Why not use imports as usual?

```
from from lightly_studio.core.start_gui import start_gui_background
```
etc

lightly_studio/src/lightly_studio/db_manager.py
```diff
@@ -74,7 +78,10 @@ def session(self) -> Generator[Session, None, None]:
             # Commit the persistent session to ensure it sees the latest data changes.
             # This prevents the persistent session from having stale data when it's used
             # after operations in short-lived sessions have modified the database.
-            self.get_persistent_session().commit()
+            # Only commit from the owning thread to avoid cross-thread session access.
```
michal-lightly: I don't see how this helps cross-thread session access. Also it is not clear to me when do we expect cross-thread session access.

Note that the GUI should never use the persistent session (this is a wishful "should"), it should be used only from the setup script.

Also I'm worried this undoes the "foreign key" fix. One of the issues was that the persistent session was not committed when the gui started (and used an ephemeral session). I.e. we wanted to commit the persistent session in that case.

lightly_studio/src/lightly_studio/__init__.py
MalteEbner: Because our styleguide tells to not import functions directly, but rather their module. Should we break that rule in the init file?

lightly_studio/src/lightly_studio/api/server.py
MalteEbner: I inlined it instead.

lightly_studio/src/lightly_studio/core/start_gui.py
MalteEbner: You are right, I removed it.

lightly_studio/src/lightly_studio/db_manager.py
MalteEbner: I removed it. Let's not do any changes/protections for now, but only once we see bugs/problems occurring.

lightly_studio/src/lightly_studio/core/start_gui.py
```diff
@@ -44,11 +48,62 @@ def _validate_has_samples() -> None:
 
 
 def start_gui() -> None:
-    """Launch the web interface for the loaded dataset."""
+    """Launch the web interface for the loaded dataset.
+
+    This call blocks until the server stops.
+    """
```
michal-lightly, line 24: Optional: It is strange that `start_gui` logic is implemented in `server.py`, but `start_gui_background` is in this file. I think it would be cleaner to move the logic here.

lightly_studio/src/lightly_studio/core/start_gui.py
```diff
@@ -44,11 +48,62 @@ def _validate_has_samples() -> None:
...
+
+_GUI_BACKGROUND_STATE: _GuiBackgroundState | None = None
+
+
+def start_gui_background() -> None:
+    """Launch the web interface in a background thread."""
+    global _GUI_BACKGROUND_STATE  # noqa: PLW0603
+    # TODO(Malte, 01/26): Handle start when a background server is already running.
+
+    _validate_has_samples()
```
michal-lightly, line 50: Minor: Would you mind moving the function to the bottom of the file in this PR?

lightly_studio/src/lightly_studio/__init__.py
michal-lightly: Yes, absolutely. We do it in all init files. It does not make sense to adhere to the rule, exposing the functions is the point.

lightly_studio/tests/core/test_start_gui.py
```diff
@@ -16,6 +18,40 @@
...
+            while not self.should_exit:
+                if self.stop_event.wait(timeout=0.01):
+                    break
+        else:
+            self.stop_event.wait(timeout=1.0)
+
+
+@pytest.fixture
+def reset_gui_background_state() -> Generator[None, None, None]:
+    start_gui_module._GUI_BACKGROUND_STATE = None
```
michal-lightly: The global state is going to cause issues when running tests in parallel (we don't do that in prod, but I do it locally). Also, there's a cleaner way in general to avoid corruption between the tests, even in serial setting.

We need to patch the _GUI_BACKGROUND_STATE variable so that every test has a separate instance. It is easy, we do something similar already in `patch_dataset` to get around global persistent session and embedding manager.

lightly_studio/tests/core/test_start_gui.py
```diff
@@ -69,3 +105,54 @@ def test_start_gui__empty_datasets(
...
+        target=start_gui_module,
+        attribute="_validate_has_samples",
+    )
+    stop_event = threading.Event()
+    fake_server = FakeUvicornServer(stop_event=stop_event)
+    mock_server = mocker.patch.object(
+        target=start_gui_module,
+        attribute="Server",
+    )
+    mock_server.return_value.create_uvicorn_server.return_value = fake_server
```
michal-lightly: Optional: Do we need to patch the whole Server class? Isn't it enough to just do thus?
```py
mock_create_uvicorn_server = mocker.patch.object(
        target=Server,
        attribute="create_uvicorn_server",
        return_value=fake_server
)
```

lightly_studio/docs/docs/tutorials/parallel_python_gui.md
```diff
@@ -0,0 +1,54 @@
+# Parallel Python and GUI
+
+Use the background GUI helpers to keep the web UI running while your Python code
+continues to execute. This is especially useful in notebooks where you want to
```
michal-lightly: Optional:
```suggestion
You can run the web UI in the background and continue executing Python code from your main script.
This is especially useful in notebooks where you want to
```

lightly_studio/docs/docs/tutorials/parallel_python_gui.md
```diff
@@ -0,0 +1,54 @@
...
+keep exploring or updating the dataset after the GUI starts. You can explore the dataset interactively in the GUI, then decided to add some metadata or do selections in Python, and then continue exploring the updated dataset in the GUI without having to restart it.
```
michal-lightly: Typo:
```suggestion
keep exploring or updating the dataset after the GUI starts. You can explore the dataset interactively in the GUI, then decide to add some metadata or do selections in Python, and then continue exploring the updated dataset in the GUI without having to restart it.
```

lightly_studio/docs/docs/tutorials/parallel_python_gui.md
```diff
@@ -0,0 +1,54 @@
...
+
+### Cell 1: Setup
+```python
+import lightly_studio as ls
+from lightly_studio.utils import download_example_dataset
+```
+
+### Cell 2: Create a dataset from COCO data
+```python
+dataset_path = download_example_dataset(download_dir="dataset_examples")
```
michal-lightly: Very optional: `ls.utils.download_example_dataset` also works without an extra import.

michal-lightly (changes requested):

lightly_studio/tests/core/test_start_gui.py
MalteEbner: I switched to patching

lightly_studio/src/lightly_studio/core/start_gui.py
MalteEbner: I removed the unused `server.py:Server.start()`

lightly_studio/tests/core/test_start_gui.py
MalteEbner: Yes, I switched to that.

lightly_studio/docs/docs/tutorials/parallel_python_gui.md
michal-lightly: This was resolved without a comment, but the change was not made. I assume you decided not to the the optional suggestion? (It's ok.)

michal-lightly (approved):

lightly_studio/docs/docs/tutorials/parallel_python_gui.md
MalteEbner: It was, but ad additional text. I now reworked it again using your sentence.

PR #363, opened 2026-01-07
Add sample details toolbar tooltip

LeonardoRosaa:
## What has changed and why?

Added sample details toolbar tooltip. The component should consider a label and a keyboard shortcut key.

## How has it been tested?

https://github.com/user-attachments/assets/a913334c-4a9e-4155-a030-69e20b6a21f8

- [ ] Yes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,11 +9,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 - Added `VideoDataset` class.
+- Added Captions support for videos.
 - Allow creating tags from all samples matching the current filters when no samples are explicitly selected.
 - Added notebook/Colab support and usage snippet to the docs.
 - Added image similarity search via drag-and-drop, file upload, or clipboard paste.
 - Added similarity score display for images and videos when using embedding-based search.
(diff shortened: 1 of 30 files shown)


lightly_studio_view/src/lib/components/SampleDetails/SampleDetailsToolbar/SampleDetailsToolbar.svelte
```diff
@@ -18,24 +19,28 @@
 
 <Card>
     <CardContent>
-        <button
-            type="button"
-            aria-label="Toggle eraser"
-            onclick={() => (isEraser = !isEraser)}
-            class={`flex
+        <SampleDetailsToolbarTooltip label="Eraser Tool">
```
IgorSusmelj, line 22: Will you create a follow up issue or todo to handle the shortcuts? We should connect them to the shortcuts hook we have that is linked to the settings. 

Another approach for later would be to just show the shortcut key in brackets.

Now:
`Eraser Tool`

Future
`Eraser tool [e]` or `[e] Eraser tool` to indicate which button to press based on the settings

LeonardoRosaa: Agreed. We can address shortcuts in follow-up issues.

PR #367, opened 2026-01-08
Add basic Annotation class

lukas-lightly:
## What has changed and why?
The class is very basic, meant to be inherited by specialized annotation classes for object detection or selection. This will come in a later commits.

## How has it been tested?
Added a test.

- [ ] Yes

### DIFF
--- a/lightly_studio/src/lightly_studio/core/annotation.py
+++ b/lightly_studio/src/lightly_studio/core/annotation.py
@@ -1,0 +1,41 @@
+"""Interface for annotations."""
+
+from typing import cast
+
+from sqlalchemy.orm import object_session
+from sqlmodel import Session, col
+
+from lightly_studio.core.db_field import DBField
+from lightly_studio.models.annotation.annotation_base import AnnotationBaseTable
+
+
+class Annotation:
+    """Class for annotation."""
+
+    confidence = DBField(col(AnnotationBaseTable.confidence))
+    # TODO(lukas 1/2026): type should be ideally (also) encoded in Python's type system.
+    type = DBField(col(AnnotationBaseTable.annotation_type))
+    label = DBField(col(AnnotationBaseTable.annotation_label))
+
+    def __init__(self, inner: AnnotationBaseTable) -> None:
+        """Initialize the Annotation.
+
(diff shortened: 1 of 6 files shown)


lightly_studio/src/lightly_studio/core/annotation.py
```diff
@@ -0,0 +1,41 @@
...
+
+    def __init__(self, inner: AnnotationBaseTable) -> None:
+        """Initialize the Annotation.
+
+        Args:
+            inner: The AnnotationBaseTable SQLAlchemy model instance.
+        """
+        self.inner = inner
+
+    def get_object_session(self) -> Session:
```
michal-lightly, line 28: Note: It is probably for a separate PR, but it seems to me that it would make sense to move the implementation to `_DBFieldOwner` that should become a) `DBFieldOwner` and b) an ABC instead of a protocol.

lightly_studio/src/lightly_studio/core/sample.py
```diff
@@ -250,6 +215,14 @@ def captions(self, captions: Iterable[str]) -> None:
                 ],
             )
 
+    @property
+    def annotations(self) -> list[Annotation]:
+        """Returns all annotations."""
+        annotations = []
+        for annotation in self.sample_table.annotations:
+            annotations.append(Annotation(inner=annotation))
+        return annotations
```
michal-lightly, line 224: Minor: More pythonic is a list comprehension:
```py
return [Annotation(inner=annotation) for annotation in self.sample_table.annotations]
```

lightly_studio/src/lightly_studio/core/annotation.py
lukas-lightly: Let's do it later.

lightly_studio/src/lightly_studio/core/sample.py
lukas-lightly: I thought about it too, but there will later be a `match annotation.type:` creating different Python classes based on the annotation type and this is not well-suited for list comprehension. If I use it now, I'd most likely remove it later.

PR #373, opened 2026-01-08
Add bounding box button to toolbar

LeonardoRosaa:
## What has changed and why?

Implemented bounding box creation via the toolbar.

Key changes:

- Moved bounding box creation to the toolbar.
- The user creates an annotation, then enters the annotation label. A default class is created.
- Used the Svelte Context API to manage state, since we need to share state deep in the component tree.
- Removed the eraser and settings icons from the toolbar, as the segmentation mask isn’t working for now due to the toolbar changes.

## How has it been tested?

https://github.com/user-attachments/assets/4cc23144-76af-42f4-bcfd-de50333183a8

- [ ] Yes


lightly_studio_view/src/lib/components/SampleDetails/SampleDetailsSidePanel/SampleDetailsSidePanel.svelte
```diff
@@ -44,6 +36,23 @@
     }: Props = $props();
 
     const tags = $derived(sample.tags.map((t) => ({ tagId: t.tag_id, name: t.name })) ?? []);
+    const annotationLabelContext = useAnnotationLabelContext();
+
+    // Auto-scroll to selected annotation
+    $effect(() => {
+        const annotationId =
+            annotationLabelContext.annotationId ?? annotationLabelContext.lastCreatedAnnotationId;
+        if (annotationId) {
```
LeonardoRosaa, line 45: Pushed it back since it was removed during the refactor.

ikondrat (approved):
lgtm

thanks
I think now it is coming to proper editor 👍

PR #376, opened 2026-01-09
Do not preload tags when hovering on links

horatiualmasan:
## What has changed and why?

Group tags by collection id in the global storage. This will avoid overwriting the tags when hovering the mouse on menu items that trigger data preload.

## How has it been tested?

Manual tests

https://github.com/user-attachments/assets/30aa13e3-8307-474a-9851-223af503850f

- [ ] Yes

### DIFF
--- a/lightly_studio/src/lightly_studio/api/routes/video_frames_media.py
+++ b/lightly_studio/src/lightly_studio/api/routes/video_frames_media.py
@@ -2,13 +2,20 @@
 
 from __future__ import annotations
 
+import asyncio
 import io
+import os
+import threading
+from collections import OrderedDict
 from collections.abc import Generator
+from concurrent.futures import ThreadPoolExecutor
 from typing import Any, cast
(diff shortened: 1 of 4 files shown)


mihnea-necsulescu:
I think the root issue here is the global tags store 
https://github.com/lightly-ai/lightly-studio/blob/main/lightly_studio_view/src/lib/hooks/useGlobalStorage.ts#L49
they'd just get overwritten on every load - grouping them by collection ID should properly fix this, but that's a bigger effort and could be done as a follow-up

a quick fix would be to overwrite the default hover preload behavior for the menu items, and only load when actually clicking
i.e., just adding `data-sveltekit-preload-data="tap"` to the [MenuItem button](https://github.com/lightly-ai/lightly-studio/blob/main/lightly_studio_view/src/lib/components/MenuItem/MenuItem.svelte#L29) should fix this with just a line

PR #391, opened 2026-01-12
Introduce ObjectDetectionAnnotation class

lukas-lightly:
## What has changed and why?
The first class of this kind, there'll be more soon

## How has it been tested?
Modified existing test.

- [ ] Yes

### DIFF
--- a/lightly_studio/Makefile
+++ b/lightly_studio/Makefile
@@ -30,6 +30,7 @@ build: clean build-lightly_studio_view
 build-lightly_studio_view: export-schema export-version
 ifeq ($(OS),Windows_NT)
 	@powershell -Command "$$env:PUBLIC_POSTHOG_KEY='$(PUBLIC_POSTHOG_KEY)'; make -C ../lightly_studio_view build"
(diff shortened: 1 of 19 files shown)


lightly_studio/src/lightly_studio/core/annotation.py
```diff
@@ -1,29 +1,25 @@
...
 from lightly_studio.core.db_field import DBField
-from lightly_studio.models.annotation.annotation_base import AnnotationBaseTable
+from lightly_studio.models.annotation.annotation_base import AnnotationBaseTable, AnnotationType
+from lightly_studio.models.annotation.object_detection import ObjectDetectionAnnotationTable
 
 
 class Annotation:
     """Class for annotation."""
 
-    confidence = DBField(col(AnnotationBaseTable.confidence))
```
lukas-lightly, line 15: `DBField` accesses `self.inner`, and we can't imitate inheritance with it: self.inner cannot point both to `AnnotationBaseTable` and `ObjectDetectionAnnotationTable` at the same time.

Instead, the removed fields are defined using `@property`.

ikondrat (commented):
lgtm

PR #396, opened 2026-01-12
Deep copy resolver + test endpoint

mihnea-necsulescu:
## What has changed and why?

Added a deep copy root collection resolver. This approach uses explicit and type-specific copy functions for each entity type. Other more "elegant" solutions (e.g., reflection, schema introspection) may reduce the number of lines but would introduce complexity, negatively affect readability, debuggability and clarity, would make foreign key remapping difficult to both implement and maintain, or would be DB-specific (e.g., DuckDB export json -> process and remap ids -> insert json could work, but could not be transferred to Postgres, for example).

## How has it been tested?

manually with the provided test endpoint + new tests

- [ ] Yes

### DIFF
--- a/lightly_studio/src/lightly_studio/resolvers/collection_resolver/deep_copy.py
+++ b/lightly_studio/src/lightly_studio/resolvers/collection_resolver/deep_copy.py
@@ -416,7 +416,6 @@ def _copy_metadata(
                 "custom_metadata_id": uuid4(),
                 "sample_id": ctx.sample_map[old_meta.sample_id],
             },
-            deep=True,  # Deep copy nested data and metadata_schema dicts
         )
         session.add(new_meta)
 
@@ -526,7 +525,9 @@ def _copy_with_updates(
     Args:
         entity: Source entity to copy.
         updates: Fields to override (e.g., remapped IDs).
-        deep: If True, deep copy nested structures (dicts, lists).
+        deep: If True, apply copy.deepcopy() after model_dump(). Only needed
+            for non-JSON mutable fields; JSON-typed fields are already
+            deep copied by Pydantic's serialization.
         exclude: Fields to exclude from copy. Excluded fields will use model defaults.
                  Defaults to _EXCLUDE_ON_COPY (created_at, updated_at).
 
--- a/lightly_studio/tests/resolvers/collection_resolver/test_deep_copy.py
+++ b/lightly_studio/tests/resolvers/collection_resolver/test_deep_copy.py
@@ -186,6 +186,64 @@ def test_deep_copy__with_metadata(test_db: Session) -> None:
     assert copied_metadata.data["temperature"] == 30
 
 
+def test_deep_copy__with_nested_metadata(test_db: Session) -> None:
+    """Verify nested metadata structures are properly deep copied.
+
+    This test ensures that modifying nested dicts/lists in copied metadata
+    does not affect the original metadata (i.e., model_dump() properly
+    deep copies JSON fields).
+    """
+    # Arrange
+    original = create_collection(session=test_db, collection_name="original")
+    img = create_image(
+        session=test_db, collection_id=original.collection_id, file_path_abs="/test.png"
+    )
+
+    # Set metadata with nested structures
+    img.sample["config"] = {"threshold": 0.5, "options": {"enabled": True, "mode": "auto"}}
+    img.sample["some_list"] = ["el1", "el2"]
+
+    # Act
+    copied = collection_resolver.deep_copy(
+        session=test_db,
+        root_collection_id=original.collection_id,
+        copy_name="copied",
+    )
+
+    # Get copied metadata
+    copied_samples = sample_resolver.get_filtered_samples(
+        session=test_db,
+        filters=SampleFilter(collection_id=copied.collection_id),
+    )
+    copied_sample = copied_samples.samples[0]
+    copied_metadata = metadata_resolver.get_by_sample_id(
+        session=test_db, sample_id=copied_sample.sample_id
+    )
+    assert copied_metadata is not None
+
+    # Assert - nested data was copied correctly
+    assert copied_metadata.data["config"]["threshold"] == 0.5
+    assert copied_metadata.data["config"]["options"]["enabled"] is True
+    assert copied_metadata.data["config"]["options"]["mode"] == "auto"
+    assert copied_metadata.data["some_list"] == ["el1", "el2"]
+
+    # Act - modify nested values in the copy
+    copied_metadata.data["config"]["threshold"] = 0.9
+    copied_metadata.data["config"]["options"]["enabled"] = False
+    copied_metadata.data["config"]["options"]["mode"] = "manual"
+    copied_metadata.data["some_list"].append("el3")
+
+    # Assert - original metadata is unchanged
+    original_metadata = metadata_resolver.get_by_sample_id(
+        session=test_db, sample_id=img.sample.sample_id
+    )
+    assert original_metadata is not None
+    assert original_metadata.data["config"]["threshold"] == 0.5
+    assert original_metadata.data["config"]["options"]["enabled"] is True
+    assert copied_metadata.data["config"]["options"]["mode"] == "manual"
+    assert original_metadata.data["some_list"] == ["el1", "el2"]
+
+
 def test_deep_copy__with_embeddings(test_db: Session) -> None:
     # Arrange
     original = create_collection(session=test_db, collection_name="original")


lightly_studio/src/lightly_studio/resolvers/collection_resolver/deep_copy.py
```diff
@@ -0,0 +1,510 @@
...
+
+    for old_label in labels:
+        new_id = uuid4()
+        ctx.annotation_label_map[old_label.annotation_label_id] = new_id
+
+        new_label = AnnotationLabelTable(
+            annotation_label_id=new_id,
+            dataset_id=new_dataset_id,
+            annotation_label_name=old_label.annotation_label_name,
+            created_at=str(datetime.now(timezone.utc)),
```
mihnea-necsulescu: is there a reason why the [AnnotationLabelTable.created_at](https://github.com/lightly-ai/lightly-studio/blob/main/lightly_studio/src/lightly_studio/models/annotation_label.py#L42) is a `str` instead of `datetime`, or can I add a todo comment to address this?

lightly_studio/src/lightly_studio/api/routes/api/collection.py
```diff
@@ -133,3 +133,18 @@ def has_embeddings(
     return embedding_utils.collection_has_embeddings(
         session=session, collection_id=collection.collection_id
     )
+
+
+@collection_router.post("/collections/{collection_id}/deep-copy-test")
```
michal-lightly, line 139: I assume this is temporary? Let's add a TODO comment to remove.

lightly_studio/src/lightly_studio/resolvers/collection_resolver/deep_copy.py
```diff
@@ -0,0 +1,510 @@
...
+    tag_map: dict[UUID, UUID] = field(default_factory=dict)
+    annotation_label_map: dict[UUID, UUID] = field(default_factory=dict)
+
+
+def deep_copy(
+    session: Session,
+    root_collection_id: UUID,
+    copy_name: str,
+) -> CollectionTable:
+    """Deep copy a root collection with all related entities.
```
michal-lightly, line 69: I was expecting this function to work recursively, copying collection by collection from top to bottom. Is the current approach a good alternative?

lightly_studio/src/lightly_studio/resolvers/collection_resolver/deep_copy.py
```diff
@@ -0,0 +1,510 @@
...
+        root_collection_id: Root collection ID to copy.
+        copy_name: Name for the new collection.
+
+    Returns:
+        The newly created root collection.
+    """
+    ctx = DeepCopyContext()
+
+    # 1. Copy collection hierarchy.
+    hierarchy = collection_resolver.get_hierarchy(session, root_collection_id)
```
michal-lightly: Can you please add keyword arguments everywhere in this function?

lightly_studio/src/lightly_studio/resolvers/collection_resolver/deep_copy.py
```diff
@@ -0,0 +1,510 @@
...
+    _copy_metadata(session, old_sample_ids, ctx)
+    _copy_embeddings(session, old_sample_ids, ctx)
+    session.flush()
+
+    # 5. Copy link tables.
+    _copy_sample_tag_links(session, old_sample_ids, ctx)
+    _copy_annotation_tag_links(session, old_sample_ids, ctx)
+    _copy_sample_group_links(session, old_sample_ids, ctx)
+
+    return root
```
michal-lightly, line 118: This function needs to be updated everytime we add a table. Let's enforce this in code.

One idea how to do this: Add an assert at the top of this function on the number of tables in the DB, with an explanatory comment why it is there.

lightly_studio/src/lightly_studio/resolvers/collection_resolver/deep_copy.py
```diff
@@ -0,0 +1,510 @@
...
+    ctx: DeepCopyContext,
+) -> CollectionTable:
+    """Copy collection hierarchy, maintaining parent-child relationships."""
+    root: CollectionTable | None = None
+    old_root_name = hierarchy[0].name
+
+    # Generate new UUIDs for all collections and build the mapping.
+    for old_coll in hierarchy:
+        new_id = uuid4()
+        ctx.collection_map[old_coll.collection_id] = new_id
```
michal-lightly: Optional: This is more pythonic:
```suggestion
    ctx.collection_map = {
        old_coll.collection_id: uuid4() for old_coll in hierarchy
    }
```

lightly_studio/src/lightly_studio/resolvers/collection_resolver/deep_copy.py
```diff
@@ -0,0 +1,510 @@
...
+
+        new_coll = CollectionTable(
+            collection_id=new_id,
+            name=derived_name,
+            sample_type=old_coll.sample_type,
+            parent_collection_id=new_parent_id,
+            group_component_name=old_coll.group_component_name,
+            group_component_index=old_coll.group_component_index,
+            created_at=datetime.now(timezone.utc),
+            updated_at=datetime.now(timezone.utc),
```
michal-lightly: Filled automatically, can be omitted.

michal-lightly: Also here is a maintenance risk. If we add a table column that has a default, forgetting to add the field here will lead to an inconsistent copy.

Not sure how to address this. One option would be to copy all the values from the old object and then only update those that are necessary. WDYT?

lightly_studio/src/lightly_studio/resolvers/collection_resolver/deep_copy.py
```diff
@@ -0,0 +1,510 @@
...
+        )
+        session.add(new_coll)
+        session.flush()  # Flush each collection so it's visible for FK checks.
+
+        if root is None:
+            root = new_coll
+
+    return root  # type: ignore[return-value]
```
michal-lightly: Why do we need the type ignore?

lightly_studio/src/lightly_studio/resolvers/collection_resolver/deep_copy.py
```diff
@@ -0,0 +1,510 @@
...
+
+def _copy_samples(
+    session: Session,
+    old_collection_ids: list[UUID],
+    ctx: DeepCopyContext,
+) -> None:
+    """Copy all samples, remapping collection_id to new collections."""
+    samples = session.exec(
+        select(SampleTable).where(col(SampleTable.collection_id).in_(old_collection_ids))
+    ).all()
```
michal-lightly, line 178: Does this return a generator or all is loaded in memory? A generator would be sufficient, but I'm not sure if adding new samples would mess with it.

lightly_studio/src/lightly_studio/resolvers/collection_resolver/deep_copy.py
michal-lightly: It is an oversight, we should change it.

lightly_studio/src/lightly_studio/resolvers/collection_resolver/deep_copy.py
michal-lightly: AI suggested model_dump + update. There's also model_copy(update=...), but there are some caveats with that one.

lightly_studio/tests/resolvers/collection_resolver/test_deep_copy.py
```diff
@@ -0,0 +1,239 @@
...
+    create_collection,
+    create_embedding_model,
+    create_image,
+    create_sample_embedding,
+)
+
+
+def test_deep_copy__empty_collection(test_db: Session) -> None:
+    # Arrange
+    original = create_collection(test_db, collection_name="original")
```
michal-lightly: Please keep the keywords. It makes the code more readable.

lightly_studio/tests/resolvers/collection_resolver/test_deep_copy.py
```diff
@@ -0,0 +1,239 @@
+"""Tests for deep_copy resolver."""
```
michal-lightly, line 1: We should discuss the testing strategy.

We have a separate function to copy every table. These tests do not even invoke all the functions. On the other hand, making a comprehensive test copying all sample types on various levels in the hierarchy can be laborious. I suggest we discuss it.

lightly_studio/src/lightly_studio/resolvers/collection_resolver/deep_copy.py
mihnea-necsulescu: that's how I initially wanted to do it, but there are some aspects that made this approach more appealing to me:
- it couldn't be purely recursive because of the many to many link tables that would have to be inserted in batch at the end, so it would become a hybrid approach - I think it's clearer to pick a single way to do things (when possible) and stick to it
- the recursive version would probably have to have a "generic" function that copies type specific sample tables based on the sample type, which would translate to a big if/else chain which I personally don't like - the current approach seems cleaner to me: just call all the copy functions -> empty results from the select statements would be no-ops

lightly_studio/src/lightly_studio/resolvers/collection_resolver/deep_copy.py
mihnea-necsulescu: did this with hardcoded values for now, as this PR is getting quite big. I'm thinking as a future improvement, we could add `DeepCopyable` and `ExcludeFromDeepCopy` classes and appropriately extend them in every model and compute the number of tables to be copied based on them

lightly_studio/src/lightly_studio/resolvers/collection_resolver/deep_copy.py
mihnea-necsulescu: added a helper function that relies on model_dump, updates, and model reconstruction + excluded the timestamps from the dump so they can be automatically recomputed

lightly_studio/src/lightly_studio/resolvers/collection_resolver/deep_copy.py
mihnea-necsulescu: forgot about it.. root is declared as CollectionTable | None; added an assert right above

lightly_studio/src/lightly_studio/resolvers/collection_resolver/deep_copy.py
mihnea-necsulescu: all() loads everything in memory. generators could be ok, considering we're adding samples with different collection ids that won't be matched, but considering we're using DuckDB, who knows what may happen.. I'd keep the all() and do batching eventually, if we run into bottlenecks - left a todo

lightly_studio/tests/resolvers/collection_resolver/test_deep_copy.py
mihnea-necsulescu: yes, I only added some minimal tests to cover the basic functionality, as testing everything would take a long time. I'm not sure what would be the proper way to fully test this - maybe with groups? we could construct a group collection, and every group could have a component for each sample type (then also make sure images have all annotation types, etc.); but I think that whatever we go for, it should be done as part of a follow-up/tech debt issue

lightly_studio/src/lightly_studio/resolvers/collection_resolver/deep_copy.py
```diff
@@ -0,0 +1,552 @@
...
+
+_TOTAL_TABLES_COUNT = _COPIED_TABLES_COUNT + _NOT_COPIED_TABLES_COUNT
+
+T = TypeVar("T", bound=SQLModel)
+
+# Fields to exclude when copying - these have default_factory and should be regenerated.
+_EXCLUDE_ON_COPY: set[str] = {"created_at", "updated_at"}
+
+
+def _copy_with_updates(
```
michal-lightly: Let's move the helper functions below.

lightly_studio/src/lightly_studio/resolvers/collection_resolver/deep_copy.py
```diff
@@ -0,0 +1,552 @@
...
+) -> T:
+    """Create a copy of an entity with specified field updates.
+
+    Uses model_dump() to extract field values, then reconstructs a new instance.
+    This ensures new fields added to models are automatically included in copies.
+
+    Args:
+        entity: Source entity to copy.
+        updates: Fields to override (e.g., remapped IDs).
+        deep: If True, deep copy nested structures (dicts, lists).
```
michal-lightly: Don't we want this to always be True?

lightly_studio/src/lightly_studio/resolvers/collection_resolver/deep_copy.py
```diff
@@ -0,0 +1,552 @@
...
+    if exclude is None:
+        exclude = _EXCLUDE_ON_COPY
+    data = entity.model_dump(exclude=exclude)
+    if deep:
+        data = copy.deepcopy(data)
+    data.update(updates)
+    return type(entity)(**data)
+
+
+def _verify_table_coverate() -> None:
```
michal-lightly: Typo:
```suggestion
def _verify_table_coverage() -> None:
```

lightly_studio/src/lightly_studio/resolvers/collection_resolver/deep_copy.py
```diff
@@ -0,0 +1,552 @@
...
+            {
+                "collection_id": new_id,
+                "name": derived_name,
+                "parent_collection_id": new_parent_id,
+            },
+        )
+        session.add(new_coll)
+        session.flush()  # Flush each collection so it's visible for FK checks.
+
+        if root is None:
```
michal-lightly, line 162: Minor: Let's add a comment. This stores the new root; it works because the root is the first collection. We should perhaps hocument that we expect this from the `hierarchy` argument.

lightly_studio/src/lightly_studio/resolvers/collection_resolver/deep_copy.py
```diff
@@ -0,0 +1,552 @@
...
+
+
+def _copy_annotation_labels(
+    session: Session,
+    root_collection_id: UUID,
+    ctx: DeepCopyContext,
+) -> None:
+    """Copy annotation labels (belong to root collection only)."""
+    labels = session.exec(
+        select(AnnotationLabelTable).where(AnnotationLabelTable.dataset_id == root_collection_id)
```
michal-lightly, line 225: Optional: I think we should call `root_collection_id` rather `dataset_id` everywhere. Don't feel strongly though.

lightly_studio/src/lightly_studio/resolvers/collection_resolver/deep_copy.py
```diff
@@ -0,0 +1,552 @@
...
+            session.add(new_inst_seg)
+    elif annotation_type.value == "semantic_segmentation":
+        old_sem_seg = session.get(SemanticSegmentationAnnotationTable, old_sample_id)
+        if old_sem_seg:
+            new_sem_seg = _copy_with_updates(
+                old_sem_seg,
+                {"sample_id": new_sample_id},
+            )
+            session.add(new_sem_seg)
+
```
michal-lightly, line 400: This is missing an `else` with raising an error

lightly_studio/src/lightly_studio/resolvers/collection_resolver/deep_copy.py
```diff
@@ -0,0 +1,552 @@
...
+    embeddings = session.exec(
+        select(SampleEmbeddingTable).where(col(SampleEmbeddingTable.sample_id).in_(old_sample_ids))
+    ).all()
+
+    for old_emb in embeddings:
+        new_emb = _copy_with_updates(
+            old_emb,
+            {
+                "sample_id": ctx.sample_map[old_emb.sample_id],
+                "embedding": old_emb.embedding.copy(),
```
michal-lightly: Deep copy does not handle this?

lightly_studio/src/lightly_studio/resolvers/collection_resolver/deep_copy.py
```diff
@@ -0,0 +1,552 @@
...
+        select(AnnotationTagLinkTable).where(
+            col(AnnotationTagLinkTable.annotation_sample_id).in_(old_sample_ids)
+        )
+    ).all()
+
+    for old_link in links:
+        if (
+            old_link.annotation_sample_id is not None
+            and old_link.tag_id is not None
+            and old_link.tag_id in ctx.tag_map
```
michal-lightly, line 486: Why do we need the checks here? Also in the function above.

lightly_studio/tests/resolvers/collection_resolver/test_deep_copy.py
michal-lightly: In a follow-up sounds good.

I think it would be sufficient to create a dataset that uses all the tables, and make a deep copy - even without asserts. That will at least run all the code. Good point that we need groups for that (or use multiple dataset to copy).

Can you create the issue?

lightly_studio/src/lightly_studio/resolvers/collection_resolver/deep_copy.py
michal-lightly: I think hardcoded values are great for our usecase

lightly_studio/src/lightly_studio/resolvers/collection_resolver/deep_copy.py
mihnea-necsulescu: no, if that was always true, copy.deepcopy() would also be called (even when not necessary) which is a costly operation and would make this much slower than needed. it's currently only true when copying metadata models because of the data and metadata_schema dictionaries, but I'm now thinking if it's even needed for those - I think model_dump may be sufficient, but I'm not entirely sure, so I'd keep it for now, just in case

lightly_studio/src/lightly_studio/resolvers/collection_resolver/deep_copy.py
mihnea-necsulescu: you're right, I saw that this is generated as a numpy array, but missed that it's saved as a float list in the db - removed it

lightly_studio/src/lightly_studio/resolvers/collection_resolver/deep_copy.py
mihnea-necsulescu: mypy was complaining because the ids are defined as optionals in AnnotationTagLinkTable
(error: Invalid index type "UUID | None" for "dict[UUID, UUID]"; expected type "UUID"  [index])

lightly_studio/src/lightly_studio/resolvers/collection_resolver/deep_copy.py
mihnea-necsulescu: I see that it was previously called root_collection_id here, too, but it was renamed back - I don't know the context, though

lightly_studio/tests/resolvers/collection_resolver/test_deep_copy.py
mihnea-necsulescu: created and added to triage

lightly_studio/src/lightly_studio/resolvers/collection_resolver/deep_copy.py
michal-lightly: Some context is that after the dataset->collection rename, we had some places that used `dataset_id` and `root_collection_id`. I tried to unify it to just use `dataset_id`. I'm not sure though if all the places were updated.

lightly_studio/src/lightly_studio/resolvers/collection_resolver/deep_copy.py
michal-lightly: So this is actually working without `deep=True`?

michal-lightly (approved):
LGTM, thanks!

lightly_studio/src/lightly_studio/resolvers/collection_resolver/deep_copy.py
mihnea-necsulescu: no, because it's just a list of primitives and model_dump will create a new list in those cases

it was actually not needed for the metadata data dict either, since the values are stored as JSON - removed the deep=True for that case, too (it's currently not used anymore, but it may be useful in the future, so it's still there) ,and also added a new unit test to confirm this

PR #400, opened 2026-01-13
Update video e2e tests

horatiualmasan:
## What has changed and why?

Add e2e tests to check the filtering functionality for videos.
Add e2e tests for video frames. Grid scroll and filtering.

## How has it been tested?

New tests

- [ ] Yes


lightly_studio_view/e2e/pages/video-frames-page.ts
```diff
@@ -0,0 +1,41 @@
...
+
+    async clickLabel(label: string) {
+        const responsePromise = this.page.waitForResponse(
+            (response) => response.url().includes('/frame') && response.status() === 200
+        );
+
+        const labelMenuItem = this.getLabelsMenuItem(label);
+        await labelMenuItem.click();
+
+        await responsePromise;
```
LeonardoRosaa, line 34: Why don’t we invoke line 27 right here?

lightly_studio_view/e2e/pages/video-frames-page.ts
```diff
@@ -0,0 +1,41 @@
...
+    constructor(public readonly page: Page) {
+        this.page = page;
+    }
+
+    async goto() {
+        await this.page.goto('/');
+        await this.page.getByTestId('navigation-menu-frames').click();
+
+        // Wait for frame grid items to be visible
+        await expect(this.getVideoFrames().first()).toBeVisible({
```
LeonardoRosaa, line 13: I’m not sure about this timeout. From my point of view, it is too long.

horatiualmasan: It is the same we use for the other pages and also within the tests. I can add a todo to check if it can be lowered everywhere

lightly_studio_view/e2e/pages/video-frames-page.ts
horatiualmasan: I used the existing implementation from the other pages. I think it would be possible to miss the  response if you invoke after the click.

lightly_studio_view/e2e/pages/video-frames-page.ts
LeonardoRosaa: Okay, let's do that.

PR #403, opened 2026-01-14
Move *Annotation classes under core/annotation

lukas-lightly:
## What has changed and why?
We usually group similar files in the same directories.

## How has it been tested?
Existing tests

- [ ] Yes


lightly_studio/src/lightly_studio/core/sample.py
```diff
@@ -15,8 +15,8 @@
 from lightly_studio.resolvers import caption_resolver, metadata_resolver, tag_resolver
 
 from .annotation import Annotation
-from .instance_segmentation import InstanceSegmentationAnnotation
-from .object_detection import ObjectDetectionAnnotation
+from .annotation.instance_segmentation import InstanceSegmentationAnnotation
+from .annotation.object_detection import ObjectDetectionAnnotation
```
JonasWurst, line 19: general question: why should we use relative import here?

lukas-lightly: It's mostly for convenience and it's making it clear that the imports are "not too far away". But I don't feel strongly about this.

PR #407, opened 2026-01-14
Add ClassificationAnnotation class

lukas-lightly:
## What has changed and why?
Another type of annotation implemented as a Python type.

## How has it been tested?
Added a test

- [ ] Yes

### DIFF
--- a/lightly_studio/src/lightly_studio/core/annotation/classification.py
+++ b/lightly_studio/src/lightly_studio/core/annotation/classification.py
@@ -1,0 +1,22 @@
+"""Interface for classification annotations."""
+
+from lightly_studio.core.annotation import Annotation
+from lightly_studio.models.annotation.annotation_base import (
+    AnnotationBaseTable,
+    AnnotationType,
+)
+
+
+class ClassificationAnnotation(Annotation):
+    """Class for classification annotations."""
+
+    def __init__(self, annotation_base: AnnotationBaseTable) -> None:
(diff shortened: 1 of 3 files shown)


lightly_studio/src/lightly_studio/core/annotation/classification.py
```diff
@@ -0,0 +1,22 @@
...
+    def __init__(self, annotation_base: AnnotationBaseTable) -> None:
+        """Initialize the Annotation.
+
+        Args:
+            annotation_base: The AnnotationBaseTable SQLAlchemy model instance.
+        """
+        if annotation_base.annotation_type != AnnotationType.CLASSIFICATION:
+            raise ValueError("Expected annotation type: classification")
+
+        super().__init__(annotation_base)
```
michal-lightly, line 22: Why did SemanticSegmentationAnnotation store `inner` but this class does not need it?

lukas-lightly: Object detection, semantic segmentation and instance segmentation all have an additional table in the database, see the DB model: https://github.com/lightly-ai/lightly-studio/blob/39c06ceaad094fd50e7daf107f8cea33269a8abb/lightly_studio/src/lightly_studio/models/annotation/annotation_base.py#L88-107

But classification doesn't have any additional data, it's just a record of association, i.e. "these samples are cats".

PR #411, opened 2026-01-14
Fix missing few shot classifier link in the menu

michal-lightly:
## What has changed and why?

Fix missing few shot classifier link in the menu.

## How has it been tested?

Manually.

- [ ] Yes
- [x] Not needed (internal change) - the bug was not in a previous release

### DIFF
--- a/lightly_studio_view/e2e/pages/index.ts
+++ b/lightly_studio_view/e2e/pages/index.ts
@@ -4,3 +4,4 @@ export { SamplesPage } from './samples-page';
 export { SampleDetailsPage } from './sample-details-page';
 export { CaptionsPage } from './captions-page';
 export { VideosPage } from './videos-page';
+export { VideoFramesPage } from './video-frames-page';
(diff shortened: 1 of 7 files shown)


lightly_studio_view/src/lib/components/Header/Header.svelte
```diff
@@ -21,7 +21,7 @@
     const { settingsStore } = useSettings();
 
     const hasEmbeddingsQuery = useHasEmbeddings({ collectionId: collection.collection_id });
-    const hasEmbeddings = $derived(!!hasEmbeddingsQuery.data);
+    const hasEmbeddings = $derived(!!$hasEmbeddingsQuery.data);
```
michal-lightly, line 24: This is the fix - the `$` subscribes to the Svelte store instead of just using the value.

lightly_studio_view/src/lib/components/Header/Header.test.ts
```diff
@@ -78,11 +78,13 @@ describe('Header', () => {
...
-            isLoading: readable(false),
-            error: readable(null)
-        });
+        useHasEmbeddings.mockReturnValue(
+            readable({
+                data: true,
+                isLoading: false,
+                error: null
+            })
+        );
```
michal-lightly, line 87: Properly mock the store.

PR #412, opened 2026-01-14
Implement brush tool for instance segmentation

LeonardoRosaa:
## What has changed and why?

Introduces the brush tool for instance segmentation annotations. Users can create and edit instance segmentation annotations.

Also adds a set of unit tests.

## How has it been tested?

https://github.com/user-attachments/assets/204f054d-51e5-4337-9ce8-70b1fdf41ee1

- [x] Yes

### DIFF
--- a/lightly_studio/src/lightly_studio/core/dataset.py
+++ b/lightly_studio/src/lightly_studio/core/dataset.py
@@ -34,44 +34,7 @@ T = TypeVar("T", bound=Sample)
 
 
 class Dataset(Generic[T], ABC):
-    """A LightlyStudio Dataset.
-
-    It can be created or loaded using one of the static methods:
-    ```python
-    dataset = ImageDataset.create()
-    dataset = ImageDataset.load()
-    dataset = ImageDataset.load_or_create()
-    ```
-
-    Samples can be added to the dataset using various methods:
-    ```python
-    dataset.add_images_from_path(...)
-    dataset.add_samples_from_yolo(...)
-    dataset.add_samples_from_coco(...)
-    dataset.add_samples_from_coco_caption(...)
-    dataset.add_samples_from_labelformat(...)
-    dataset.add_videos_from_path(...)
-    ```
-
-    The dataset samples can be queried directly by iterating over it or slicing it:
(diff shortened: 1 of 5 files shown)


lightly_studio_view/src/lib/components/SampleAnnotation/utils.ts
```diff
@@ -124,3 +124,51 @@ export const withAlpha = (color: string, alpha: number) =>
         const [r, g, b] = c.split(',').map(Number);
         return `rgba(${r}, ${g}, ${b}, ${alpha})`;
     });
+
+export const applyBrushToMask = (
```
ikondrat, line 128: I would add explanation in jsdoc what you are doing here

lightly_studio_view/src/lib/components/SampleAnnotation/utils.test.ts
```diff
@@ -0,0 +1,54 @@
...
+
+        const rle = encodeBinaryMaskToRLE(mask);
+
+        expect(rle).toEqual([0]);
+    });
+
+    it('decodes a RLE mask', () => {
+        const mask = decodeRLEToBinaryMask([2, 3, 1], 3, 2);
+
+        expect(mask).toEqual(new Uint8Array([0, 0, 1, 1, 1, 0]));
```
ikondrat, line 57: I would explain a bit more expected result

It would be helpfull if you will not touch this code for awhile

ikondrat (approved):
It is easy to miss smth in this size of PR,
would be better to have smaller PR



I would ask about testing @JonasWurst

lightly_studio_view/src/lib/components/SampleAnnotation/utils.ts
LeonardoRosaa: I added comments. It seems better to understand

lightly_studio_view/src/lib/components/SampleAnnotation/utils.test.ts
LeonardoRosaa: Done!

PR #415, opened 2026-01-15
Make it possible to group annotations

lukas-lightly:
## What has changed and why?
Annotations are samples and samples can be grouped into collections. Each collection can have a name, thus we can have multiple types, e.g. "ground truth" annotations and "prediction' annotations. They come fundamentally from different sources.

We've added an optional `name` parameter to
`annotation_resolver.create_many()` which allows such naming. We've also added a new function `annotation_resolver.get_all_by_collection_name()`, which gets only annotations from a given collection.

## How has it been tested?
Added a test.

- [ ] Yes


lightly_studio/src/lightly_studio/resolvers/collection_resolver/get_or_create_child_collection.py
```diff
@@ -10,8 +10,12 @@
 from lightly_studio.resolvers import collection_resolver
 
 
+def _matches_name(collection_name: str, filter_name: str | None) -> bool:
```
michal-lightly: Style: We add helper functions to the bottom of the file.

lightly_studio/src/lightly_studio/resolvers/collection_resolver/get_or_create_child_collection.py
```diff
@@ -35,7 +40,11 @@ def get_or_create_child_collection(
     collection = collection_resolver.get_by_id(session=session, collection_id=collection_id)
     if collection is None:
         raise ValueError(f"Collection with id {collection_id} not found.")
-    child_collections = [col for col in collection.children if col.sample_type == sample_type]
+    child_collections = [
+        col
+        for col in collection.children
+        if col.sample_type == sample_type and _matches_name(col.name, name)
```
michal-lightly: Style: Write out the argument keywords:
```suggestion
        if col.sample_type == sample_type and _matches_name(collection_name=col.name, filter_name=name)
```

lightly_studio/src/lightly_studio/resolvers/annotation_resolver/get_all_by_collection_name.py
```diff
@@ -0,0 +1,37 @@
...
+from sqlmodel import Session
+
+from lightly_studio.api.routes.api.validators import Paginated
+from lightly_studio.resolvers.annotations.annotations_filter import AnnotationsFilter
+from lightly_studio.resolvers.collection_resolver import get_by_name
+
+from .get_all import GetAllAnnotationsResult, get_all
+
+
+def get_all_by_collection_name(
```
michal-lightly, line 14: There is a limitation with this design for annotation tasks: The collection name is global, and has a unique constraint. That means that annotations for different collections (e.g. say different cameras around a car) cannot have the same task name.

I think for now, we should merge. But it is not a desired behaviour. One idea how to resolve it: Make collection name non-unique. I can imagine that dataset names should be unique - for that, we'd need to do something smart (can e.g. a tuple `(parent_dataset, dataset_name)` be added a unique constraint?)

lightly_studio/tests/resolvers/test_annotation_resolver.py
```diff
@@ -443,6 +443,78 @@ def test_get_all_returns_filtered_by_collection_results(
...
+        parent_collection_id=collection.collection_id,
+        annotations=[
+            AnnotationCreate(
+                parent_sample_id=image.sample_id,
+                annotation_label_id=cat_label.annotation_label_id,
+                annotation_type=AnnotationType.OBJECT_DETECTION,
+                x=12,
+                y=13,
+                width=47,
+                height=52,
```
michal-lightly, line 483: Optional: Would be  nice to add a `confidence` field since this is a prediction.

lightly_studio/src/lightly_studio/resolvers/collection_resolver/get_or_create_child_collection.py
lukas-lightly: Done

lightly_studio/tests/resolvers/test_annotation_resolver.py
lukas-lightly: Done

lightly_studio/src/lightly_studio/resolvers/annotation_resolver/get_all_by_collection_name.py
lukas-lightly: I didn't know about this constraint, but it makes sense given that we have the `collection_resolver.get_by_name()` function.

Do you suggest to fix this immediately? In the next sprint? Or later?

PR #425, opened 2026-01-16
Prepare Python interface for semantic segmentation table removal

michal-lightly:
## What has changed and why?

Prepare Python interface for semantic segmentation table removal:
* Read from SegmentationAnnotationTable in SemanticSegmentationAnnotation function

## How has it been tested?

Tested in a draft PR. In this PR, marking a test temporarily skipped.

- [ ] Yes

### DIFF
--- a/lightly_studio/src/lightly_studio/resolvers/annotation_resolver/__init__.py
+++ b/lightly_studio/src/lightly_studio/resolvers/annotation_resolver/__init__.py
@@ -11,6 +11,9 @@ from lightly_studio.resolvers.annotation_resolver.delete_annotations import (
     delete_annotations,
 )
 from lightly_studio.resolvers.annotation_resolver.get_all import get_all
+from lightly_studio.resolvers.annotation_resolver.get_all_by_collection_name import (
+    get_all_by_collection_name,
+)
 from lightly_studio.resolvers.annotation_resolver.get_all_with_payload import (
     get_all_with_payload,
 )
@@ -34,6 +37,7 @@ __all__ = [
     "delete_annotation",
     "delete_annotations",
     "get_all",
+    "get_all_by_collection_name",
     "get_all_with_payload",
     "get_by_id",
     "get_by_id_with_payload",
--- a/lightly_studio/src/lightly_studio/resolvers/annotation_resolver/create_many.py
+++ b/lightly_studio/src/lightly_studio/resolvers/annotation_resolver/create_many.py
@@ -29,6 +29,7 @@ def create_many(
     session: Session,
     parent_collection_id: UUID,
     annotations: list[AnnotationCreate],
+    collection_name: str | None = None,
 ) -> list[UUID]:
     """Create multiple annotations in bulk with their respective type-specific details.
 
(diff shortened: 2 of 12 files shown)


lightly_studio/src/lightly_studio/core/annotation/semantic_segmentation.py
```diff
@@ -18,19 +16,29 @@ class SemanticSegmentationAnnotation(Annotation):
...
+    x = DBField(col(SegmentationAnnotationTable.x))
+    """X coordinate (px) of the segmentation bounding box."""
+    y = DBField(col(SegmentationAnnotationTable.y))
+    """Y coordinate (px) of the segmentation bounding box."""
+    width = DBField(col(SegmentationAnnotationTable.width))
+    """Width (px) of the segmentation bounding box."""
+    height = DBField(col(SegmentationAnnotationTable.height))
+    """Height (px) of the segmentation bounding box."""
+    segmentation_mask = DBField(col(SegmentationAnnotationTable.segmentation_mask))
+    """Segmentation mask given as a run-length encoding."""
```
michal-lightly, line 35: We will also have the bounding box.

lightly_studio/src/lightly_studio/core/sample.py
```diff
@@ -225,17 +225,22 @@ def annotations(self) -> list[Annotation]:
         """Returns all annotations."""
         annotations: list[Annotation] = []
         for annotation in self.sample_table.annotations:
-            if annotation.object_detection_details is not None:
+            if annotation.annotation_type == AnnotationType.OBJECT_DETECTION:
+                assert annotation.object_detection_details is not None, (
```
LeonardoRosaa, line 229: Should we add tests to cover these assertions?

LeonardoRosaa (approved):

lightly_studio/src/lightly_studio/core/sample.py
lukas-lightly: Note that each `*Annotation` class checks the type in its `__init__()` method:
* https://github.com/lightly-ai/lightly-studio/blob/059fa52ce98205ab6191ad1bb2aa0e89fb818c53/lightly_studio/src/lightly_studio/core/annotation/instance_segmentation.py#L45
* https://github.com/lightly-ai/lightly-studio/blob/059fa52ce98205ab6191ad1bb2aa0e89fb818c53/lightly_studio/src/lightly_studio/core/annotation/object_detection.py#L37
*  .... and the two others, too

This refactoring make this check performed twice. Should we remove one of the checks?

michal-lightly: I would keep two checks.

The sample_type check has to be sample.py because that’s the only way to distinguish between instance and semantic segmentation, and also classification. I changed it for object detection just for consistency.

And I think it is OK that is is also in specific classes. Because the classes "don't know" where they are called from.

Or do you have a suggestion?

michal-lightly: @LeonardoRosaa The asserts can't be broken, unless our code is broken (impossible to create an anootation that way). We already test that the asserts pass implicitly.

lukas-lightly: I see now, so my preference would be to keep the class checks - as you say you don't know where else you call it from. However, after the latest refactoring you can't distinguish instance and semantic segmentation in another way, so there's some redundant checks.

PR #432, opened 2026-01-19
API for getting all annotations from a collection

lukas-lightly:
## What has changed and why?
We want to expose creating and reading annotations collections.

Also added support to `annotations_service for` setting annotation collection name, e.g. you can have "ground truth" and "prediction" annotation collections.

## How has it been tested?
Added tests.

- [ ] Yes

### DIFF
--- a/lightly_studio_view/src/lib/components/TagCreateDialog/TagCreateDialog.svelte
+++ b/lightly_studio_view/src/lib/components/TagCreateDialog/TagCreateDialog.svelte
@@ -102,16 +102,6 @@
         tag_ids: $tagsSelected.size > 0 ? Array.from($tagsSelected) : undefined
     });
 
-    const sampleText = $derived.by(() => {
-        if (tagKind == 'sample') {
-            return itemsSelected.size > 1 ? 'samples' : 'sample';
-        }
-
-        if (tagKind == 'annotation') {
-            return itemsSelected.size > 1 ? 'annotations' : 'annotation';
-        }
-    });
-
     // setup global selection state
     const {
         getSelectedSampleIds,
@@ -133,6 +123,16 @@
         return ['samples', 'videos', 'video_frames'].includes(gridType)
             ? $selectedSampleIds
             : $selectedSampleAnnotationCropIds[collectionId];
+    });
+
+    const sampleText = $derived.by(() => {
+        if (tagKind == 'sample') {
+            return itemsSelected.size > 1 ? 'samples' : 'sample';
+        }
+
+        if (tagKind == 'annotation') {
+            return itemsSelected.size > 1 ? 'annotations' : 'annotation';
+        }
     });
     // setup initial dialog state
     let isDialogOpened = $state(false);
--- a/lightly_studio_view/src/lib/schema.d.ts
+++ b/lightly_studio_view/src/lib/schema.d.ts
@@ -2499,22 +2499,6 @@ export interface components {
             nextCursor?: number | null;
         };
         /**
-         * InstanceSegmentationAnnotationView
-         * @description API response model for instance segmentation annotations.
-         */
-        InstanceSegmentationAnnotationView: {
-            /** X */
-            x: number;
-            /** Y */
-            y: number;
-            /** Width */
-            width: number;
-            /** Height */
-            height: number;
-            /** Segmentation Mask */
-            segmentation_mask?: number[] | null;
-        };
-        /**
          * IntRange
          * @description Defines a range of integer-point values.
          */
@@ -2854,6 +2838,22 @@ export interface components {
             samples: {
                 [key: string]: string[];
             };
+        };
+        /**
+         * SegmentationAnnotationView
+         * @description API response model for instance segmentation annotations.
+         */
+        SegmentationAnnotationView: {
+            /** X */
+            x: number;
+            /** Y */
+            y: number;
+            /** Width */
+            width: number;
+            /** Height */
+            height: number;
+            /** Segmentation Mask */
+            segmentation_mask?: number[] | null;
         };
         /**
          * SelectionRequest


lightly_studio/src/lightly_studio/resolvers/annotation_resolver/get_all_by_collection_name.py
```diff
@@ -16,18 +16,18 @@ def get_all_by_collection_name(
     collection_name: str,
     pagination: Paginated | None = None,
 ) -> GetAllAnnotationsResult:
-    """Get all annotations for a collection by its name.
+    """Get all annotations from a annotation collection, given by its name.
```
lukas-lightly, line 19: I fixed docstrings in this file, because each mention of "collection" was ambiguous.

lightly_studio/src/lightly_studio/resolvers/annotations/annotations_filter.py
```diff
@@ -21,6 +22,7 @@ class AnnotationsFilter(BaseModel):
         description="Types of annotation to filter (e.g., 'object_detection')",
     )
     collection_ids: list[UUID] | None = Field(default=None, description="List of collection UUIDs")
+    collection_names: list[str] | None = Field(default=None, description="List of collection names")
```
lukas-lightly, line 25: I wonder whether we should pair annotation collection name with parent collection ID.

I assume that we can have multiple parent collections whose annotations are grouped into "predictions" and "ground truth", so the name won't be unique.

lightly_studio/src/lightly_studio/resolvers/annotations/annotations_filter.py
```diff
@@ -33,6 +35,13 @@ class AnnotationsFilter(BaseModel):
         default=None, description="List of sample UUIDs to filter annotations by"
     )
 
+    @model_validator(mode="after")
+    def check_collection_identifiers(self) -> AnnotationsFilter:  # noqa: N804
```
lukas-lightly, line 39: I'll add a test for this, but in the next PR as this one is already too big IMO.

lightly_studio/src/lightly_studio/resolvers/annotations/annotations_filter.py
michal-lightly: We should. Do you want to do this in this PR or a follow-up?

lightly_studio/src/lightly_studio/resolvers/annotations/annotations_filter.py
michal-lightly: Optional: I would not mind if we leave out the check. All conditions in the filter are combined with `and`, it is a valid behaviour to return an empty set.

lightly_studio/src/lightly_studio/api/routes/api/annotation.py
```diff
@@ -130,6 +130,32 @@ def read_annotations(
...
+)
+def read_annotations_by_collection_name(
+    annotation_collection_name: str,
+    session: SessionDep,
+    pagination: Annotated[PaginatedWithCursor, Depends()],
+    annotation_label_ids: Annotated[list[UUID] | None, Query()] = None,
+    tag_ids: Annotated[list[UUID] | None, Query()] = None,
+) -> GetAllAnnotationsResult:
+    """Retrieve a list of annotations from the database."""
+    return annotation_resolver.get_all_by_collection_name(
```
michal-lightly, line 145: I suggest to remove `get_all_by_collection_name` resolver function and just use `get_all` with properly set filter. WDYT?

lightly_studio/src/lightly_studio/api/routes/api/annotation.py
```diff
@@ -130,6 +130,32 @@ def read_annotations(
     )
 
 
+@annotations_router.get(
+    "/annotations/collection-name/{annotation_collection_name}",
```
michal-lightly, line 134: I am not sure if this endpoint fits well with the other endpoints here. Let's discuss in person.

horatiualmasan, line 134: Do we need a new endpoint? We could use `read_annotations` and add the `annotation_collection_name` as an optional parameter, and then as Michal suggested pass it in the filter

lightly_studio/src/lightly_studio/api/routes/api/annotation.py
lukas-lightly: I've done that locally, we can rely on `AnnotationsFilter` instead.

lightly_studio/src/lightly_studio/resolvers/annotations/annotations_filter.py
lukas-lightly: > We should. Do you want to do this in this PR or a follow-up?

I think it makes sense to do this in a followup. We need to drop the uniqueness requirement on annotation collection name first, before it makes sense to include the parent collection ID. Currently the annotation collection name has to be unique, so it's sufficient to identify the collection.

lightly_studio/src/lightly_studio/api/routes/api/annotation.py
lukas-lightly: Let's jump on a call.

We have two disjoint use cases:
* querying annotation collection by ID
* querying annotation collection by name and parent collection ID
In Rust I could consume an enum (sum type) as an argument, something like:
```rust
enum Query {
  ById(UUID),
  ByName(String, UUID)
}
```

To emulate this in a HTTP API, it's best IMO to have two separate endpoints, but I'm open to suggestions.

michal-lightly: The alternative would be to remove the collection ID from the arguments and pass the filter directly instead. IMO that would be the cleanest, however it would require changes to the current endpoint calls.

michal-lightly: In addition, there is the parallel "with payload" endpoint, broadening the "impact radius" of changes.

lukas-lightly:
Closing after offline discussion. `read_annotations_by_collection_name` endpoint won't be needed, as the UI always knows the annotation collection ID.

PR #436, opened 2026-01-20
Add video tag e2e test

horatiualmasan:
## What has changed and why?

Added new test for video tags
Updated `scroll to the bottom of the grid container` flaky test 

## How has it been tested?

test runs on ci

- [ ] Yes


.github/workflows/end2end_test.yml
```diff
@@ -293,7 +293,7 @@ jobs:
         if: failure()
         with:
           name: Upload Playwright Report
-          path: lightly_studio_view/playwright-report-${{ github.job }}
+          path: lightly_studio_view/playwright-report
```
horatiualmasan, line 296: Reports are still generated in playwright-report folder. With the old folder name no report is uploaded on test failure.

lightly_studio_view/e2e/pages/videos-page.ts
```diff
@@ -38,4 +43,60 @@ export class VideosPage {
...
+        for (let i = 0; i < count; i++) {
+            const isChecked = await boxes.nth(i).isChecked();
+            if (isChecked) {
+                selectedCount++;
+            }
+        }
+        return selectedCount;
+    }
+
+    async createTag(tagName: string): Promise<void> {
```
mihnea-necsulescu, line 60: the tag-related functions below could be extracted to a tag-utils class and then we could have all pages using tags encapsulate it, similarly to how caption-utils is done 

this could also be done in a follow-up, though, but in that case, we could rewrite this function to how it's done in annotations-page, as that looks more readable

lightly_studio_view/e2e/videos/video-grid.e2e-test.ts
```diff
@@ -4,13 +4,27 @@ import { youtubeVisVideosDataset } from './fixtures/youtubeVisVideosDataset';
...
         await gridContainer.evaluate((el) => {
-            el.scrollTop = el.scrollHeight;
+            el.scrollTo({ top: el.scrollHeight, behavior: 'auto' });
         });
 
+        // Wait for the API response
+        await responsePromise;
+
+        // Wait a bit for the grid to update
+        await page.waitForTimeout(100);
```
mihnea-necsulescu, line 26: I don't think this is needed, no? we're already waiting for the promise to complete, then we also have a big timeout below

horatiualmasan: This is the flaky test. I added a extra timeout to be sure, I would leave it for now to see if it fixes the test

lightly_studio_view/e2e/pages/videos-page.ts
horatiualmasan: I would do it in a follow up once the video tests are done, then we can clean up the duplicated code

mihnea-necsulescu (approved):

PR #439, opened 2026-01-20
feat: reducing amount of issues reported by svelte check from 95 to 68

ikondrat:
## What has changed and why?

This PR fixes issues reported by svelte-check

`npm run check` command

## How has it been tested?

By running `npm run check` and testing it locally as well 

- [ ] Yes

### DIFF
--- a/.claude/instructions.md
+++ b/.claude/instructions.md
@@ -1,4 +1,4 @@
-# Project Instructions
+Project Instructions
 
 Always follow the coding guidelines when working on this project:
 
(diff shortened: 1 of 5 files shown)


lightly_studio_view/package.json
```diff
@@ -90,10 +91,11 @@
         "d3-zoom": "^3.0.0",
         "embedding-atlas": "^0.10.0",
         "lodash": "^4.17.21",
```
LeonardoRosaa: Could we remove the `lodash`, or should we keep both?

ikondrat: yes we need to remove it. cleaned it up

PR #448, opened 2026-01-21
feat: reduce ammount of issues to 32

ikondrat:
## What has changed and why?

Fixed various issues related to type mismatch and just errors in the code

## How has it been tested?

Tested by running `npm run check` and running app locally with `make -C lightly_studio start-e2e`

- [ ] Yes

### DIFF
--- a/.github/workflows/end2end_test.yml
+++ b/.github/workflows/end2end_test.yml
@@ -293,7 +293,7 @@ jobs:
         if: failure()
         with:
           name: Upload Playwright Report
-          path: lightly_studio_view/playwright-report-${{ github.job }}
+          path: lightly_studio_view/playwright-report
           retention-days: 5
 
       - &cleanup-stop-server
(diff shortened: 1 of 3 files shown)


lightly_studio_view/src/lib/components/SampleDetails/SampleDetailsSidePanel/SampleDetailsSidePanel.svelte
```diff
@@ -36,7 +36,16 @@
         metadataItem
     }: Props = $props();
 
-    const tags = $derived(sample.tags.map((t) => ({ tagId: t.tag_id, name: t.name })) ?? []);
+    const tags = $derived(
+        sample.tags
```
LeonardoRosaa, line 40: Sorry, I didn't mean why this changed. The `tag_id` can't be null, right? If so, I think we can keep the previous version.

LeonardoRosaa: Ok. It seems `tag_id` can be null now.

PR #449, opened 2026-01-21
Add youtubevis loading from labelformat to add_videos- Prepare helpers

JonasWurst:
## What has changed and why?

This is the first PR to enable loading from youtubevis format via labelformat. It is adding the helpers that will be required in the add_video

## How has it been tested?

tests are implemented

- [ ] Yes


lightly_studio/src/lightly_studio/core/add_videos.py
```diff
@@ -173,6 +188,79 @@ def load_into_dataset_from_paths(
...
+        input_labels: The labelformat input containing video annotations.
+    """
+    videos = video_resolver.get_all_by_collection_id_with_frames(
+        session=session, collection_id=dataset_id
+    )
+    if not videos:
+        logger.warning("No videos found in dataset. Skipping annotation load.")
+        return
+
+    video_name_to_video = {Path(video.file_name).stem: video for video in videos}
```
michal-lightly: What if the user has a folder with `my_video.mp4` and `my_video.mov`?

lightly_studio/src/lightly_studio/core/add_videos.py
```diff
@@ -173,6 +188,79 @@ def load_into_dataset_from_paths(
...
+    input_labels: ObjectDetectionTrackInput | InstanceSegmentationTrackInput,
+) -> None:
+    """Load video frame annotations from a labelformat input into the dataset.
+
+    Args:
+        session: The database session.
+        dataset_id: The ID of the video dataset to load annotations into.
+        input_labels: The labelformat input containing video annotations.
+    """
+    videos = video_resolver.get_all_by_collection_id_with_frames(
```
michal-lightly: We should not load all frames in memory (or do we do it already elsewhere and optimising here does not make a difference?). From labelformat we get labels per video, so we can load frames video-by-video in the loop.

Moreover, the `get_all_by_collection_id_with_frames` is deprecated (there's a todo).

lightly_studio/src/lightly_studio/core/add_videos.py
```diff
@@ -292,3 +380,110 @@ def _get_frame_rotation_deg(frame: AVVideoFrame) -> int:
...
+    Args:
+        session: The database session.
+        dataset_id: The ID of the root collection the labels belong to.
+        input_labels: The labelformat input containing the categories.
+
+    Returns:
+        dict[int, UUID]: A mapping from category IDs to annotation label IDs.
+    """
+    label_map = {}
+    for category in tqdm(
```
michal-lightly: Let's not use tqdm for categories. It will be a short list.

lightly_studio/src/lightly_studio/core/add_videos.py
```diff
@@ -292,3 +380,110 @@ def _get_frame_rotation_deg(frame: AVVideoFrame) -> int:
     if matrix[0, 1] < 0:
         return 90
     return 270
+
+
+def _create_label_map(
```
michal-lightly: Isn't this function the same as `add_samples._create_label_map`? Let's move it to a common location and reuse.

lightly_studio/src/lightly_studio/core/add_videos.py
```diff
@@ -292,3 +380,110 @@ def _get_frame_rotation_deg(frame: AVVideoFrame) -> int:
...
+            segmentation = obj.segmentations[idx]
+            if segmentation is None:
+                continue
+
+            segmentation_rle: None | list[int] = None
+            if isinstance(segmentation, MultiPolygon):
+                box = segmentation.bounding_box().to_format(BoundingBoxFormat.XYWH)
+            elif isinstance(segmentation, BinaryMaskSegmentation):
+                box = segmentation.bounding_box.to_format(BoundingBoxFormat.XYWH)
+                segmentation_rle = segmentation._rle_row_wise  # noqa: SLF001
```
michal-lightly: There is a public function now, no need for noqa.

michal-lightly: Also, this bit is shared with `add_samples.py`. Let's factor it to a function.

lightly_studio/tests/core/test_add_videos.py
```diff
@@ -252,3 +280,334 @@ def test__configure_stream_threading__min_1_thread(mocker: MockerFixture) -> Non
     # Should use at least 1
     assert video_stream.codec_context.thread_type == ThreadType.AUTO
     assert video_stream.codec_context.thread_count == 1
+
+
+def test_create_label_map(db_session: Session) -> None:
```
michal-lightly: Let's keep the order the same as in the tested file. I would expect the main function to be tested first.

And this test can be removed when the function is factored out.

lightly_studio/src/lightly_studio/core/add_videos.py
```diff
@@ -173,6 +188,79 @@ def load_into_dataset_from_paths(
...
+            )
+
+        if video_annotation.video.number_of_frames != len(video.frames):
+            raise ValueError(
+                f"Number of frames in annotation ({video_annotation.video.number_of_frames}) "
+                f"does not match number of frames in video ({len(video.frames)}) "
+                f"for video {video.file_name}"
+            )
+
+        frame_number_to_id = {idx: frame.sample_id for idx, frame in enumerate(video.frames)}
```
michal-lightly: This code might not behave properly - the frames in `video.frames` are not guaranteed to be in order. It is necessary to use `frame.frame_number`.

lightly_studio/src/lightly_studio/core/add_videos.py
```diff
@@ -292,3 +380,110 @@ def _get_frame_rotation_deg(frame: AVVideoFrame) -> int:
...
+    frames_collection_id: UUID,
+    frame_number_to_id: dict[int, UUID],
+    video_annotation: VideoInstanceSegmentationTrack,
+    label_map: dict[int, UUID],
+) -> list[AnnotationCreate]:
+    """Process instance segmentation annotations for a single video."""
+    annotations = []
+    for idx in range(video_annotation.video.number_of_frames):
+        for obj in video_annotation.objects:
+            segmentation = obj.segmentations[idx]
```
michal-lightly: Optional:
```suggestion
    for frame_number, frame_id in frame_number_to_id.items():
        for obj in video_annotation.objects:
            segmentation = obj.segmentations[frame_number]
```

michal-lightly: Same in _process_video_annotations_object_detection

lightly_studio/src/lightly_studio/core/add_videos.py
JonasWurst: Yes, that is an issue I was thinking about. However when we stick to the youtube-vis format, we can not do anything here.

The only option is to warn the user, I intended to add this warning to the loading of the videos.

JonasWurst: here, we assume that only one of the two exists

JonasWurst: I added two comments explaining it, and made the method overall private

lightly_studio/src/lightly_studio/core/add_videos.py
JonasWurst: I planned to do those refactorings in follow up pr. happy to add it now already

lightly_studio/src/lightly_studio/core/labelformat_helpers.py
```diff
@@ -0,0 +1,123 @@
...
+from labelformat.model.object_detection import ObjectDetectionInput
+from labelformat.model.object_detection_track import ObjectDetectionTrackInput
+from sqlmodel import Session
+
+from lightly_studio.models.annotation.annotation_base import AnnotationCreate, AnnotationType
+from lightly_studio.models.annotation_label import AnnotationLabelCreate
+from lightly_studio.resolvers import annotation_label_resolver
+
+
+def get_annotation_create_segmentation(
```
michal-lightly: Optional:
```
get_segmentation_annotation_create
```

lightly_studio/src/lightly_studio/core/labelformat_helpers.py
```diff
@@ -0,0 +1,123 @@
...
+    )
+
+
+def create_label_map(
+    session: Session,
+    dataset_id: UUID,
+    input_labels: ObjectDetectionInput
+    | InstanceSegmentationInput
+    | ObjectDetectionTrackInput
+    | InstanceSegmentationTrackInput,
```
michal-lightly: Minor: There is a clever way to type-annotate such that we don't have to change this in the future. We can use structural typing (Protocols):
```py
class _LabelsWithCategories(Protocol):
    def get_categories(self) -> Iterable[Category]:
        ...
```

Side note: We should also independently improve the class structure in labelformat.

lightly_studio/tests/core/test_labelformat_helpers.py
```diff
@@ -0,0 +1,35 @@
...
+
+from lightly_studio.core import labelformat_helpers
+from tests import helpers_resolvers
+from tests.helpers_labelformat import get_labelformat_input_obj_det
+
+
+def test_create_label_map(db_session: Session) -> None:
+    # Test the creation of new labels and re-use of existing labels
+    collection_id = helpers_resolvers.create_collection(session=db_session).collection_id
+    label_input = get_labelformat_input_obj_det(filename="image.jpg", category_names=["dog", "cat"])
```
michal-lightly: Minor: It is cleaner to define a custom label input in the test. With the protocol change, you can literally do this, even inside the function:
```py
    class TestLabels1:
        def get_categories(self) -> Iterable[Category]:
            return [Category(...), ...]
```

lightly_studio/tests/helpers_labelformat.py
```diff
@@ -0,0 +1,43 @@
+from __future__ import annotations
```
michal-lightly: Do we need this file?

lightly_studio/src/lightly_studio/core/labelformat_helpers.py
michal-lightly: Update: This also simplifies testing, we don't need the helper.

lightly_studio/tests/core/test_labelformat_helpers.py
```diff
@@ -0,0 +1,35 @@
...
+        input_labels=label_input_2,
+    )
+
+    assert len(label_map_1) == 2  # dog and cat
+    assert len(label_map_2) == 3  # dog, cat and bird
+
+    # Compare label IDs for:
+    assert label_map_2[0] == label_map_1[0]  # dog exists already
+    assert label_map_2[1] == label_map_1[1]  # cat exists already
+    assert label_map_2[2] not in label_map_1.values()  # bird is new
```
michal-lightly, line 166: You can test equality for categories because they are frozen dataclasses. So e.g.:

```py
assert label_map_1 == {1: Category(id=1, name="dog"), 2: Category(id=2, name="cat")}
```

lightly_studio/tests/core/test_labelformat_helpers.py
```diff
@@ -0,0 +1,35 @@
...
+    collection_id = helpers_resolvers.create_collection(session=db_session).collection_id
+    label_input = get_labelformat_input_obj_det(filename="image.jpg", category_names=["dog", "cat"])
+
+    label_map_1 = labelformat_helpers.create_label_map(
+        session=db_session,
+        dataset_id=collection_id,
+        input_labels=label_input,
+    )
+
+    label_input_2 = get_labelformat_input_obj_det(
```
michal-lightly: Optional: I would drop the second test - it does not test anything new.

JonasWurst: It does test something: this will ensure existing classes are reused and we are not defining duplicated classes

lightly_studio/tests/core/test_labelformat_helpers.py
JonasWurst: The test here is different. It verifies that the db UUIDs are actually the same --> no duplicates created 
Will keep as is

lightly_studio/tests/helpers_labelformat.py
JonasWurst: not anymore, I removed it.

lightly_studio/tests/core/test_labelformat_helpers.py
michal-lightly: Ah ok, I see.

We could also do
```py
dog_label = annotation_label_resolver.get_by_label_name(..., label_name="dog")`
...
assert label_map_1 = {0: dog_label.annotation_label_id, ...}
```
To be more explicit. But consider this as optional.

lightly_studio/tests/core/test_labelformat_helpers.py
```diff
@@ -0,0 +1,166 @@
...
+from labelformat.model.category import Category
+from labelformat.model.multipolygon import MultiPolygon
+from sqlmodel import Session
+
+from lightly_studio.core import labelformat_helpers
+from lightly_studio.models.annotation.annotation_base import AnnotationType
+from tests import helpers_resolvers
+
+
+def test_get_segmentation_annotation_create_with_multipolygon() -> None:
```
michal-lightly: Super minor: Style
```suggestion
def test_get_segmentation_annotation_create__with_multipolygon() -> None:
```

lightly_studio/tests/core/test_labelformat_helpers.py
```diff
@@ -0,0 +1,166 @@
...
+    assert annotation.annotation_label_id == annotation_label_id
+    assert annotation.annotation_type == AnnotationType.INSTANCE_SEGMENTATION
+    assert annotation.x == 10
+    assert annotation.y == 20
+    assert annotation.width == 20
+    assert annotation.height == 20
+    assert annotation.segmentation_mask is None  # MultiPolygon doesn't have RLE
+
+
+def test_get_segmentation_annotation_create_with_binary_mask() -> None:
```
michal-lightly:
```suggestion
def test_get_segmentation_annotation_create__with_binary_mask() -> None:
```

lightly_studio/tests/core/test_labelformat_helpers.py
```diff
@@ -0,0 +1,166 @@
...
+    """Test creating segmentation annotation from BinaryMaskSegmentation."""
+    parent_sample_id = uuid4()
+    annotation_label_id = uuid4()
+
+    # Create a simple binary mask using RLE encoding
+    # RLE format: [start, length, start, length, ...]
+    # This represents a 2x2 mask in the top-left of a 4x4 image
+    binary_mask = BinaryMaskSegmentation(
```
michal-lightly: Minor: It is better to use the intended public interface - either `BinaryMaskSegmentation.from_rle` or `BinaryMaskSegmentation.from_binary_mask`

lightly_studio/tests/core/test_labelformat_helpers.py
```diff
@@ -0,0 +1,166 @@
...
+        _rle_row_wise=[0, 2, 4, 2],
```
michal-lightly: Nitpick: I'd expect it to be `0, 2, 2, 2, 10`. But why not pick a simpler example? E.g. 2x2 image with `[3, 1]` - mask for the bottom right pixel

lightly_studio/tests/core/test_labelformat_helpers.py
```diff
@@ -0,0 +1,166 @@
...
+    )
+
+    assert annotation.parent_sample_id == parent_sample_id
+    assert annotation.annotation_label_id == annotation_label_id
+    assert annotation.annotation_type == AnnotationType.SEMANTIC_SEGMENTATION
+    assert annotation.x == 0
+    assert annotation.y == 0
+    assert annotation.width == 2
+    assert annotation.height == 2
+    assert annotation.segmentation_mask is not None  # BinaryMaskSegmentation has RLE
```
michal-lightly: We know the rle, we can assert it (by writing it out).

lightly_studio/tests/core/test_labelformat_helpers.py
```diff
@@ -0,0 +1,166 @@
...
+    assert annotation.annotation_label_id == annotation_label_id
+    assert annotation.annotation_type == AnnotationType.OBJECT_DETECTION
+    assert annotation.x == 10  # Converted to int and XYWH format
+    assert annotation.y == 20
+    assert annotation.width == 20
+    assert annotation.height == 20
+    assert annotation.confidence is None
+
+
+def test_get_object_detection_annotation_create_with_confidence() -> None:
```
michal-lightly:
```suggestion
def test_get_object_detection_annotation_create__with_confidence() -> None:
```

lightly_studio/tests/core/test_labelformat_helpers.py
JonasWurst: I will keep it as is. Thanks for the good input :)

lightly_studio/tests/core/test_labelformat_helpers.py
JonasWurst: wiIll keep as is. will consider in future

PR #469, opened 2026-01-23
Download using temp files to prevent corrupted files

mihnea-necsulescu:
## What has changed and why?

Previously, when the model downloads failed for some reason, they would still be persisted while corrupted, which would cause issues. To fix those issues, you'd have to manually deleted the corrupted files so the models could be properly re-downloaded. 
The download logic is now changed to use temporary files: download to a temporary file and only move it to the right location if the download succeeded; if not, delete it.

## How has it been tested?

Unit tests + manually
To test manually:
1. Delete PE model from your temp directory (if any)
2. Run the video indexing script with embedding turned on
3. ctrl + c as the model is downloading
4. Check your temp directory: no model should be there
5. Run the indexing script again and allow the model to download
6. Check your temp directory: the model should be there
7. Run the indexing script again: the download should not be triggered now

- [ ] Yes

### DIFF
--- a/lightly_studio_view/src/lib/components/TagCreateDialog/TagCreateDialog.svelte
+++ b/lightly_studio_view/src/lib/components/TagCreateDialog/TagCreateDialog.svelte
@@ -114,7 +114,7 @@
         clearSelectedSampleAnnotationCrops,
         clearSelectedSamples
     } = useGlobalStorage();
-    const selectedSampleIds = getSelectedSampleIds(collectionId);
+    const selectedSampleIds = $derived(getSelectedSampleIds(collectionId));
     const clearItemsSelected = $derived(
         ['samples', 'videos', 'video_frames'].includes(gridType)
             ? () => clearSelectedSamples(collectionId)


lightly_studio/src/lightly_studio/dataset/file_utils.py
```diff
@@ -13,22 +15,37 @@
 
 
 def download_file_if_does_not_exist(url: str, local_filename: Path) -> None:
-    """Download a file from a URL if it does not already exist locally."""
+    """Download a file from a URL if it does not already exist locally.
+
+    downloads to temp file first, then moves to final location only on success.
+    This prevents corrupted partial downloads from persisting.
```
michal-lightly:
```suggestion
    Downloads to a temp file first, then moves to the final location only on success.
    This prevents corrupted partial downloads.
```

lightly_studio/src/lightly_studio/dataset/file_utils.py
```diff
@@ -13,22 +15,37 @@
...
-            # Raise an error for bad status codes
-            r.raise_for_status()
-            with open(local_filename, "wb") as f:
-                shutil.copyfileobj(r.raw, f)
-    except Exception:
-        # If download fails, remove any partial file to allow retry.
-        if local_filename.exists():
-            local_filename.unlink()
-        raise
+        with tempfile.NamedTemporaryFile(dir=dir_name, delete=False, mode="wb") as tmp_file:
```
michal-lightly, line 35: Minor: Is there a reason to create the temp file in `dir_name`? We don't care where the temp download is located.

lightly_studio/src/lightly_studio/dataset/file_utils.py
```diff
@@ -13,22 +15,37 @@
...
+            tmp_path = tmp_file.name
+            logger.info(f"Downloading {url} to {local_filename}")
+            with requests.get(url, stream=True, timeout=30) as r:
+                r.raise_for_status()
+                shutil.copyfileobj(r.raw, tmp_file)
+        os.replace(tmp_path, local_filename)
+    finally:
+        # Clean up temp file if it still exists (download failed or move failed)
+        if tmp_path and os.path.exists(tmp_path):
```
michal-lightly: Minor: It is better to be explicit in `falsey` conditions:
```py
if tmp_path is not None and os.path.exists(tmp_path):
```

lightly_studio/src/lightly_studio/dataset/file_utils.py
```diff
@@ -13,22 +15,37 @@
...
-        if local_filename.exists():
-            local_filename.unlink()
-        raise
+        with tempfile.NamedTemporaryFile(dir=dir_name, delete=False, mode="wb") as tmp_file:
+            tmp_path = tmp_file.name
+            logger.info(f"Downloading {url} to {local_filename}")
+            with requests.get(url, stream=True, timeout=30) as r:
+                r.raise_for_status()
+                shutil.copyfileobj(r.raw, tmp_file)
+        os.replace(tmp_path, local_filename)
```
michal-lightly, line 41: Shouldn't this be inside the with statement?

lightly_studio/src/lightly_studio/dataset/file_utils.py
michal-lightly: Minor: It is better to be explicit in `falsey` conditions:
```py
if tmp_path is not None and os.path.exists(tmp_path):
```

lightly_studio/src/lightly_studio/dataset/file_utils.py
```diff
@@ -13,22 +15,37 @@
...
+    finally:
+        # Clean up temp file if it still exists (download failed or move failed)
+        if tmp_path and os.path.exists(tmp_path):
+            try:
+                os.remove(tmp_path)
+            except OSError as e:
+                logger.warning(f"Failed to clean up temp file {tmp_path}: {e}")
```
michal-lightly, line 48: Let's also raise if the dest file does not exist

lightly_studio/tests/dataset/test_file_utils.py
```diff
@@ -0,0 +1,133 @@
...
+
+
+def _mock_response(mocker: MockerFixture) -> MagicMock:
+    """Helper to create a mock requests response."""
+    mock_resp = MagicMock()
+    mock_resp.raise_for_status = MagicMock()
+    mock_resp.raw = io.BytesIO(b"test")
+    mock_resp.__enter__ = MagicMock(return_value=mock_resp)
+    mock_resp.__exit__ = MagicMock(return_value=False)
+    mocker.patch("lightly_studio.dataset.file_utils.requests.get", return_value=mock_resp)
```
michal-lightly: It is preferable to use `object.patch`. It needs an import but only the end function name is a string.
```suggestion
    mocker.object.patch(requests, "get", return_value=mock_resp)
```

michal-lightly: Everywhere in this file. I did not have a closer look at the rest of the tests yet.

lightly_studio/src/lightly_studio/dataset/file_utils.py
mihnea-necsulescu: without it, dir would default to /tmp, and if that is mounted on a different filesystem, os.replace would crash

this shouldn't currently be a problem, since `local_filename` would be in the temp directory for all usages of this function, but I'd still like to keep it explicit; it may be used for other purposes in the future

lightly_studio/src/lightly_studio/dataset/file_utils.py
mihnea-necsulescu: the file will have to be closed before moving it - this works on Mac, but apparently it would cause issues on Windows; also found this: https://stackoverflow.com/questions/50569253/trying-to-replace-original-file-with-created-temp-file

lightly_studio/src/lightly_studio/dataset/file_utils.py
mihnea-necsulescu: if it does not exist, os.replace will raise an exception that will be propagated - that should be sufficient, no?

lightly_studio/src/lightly_studio/dataset/file_utils.py
michal-lightly: Still the `with` statements is suspicious. What gets cleaned-up when the statement is exited?

It seems to me that either:
* `os.replace` should be inside because the temp file might not exist
* or `NamedTemporaryFile` just generates a filename - then it does not need to be in a with statement

lightly_studio/src/lightly_studio/dataset/file_utils.py
michal-lightly: That seems sufficient.

lightly_studio/src/lightly_studio/dataset/file_utils.py
mihnea-necsulescu: > What gets cleaned-up when the statement is exited?

the file descriptor for the temporary file; NamedTemporaryFile is just a higher-level interface that still uses `mkstemp()` under the hood - it would be equivalent to
```python
fd, filename = tempfile.mkstemp()
os.close(fd)
```

michal-lightly: Makes sense, thanks for explaining!

PR #470, opened 2026-01-28
Add Sample.add_annotation() method

lukas-lightly:
## What has changed and why?
Supports only classification annotation right now, the rest can come later.

## How has it been tested?
Added a test

- [x] Yes


lightly_studio/src/lightly_studio/models/annotation/annotation_base.py
```diff
@@ -93,14 +94,16 @@ class AnnotationBaseTable(SQLModel, table=True):
...
     """Input model for creating annotations."""
 
     """ Required properties for all annotations. """
     annotation_label_id: UUID
     annotation_type: AnnotationType
     confidence: Optional[float] = None
-    parent_sample_id: UUID
+    # Note: If initially set to None, it has to be not None when passed to
+    # `annotation_resolver.create_many()`.
+    parent_sample_id: Optional[UUID] = None
```
lukas-lightly: This part is a bit ugly but I don't have another idea besides creating a duplicate type `T` with `parent_sample_id` missing. `Sample.add_annotation` would then accept `T` and convert `T` to `AnnotationCreate` by adding the `parent_sample_id` field.

This seems like a lot of work. We would encode everything in types and it would be cleaner in a sense, but I don't think it's worth it.

michal-lightly: I would suggest do it, it is not a lot of work, no? We would make `AnnotationCreate` inherit from `T`. The concern is that AnnotationCreate is used on the interface between FE and BE, having a different class for python interface is ok.

We just need a good name for T (and potentially rename AnnotationCreate).

We do something similar btw. in the create_many.py files for image and video - those classes are internal though.

WDYT?

lukas-lightly: I've added `AnnotationCreateWithParent` which inherits from `AnnotationCreate` and I remove `parent_sample_id` from `AnnotationCreate`. PTAL.

There is now only the `ClassificationCreate(AnnotationCreate)` type which is very simple. Things get only interesting with ojbect detection annotation, which I'll add in the followup PR.

lightly_studio/tests/core/test_image_sample.py
```diff
@@ -423,3 +423,35 @@ def test_annotations_semantic_segmentation(
         assert annotations[0].label == "road"
         assert annotations[0].confidence == pytest.approx(0.8)
         assert annotations[0].segmentation_mask == [0, 0, 1, 1, 0, 0]
+
+    def test_add_annotation_classification(
+        self,
+        test_db: Session,
+    ) -> None:
+        from lightly_studio.models.annotation.annotation_base import ClassificationCreate
```
michal-lightly: Import should be at the file top

lukas-lightly: Done

PR #473, opened 2026-01-28
Add *Create types for annotations

lukas-lightly:
## What has changed and why?
We've added ClassificationCreate previously, the remaining ones where: object detection, instance segmentation and semantic segmentation.

## How has it been tested?
Added tests.

- [ ] Yes

### DIFF
--- a/lightly_studio/src/lightly_studio/core/annotation/annotation_create.py
+++ b/lightly_studio/src/lightly_studio/core/annotation/annotation_create.py
@@ -10,7 +10,7 @@ from pydantic import BaseModel
 from lightly_studio.models.annotation.annotation_base import AnnotationCreate, AnnotationType
 
 
-class ToAnnotationCreateModel(Protocol):
+class CreateAnnotation(Protocol):
     """Protocol from converting to AnnotationCreate."""
 
     def to_annotation_create(self, parent_sample_id: UUID) -> AnnotationCreate:
@@ -18,14 +18,14 @@ class ToAnnotationCreateModel(Protocol):
         ...
 
 
-class AnnotationCreateBase(BaseModel):
+class CreateAnnotationBase(BaseModel):
     """Base model for creating annotations."""
 
     annotation_label_id: UUID
     confidence: float | None = None
 
 
-class ClassificationCreate(AnnotationCreateBase):
+class CreateClassification(CreateAnnotationBase):
     """Input model for creating classification annotations."""
 
     def to_annotation_create(self, parent_sample_id: UUID) -> AnnotationCreate:
@@ -38,7 +38,7 @@ class ClassificationCreate(AnnotationCreateBase):
         )
 
 
-class ObjectDetectionCreate(AnnotationCreateBase):
+class CreateObjectDetection(CreateAnnotationBase):
     """Input model for creating object detection annotations."""
 
     x: int
@@ -60,7 +60,7 @@ class ObjectDetectionCreate(AnnotationCreateBase):
         )
 
 
-class InstanceSegmentationCreate(AnnotationCreateBase):
+class CreateInstanceSegmentation(CreateAnnotationBase):
     """Input model for creating instance segmentation annotations."""
 
     x: int
@@ -84,7 +84,7 @@ class InstanceSegmentationCreate(AnnotationCreateBase):
         )
 
 
-class SemanticSegmentationCreate(AnnotationCreateBase):
+class CreateSemanticSegmentation(CreateAnnotationBase):
     """Input model for creating semantic segmentation annotations."""
 
     x: int
--- a/lightly_studio/src/lightly_studio/core/annotation/__init__.py
+++ b/lightly_studio/src/lightly_studio/core/annotation/__init__.py
@@ -1,17 +1,17 @@
 from .annotation import Annotation
 from .annotation_create import (
-    ClassificationCreate,
-    InstanceSegmentationCreate,
-    ObjectDetectionCreate,
-    SemanticSegmentationCreate,
-    ToAnnotationCreateModel,
+    CreateAnnotation,
+    CreateClassification,
+    CreateInstanceSegmentation,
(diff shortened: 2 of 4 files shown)


lightly_studio/src/lightly_studio/models/annotation/annotation_base.py
```diff
@@ -124,6 +124,38 @@ class ClassificationCreate(AnnotationCreate):
...
+
+class InstanceSegmentationCreate(AnnotationCreate):
+    """Input model for creating instance segmentation annotations."""
+
+    annotation_type: AnnotationType = AnnotationType.INSTANCE_SEGMENTATION
+    x: int
+    y: int
+    width: int
+    height: int
+    segmentation_mask: List[int]
```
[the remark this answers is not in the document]
lukas-lightly: Done

lightly_studio/src/lightly_studio/models/annotation/annotation_base.py
```diff
@@ -124,6 +124,38 @@ class ClassificationCreate(AnnotationCreate):
...
+
+class SemanticSegmentationCreate(AnnotationCreate):
+    """Input model for creating semantic segmentation annotations."""
+
+    annotation_type: AnnotationType = AnnotationType.SEMANTIC_SEGMENTATION
+    x: int
+    y: int
+    width: int
+    height: int
+    segmentation_mask: List[int]
```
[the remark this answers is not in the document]
lukas-lightly: Done

lightly_studio/src/lightly_studio/models/annotation/annotation_base.py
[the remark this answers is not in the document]
lukas-lightly: Actually reverted the fix, as it breaks SQLAlchemy. The file is not using `from __future__ import annotations` for a reason.

lightly_studio/src/lightly_studio/models/annotation/annotation_base.py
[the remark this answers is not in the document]
lukas-lightly: Same as above

lightly_studio/src/lightly_studio/models/annotation/annotation_base.py
```diff
@@ -124,6 +124,38 @@ class ClassificationCreate(AnnotationCreate):
...
 
 
+class ObjectDetectionCreate(AnnotationCreate):
+    """Input model for creating object detection annotations."""
+
+    annotation_type: AnnotationType = AnnotationType.OBJECT_DETECTION
+    x: int
+    y: int
+    width: int
+    height: int
```
michal-lightly: We are breaking the Liskov substitution principle here: `ann.x = None` is allowed for `AnnotationCreate` instances, but not for `ObjectDetectionCreate` instances.

I am actually surprised that the syntax is allowed (to re-define parent attributes with a different type) - but it also makes sense since python type-checking is "an addon".

There is a clean solution:
* [Optional] Move the `ClassificationCreate` etc. classes to the `lightly_studio.core` module
* Create a `ToAnnotationCreate` protocol (name tbd)
* Implement `to_annotation_create(self) -> AnnotationCreate` for all of them

michal-lightly: Note that it should be expressible in a protocol that class attributes are mandatory (confidence, annotation_type, annotation_label_id) - I am not 100% sure though. If it is not the case than we might need an ABC or combine ABC with a protocol, or redefine it for children (open to suggestions).

lukas-lightly: Nice catch, it's weird this is not caught by mypy. I'll rework it.

Once we do this, we can drop `AnnotationCreateWithParent` (I think), as it's only needed, because we wanted to inherit from `AnnotationCreate` here but this will be dropped.

lukas-lightly: I've refactored it. As mentioned, I dropped `AnnotationCreateWithParent` and the `*Create` classes have been moved under `lightly_studio.core`.

lightly_studio/src/lightly_studio/core/annotation/annotation_create.py
```diff
@@ -0,0 +1,108 @@
...
+
+from typing import Protocol
+from uuid import UUID
+
+from pydantic import BaseModel
+
+from lightly_studio.models.annotation.annotation_base import AnnotationCreate, AnnotationType
+
+
+class ToAnnotationCreateModel(Protocol):
```
michal-lightly: Optional: We could call it `ToAnnotationCreate` or `ToAnnotationCreateProtocol`.

Honestly, it is unusual to call protocols with -Protocol suffix in python, usually they are -able etc. However that makes protocol naming very difficult. I would not mind going against the convention in this case.

lightly_studio/src/lightly_studio/core/annotation/annotation_create.py
```diff
@@ -0,0 +1,108 @@
...
+
+
+class AnnotationCreateBase(BaseModel):
+    """Base model for creating annotations."""
+
+    annotation_label_id: UUID
+    confidence: float | None = None
+
+
+class ClassificationCreate(AnnotationCreateBase):
```
michal-lightly: Maybe we should call this `CreateClassification` to make it aligned with the group creation design. I'll reach out in person to discuss.

lightly_studio/src/lightly_studio/core/annotation/annotation_create.py
lukas-lightly: Done, used `CreateAnnotation`.

lightly_studio/src/lightly_studio/core/annotation/annotation_create.py
lukas-lightly: Done

PR #476, opened 2026-01-29
Add AnnotationLabel class

lukas-lightly:
## What has changed and why?
We need to specify a label when adding an annotation, this class will be referenced when calling `Sample.add_annotation()`.

## How has it been tested?
Added a test

- [x] Yes

### DIFF
--- a/lightly_studio/tests/core/test_dataset.py
+++ b/lightly_studio/tests/core/test_dataset.py
@@ -437,7 +437,6 @@ class TestDataset:
         dataset = ImageDataset.create(name="test_dataset")
         label = dataset.add_annotation_label("cat")
         assert label.name == "cat"
-        assert isinstance(label.id, UUID)
 
         # Adding same label again should return the same label
         label2 = dataset.add_annotation_label("cat")


lightly_studio/src/lightly_studio/core/dataset.py
```diff
@@ -198,6 +201,27 @@ def compute_similarity_metadata(
             metadata_name=metadata_name,
         )
 
+    def add_annotation_label(self, name: str) -> AnnotationLabel:
```
michal-lightly, line 204: Optional: Could be `get_or_create_annotation_label`. Not feeling strongly about it.

michal-lightly: Side note: For the future, will we also need `get_annotation_labels` at some point?

I had a discussion with Guarin who tested LightlyStudio for the hackathon. We will likely need to make labels tied to annotation tasks instead of datasets.

lightly_studio/tests/core/test_dataset.py
```diff
@@ -430,6 +430,25 @@ def test_compute_similarity_metadata(
...
         assert enriched_samples[2].metadata[metadata_name] == 1.0
 
+    def test_add_annotation_label(
+        self,
+        patch_collection: None,  # noqa: ARG002
+    ) -> None:
+        dataset = ImageDataset.create(name="test_dataset")
+        label = dataset.add_annotation_label("cat")
+        assert label.name == "cat"
+        assert isinstance(label.id, UUID)
```
michal-lightly: Let's remove, we don't need to test what is checked by mypy

```suggestion
```

PR #485, opened 2026-02-01
Igor lig 8454 ghosting effect after stopping brush painting

IgorSusmelj:
## What has changed and why?

- before this PR we had a few bugs when drawing segmentation masks or erasing them
  - we did only paint locations when there was a mouse move event. If the mouse is moved fast this could result in a "jittered" drawing rather than nice brush strokes --> has been fixed by interpolating between the mouse move locations (as it's commonly done in other painting apps)
  - if the user would draw and then release the mouse button while still moving the mouse around there could have been a weird ghosting effect which could confuse the user as it's not clear if we're still drawing or not --> the bug was that we only checked for the mouse up event on the current annotation but if the user would move the mouse very quickly it could be that we're already on another object. We now propagate the mouse up event properly.
  - we did fix both bugs for the brush as well as for the eraser

Note that the mouse pointer in the video is delayed and smoothed from the recording software

https://github.com/user-attachments/assets/315d4c6b-743e-40ed-9948-d62f99915bde

## How has it been tested?

- manual testing

- [X] Yes


lightly_studio_view/src/lib/components/SampleAnnotation/utils.ts
```diff
@@ -178,3 +178,36 @@ export const decodeRLEToBinaryMask = (rle: number[], width: number, height: numb
 
     return mask;
 };
+
+/**
+ * Interpolate points along a line between two points.
+ * Uses linear interpolation to ensure continuous brush strokes
+ * even when mouse events are spaced far apart.
+ */
+export const interpolateLineBetweenPoints = (
```
ikondrat, line 187: we need test for this function as well

lightly_studio_view/src/lib/components/SampleDetails/SampleEraserRect/SampleEraserRect.svelte
```diff
@@ -190,7 +191,10 @@
         const point = getImageCoordsFromMouse(e, interactionRect, sample.width, sample.height);
         if (!point) return;
 
+        (e.currentTarget as Element).setPointerCapture(e.pointerId);
```
IgorSusmelj: I think we can leave those out. It was proposed by AI to make it possible to leave the canvas and go in again when drawing. I don't see this as necessary though. And a more elegant solution might be to capture the mouse events on the parent element.

ikondrat (approved):
I would just add test for interpolation, otherwise looks good to me

PR #489, opened 2026-02-02
Update docs for annotations

lukas-lightly:
## What has changed and why?
In order to make the API easier to use, I also:
* Put all relevant classes under `lightly_studio.core.annotation`. It makes it easier to import.
* Changed `annotation_label_name` to just `label` for all `Create*` classes. The former is an internal field name.
* Added docstrings to `Create*` classes.

## How has it been tested?
Rendered docs manually.

- [ ] Yes


lightly_studio/docs/docs/index.md
```diff
@@ -498,7 +532,7 @@ for pred in predictions:
     # Add the prediction with confidence
     annotation_resolver.create_many(
         session=dataset.session,
-        collection_id=dataset.dataset_id,
+        parent_collection_id=dataset.dataset_id,
```
lukas-lightly, line 555: The fixes here and below are by AI, it noticed that we didn't fix the code examples after renaming things in the API.

lightly_studio/docs/docs/index.md
```diff
@@ -459,6 +459,40 @@ for sample in dataset:
...
+from lightly_studio.core.annotation import CreateSemanticSegmentation
+
+sample.add_annotation(CreateSemanticSegmentation(
+    label="car",
+    confidence=0.85,
+    x=0,
+    y=0,
+    width=100,
+    height=100,
+    segmentation_mask=[0, 1, 1, 0, 0, 1],
```
michal-lightly: Segmentation mask is in a run-length encoding:
- sum of the numbers must be the number of pixels in the image
- no elements except the first number can be zeros
- the bounding box should match the segmentation
    - Here we should make the interface more user friendly and compute the bbox for the user. There already is a helper in labelformat.

lightly_studio/src/lightly_studio/core/annotation/annotation_create.py
```diff
@@ -62,9 +64,13 @@ class CreateObjectDetection(CreateAnnotationBase):
     """Input model for creating object detection annotations."""
 
     x: int
+    """X coordinate (px) of the object detection bounding box."""
```
michal-lightly: It is important to mention that (x, y) are the top left corner coordinates. We should mention "left" in this comment.

lightly_studio/docs/docs/index.md
michal-lightly: We should also explain the encoding in the text, or at least link to a resource. (I think we have to describe it since there are some choices - it is row-wise, starting with the count of zeros (ones mark the object). There is some info here: https://github.com/lightly-ai/labelformat/blob/main/src/labelformat/model/binary_mask_segmentation.py

lightly_studio/src/lightly_studio/core/annotation/annotation_create.py
lukas-lightly: Done

lightly_studio/docs/docs/index.md
lukas-lightly: I've fixed the mask and the dimensions, however I haven't computed the bbox for the user. I also haven't explained the encoding yet.

lukas-lightly: I've fixed the bounding box example, so that it is consistent. I've also added an explanation of the binary mask encoding. I didn't use the helper function yet, I need to first write code for it.

lukas-lightly:
Current segmentation example with Binary Mask encoding:
image

lightly_studio/docs/docs/index.md
```diff
@@ -459,6 +459,60 @@ for sample in dataset:
...
+
+You can add annotations to samples using the `add_annotation` method, the following example shows how to create an object detection annotation.
+
+```python
+from lightly_studio.core.annotation import CreateObjectDetection
+
+# Add an object detection annotation to a sample
+sample.add_annotation(CreateObjectDetection(
+    label="car",
+    confidence=0.9,
```
michal-lightly: Minor:
```suggestion
    confidence=0.9,  # optional
```

lightly_studio/docs/docs/index.md
```diff
@@ -459,6 +459,60 @@ for sample in dataset:
...
+
+For example, consider a 2x4 mask:
+```
+[[0, 1, 1, 0],
+ [1, 1, 1, 1]]
+```
+Flattened row-wise: `[0, 1, 1, 0, 1, 1, 1, 1]`.
+
+There are 4 sequences of identical bits: one 0, two 1s, one 0 and four 1s. The resulting `segmentation_mask` is `[1, 2, 1, 4]`.
+
```
michal-lightly, line 515: Very nicely explained!

lukas-lightly: AI might have been involved for parts of it :)

PR #491, opened 2026-02-02
Update e2e video frame caption test

horatiualmasan:
## What has changed and why?
Wait for API responses and change the caption text change effect.

## How has it been tested?

Manual test locally
image

Repeated CI e2e tests

- [ ] Yes

### DIFF
--- a/lightly_studio_view/e2e/caption-utils.ts
+++ b/lightly_studio_view/e2e/caption-utils.ts
@@ -36,6 +36,13 @@ export class CaptionUtils {
         await expect(this.page.getByTestId('caption-field')).toHaveCount(captionCountBefore + 1);
     }
 
+    /**
+     * Add a caption on the captions page (grid of caption items).
+     * Unlike addCaption, this waits for the POST /api/samples/list response after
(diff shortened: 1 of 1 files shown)


lightly_studio_view/e2e/caption-utils.ts
```diff
@@ -36,6 +36,22 @@ export class CaptionUtils {
         await expect(this.page.getByTestId('caption-field')).toHaveCount(captionCountBefore + 1);
     }
 
+    async addCaptionInCaptionPage(addButtonIndex: number = 0) {
```
michal-lightly, line 46: What is the difference between `addCaption` and `addCaptionInCaptionPage`? Do we need both?

horatiualmasan: When adding a caption in the caption page the sample list is retrieved again from the backend and will update the grid. In that case we should also wait for the  samples/list api response. When adding from the details panel it is not needed.

michal-lightly: Let's add that to a docstring please

PR #495, opened 2026-02-03
Add delete dataset resolver

mihnea-necsulescu:
## What has changed and why?

Added new delete dataset resolver.

## How has it been tested?

Manual + new tests

- [ ] Yes


lightly_studio/src/lightly_studio/resolvers/collection_resolver/delete_dataset.py
```diff
@@ -0,0 +1,288 @@
...
+
+    Args:
+        session: Database session.
+        root_collection_id: Root collection ID to delete.
+
+    Raises:
+        ValueError: If the collection is not a root collection.
+    """
+    # Ensure all tables are handled - fails if new tables were added without updating this function.
+    table_coverage_utils.verify_table_coverage()
```
lukas-lightly, line 49: This this mean that if I add a field and a table to the dataset and forget to update this function, we'll get a failed test?

mihnea-necsulescu: only if the table count changes; adding/removing fields should not cause any issues

PR #503, opened 2026-02-05
Decouple DuckDB hardcoding

mihnea-necsulescu:
## What has changed and why?

Set the foundation for adding support for Postgres.

## How has it been tested?

Manually (everything should run normally) + new tests

- [ ] Yes


lightly_studio/src/lightly_studio/db_manager.py
```diff
@@ -33,12 +53,23 @@ def __init__(
...
+            self._engine_url = engine_url
+        elif LIGHTLY_STUDIO_DATABASE_URL is not None:
+            self._engine_url = LIGHTLY_STUDIO_DATABASE_URL
+        else:
+            self._engine_url = "duckdb:///lightly_studio.db"
+
+        self._backend = _detect_backend_from_url(self._engine_url)
+
+        # Only cleanup for DuckDB.
+        if cleanup_existing and self._backend == DatabaseBackend.DUCKDB:
```
michal-lightly, line 72: Please add a todo and create an issue for clearing the postgres db too. I'm sure there is an easy way to do that.

lightly_studio/src/lightly_studio/db_manager.py
```diff
@@ -130,17 +169,30 @@ def set_engine(engine: DatabaseEngine) -> None:
...
+    Helper function to set up the database engine. Supports both DuckDB (default)
+    and PostgreSQL via the engine_url parameter or LIGHTLY_STUDIO_DATABASE_URL env var.
 
     Args:
-        db_file: The path to the DuckDB file. If None, uses a default, see DatabaseEngine class.
-        cleanup_existing: If True, deletes the pre-existing database file if a file database
-            is used.
+        db_file: Path to DuckDB file. Ignored if engine_url is provided.
+        cleanup_existing: If True, deletes existing database file (DuckDB only).
+        engine_url: Full database URL. Takes precedence over db_file.
```
michal-lightly: Minor: This is not true, only one of the two can be provided. Precedence does not make sense.

lightly_studio/src/lightly_studio/db_manager.py
mihnea-necsulescu: oh I thought we only wanted this for the OSS version - updated and added an issue

michal-lightly: We can let OSS users use postgres too, if they want to set it up.

lightly_studio/src/lightly_studio/db_manager.py
michal-lightly: Two lines above too, please :)

PR #506, opened 2026-02-05
Allow embedding search for deep copied datasets

mihnea-necsulescu:
## What has changed and why?

Embedding text search was not working for deep copied datasets because the embedding model was not copied. It now gets properly deep copied, and the copied sample embeddings correctly references the copy. Some of the embedding model resolver's functions have also been updated to allow multiple embedding models in the DB with the same hash.

## How has it been tested?

Manually + new tests.
To test locally:
1. Deep copy a dataset (POST http://localhost:8001/api/collections/{dataset_id}/deep-copy
2. Navigate to the copied dataset - text search should be available 
3. (optional) delete the original dataset - this was failing before due to a FK error (DELETE http://localhost:8001/api/collections/{dataset_id}/delete-dataset)

- [ ] Yes

### DIFF
--- a/lightly_studio/docs/docs/api/sample.md
+++ b/lightly_studio/docs/docs/api/sample.md
@@ -8,12 +8,12 @@
 
 ## ImageSample
 
-::: lightly_studio.core.image_sample
+::: lightly_studio.core.image.image_sample
     options:
         members: [ImageSample]
 
 ## VideoSample
 
-::: lightly_studio.core.video_sample
+::: lightly_studio.core.video.video_sample
     options:
         members: [VideoSample]


lightly_studio/src/lightly_studio/resolvers/embedding_model_resolver.py
```diff
@@ -60,13 +62,18 @@ def get_by_id(session: Session, embedding_model_id: UUID) -> EmbeddingModelTable
...
+    embedding_model_hash: str,
+    collection_id: UUID | None = None,
+) -> EmbeddingModelTable | None:
+    """Retrieve a single embedding model by hash, optionally filtered by collection."""
+    query = select(EmbeddingModelTable).where(
+        EmbeddingModelTable.embedding_model_hash == embedding_model_hash
+    )
+    if collection_id is not None:
+        query = query.where(EmbeddingModelTable.collection_id == collection_id)
+    return session.exec(query).first()
```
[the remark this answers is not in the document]
mihnea-necsulescu: classifier flows will not work for deep copied datasets, but that will be addressed in a separate PR, as the fix is non-trivial and out of scope for the current fix of enabling embedding search for copies

lightly_studio/src/lightly_studio/resolvers/collection_resolver/deep_copy.py
```diff
@@ -357,8 +386,12 @@ def _copy_annotations(
         )
         session.add(new_ann)
 
-        # Copy annotation-type-specific details.
-        _copy_annotation_details(session, old_ann.sample_id, new_sample_id, old_ann.annotation_type)
+        # Copy annotation-type-specific details only for non-classification annotations,
+        # which have no details table.
```
michal-lightly: Why was this changed?

mihnea-necsulescu: `_copy_annotation_details()` was previously called for classification annotations, too, which don't have a details table, and this resulted in a ValueError (https://github.com/lightly-ai/lightly-studio/blob/main/lightly_studio/src/lightly_studio/resolvers/collection_resolver/deep_copy.py#L390-L391)

I could have just removed that `else` clause, but I'd rather keep it explicit

also added a test covering this

michal-lightly: Minor: I think it is cleaner to move the `if` inside `_copy_annotation_details`. The function should be responsible for knowing that for classifications there are no details to copy.

michal-lightly: I.e. let's keep the ValueError, but add
```py
elif AnnotationType.CLASSIFICATION:
    pass
```

lightly_studio/tests/resolvers/collection_resolver/test_deep_copy.py
```diff
@@ -314,6 +332,61 @@ def test_deep_copy__with_embeddings(test_db: Session) -> None:
...
 
 
+def test_deep_copy__can_delete_original_after_copy(test_db: Session) -> None:
+    """Verify deleting original collection after deep copy doesn't cause FK errors."""
+    # Arrange
+    original = create_collection(session=test_db, collection_name="original")
+    original_collection_id = original.collection_id
+    img = create_image(
+        session=test_db, collection_id=original.collection_id, file_path_abs="/a.png"
+    )
```
michal-lightly, line 342: Optional: Maybe also let's add an annotation to this test?

PR #507, opened 2026-02-05
feat: potential fix for caching issue

ikondrat:
## What has changed and why?

This is potential fix for caching issue with app. 
https://linear.app/lightly/issue/LIG-8419/ux-caching-issue-figure-out

The problem is I reproduced it but don't know how to test it properly + I am not fully convinced what it is actualy caused by cached version.json

## How has it been tested?

What you can check what after runing test version of app locally
with
```
make -C lightly_studio start-e2e
```
You have non-caching headers

Screenshot 2026-02-06 at 12 27 01

- [ ] Yes


lightly_studio/src/lightly_studio/api/routes/webapp.py
```diff
@@ -45,6 +45,18 @@ async def serve_static_webapp_files_or_default_index_file(
                 status_code=HTTP_STATUS_NOT_FOUND,
                 detail=f"File '{path}' not found",
             )
+
+        # Special handling for version.json: never cache it
+        if path.endswith("version.json"):
+            return FileResponse(
+                file_path,
+                headers={
```
mihnea-necsulescu: maybe we should also add these headers to the index file?

mihnea-necsulescu (approved):

PR #510, opened 2026-02-06
Remove unused annotation tag methods

michal-lightly:
## What has changed and why?

Remove unused annotation tag methods.

Note that one more PR will be needed to finally remove the AnnotationTagLinkTable.

## How has it been tested?

Existing tests.

- [ ] Yes

### DIFF
--- a/ai_guidelines/python.md
+++ b/ai_guidelines/python.md
@@ -67,7 +67,7 @@ Use the following format: `# TODO({name}, {mm}/{yyyy}): Blah blah`
 
 ## Comments
 
-Prefer full sentences and punctuation. Final full stop can be omitted for a single sentence.
+Prefer properly formatted comments with a leading capital and punctuation. Final full stop may be omitted for a single sentence.
 
 ```python
 # This is a proper comment. It spans multiple sentences.


lightly_studio/src/lightly_studio/resolvers/collection_resolver/export.py
```diff
@@ -150,6 +150,20 @@ def _build_export_query(  # noqa: C901
...
+                .join(SampleTable, col(AnnotationBaseTable.sample_id) == col(SampleTable.sample_id))
+                .where(
+                    col(SampleTable.tags).any(
+                        and_(
+                            TagTable.kind == "annotation",
+                            col(TagTable.tag_id).in_(include.tag_ids),
+                        )
+                    )
+                )
+            )
```
michal-lightly, line 165: Needed to use a subquery to check annotation tags in export.

lightly_studio/src/lightly_studio/resolvers/collection_resolver/export.py
```diff
@@ -150,6 +150,20 @@ def _build_export_query(  # noqa: C901
     # include tags or sample_ids or annotation_ids from result
     if include:
         if include.tag_ids:
+            # Subquery: find parent_sample_ids that have annotations with matching tags
+            annotation_tag_subquery = (
+                select(AnnotationBaseTable.parent_sample_id)
```
lukas-lightly, line 155: The `annotation_tag_subquery` here and on line 222 seem to be identical. (I did an eye-test, so I might have missed some important difference).

Is it possible to have this as a constant or a static function?

michal-lightly: I was thinking the same. But there is a difference inside, a different variable is used.

It is possible to create a function, but I suggest to move forward with duplication. This is quite an old code and I suspect we will have to replace it completely when we rework the export functionality.

PR #512, opened 2026-02-09
Add docs example for from_binary_mask() method

lukas-lightly:
## What has changed and why?
The newly added methods simplify the usage and should be in the examples.

## How has it been tested?
Rendered and inspected manually.

- [ ] Yes

### DIFF
--- a/lightly_studio/docs/docs/index.md
+++ b/lightly_studio/docs/docs/index.md
@@ -477,7 +477,32 @@ sample.add_annotation(CreateObjectDetection(
 ))
 ```
 
-There are also `CreateClassification`, `CreateInstanceSegmentation`, and `CreateSemanticSegmentation` classes for other annotation types. For example, to add a semantic segmentation annotation:
+There are also `CreateClassification`, `CreateInstanceSegmentation`, and `CreateSemanticSegmentation` classes for other annotation types.
+
+For segmentation annotations, it is recommended to use the `from_binary_mask` method, which automatically handles the bounding box and mask encoding from a 2D numpy array:
+
+```python
+import numpy as np
+from lightly_studio.core.annotation import CreateSemanticSegmentation
+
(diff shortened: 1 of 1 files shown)


lightly_studio/docs/docs/index.md
```diff
@@ -477,7 +477,32 @@ sample.add_annotation(CreateObjectDetection(
...
+There are also `CreateClassification`, `CreateInstanceSegmentation`, and `CreateSemanticSegmentation` classes for other annotation types.
+
+For segmentation annotations, it is recommended to use the `from_binary_mask` method, which automatically handles the bounding box and mask encoding from a 2D numpy array:
+
+```python
+import numpy as np
+from lightly_studio.core.annotation import CreateSemanticSegmentation
+
+# A 2D numpy array representing the binary mask (1 for foreground, 0 for background)
+mask = np.array([
```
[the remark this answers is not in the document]
lukas-lightly: The rest of the code base never does this for `np.array()`. I will leave this as is.

lightly_studio/docs/docs/index.md
```diff
@@ -497,6 +522,9 @@ sample.add_annotation(CreateSemanticSegmentation(
 
 For segmentation annotations (`CreateSemanticSegmentation`, `CreateInstanceSegmentation`), the `segmentation_mask` is expected to be a list of integers representing the binary mask in a row-wise Run-Length Encoding (RLE) format.
```
michal-lightly, line 523: Do you plan to change `CreateSemanticSegmentation` to not accept x,y,w,h and rather compute it from the mask? I think that's the preferred solution, otherwise we should also document that x,y,w,h must match the binary mask.

lukas-lightly: I noticed some redundancy there but it didn't occur to me to fix the API. Let me do that in a followup.

PR #518, opened 2026-02-09
Make erase mode mask more transparent while drawing

LeonardoRosaa:
## What has changed and why?

Reduce mask opacity while the user is erasing so it’s easier to correct masks, while keeping the normal opacity when not erasing to ensure the selected annotation remains clearly highlighted.

## How has it been tested?

https://github.com/user-attachments/assets/2ee310ec-338b-4b23-9560-df740b2bc132

- [x] Yes

### DIFF
--- a/lightly_studio/docs/docs/api/sample.md
+++ b/lightly_studio/docs/docs/api/sample.md
@@ -8,12 +8,12 @@
 
 ## ImageSample
 
-::: lightly_studio.core.image_sample
+::: lightly_studio.core.image.image_sample
     options:
         members: [ImageSample]
 
 ## VideoSample
 
-::: lightly_studio.core.video_sample
+::: lightly_studio.core.video.video_sample
     options:
         members: [VideoSample]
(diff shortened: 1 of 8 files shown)


lightly_studio_view/src/lib/components/SampleDetails/SampleEraserRect/SampleEraserRect.svelte
```diff
@@ -175,7 +175,7 @@
         segmentation={previewRLE}
         width={sample.width}
         colorFill={withAlpha(drawerStrokeColor, 0)}
-        opacity={0.85}
+        opacity={0.35}
```
JonasWurst, line 178: Did you check it for various colors on various backgrounds if this is visible enough?

LeonardoRosaa: I tested with a few different colors, and it seems visible enough.

PR #520, opened 2026-02-09
Enable classifier operations on deep copied datasets

mihnea-necsulescu:
## What has changed and why?

With the addition of deep copied datasets, embedding models could not be fetched by model hash only, as original and duplicated datasets would have the same embedding model (with the same hash). Therefore, the 
`get_by_model_hash()` had to be updated to receive a collection id. In a first phase, this was done only to enable embedding text search for duplicated datasets. This PR also addressed classifier operations (which also use `get_by_model_hash()`), by passing the collection id throughout the entire classifier flows - without this, the wrong embedding model would be picked for duplicated datasets (i.e., the one for the original dataset); creating classifiers, for example, would fail on duplicated datasets with a "no samples found" error, because the wrong embeddings would end up being used

## How has it been tested?

Manually (all classifier flows should work on duplicated datasets too) + updated tests

- [ ] Yes


lightly_studio/src/lightly_studio/api/routes/api/classifier.py
```diff
@@ -347,14 +354,17 @@ class GetAllClassifiersResponse(BaseModel):
 
 
 @classifier_router.get("/classifiers/get_all_classifiers")
-def get_all_classifiers() -> GetAllClassifiersResponse:
-    """Get all active classifiers.
+def get_all_classifiers(collection_id: UUID) -> GetAllClassifiersResponse:
```
mihnea-necsulescu, line 357: this was updated because all classifiers were present on all datasets (e.g., classifiers created on a deep copy collection would also show up for original collections and vice versa) - that does not seem correct to me, but I'll revert if I'm mistaken

michal-lightly: Potentially, it might make sense to link a classifier to dataset_id + sample_type. But for now, let's keep collection_id, I don't think it is a feature used much anyway.

lightly_studio/src/lightly_studio/resolvers/embedding_model_resolver.py
```diff
@@ -65,17 +65,14 @@ def get_by_id(session: Session, embedding_model_id: UUID) -> EmbeddingModelTable
 def get_by_model_hash(
     session: Session,
     embedding_model_hash: str,
-    collection_id: UUID | None = None,
+    collection_id: UUID,
```
michal-lightly: Optional: Would you mind swapping the two parameters? session > collection > hash

michal-lightly (approved):

PR #521, opened 2026-02-09
Allow editing of newly created bounding boxes

LeonardoRosaa:
## What has changed and why?

Allow resizing and adjusting a bounding box immediately after it is drawn instead of starting a new bounding box. Users can start a new bounding box by clicking outside of the current one.

## How has it been tested?

https://github.com/user-attachments/assets/46a1af78-aa72-42eb-934c-f557ff0b9e4a

- [x] Yes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Enabled Selection for videos in the GUI.
 - Added `from_binary_mask()` method to CreateInstanceSegmentation and CreateSemanticSegmentation.
 - Group samples can be loaded in python UI. Note that GUI viewing is not supported yet.
+- Added `from_rle_mask()` method to CreateInstanceSegmentation and CreateSemanticSegmentation.
 
 ### Changed
 - Embedding plot doesn't require a license key anymore.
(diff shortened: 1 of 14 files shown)


lightly_studio_view/src/lib/components/SampleDetails/SampleObjectDetectionRect/SampleObjectDetectionRect.svelte
```diff
@@ -185,6 +190,7 @@
 
             setLastCreatedAnnotationId(newAnnotation.sample_id);
             setAnnotationId(newAnnotation.sample_id);
+            setCurrentBoundingBox(getBoundingBox(newAnnotation));
```
IgorSusmelj, line 194: If you add these two lines we can directly move the box. Because currently, I have to move the mouse pointer away and then back on the edge where I stopped drawing. It's only a minor thing. I leave it to you whether you want to implement it or not.
```suggestion
shouldDisableInteraction = true;
hoverbbox = true;
```

IgorSusmelj (approved):
LGTM

Left a minor suggestion. I tested it locally and it works.

lightly_studio_view/src/lib/components/SampleDetails/SampleObjectDetectionRect/SampleObjectDetectionRect.svelte
LeonardoRosaa: Nice suggestion.

It is done, thanks.

PR #523, opened 2026-02-10
Add PostgreSQL as a supported database backend

mihnea-necsulescu:
## What has changed and why?

Added support for PostgreSQL. LightlyStudio can now run on Postgres (without embedding plot and text search for now).

## How has it been tested?

Manually. To test manually:
1. Run a Postgres Docker container, e.g.:
    ```
    docker run -d --name postgres-test \
      -e POSTGRES_USER=lightly \
      -e POSTGRES_PASSWORD=lightly \
      -e POSTGRES_DB=lightly_studio \
      -p 5432:5432 \
      postgres:18-alpine
    ```
2. Set the `LIGHTLY_STUDIO_DATABASE_URL` env var (e.g., `LIGHTLY_STUDIO_DATABASE_URL=postgresql://lightly:lightly@localhost:5432/lightly_studio`)
3. Run LightlyStudio - everything should work as expected, except for the embedding plot and embedding text search, which will be resolved in a future PR)

- [ ] Yes
- [x] Not needed (internal change) - will update once everything is fully functional

### DIFF
--- a/lightly_studio_view/src/lib/components/SampleDetails/ImageDetails.svelte
+++ b/lightly_studio_view/src/lib/components/SampleDetails/ImageDetails.svelte
@@ -10,6 +10,7 @@
     import SampleDetailsBreadcrumb from './SampleDetailsBreadcrumb/SampleDetailsBreadcrumb.svelte';
     import { page } from '$app/state';
     import type { ImageView } from '$lib/api/lightly_studio_local';
+    import SampleDetailsNavigation from './SampleDetailsNavigation/SampleDetailsNavigation.svelte';
 
     const {
         sampleId,
@@ -59,6 +60,7 @@
     >
         {#if children}
             {@render children()}
+            <SampleDetailsNavigation />
         {/if}
         {#snippet breadcrumb({ collection: rootCollection })}
             <SampleDetailsBreadcrumb {rootCollection} {sampleIndex} />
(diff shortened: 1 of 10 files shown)


lightly_studio/src/lightly_studio/resolvers/metadata_resolver/json_utils.py
```diff
@@ -0,0 +1,89 @@
...
+    Returns:
+        A raw SQL expression string (e.g. ``"... IS NOT NULL"``).
+    """
+    backend = db_manager.get_backend()
+    if backend == DatabaseBackend.POSTGRESQL:
+        return f"{_build_pg_json_accessor(column=column, field=field)} IS NOT NULL"
+    return f"json_extract({column}, '$.{field}') IS NOT NULL"
+
+
+def _build_pg_json_accessor(column: str, field: str, *, cast_to_float: bool = False) -> str:
```
lukas-lightly, line 59: Is casting to float the only casting we ever do?

lightly_studio/src/lightly_studio/resolvers/metadata_resolver/json_utils.py
```diff
@@ -0,0 +1,89 @@
...
+    """Build a PostgreSQL JSON accessor expression from a dot-separated field path.
+
+    Converts paths like ``dict.key`` to PostgreSQL ``->``/``->>`` operator chains.
+
+    Args:
+        column: The fully-qualified column reference (e.g. ``metadata.data``).
```
lukas-lightly, line 65: I think we should also add an example with array access.

lightly_studio/src/lightly_studio/resolvers/metadata_resolver/json_utils.py
mihnea-necsulescu: at this point, yes

mihnea-necsulescu:
> Please consider splitting a longer PR such as this one into two in the future. It'd have been easier to review smaller parts, especially when there's a function such as `_build_pg_json_accessor()` with parsing logic.

I agree, but in this case, I really didn't think it was that long - most changes are in the test and uv.lock files; it also made sense to have everything here because json_utils is being called in 3 places only, and it also allows testing in this single PR

will keep this in mind, though, thanks!

PR #524, opened 2026-02-10
Igor lig 8567 segmentation labeling disable recentering in annoationview

IgorSusmelj:
## What has changed and why?

Fixes UX regressions in annotation-details mask editing and focus behaviour.
- Update zoomable container to allow conditional zooming based on events
  - we add `autoFocusEnabled` to allow disabling the autofocus / autozooming fully. We could drop this and only rely on the key. The key changes if we change the annotation for example. But then we need to be more careful when using the container in the future as the key would always need to be meaningful
  - we also add `autoFocusKey` as a mechanism to decide when to autofocus again. So we can autofocus when switching to the next annotation in annotations detail view while still being in edit mode for example
- Prevent repeated re-centering/jump zoom while editing segmentation masks (brush/eraser).
- Keep initial auto-focus when opening a new annotation-details item (next/prev, arrow keys, grid click).
- Ensure annotation-details always has its single annotation selected so brush/eraser can edit immediately (no “no annotation selected” error).

https://github.com/user-attachments/assets/28974b36-4d39-4033-bd9e-593eea15245f

## How has it been tested?

- Manual testing
- Additional unit and e2e tests for frontend

- [X] Yes

### DIFF
--- a/lightly_studio_view/src/lib/hooks/useCreateAnnotation/useCreateAnnotation.ts
+++ b/lightly_studio_view/src/lib/hooks/useCreateAnnotation/useCreateAnnotation.ts
@@ -2,14 +2,25 @@ import {
     type AnnotationCreateInput,
     type CreateAnnotationResponse
 } from '$lib/api/lightly_studio_local';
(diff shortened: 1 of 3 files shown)


ikondrat (approved):
lgtm

PR #527, opened 2026-02-10
feat: introduce resolver to return groups

ikondrat:
## What has changed and why?

This PR introduces resolver to fetch groups by collection id. 
Initial implementation will include filtering by tags and pagination. 

## How has it been tested?

By unit tests

## Did you update [CHANGELOG.md](../CHANGELOG.md)?

- [] Yes
- [x] Not needed (internal change)

PR #530, opened 2026-02-10
feat: introduce endpoint to fetch groups

ikondrat:
## What has changed and why?

NOTE: This PR requires [groups resolver PR](https://github.com/lightly-ai/lightly-studio/pull/527/changes#diff-924e40afdb8c35c327f291a787d8c4de7e220b46633cff35aa616ebc0a85a908) and will be merged only after mentioned.

We need to get information about available groups. 
This PR introduces endpoint to fetch groups as collection, with typical for sample filters.
Initial implementation should work for filtering with tags and pagination

## How has it been tested?

By unit tests

- [ ] Yes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Fixed annotation details mask editing to keep focus stable without annoying recentering after every edit and always select the shown annotation.
+- Fixed instance-segmentation brush/eraser edits occasionally being applied to the wrong sample after navigating between samples.
 
 ### Security
 
--- a/lightly_studio_view/src/lib/components/SampleDetails/SampleEraserRect/SampleEraserRect.svelte
+++ b/lightly_studio_view/src/lib/components/SampleDetails/SampleEraserRect/SampleEraserRect.svelte
@@ -59,11 +59,13 @@
     const { createAnnotation } = useCreateAnnotation({
         collectionId
     });
-    const { finishErase } = useSegmentationMaskEraser({
-        collectionId,
-        sample,
-        refetch: annotationLabelContext.isOnAnnotationDetailsView ? undefined : refetch
-    });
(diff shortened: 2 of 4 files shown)


lightly_studio/src/lightly_studio/api/routes/api/group.py
```diff
@@ -0,0 +1,49 @@
...
+from pydantic import BaseModel, Field
+from typing_extensions import Annotated
+
+from lightly_studio.api.routes.api.validators import Paginated, PaginatedWithCursor
+from lightly_studio.db_manager import SessionDep
+from lightly_studio.models.group import GroupViewsWithCount
+from lightly_studio.resolvers import group_resolver
+from lightly_studio.resolvers.group_resolver.group_filter import GroupFilter
+
+group_router = APIRouter(prefix="/collections/{collection_id}/group", tags=["group"])
```
michal-lightly: Optional side note: Note that we are not consistent in how we define the routers. I wonder if collection_id makes sense in the common prefix, some endpoints don't need it (e.g. we have `read_image` for images that does not use the collection id).

lightly_studio/src/lightly_studio/api/routes/api/group.py
```diff
@@ -0,0 +1,49 @@
...
+class ReadGroupsRequest(BaseModel):
+    """Request body for reading groups."""
+
+    filter: GroupFilter | None = Field(None, description="Filter parameters for groups")
+
+
+@group_router.post("/", response_model=GroupViewsWithCount)
+def get_all_groups(
+    session: SessionDep,
+    collection_id: Annotated[UUID, Path(title="collection Id")],
```
michal-lightly: Same as in the parent PR, collection_id can be removed.

lightly_studio/src/lightly_studio/api/routes/api/group.py
ikondrat: you are right let's drop it here

michal-lightly (approved):

PR #544, opened 2026-02-12
Bugfix: Active drawing tool disappears when switching image

IgorSusmelj:
## What has changed and why?

- Fixes bug: When navigating in sample details view with an active tool (e.g. brush for segmentations) the tool would disappear sometimes when navigating to the prev/next frame

## How has it been tested?

- Additional tests
- Extensive manual testing

- [X] Yes

### DIFF
--- a/lightly_studio/docs/docs/index.md
+++ b/lightly_studio/docs/docs/index.md
@@ -502,7 +502,7 @@ sample.add_annotation(
 )
 ```
 
-Alternatively, you can provide the mask encoding (see [Binary Mask Format](#binary-mask-format)) using the `from_rle_mask` method:
+Alternatively, you can provide the mask encoding using the `from_rle_mask` method:
 
 ```python
 from lightly_studio.core.annotation import CreateSemanticSegmentation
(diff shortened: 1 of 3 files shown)


lightly_studio_view/src/lib/components/SampleDetails/SampleDetailsNavigation/SampleDetailsNavigation.svelte
```diff
@@ -68,5 +57,3 @@
         isDrawing={context?.isDrawing}
     />
 {/if}
-
-<svelte:window onkeydown={handleKeyDownEvent} />
```
ikondrat, line 72: so if I understood correctly it was caused by duplicated event handling, right?

IgorSusmelj: Yes, the duplicate even handling was part of the issue. There was also an issue that we sometimes trigger a remount which resets the mount (I removed the onmount in this PR). 
I did not notice this before this PR...

ikondrat:
For sample details the issue is solved but it looks like we have the same issue for annotation details

ikondrat (approved):

PR #549, opened 2026-02-13
Bugfix: Embedding plot bug where the lasso and square selection stay visible

IgorSusmelj:
## What has changed and why?

This PR fixes embedding plot selection UX:

- After lasso/rectangle selection is applied, the selection geometry is cleared from the plot.
- Pressing Escape resets the current selection when the embedding plot is open.

https://github.com/user-attachments/assets/6cbf9663-b2fa-4bd4-aed6-604a6f0d67ee

## How has it been tested?

- Additional unit test
- Manual testing

- [X] Yes

### DIFF
--- a/lightly_studio_view/src/lib/components/Video/Video.svelte
+++ b/lightly_studio_view/src/lib/components/Video/Video.svelte
@@ -1,6 +1,7 @@
 <script lang="ts">
     import { PUBLIC_VIDEOS_FRAMES_MEDIA_URL, PUBLIC_VIDEOS_MEDIA_URL } from '$env/static/public';
     import type { FrameView, VideoFrameView, VideoView } from '$lib/api/lightly_studio_local';
+    import { findFrame } from '$lib/utils/frame';
     import { onMount } from 'svelte';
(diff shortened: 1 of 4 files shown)


lightly_studio_view/src/lib/components/PlotPanel/usePlotData/usePlotData.ts
```diff
@@ -61,6 +63,16 @@ export function usePlotData({
         }
     }
 
+    if (highlightedSampleIds.length > 0) {
+        const highlightedSampleIdSet = new Set(highlightedSampleIds);
+        category = category.map((cat, index) => {
+            if (cat !== 1) {
+                return cat;
+            }
+            return highlightedSampleIdSet.has(sampleIds[index]) ? 2 : cat;
```
ikondrat: it is hard to understand the fix here
I would add few comments

and simplify the conditions

smth like 
```
return cat === 1 && highlightedSampleIdSet.has(sampleIds[index]) ? 2 : 1; 
```

PR #555, opened 2026-02-14
Feat: Add support for bulk selection using shift click

IgorSusmelj:
## What has changed and why?

- Adds support for bulk selection in the grid views (samples, annotations, videos, frames)
- Adds new tests

Other changes:
 - We now clear the selection when switching views
   - Before this PR: After selecting items in image grid and then going to annotations and back one would still have the same image items selected. 
   - After this PR: We clear the selection and whenever you switch the view (images <--> annotations, videos, frames..)

I also fixed some other issues around the grid views:
- When doing bulk selection with shift + click we no longer have a blue film on the items (see image below) 
image

- When pressing escape after selecting an item we no longer have the blueish border around it (see image below of what happened before this change)
image

Video of all the changes in action
https://github.com/user-attachments/assets/cb0e89da-9dfa-4ddd-94f9-a6f6e8271000

## How has it been tested?

- New E2E tests
- Manual testing for image and video datasets

- [X] Yes

### DIFF
--- a/lightly_studio_view/src/lib/components/Samples/Samples.svelte
+++ b/lightly_studio_view/src/lib/components/Samples/Samples.svelte
@@ -7,7 +7,7 @@
     import { useTags } from '$lib/hooks/useTags/useTags';
     import { routeHelpers } from '$lib/routes';
     import { onMount } from 'svelte';
-    import { get, type Readable } from 'svelte/store';
+    import type { Readable } from 'svelte/store';
     import {
         isNormalModeParams,
         useImagesInfinite,
@@ -209,7 +209,7 @@
     }) {
         selectionAnchorSampleId = selectRangeByAnchor({
             sampleIdsInOrder: samples.map((sample) => sample.sample_id),
-            selectedSampleIds: get(selectedSampleIds),
+            selectedSampleIds: $selectedSampleIds,
             clickedSampleId: sampleId,
             clickedIndex: index,
             shiftKey,


lightly_studio_view/src/lib/utils/selectRangeByAnchor.ts
```diff
@@ -0,0 +1,37 @@
...
+    sampleIdsInOrder: string[];
+    selectedSampleIds: Set<string>;
+    clickedSampleId: string;
+    clickedIndex: number;
+    shiftKey: boolean;
+    anchorSampleId: string | null;
+    onSelectSample: (sampleId: string) => void;
+};
+
+export function selectRangeByAnchor({
```
LeonardoRosaa, line 27: I'd suggest adding a few comments since the code isn't trivial

lightly_studio_view/src/routes/datasets/[dataset_id]/[collection_type]/[collection_id]/frames/+page.svelte
```diff
@@ -37,13 +39,37 @@
...
+        index,
+        shiftKey
+    }: {
+        sampleId: string;
+        index: number;
+        shiftKey: boolean;
+    }) {
+        selectionAnchorSampleId = selectRangeByAnchor({
+            sampleIdsInOrder: items.map((item) => item.sample_id),
+            selectedSampleIds: get(selectedSampleIds),
```
Comment (LeonardoRosaa) [inline lightly_studio_view/src/routes/datasets/[dataset_id]/[collection_type]/[collection_id]/frames/+page.svelte]: Could you use `$selectedSampleIds` instead of `get(selectedSampleIds)`?

PR #557, opened 2026-02-14
Bugfix: Empty captions were added after caption creation

IgorSusmelj:
## What has changed and why?

- Adds a new component for caption creation
- New component does not create empty captions. And use can start typing right away and finish with the "tick" button and a mouse click or by pressing 'Enter" on the keyboard.
- While testing I found another bug: Pressing the 'Spacebar' was not propagated to the input field properly so captions could ever be a single word. I fixed that and made sure we test for this.
- Additionally, limit that only 1 new caption draft can be active. Also added a test for this.

Note that we already have a component for caption editing/showing. I decided to make a new component only for the flow of creating a new caption from scratch. 

https://github.com/user-attachments/assets/7da67794-7a97-4e5d-80af-b3285b9ae936

## How has it been tested?

- New unit tests
- Manual testing

- [X] Yes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,9 +9,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Add “View Video” button in the frame details view to open video details for the selected frame.
 - Added `Escape` shortcut support in the embedding plot to clear the current selection.
+- Added Shift+click range selection in grid views.
 
 ### Changed
+
+- Selection now resets when switching between grid views, while filters persist.
 
 ### Deprecated
 
@@ -21,6 +25,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Fixed right-click `Copy image` in grid and detail views to copy images from the GUI.
 - Fixed annotation details mask editing to keep focus stable without annoying recentering after every edit and always select the shown annotation.
 - Fixed embedding plot UI stability and improved legend/control layout for narrow windows.
 - Fixed instance-segmentation brush/eraser edits occasionally being applied to the wrong sample after navigating between samples.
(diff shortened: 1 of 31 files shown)


lightly_studio_view/e2e/caption-utils.ts
```diff
@@ -27,24 +27,35 @@ export class CaptionUtils {
         return this.getNthCaption(index).getByTestId('caption-input').inputValue();
     }
 
-    // TODO(Michal, 12/2025): Change the function signature once it is not possible to
```
IgorSusmelj, line 30: I assume the todo is resolved with this change. We now create a new caption and can directly edit it before submitting it.

lightly_studio_view/src/lib/components/CaptionField/CreateCaptionField.svelte
```diff
@@ -0,0 +1,122 @@
...
+        if (onCreatingCaptionChange) {
+            onCreatingCaptionChange(isOpen);
+            return;
+        }
+        internalIsCreatingCaption = isOpen;
+    };
+
+    $effect(() => {
+        if (!isCreatingCaption || !newCaptionInput) return;
+        void tick().then(() => newCaptionInput?.focus());
```
ikondrat: 🤔 why you need tick here?

ikondrat (approved):
lgtm

I can create captions with it 

but I am not sure I know the full flow for captions creation, maybe it is better to have a check by @JonasWurst

lightly_studio_view/src/lib/components/CaptionField/CreateCaptionField.svelte
IgorSusmelj: Looking into it

IgorSusmelj: Got rid if it :)

PR #561, opened 2026-02-16
Enable embedding text search on Postgres

mihnea-necsulescu:
## What has changed and why?

Embedding text search was not functional out of the box for Postgres, because Postgres does not support vector operations on arrays. This was fixed by: 
- adding the pgvector extension 
- changing the sample embedding type to `vector` (for Postgres only)
- abstracting the cosine distance functionality behind a dialect-specific `cosine_distance`

## How has it been tested?

Manually + existing tests still pass.
To test manually:
1. Run a pgvector container, e.g.:
    ```
    docker run -d --name pgvector-test \
      -e POSTGRES_USER=lightly \
      -e POSTGRES_PASSWORD=lightly \
      -e POSTGRES_DB=lightly_studio \
      -p 5432:5432 \
      pgvector/pgvector:pg17
    ```
2. Set the `LIGHTLY_STUDIO_DATABASE_URL` env var, e.g., `LIGHTLY_STUDIO_DATABASE_URL=postgresql://lightly:lightly@localhost:5432/lightly_studio`
3. Run Studio with Postgres: text search should work.
4. Also run Studio with DuckDB: text search should still work.

- [ ] Yes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fixed sample-details navigation so keyboard and button navigation keep active tool behavior deterministic across samples.
 - Fixed embedding plot selection UX so rectangle/lasso overlays disappear after selection while selected samples remain highlighted.
 - Fixed outdated `VideoDataset` import path in README and docs quickstart examples.
+- Fixed caption creation UX in edit mode: clicking `+` now opens a focused input draft, captions are created only on explicit save/Enter, and spaces in the draft input are handled correctly.
 
 ### Security
 
--- a/lightly_studio_view/e2e/caption-utils.ts
+++ b/lightly_studio_view/e2e/caption-utils.ts
@@ -17,63 +17,35 @@ export class CaptionUtils {
         return this.page.getByTestId('caption-field').nth(index);
     }
 
-    // Get text content of the nth caption. Useful for checking the caption in view mode.
     getNthCaptionText(index: number) {
         return this.getNthCaption(index).textContent();
     }
 
-    // Get input value of the nth caption. Useful for checking the caption in edit mode.
     getNthCaptionInput(index: number) {
         return this.getNthCaption(index).getByTestId('caption-input').inputValue();
     }
 
-    // TODO(Michal, 12/2025): Change the function signature once it is not possible to
-    // add a caption without text.
-    async addCaption(addButtonIndex: number = 0) {
+    async addCaption(addButtonIndex: number = 0, text: string = 'new caption') {
         const captionCountBefore = await this.getCaptionCount();
         await this.page.getByTestId('add-caption-button').nth(addButtonIndex).click();
-        // Wait until the new caption field appears
+        const newCaptionInput = this.page.getByTestId('new-caption-input').nth(0);
+        await expect(newCaptionInput).toBeVisible();
+        await newCaptionInput.fill(text);
+        await newCaptionInput.press('Enter');
         await expect(this.page.getByTestId('caption-field')).toHaveCount(captionCountBefore + 1);
     }
 
-    /**
-     * Add a caption on the captions page (grid of caption items).
-     * Unlike addCaption, this waits for the POST /api/samples/list response after
-     * clicking add, so the grid has refetched and the new caption is in the list
-     * before proceeding. Use this when adding a caption on the captions page to
-     * avoid flaky tests from asserting or interacting before the list updates.
-     */
-    async addCaptionInCaptionPage(addButtonIndex: number = 0) {
-        const captionCountBefore = await this.getCaptionCount();
-
-        const samplesListPromise = this.page.waitForResponse(
-            (response) =>
-                response.request().method() === 'POST' &&
-                response.url().includes('/api/samples/list') &&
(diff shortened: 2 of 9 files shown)


lightly_studio/src/lightly_studio/resolvers/similarity_utils.py
```diff
@@ -5,9 +5,10 @@
 from typing import Any
 from uuid import UUID
 
-from sqlalchemy import ColumnElement, func
+from sqlalchemy import ColumnElement
 from sqlmodel import Session, col, select
 
+from lightly_studio.db_types import cosine_distance
```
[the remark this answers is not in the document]
michal-lightly: Please implement this comment. The rest is import from sqlalchemy where we are a bit lax (I think we should though chnage the imports to follow the guideline also there at some point).

lightly_studio/src/lightly_studio/db_vector.py
```diff
@@ -0,0 +1,63 @@
+"""Dialect-aware database types and functions.
```
michal-lightly: Minor: I'd suggest to call the file `db_vector_type.py`. No need to mix more types in one file.

lightly_studio/src/lightly_studio/db_types.py
```diff
@@ -0,0 +1,63 @@
...
+    impl = ARRAY(Float)
+    cache_ok = True
+
+    def load_dialect_impl(self, dialect: Dialect) -> TypeEngine[Any]:
+        """Return the dialect-specific type: pgvector VECTOR or ARRAY(Float)."""
+        if dialect.name == "postgresql":
+            from pgvector.sqlalchemy import Vector
+
+            return dialect.type_descriptor(Vector())
+        return dialect.type_descriptor(ARRAY(Float))
```
michal-lightly: This logic does not follow the docstring.
* Either there should be if-else logic by `dialect.name` with error on unknown type
* Or the docstring should say that it is x for postgres and ARRAY for everything else.

In my opinion, the first option is cleaner (easier to maintain).

lightly_studio/src/lightly_studio/db_vector.py
```diff
@@ -0,0 +1,63 @@
...
+
+class cosine_distance(GenericFunction[float]):  # noqa: N801
+    """Cosine distance function that compiles to dialect-specific SQL.
+
+    Uses the <=> operator on both DuckDB and PostgreSQL (pgvector).
+    PostgreSQL requires explicit ::vector casts on both operands.
+    """
+
+    type = Float()
+    inherit_cache = True
```
michal-lightly, line 53: What does this line do?

lightly_studio/tests/test_db_types.py
```diff
@@ -0,0 +1,31 @@
+"""Tests for db_types module."""
+
+from __future__ import annotations
+
+from sqlalchemy import ARRAY, Float, create_mock_engine
```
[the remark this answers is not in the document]
michal-lightly: Optional: I think we can also here do `import sqlalchemy` and then `sqlalchemy.create_mock_engine(...)`, or do you see why not?

lightly_studio/tests/test_db_types.py
```diff
@@ -0,0 +1,39 @@
...
+    # TODO(Mihnea, 02/2026): Remove the skip once we deprecate support for Python 3.8.
+    @pytest.mark.skipif(
+        sys.version_info < (3, 9),
+        reason="pgvector is only installed for Python >= 3.9",
+    )
+    def test_load_dialect_impl__postgresql(self) -> None:
+        """VectorType returns pgvector VECTOR for PostgreSQL dialect."""
+        from pgvector.sqlalchemy import Vector
+
+        # Use a mock engine since we don't have a test Postgres connection yet.
```
michal-lightly: Should there be a TODO to change this in the future?

michal-lightly (changes requested):

lightly_studio/src/lightly_studio/db_vector.py
mihnea-necsulescu: SQLAlchemy caches compiled SQL, but not for custom types/functions, because it doesn't know if it should, so it's up to the user to enable it

from what I read, this would only be problematic if they had custom initializers that would affect the instance level state (didn't look too much into it cause it's not our case); without them, though, caching can just be enabled

lightly_studio/tests/test_db_types.py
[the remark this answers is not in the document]
mihnea-necsulescu: did this (also for `column`, too)

lightly_studio/src/lightly_studio/db_vector.py
mihnea-necsulescu: renamed it to `db_vector_type`, but then renamed it to just `db_vector` - it contains both a type and an operation + I'll also need to add another generic function for https://linear.app/lightly/issue/LIG-8683/2d-embedding-caching-hash-computation-only-hash-first-embedding, which makes sense to be added here, too

michal-lightly: Even better :)

michal-lightly: The comment can be now adjusted.

PR #562, opened 2026-02-16
API endpoint for exporting instance segmentations

lukas-lightly:
## What has changed and why?
We need an API endpoint for exporting instance segmentations, in order to support it in the GUI.

## How has it been tested?
Added a test.

- [ ] Yes

### DIFF
--- a/lightly_studio_view/src/lib/components/NavigationMenu/utils.ts
+++ b/lightly_studio_view/src/lib/components/NavigationMenu/utils.ts
@@ -5,16 +5,17 @@ import { Image, WholeWord, Video, Frame, ComponentIcon, LayoutDashboard } from '
 import type { BreadcrumbLevel, NavigationMenuItem } from './types';
 
 export function getMenuItem(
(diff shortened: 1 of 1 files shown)


lightly_studio/src/lightly_studio/api/routes/api/export.py
```diff
@@ -108,6 +108,49 @@ def export_collection_captions(
     )
 
 
+@export_router.get("/export/instance-segmentations")
+def export_collection_instance_segmentations(
```
horatiualmasan: This is a very close copy of `export_collection_annotations`, do you think it make sense to add a annotation type parameter to `/export/annotations` and use the same endpoint?

horatiualmasan (approved):
LGTM

lightly_studio/src/lightly_studio/api/routes/api/export.py
lukas-lightly: Done, I added a annotation type parameter.

PR #564, opened 2026-02-16
feat: return group_snapshot and sample_count information

ikondrat:
## What has changed and why?

For every group we need to get group snapshot information and sample_count. 

**group snapshot**
It should provide information as image if the first component in the group is image 
It should provide information about video if the first component is video.

## How has it been tested?
By unit tests

- [ ] Yes

### DIFF
--- a/lightly_studio/src/lightly_studio/models/group.py
+++ b/lightly_studio/src/lightly_studio/models/group.py
@@ -1,6 +1,6 @@
 """Group table definition."""
 
-from typing import List, Optional
+from typing import List, Optional, Union
 from uuid import UUID
 
 from pydantic import BaseModel, ConfigDict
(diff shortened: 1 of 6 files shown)


lightly_studio/src/lightly_studio/models/group.py
```diff
@@ -48,3 +51,8 @@ class GroupViewsWithCount(BaseModel):
     samples: List[GroupView] = PydanticField(..., alias="data")
     total_count: int
     next_cursor: Optional[int] = PydanticField(None, alias="nextCursor")
+
+
+# Import at the bottom to avoid circular imports
```
ikondrat, line 56: Would be great to have better solution for circular imports

PR #567, opened 2026-02-16
feat: return amount of samples per group

ikondrat:
## What has changed and why?

We need to proved amount of samples per group. To show it for UI.
It will be reused in coming PRs in the get_all.py resolver

## How has it been tested?

By unit tests

- [ ] Yes

### DIFF
--- a/lightly_studio/src/lightly_studio/resolvers/group_resolver/get_group_sample_counts.py
+++ b/lightly_studio/src/lightly_studio/resolvers/group_resolver/get_group_sample_counts.py
@@ -2,7 +2,7 @@
 
 from __future__ import annotations
 
-from typing import Sequence
+from collections.abc import Sequence
 from uuid import UUID
 
 from sqlmodel import Session, col, func, select


lightly_studio/src/lightly_studio/resolvers/group_resolver/get_group_sample_counts.py
```diff
@@ -0,0 +1,40 @@
...
+    """Get the count of samples for each group.
+
+    Args:
+        session: Database session for executing queries.
+        group_sample_ids: List of group sample IDs to count samples for.
+
+    Returns:
+        Dictionary mapping group sample_id to the count of samples in that group.
+    """
+    if not group_sample_ids:
```
LeonardoRosaa, line 26: Should we throw an error or None?

ikondrat: Can you ellaborate a bit why it would be better than empty dict?

LeonardoRosaa: I’d say we should throw an error if we always expect `group_sample_ids` to be non-empty, and make that a hard rule. Otherwise, we could return `None`, since it clearly indicates that no result was found.

ikondrat: alright let's make the ValueError

PR #569, opened 2026-02-16
2d-Embedding caching hash computation: Only hash first embedding

mihnea-necsulescu:
## What has changed and why?

Hash just on the SampleEmbeddingTable.embedding[0], i.e. the first dimension instead of the whole embedding. 

## How has it been tested?

Manually + embedding plot should now work for both DuckDB and Postgres

- [ ] Yes


MalteEbner (commented):
Code looks good to me.

Can you add a quick benchmark of the embedding hash computation with and without this PR? 

You can vibe-code the benchmark and run it e.g. on 50k samples with a 512-dim embedding each.

 Similar e.g. to https://github.com/lightly-ai/lightly-studio/pull/543

MalteEbner (approved):
LGTM² (Looks good to me, let's get this merged)

PR #590, opened 2026-02-19
Introduce ObjectTrack model

horatiualmasan:
## What has changed and why?

New model for object tracks

## How has it been tested?

tests will follow when adding resolvers

- [ ] Yes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - Selection now resets when switching between grid views, while filters persist.
+- Embedding model weights are now saved to the user cache by default. The cache dir can be changed via `LIGHTLY_STUDIO_MODEL_CACHE_DIR`.
 
 ### Deprecated
 
--- a/lightly_studio/.env.example
+++ b/lightly_studio/.env.example
@@ -14,6 +14,9 @@ LIGHTLY_STUDIO_LICENSE_KEY="your_license_key_here"
 # Embedding model if using LightlyEdge
 # EDGE_MODEL_PATH="/path/to/lightly_model.tar"
 
+# Embedding model cache dir, defaults to user/.cache/lightly-studio. 
+# LIGHTLY_STUDIO_MODEL_CACHE_DIR="/Users/username/.cache/lightly-studio"
+
 # Example script dataset paths
 EXAMPLES_DATASET_PATH="dataset_examples/coco_subset_128_images/images"
 EXAMPLES_YOLO_YAML_PATH="dataset_examples/road_signs_yolo/data.yaml"
--- a/lightly_studio/src/lightly_studio/dataset/env.py
+++ b/lightly_studio/src/lightly_studio/dataset/env.py
@@ -1,5 +1,6 @@
 """Initialize environment variables for the dataset module."""
 
+from pathlib import Path
 from typing import Optional
 
 from environs import Env
@@ -8,6 +9,9 @@ env = Env()
 env.read_env()
 LIGHTLY_STUDIO_EMBEDDINGS_MODEL_TYPE: str = env.str(
     "LIGHTLY_STUDIO_EMBEDDINGS_MODEL_TYPE", "MOBILE_CLIP"
+)
+LIGHTLY_STUDIO_MODEL_CACHE_DIR: Path = env.path(
+    "LIGHTLY_STUDIO_MODEL_CACHE_DIR", Path.home() / ".cache" / "lightly-studio"
 )
 LIGHTLY_STUDIO_EDGE_MODEL_FILE_PATH: str = env.str("EDGE_MODEL_PATH", "./lightly_model.tar")
 LIGHTLY_STUDIO_PROTOCOL: str = env.str("LIGHTLY_STUDIO_PROTOCOL", "http")
--- a/lightly_studio/src/lightly_studio/dataset/mobileclip_embedding_generator.py
+++ b/lightly_studio/src/lightly_studio/dataset/mobileclip_embedding_generator.py
@@ -2,7 +2,6 @@
 
 from __future__ import annotations
 
-import tempfile
 from pathlib import Path
 from typing import Callable
 from uuid import UUID
@@ -15,6 +14,7 @@ from PIL import Image
 from torch.utils.data import DataLoader, Dataset
 from tqdm import tqdm
 
+from lightly_studio.dataset.env import LIGHTLY_STUDIO_MODEL_CACHE_DIR
 from lightly_studio.models.embedding_model import EmbeddingModelCreate
 from lightly_studio.vendor import mobileclip
 
@@ -157,7 +157,7 @@ class MobileCLIPEmbeddingGenerator(ImageEmbeddingGenerator):
 
 
 def _get_cached_mobileclip_checkpoint() -> Path:
-    file_path = Path(tempfile.gettempdir()) / f"{MODEL_NAME}.pt"
+    file_path = LIGHTLY_STUDIO_MODEL_CACHE_DIR / f"{MODEL_NAME}.pt"
     file_utils.download_file_if_does_not_exist(
         url=MOBILECLIP_DOWNLOAD_URL,
         local_filename=file_path,
--- a/lightly_studio/src/lightly_studio/dataset/perception_encoder_embedding_generator.py
+++ b/lightly_studio/src/lightly_studio/dataset/perception_encoder_embedding_generator.py
@@ -15,6 +15,7 @@ from PIL import Image
 from torch.utils.data import DataLoader, Dataset
 from tqdm import tqdm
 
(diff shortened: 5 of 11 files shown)


lightly_studio/src/lightly_studio/models/annotation/annotation_base.py
```diff
@@ -88,6 +93,12 @@ class AnnotationBaseTable(SQLModel, table=True):
         sa_relationship_kwargs={"lazy": "select"},
     )
 
+    """ The track this annotation belongs to, if any. """
+    object_track: Mapped[Optional["ObjectTrackTable"]] = Relationship(
+        back_populates="annotations",
+        sa_relationship_kwargs={"lazy": "select"},
```
michal-lightly: I think we can make it eager. It is essentially just one int. Otherwise seemingly one query might turn under the hood into a set of many queries if we iterate over annotations and access the track id. WDYT?

lightly_studio/src/lightly_studio/models/annotation/annotation_base.py
```diff
@@ -145,6 +161,10 @@ def from_annotation_table(cls, annotation: "AnnotationBaseTable") -> "Annotation
             annotation_type=annotation.annotation_type,
             confidence=annotation.confidence,
             created_at=annotation.created_at,
+            object_track_id=annotation.object_track_id,
+            object_track_number=annotation.object_track.object_track_number
```
michal-lightly, line 164: E.g. here, this might trigger a new query.

lightly_studio/src/lightly_studio/models/annotation/object_track.py
```diff
@@ -0,0 +1,77 @@
...
+
+from typing import TYPE_CHECKING
+from uuid import UUID, uuid4
+
+from pydantic import BaseModel, ConfigDict
+from sqlalchemy import UniqueConstraint
+from sqlalchemy.orm import Mapped
+from sqlmodel import Field, Relationship, SQLModel
+
+if TYPE_CHECKING:
```
michal-lightly: Can you try to write this without this if clause? I was able to completely eliminate it for models outside annotations. I would timebox it to 10 minutes.

lightly_studio/src/lightly_studio/models/annotation/object_track.py
```diff
@@ -0,0 +1,77 @@
...
+    """
+
+    __tablename__ = "object_track"
+    __table_args__ = (
+        # Object tracks should should have unique object_track_number per parent sample
+        UniqueConstraint("object_track_number", "parent_sample_id", name="unique_object_track"),
+    )
+    object_track_id: UUID = Field(default_factory=uuid4, primary_key=True)
+
+    """Numeric identifier for the object track, scoped to the parent sample."""
```
michal-lightly: Docstring in python are written below the symbol they document. (For normal comments, we write it at the top.)

lightly_studio/src/lightly_studio/models/annotation/object_track.py
```diff
@@ -0,0 +1,77 @@
...
+    object_track_number: int
+
+    """The sample that this object track belongs to."""
+    parent_sample_id: UUID = Field(foreign_key="sample.sample_id")
```
michal-lightly: I see a potential issue with this.

If we want the object number to be unique per video, this works well. But what if we have multiple videos of the same scene, and we want to identify an object across videos? What would be the parent sample id?

We could solve it with groups, but it creates some issues (loading might get complex, users have to create artificial groups).

The core reason this is introduced is the uniqueness constraint - we need to think what the object number should be unique for (or just drop the unique constraint).

lightly_studio/src/lightly_studio/models/annotation/object_track.py
```diff
@@ -0,0 +1,77 @@
...
+
+    """The label for all annotations in this track."""
+    annotation_label_id: UUID = Field(foreign_key="annotation_label.annotation_label_id")
```
michal-lightly: How would this work? The individual annotations already have labels.

lightly_studio/src/lightly_studio/models/annotation/object_track.py
horatiualmasan: Removed parent_sample_id and the unique constraint

lightly_studio/src/lightly_studio/models/annotation/object_track.py
horatiualmasan: removed from the object track, annotation label consistency will be handled by the resolvers if annotation is linked to a track

lightly_studio/src/lightly_studio/models/annotation/object_track.py
horatiualmasan: Didn't work within 10 min

lightly_studio/src/lightly_studio/models/annotation/annotation_base.py
```diff
@@ -88,6 +93,12 @@ class AnnotationBaseTable(SQLModel, table=True):
         sa_relationship_kwargs={"lazy": "select"},
     )
 
+    """ The track this annotation belongs to, if any. """
```
michal-lightly: Optional: I noticed that the comments are weird in this class, there are extra spaces. Can you fix it?

Also the whole file uses docstrings wrong - but that's for another PR.

lightly_studio/src/lightly_studio/models/annotation/annotation_base.py
```diff
@@ -88,6 +93,12 @@ class AnnotationBaseTable(SQLModel, table=True):
...
+    object_track: Mapped[Optional["ObjectTrackTable"]] = Relationship(
+        back_populates="annotations",
+        sa_relationship_kwargs={"lazy": "joined"},
```
michal-lightly, line 98: Just checking, what does joined vs. select mean? Joined is eager and select lazy?

I would naively expect there to be no relationship keyworkds.

lightly_studio/src/lightly_studio/models/annotation/object_track.py
```diff
@@ -0,0 +1,68 @@
...
+    # Numeric identifier for the object track.
+    object_track_number: int
+
+    # The root collection the object track belongs to.
+    dataset_id: UUID = Field(foreign_key="collection.collection_id")
+
+    annotations: Mapped[list["AnnotationBaseTable"]] = Relationship(
+        back_populates="object_track",
+        sa_relationship_kwargs={"lazy": "select"},
+    )
```
michal-lightly: I'd suggest to remove this. It is useful if we fetch ObjectTrackTable as an object and want to access all of its annotations. I think it is unlikely we'd do this without a resolver, and a resolver can construct its own query. WDYT?

So the idea is that annotation can access its track, but not the other way (one-way relationship).

This might also solve the issue with recursive imports.

lightly_studio/src/lightly_studio/resolvers/collection_resolver/deep_copy.py
```diff
@@ -361,6 +363,28 @@ def _copy_captions(
...
+        select(ObjectTrackTable).where(col(ObjectTrackTable.dataset_id).in_(old_collection_ids))
+    ).all()
+
+    for old_track in tracks:
+        new_track = _copy_with_updates(
+            old_track,
+            {
+                "object_track_id": uuid4(),
+                "object_track_number": old_track.object_track_number,
+                "collection_id": ctx.collection_map[old_track.dataset_id],
```
michal-lightly:
```suggestion
                "dataset_id": ctx.collection_map[old_track.dataset_id],
```

michal-lightly: We need to add a test with an annotation with a track to check for such issues.

michal-lightly (changes requested):

lightly_studio/src/lightly_studio/models/annotation/annotation_base.py
horatiualmasan: Yes joined is eager and select lazy. The default (also with no relationship keywords) is [lazy ](https://docs.sqlalchemy.org/en/20/orm/queryguide/relationships.html) loading.

lightly_studio/src/lightly_studio/models/annotation/annotation_base.py
horatiualmasan: Updated comments

lightly_studio/src/lightly_studio/models/annotation/object_track.py
horatiualmasan: Yes, we could leave it to a resolver. Removed relationship and also the imports are not needed anymore.

lightly_studio/src/lightly_studio/resolvers/collection_resolver/deep_copy.py
horatiualmasan: I will add it in the resolver pr once we can create a object track.

michal-lightly (approved):

PR #601, opened 2026-02-23
Add resolvers for Object Track

horatiualmasan:
## What has changed and why?

Add new resolver to work with object tracks.

## How has it been tested?

New tests

- [ ] Yes

### DIFF
--- a/lightly_studio/tests/resolvers/collection_resolver/test_deep_copy.py
+++ b/lightly_studio/tests/resolvers/collection_resolver/test_deep_copy.py
@@ -6,7 +6,7 @@ from sqlmodel import Session
 from lightly_studio.metadata.gps_coordinate import GPSCoordinate
 from lightly_studio.models.annotation.annotation_base import AnnotationType
 from lightly_studio.models.annotation.object_detection import ObjectDetectionAnnotationTable
-from lightly_studio.models.annotation.object_track import ObjectTrackCreate, ObjectTrackTable
+from lightly_studio.models.annotation.object_track import ObjectTrackCreate
 from lightly_studio.models.annotation.segmentation import SegmentationAnnotationTable
 from lightly_studio.models.collection import SampleType
 from lightly_studio.resolvers import (
@@ -585,7 +585,9 @@ def test_deep_copy__with_annotations(test_db: Session) -> None:
     assert od_detail.height == 40
     assert copied_od.object_track_id is not None
     assert copied_od.object_track_id != original_track_id
-    copied_track = test_db.get(ObjectTrackTable, copied_od.object_track_id)
+    copied_track = object_track_resolver.get_by_id(
+        session=test_db, object_track_id=copied_od.object_track_id
+    )
     assert copied_track is not None
     assert copied_track.object_track_number == 7
     assert copied_track.dataset_id == copied.collection_id


michal-lightly:
@codex review

lightly_studio/src/lightly_studio/resolvers/object_track_resolver/add_annotation_to_object_track.py
```diff
@@ -0,0 +1,43 @@
...
+
+    Args:
+        session: Database session for executing the operation.
+        annotation_id: UUID of the annotation to link.
+        object_track: The object track to link the annotation to.
+
+    Returns:
+        The updated annotation with tracking_id set.
+    """
+    annotation = annotation_resolver.get_by_id(session, annotation_id)
```
michal-lightly: Minor: Style: Please use keywords

lightly_studio/src/lightly_studio/resolvers/object_track_resolver/add_annotation_to_object_track.py
```diff
@@ -0,0 +1,43 @@
...
+        raise ValueError(f"Annotation with ID {annotation_id} not found.")
+    try:
+        return annotation_helper.update_annotation_object(
+            session=session,
+            annotation=annotation,
+            fields_to_update={"object_track_id": object_track.object_track_id},
+        )
+    except Exception:
+        session.rollback()
+        raise
```
michal-lightly: Why do we need to handle the exception? Isn't the session rolled back by the db_manager?

lightly_studio/src/lightly_studio/resolvers/object_track_resolver/create_object_track.py
```diff
@@ -0,0 +1,34 @@
...
+from sqlmodel import Session
+
+from lightly_studio.models.annotation.object_track import ObjectTrackTable
+
+
+def create_object_track(
+    session: Session,
+    object_track_number: int,
+    dataset_id: UUID,
+) -> ObjectTrackTable:
```
michal-lightly: Optional: It might be cleaner to just return the id and fetch the track separately. I suspect that we will need a bulk create method anyway (to load large datasets with tracks) and the pattern we successfully used is `create_many(..) -> list[UUID]`.

Do you think it makes sense to just introduce `create_many` already now?

lightly_studio/tests/resolvers/collection_resolver/test_deep_copy.py
```diff
@@ -571,6 +584,12 @@ def test_deep_copy__with_annotations(test_db: Session) -> None:
     assert od_detail.y == 20
     assert od_detail.width == 30
     assert od_detail.height == 40
+    assert copied_od.object_track_id is not None
+    assert copied_od.object_track_id != track.object_track_id
+    copied_track = test_db.get(ObjectTrackTable, copied_od.object_track_id)
+    assert copied_track is not None
+    assert copied_track.object_track_number == track.object_track_number
```
michal-lightly: Optional:
```suggestion
    assert copied_track.object_track_number == 7
```

lightly_studio/tests/resolvers/object_track_resolver/test_add_annotation_to_object_track.py
```diff
@@ -0,0 +1,48 @@
...
+        annotation_id=annotation.sample_id,
+        object_track=track,
+    )
+
+    assert result.object_track_id == track.object_track_id
+
+    # Verify persisted in database.
+    fetched = annotation_resolver.get_by_id(session=test_db, annotation_id=annotation.sample_id)
+    assert fetched is not None
+    assert fetched.object_track_id == track.object_track_id
```
michal-lightly: Optional: Also here I would suggest and explicit comparison: `== 1`. I would choose a number that would stand out more (i.e. that it is clear it is a test number), e.g. 99.

lightly_studio/tests/resolvers/object_track_resolver/test_create_object_track.py
```diff
@@ -0,0 +1,45 @@
...
+        object_track_number=1,
+        dataset_id=collection.collection_id,
+    )
+
+    assert track.object_track_id is not None
+    assert track.object_track_number == 1
+    assert track.dataset_id == collection.collection_id
+
+
+def test_create_object_track__multiple_tracks(test_db: Session) -> None:
```
michal-lightly: I do not understand the value of this test. What is it supposed to test?

It seems to me the same would be achieved by appending this to the test above, and it would be more readable:

```py
    track2 = object_track_resolver.create_object_track(
        session=test_db,
        object_track_number=2,
        dataset_id=collection.collection_id,
    )

    assert track2.object_track_id is not None
    assert track2.object_track_number == 2
    assert track2.dataset_id == collection.collection_id
```

lightly_studio/tests/resolvers/object_track_resolver/test_create_object_track.py
```diff
@@ -0,0 +1,45 @@
...
+    """Test creating a track and verifying returned fields."""
+    collection = create_collection(session=test_db)
+
+    track = object_track_resolver.create_object_track(
+        session=test_db,
+        object_track_number=1,
+        dataset_id=collection.collection_id,
+    )
+
+    assert track.object_track_id is not None
```
michal-lightly: I believe `object_track_id` is non-optional, so the assert can be deleted.

michal-lightly (changes requested):

lightly_studio/src/lightly_studio/resolvers/object_track_resolver/add_annotation_to_object_track.py
horatiualmasan: Yes, we can remove it. It was in the old annotation label update code. I removed it form there also.

lightly_studio/src/lightly_studio/resolvers/object_track_resolver/create_object_track.py
horatiualmasan: Yes, updated to use create_many instead of create_object_track

lightly_studio/tests/resolvers/object_track_resolver/test_add_annotation_to_object_track.py
horatiualmasan: here I compared the ids but added also a comparison to the annotation object track values to check if the relationship is ok

michal-lightly: Ah I see :)

lightly_studio/tests/resolvers/collection_resolver/test_deep_copy.py
```diff
@@ -571,6 +583,12 @@ def test_deep_copy__with_annotations(test_db: Session) -> None:
     assert od_detail.y == 20
     assert od_detail.width == 30
     assert od_detail.height == 40
+    assert copied_od.object_track_id is not None
+    assert copied_od.object_track_id != original_track_id
+    copied_track = test_db.get(ObjectTrackTable, copied_od.object_track_id)
```
michal-lightly: Optional: Now you can use `get_by_id`

PR #604, opened 2026-02-23
Create Object Tracks when loading video annotations

horatiualmasan:
## What has changed and why?

Create object tracks using the `object_track_id` from the labelformat models

## How has it been tested?

updated test

- [ ] Yes

### DIFF
--- a/lightly_studio/tests/core/video/test_add_videos.py
+++ b/lightly_studio/tests/core/video/test_add_videos.py
@@ -635,9 +635,8 @@ def test_process_video_annotations_instance_segmentation() -> None:
             ],
         ],
     )
-    # Create object track map with one track for each object
+    # Create object track map with one track for the second object and no track for the first object
     object_track_map = {
-        0: uuid4(),
         1: uuid4(),
     }
 
@@ -655,12 +654,12 @@ def test_process_video_annotations_instance_segmentation() -> None:
     assert annotations[0].segmentation_mask is None
     assert annotations[0].annotation_label_id == label_map[0]
     assert annotations[0].parent_sample_id == frame_number_to_id[0]
-    assert annotations[0].object_track_id == object_track_map[0]
+    assert annotations[0].object_track_id is None  # No track for first object
     assert annotations[1].annotation_type == "instance_segmentation"
     assert annotations[1].segmentation_mask is None
     assert annotations[1].annotation_label_id == label_map[0]
     assert annotations[1].parent_sample_id == frame_number_to_id[1]
-    assert annotations[1].object_track_id == object_track_map[0]
+    assert annotations[1].object_track_id is None  # No track for first object
     assert annotations[2].annotation_type == "instance_segmentation"
     assert annotations[2].segmentation_mask is None
     assert annotations[2].annotation_label_id == label_map[1]


lightly_studio/src/lightly_studio/core/video/add_videos.py
```diff
@@ -474,43 +484,107 @@ def _resolve_video_paths_from_labelformat(
...
+    """Create an ObjectTrackTable entry for each tracked object in the video.
+
+    Args:
+        session: Database session.
+        video_annotation: The labelformat video annotation containing objects.
+        dataset_id: UUID of the root collection (dataset).
+
+    Returns:
+        Mapping from object index (position in video_annotation.objects) to the
+        UUID of the created object track.
```
michal-lightly: This sounds like the return type should be a `list[UUID]` instead, WDYT?

lightly_studio/src/lightly_studio/core/video/add_videos.py
```diff
@@ -474,43 +484,107 @@ def _resolve_video_paths_from_labelformat(
...
+    """
+    object_track_map: dict[int, UUID] = {}
+    tracks_to_create: list[ObjectTrackCreate] = []
+    object_indices: list[int] = []
+    for obj_idx, obj in enumerate(video_annotation.objects):
+        object_track_number = getattr(obj, "object_track_id", None)
```
michal-lightly: Why do we need `getattr`? We should avoid it, we don't get the type check. We know the type of `video_annotation.objects`, it is always SingleObjectDetectionTrack or SingleInstanceSegmentationTrack and they both have `object_track_id`.

lightly_studio/src/lightly_studio/core/video/add_videos.py
```diff
@@ -474,43 +484,107 @@ def _resolve_video_paths_from_labelformat(
...
+        for obj_idx, obj in enumerate(video_annotation.objects):
             segmentation = obj.segmentations[frame_number]
             if segmentation is not None:
-                annotations.append(
-                    labelformat_helpers.get_segmentation_annotation_create(
-                        parent_sample_id=frame_id,
-                        annotation_label_id=label_map[obj.category.id],
-                        segmentation=segmentation,
-                    )
+                annotation = labelformat_helpers.get_segmentation_annotation_create(
```
michal-lightly, line 565: `object_track_id` should be added to the helper instead

michal-lightly (changes requested):
Thanks, minor comments

lightly_studio/src/lightly_studio/core/video/add_videos.py
michal-lightly: Which also means that the helper test must be slightly extended

lightly_studio/src/lightly_studio/core/video/add_videos.py
horatiualmasan: Some objects might not have the object id set, in that case a track is not created for them. It seems easier to manage this way. Updated description to make it more clear.

lightly_studio/src/lightly_studio/core/video/add_videos.py
```diff
@@ -474,43 +486,109 @@ def _resolve_video_paths_from_labelformat(
...
-                    labelformat_helpers.get_object_detection_annotation_create(
-                        parent_sample_id=frame_id,
-                        annotation_label_id=label_map[obj.category.id],
-                        box=box,
-                    )
+                annotation = labelformat_helpers.get_object_detection_annotation_create(
+                    parent_sample_id=frame_id,
+                    annotation_label_id=label_map[obj.category.id],
+                    box=box,
+                    object_track_id=object_track_map.get(obj_idx),
```
michal-lightly, line 591: Double checking, do we have a test where no track is specified and this will evaluate to None?

horatiualmasan: Updated one test to handle this case.

PR #606, opened 2026-02-23
Numeric Metadata Slider: Resolve UX and Performance Issues

MalteEbner:
### Problem
The numeric Metadata slider uses a fixed tick size of 0.01. This causes many problems:
- UX: The maximum is set to `min + 0.01*n`, which will truncate the true maximum to the next lower point with a natural number of ticks, which depends on the minimum. This will remove all samples larger than the `min+n*0.01`. E.g. if min=0.001 and max=1, then it will set the max to 0.99, from which it cannot be changed back.
- UX: For small floats, e.g. only in [0, 0.02], the tick size is far too large, the number of ticks is too small.
- Performance: For large floats, e.g. in [0, 1e6], it will generate 1e8 ticks, making the UI very slow.

The performance problem also applies to large integers, thus also introduced a fixed number of ticks for them. 

### Solution

- Use a fixed number of steps (1000 atm)
- Set a virtual maximum at the grid step above the real maximum to prevent the truncation.

Other: Delete unused `MetadataFilter/index.svelte`

## How has it been tested?

- Updated `example_metadata.py` with one large and one small float and int each.

image

- [x] Yes


lightly_studio_view/src/lib/components/CombinedMetadataDimensionsFilters/index.svelte
```diff
@@ -146,10 +168,14 @@
...
-                            step={isInteger ? 1 : 0.01}
-                            value={[value.min, value.max]}
-                            onValueCommit={handleChangeMetadata(metadataKey)}
+                            max={sliderMax}
+                            step={sliderStep}
+                            value={[value.min, value.max >= bound.max ? sliderMax : value.max]}
+                            onValueCommit={(newValues) =>
+                                handleChangeMetadata(metadataKey)([
+                                    Math.min(newValues[0], bound.max),
+                                    Math.min(newValues[1], bound.max)
```
MalteEbner: Clamp to bound.max

lightly_studio_view/src/lib/components/CombinedMetadataDimensionsFilters/index.svelte
```diff
@@ -134,6 +152,10 @@
                 {@const bound = $metadataBounds[metadataKey]}
                 {@const value = $metadataValues[metadataKey]}
                 {@const isInteger = Number.isInteger(bound.min) && Number.isInteger(bound.max)}
+                {@const sliderStep = getSliderStep(bound.min, bound.max, isInteger)}
+                {@const sliderMax = isMaxOnStepGrid(bound.min, bound.max, sliderStep)
+                    ? bound.max
+                    : bound.max + sliderStep}
```
MalteEbner: Required to remove the truncation

lightly_studio_view/src/lib/components/CombinedMetadataDimensionsFilters/index.svelte
```diff
@@ -79,6 +81,22 @@
...
+    const getSliderStep = (min: number, max: number, isInteger: boolean): number => {
+        const step = (max - min) / METADATA_SLIDER_TICKS;
+        if (step <= 0) {
+            return isInteger ? 1 : 1 / METADATA_SLIDER_TICKS;
+        }
+        if (isInteger) {
+            return Math.max(1, Math.round(step));
+        }
+        return step;
+    };
```
MalteEbner: For getting a fixed number of ticks, regardless of range.

lightly_studio_view/src/lib/components/CombinedMetadataDimensionsFilters/index.svelte
```diff
@@ -146,10 +168,14 @@
                             type="multiple"
                             class="filter-{metadataKey}"
                             min={bound.min}
-                            max={bound.max}
-                            step={isInteger ? 1 : 0.01}
-                            value={[value.min, value.max]}
-                            onValueCommit={handleChangeMetadata(metadataKey)}
+                            max={sliderMax}
+                            step={sliderStep}
+                            value={[value.min, value.max >= bound.max ? sliderMax : value.max]}
```
MalteEbner: When setting to the "virtual max" that is >= the actual max but on the grid, still show it as being the actual max.

lightly_studio_view/src/lib/components/CombinedMetadataDimensionsFilters/CombinedMetadataDimensionsFilters.svelte
```diff
@@ -9,6 +9,8 @@
     import VideoFieldBoundsFilters from '../VideoFieldBoundsFilters/VideoFieldBoundsFilters.svelte';
```
ikondrat, line 12: this component filename  conflicts with convention
https://github.com/lightly-ai/lightly-studio/blob/main/docs/coding-guidelines/frontend.md#naming-conventions

It should be 
CombinedMetadataDimensionsFilters/CombinedMetadataDimensionsFilters.svelte‎

lightly_studio_view/src/lib/components/CombinedMetadataDimensionsFilters/index.svelte
```diff
@@ -79,6 +81,22 @@
         }
         return formatFloat(value);
     };
+
+    const getSliderStep = (min: number, max: number, isInteger: boolean): number => {
```
ikondrat: I would add test for this logic within component

maybe it would be easier to move out function and write test for it

lightly_studio_view/src/lib/components/CombinedMetadataDimensionsFilters/index.svelte
ikondrat: It is not really nice practice when you define so many context variables in the tempate syntax.
it would be hard to debug if smth will not go well

I would propose to create separate component, maybe MetadataFilterItem

and split the logic to separate component
e.g.
```
{#each numericalMetadata as metadataKey}
     <MetadatFilerItem bound={$metadataBounds[metadataKey]} value={$metadataValues[metadataKey]}

File: lightly_studio_view/src/lib/components/CombinedMetadataDimensionsFilters/index.svelte (hunk 1 above)
Comment (ikondrat) [reply lightly_studio_view/src/lib/components/CombinedMetadataDimensionsFilters/index.svelte]: I would move it inside **script** section like
```
const handleValueCommit = (newValues) =>
```
...
and use 
```
onValueCommit={handleValueCommit} 
```
in the template syntax
It would be eaisier to debug and read and keeps representation part (html like syntax) cleaner

PR #618, opened 2026-02-25
Embedding caching hash optimizations

horatiualmasan:
## What has changed and why?

Embedding caching hash optimizations

## How has it been tested?

manual performance tests:
dataset with 50k images
old implementation:
old_get_2_dim
new implementation:
new_2_dim
First call takes longer as it calculates the embeddings, the following calls are cache hits and there is a noticeable improvement. 

- [ ] Yes

### DIFF
--- a/lightly_studio_view/src/lib/components/Grid/index.ts
+++ b/lightly_studio_view/src/lib/components/Grid/index.ts
@@ -1,0 +1,1 @@
+export { default as Grid } from './Grid.svelte';
--- a/lightly_studio_view/src/lib/components/GridItem/index.ts
+++ b/lightly_studio_view/src/lib/components/GridItem/index.ts
@@ -1,0 +1,1 @@
+export { default as GridItem } from './GridItem.svelte';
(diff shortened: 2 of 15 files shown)


lightly_studio/src/lightly_studio/resolvers/sample_embedding_resolver.py
MalteEbner: As this function is then not used anymore: Can you delete it and replace it with `get_hash_by_collection_id`? Also try to keep both at the same position, so that git gets that the diff is small

lightly_studio/src/lightly_studio/resolvers/sample_embedding_resolver.py
```diff
@@ -92,50 +92,48 @@ def get_all_by_collection_id(
...
-
+    sample_ids: list[UUID] = []
     hasher = hashlib.sha256()
-    hasher.update("".join(str(v) for v in first_dims_ordered).encode("utf-8"))
-    return hasher.hexdigest(), sample_ids_of_samples_with_embeddings
+    update = hasher.update
+
+    for row in rows:
+        sample_ids.append(row.sample_id)  # type: ignore[attr-defined]
+        update(str(row.first_dim).encode("utf-8"))  # type: ignore[attr-defined]
```
MalteEbner: Nitpick
```suggestion

    for row in rows:
        sample_ids.append(row.sample_id)  # type: ignore[attr-defined]
        hasher.update(str(row.first_dim).encode("utf-8"))  # type: ignore[attr-defined]
```

PR #619, opened 2026-02-25
Operator Context: operatorscope executioncontext

JonasWurst:
## What has changed and why?

Introducing the operatorscope and the executioncontext, as they are required to introduce the operators that operate on a certain context.

## How has it been tested?

test added.

- [ ] Yes

### DIFF
--- a/lightly_studio_view/src/lib/hooks/useFramesFilter/frameFilter.test.ts
+++ b/lightly_studio_view/src/lib/hooks/useFramesFilter/frameFilter.test.ts
@@ -1,0 +1,77 @@
+import { describe, it, expect, vi } from 'vitest';
+import { getFrameFilter } from './frameFilter';
+
+vi.mock('../useMetadataFilters/useMetadataFilters', () => ({
+    createMetadataFilters: vi.fn(() => [])
+}));
+
+describe('getFrameFilter', () => {
+    it('returns null when params are null', () => {
+        expect(getFrameFilter(null)).toBeNull();
+    });
+
+    it('returns null when collection_id is missing', () => {
+        expect(
+            getFrameFilter({
+                collection_id: ''
+            })
+        ).toBeNull();
+    });
+
+    it('builds minimal filter with collection_id', () => {
+        expect(
+            getFrameFilter({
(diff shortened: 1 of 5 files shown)


lightly_studio/src/lightly_studio/plugins/operator_context.py
```diff
@@ -0,0 +1,62 @@
...
+@dataclass
+class ExecutionContext:
+    """Contextual data passed to ``BaseOperator.execute()``.
+
+    ``filter=None`` means the whole collection, while a non-null filter
+    restricts the target samples.
+    """
+
+    collection_id: UUID
+    filter: ImageFilter | VideoFilter | None = None
```
horatiualmasan: As below we provide multiple scopes would it make sense to also pass all valid filters ? 
`ImageFilter | VideoFilter | VideoFrameFilter | AnnotationsFilter`

lightly_studio/src/lightly_studio/plugins/operator_context.py
```diff
@@ -0,0 +1,62 @@
...
+
+    Operators declare which scopes they support via ``BaseOperator.supported_scopes``.
+    The UI uses this to surface operators contextually by media type.
+    """
+
+    ROOT = "root"
+    """Operate on the root collection (dataset-level)."""
+
+    IMAGE = "image"
+    """Operate on images and video frames."""
```
horatiualmasan: Is this accurate? the same description for IMAGE and VIDEO_FRAME

lightly_studio/src/lightly_studio/plugins/operator_context.py
```diff
@@ -0,0 +1,62 @@
...
+
+    GROUP = "group"
+    """Operate on groups."""
+
+    CAPTION = "caption"
+    """Operate on captions."""
+
+
+def get_allowed_scopes_for_collection(
+    *, sample_type: SampleType, is_root_collection: bool
```
horatiualmasan: Optional: We can drop the star. It enforces the keyword arguments but we use them everywhere anyway.

PR #620, opened 2026-02-25
Properly cache CI e2e jobs

michal-lightly:
## What has changed and why?

Fixes an issue with e2e jobs caching.

The two e2e jobs - `end-to-end-general` and `end-to-end-videos` - were skipped if there were pre-built caches from `prepare-caches` step. This is wrong, it allows for this scenario:

1. prepare-caches succeeds, end-to-end-general is cancelled by the user -> CI failed
2. users retriggers the CI, prepare-caches is cached, end-to-end-general is skipped -> CI succeeds

The solution is to cache success of the steps that run the tests.

## How has it been tested?

Replicated the failure scenario above, e2e tests are no longer skipped.

- [ ] Yes


.github/workflows/end2end_test.yml
```diff
@@ -46,20 +63,7 @@ jobs:
...
-                'lightly_studio/src/**/*',
-                'lightly_studio/pyproject.toml',
-                'lightly_studio_view/e2e/**/*',
-                'lightly_studio_view/src/**/*',
-                'lightly_studio_view/public/**/*',
-                'lightly_studio_view/package.json',
-                'lightly_studio_view/package-lock.json'
-              )
-            }}
+          key: &cache-build-key python-build-${{ steps.file-hash.outputs.hash }}
```
mihnea-necsulescu: this doesn't seem like it's properly computed -> https://github.com/lightly-ai/lightly-studio/actions/runs/22397365416/job/64834745365?pr=620#step:4:25

I think it's because `steps.file-hash` is part of a different job (prepare-caches)?

michal-lightly: You're right, the "yaml anchor" syntax just reuses the same definition, i.e. `python-build-${{ steps.file-hash.outputs.hash }}` which references something that does not exist in the `end-to-end-general` job.

mihnea-necsulescu (approved):

PR #623, opened 2026-02-25
Introduce helper to compute navigation path

michal-lightly:
## What has changed and why?

Add a helper function to find a path from root (dataset id) to a leaf (collection id) in the collection hierarchy. It will be used to compose the navigation menu.

## How has it been tested?

New unit test.

- [ ] Yes

### DIFF
--- a/lightly_studio/src/lightly_studio/resolvers/annotation_resolver/create_many.py
+++ b/lightly_studio/src/lightly_studio/resolvers/annotation_resolver/create_many.py
@@ -71,6 +71,7 @@ def create_many(
             annotation_type=annotation_create.annotation_type,
             confidence=annotation_create.confidence,
             parent_sample_id=annotation_create.parent_sample_id,
+            object_track_id=annotation_create.object_track_id,
         )
 
         # Set other relationship details to None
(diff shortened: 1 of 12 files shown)


lightly_studio_view/src/lib/components/NavigationMenu/utils.ts
```diff
@@ -0,0 +1,24 @@
+import type { CollectionView } from '$lib/api/lightly_studio_local';
+
+/**
+ * Finds the path from root to the collection with the given targetId using DFS.
+ * Returns an array of ancestors [root, child, ..., target] or null if not found.
+ */
+export function findAncestorPath(root: CollectionView, targetId: string): CollectionView[] | null {
```
ikondrat, line 7: I am not sure how you are going to use it to render as a menu?

I have in mind this structure
```
<Menu
    ..
    subCollections=[
       collectionWithAnnotations,
       collectionWithVideos
    ]
```
I would call it diffrently then

smth like 
getSubCollectionsByCollectionId

michal-lightly: Discussed on Slack.

PR #626, opened 2026-02-26
Introduce frame filter hook

LeonardoRosaa:
## What has changed and why?

Introduced the frame filter hook to store frame filter parameters and to follow the pattern we have for videos and images.

## How has it been tested?

Unit tests.

- [ ] Yes


lightly_studio_view/src/lib/hooks/useFramesFilter/useFramesFilter.ts
```diff
@@ -0,0 +1,105 @@
...
+        sample_ids?: string[];
+        metadata_values?: MetadataValues;
+    };
+    frame_bounds?: VideoFrameFieldsBoundsView | null;
+    video_id?: string | null;
+};
+
+const filterParams = writable<VideoFrameFilterParams | null>(null);
+
+const frameFilter = derived(filterParams, ($filterParams): VideoFrameFilter | null => {
```
ikondrat: this needs to be separated to the dedicated function

smth like 
frameFilter = new Filter()
or

getFrameFilter()

with a proper test

move it out from here please

lightly_studio_view/src/lib/hooks/useFramesFilter/useFramesFilter.ts
```diff
@@ -0,0 +1,105 @@
...
+                ...params.filters,
+                sample_ids: sampleIds.length > 0 ? sampleIds : undefined
+            }
+        };
+        filterParams.set(newParams);
+    };
+
+    return {
+        filterParams,
+        frameFilter,
```
ikondrat, line 34: here we should have smth frameFilter: getFrameFilter(...)

lightly_studio_view/src/lib/hooks/useFramesFilter/useFramesFilter.test.ts
```diff
@@ -0,0 +1,207 @@
...
+}));
+
+describe('useFramesFilter', () => {
+    beforeEach(() => {
+        vi.clearAllMocks();
+        const { updateFilterParams } = useFramesFilter();
+        updateFilterParams(null);
+    });
+
+    describe('updateFilterParams and frameFilter', () => {
```
ikondrat: I would rather make separate groups for describing  frameFilter and updateFilterParams

lightly_studio_view/src/lib/hooks/useFramesFilter/useFramesFilter.ts
```diff
@@ -0,0 +1,105 @@
...
+    } else {
+        filters.frame_number = {};
+    }
+
+    const sampleFilter: SampleFilter = {};
+    sampleFilter.collection_id = $filterParams.collection_id;
+
+    const sampleIds = $filterParams.filters?.sample_ids;
+    if (sampleIds && sampleIds.length > 0) {
+        sampleFilter.sample_ids = sampleIds;
```
ikondrat: when you don't change values 

I would rather use existing normalisation tools, e.g. lodash

```
import _ from 'lodash';

const result = _.mapKeys(obj, (value, key) => _.camelCase(key));
```

LeonardoRosaa: I’d prefer to leave it as it is, also because it’s the same design pattern as in the other filters.

PR #628, opened 2026-02-26
Add video filter to video frame adjacent

LeonardoRosaa:
## What has changed and why?

Added a `VideoFrameAdjacentFilter` class that accepts filters using both the video filter and video text embeddings.

## How has it been tested?

Unit tests.

- [ ] Yes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,22 +9,39 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- Add “View Video” button in the frame details view to open video details for the selected frame.
-- Added `Escape` shortcut support in the embedding plot to clear the current selection.
-- Loading videos with annotations from youtube-vis format via `dataset.add_videos_from_youtube_vis`.
-- Added Shift+click range selection in grid views.
-- Display semantic segmentation within the GUI.
+### Changed
+
+### Deprecated
+
+### Removed
+
+### Fixed
+
+- Fixed spurious "Annotation updated successfully" toast appearing when clicking a bounding box without moving or resizing it.
+
+### Security
+
+## \[0.4.9\] - 2026-02-27
+
+### Added
+
+- Semantic segmentation labeling support:
+    - Load semantic segmentation from Pascal VOC format
+    - Visualize and label semantic segmentation in the GUI
+- Object tracking support:
+    - Loading videos with object track annotations from youtube-vis format via `dataset.add_videos_from_youtube_vis`.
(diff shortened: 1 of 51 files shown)


lightly_studio/src/lightly_studio/resolvers/video_frame_resolver/get_adjacent_video_frames.py
```diff
@@ -2,32 +2,78 @@
...
+
+        if video_collection_id is None:
+            raise ValueError("Collection ID must be provided in video_filter.sample_filter")
+
+    if filters.video_text_embedding and video_collection_id is None:
+        raise ValueError("Collection ID must be resolvable when video_text_embedding is provided.")
+
+    embedding_model_id, distance_expr = similarity_utils.get_distance_expression(
+        session=session,
+        collection_id=video_collection_id or collection_id,
```
horatiualmasan: i would only keep video_collection_id, the embedding model is used for videos only

lightly_studio/tests/resolvers/video/video_frame_resolver/test_get_adjacent_video_frames.py
```diff
@@ -151,6 +195,267 @@ def test_get_adjacent_video_frames__respects_annotation_filter(test_db: Session)
...
+        tag_name="keep",
+    )
+
+    tag_resolver.add_sample_ids_to_tag_id(
+        session=test_db, tag_id=tag.tag_id, sample_ids=[video_b.video_sample_id]
+    )
+
+    result = video_frame_resolver.get_adjacent_video_frames(
+        session=test_db,
+        sample_id=video_b.frame_sample_ids[-1],
```
horatiualmasan: I would use here the first frame, then prev should be in video_a frames but those are filtered out.

lightly_studio/tests/resolvers/video/video_frame_resolver/test_get_adjacent_video_frames.py
```diff
@@ -151,6 +195,267 @@ def test_get_adjacent_video_frames__respects_annotation_filter(test_db: Session)
...
+            helpers_resolvers.AnnotationDetails(
+                sample_id=video_b.frame_sample_ids[1],
+                annotation_label_id=label.annotation_label_id,
+            ),
+        ],
+    )
+
+    result = video_frame_resolver.get_adjacent_video_frames(
+        session=test_db,
+        sample_id=video_b.frame_sample_ids[1],
```
horatiualmasan: Same here,I would use here the first frame, then prev should be in video_a frames but those are filtered out.

PR #630, opened 2026-02-26
Update label for all annotations in a object track

horatiualmasan:
## What has changed and why?

Update label for all annotations in a object track

## How has it been tested?

New test and manual tests

- [ ] Yes


lightly_studio/src/lightly_studio/services/annotations_service/update_annotations.py
```diff
@@ -20,10 +23,40 @@ def update_annotations(
...
+
     Returns:
         List of updated annotations.
     """
-    return [
-        annotations_service.update_annotation(session, annotation_update)
-        for annotation_update in annotation_updates
-    ]
+    results: list[AnnotationBaseTable] = []
+    track_target_label_ids: dict[UUID, UUID] = {}
```
michal-lightly: The meaning of this variable is unclear. Please rename it as `x_to_y`, e.g. `track_id_to_label_id`.

lightly_studio/src/lightly_studio/services/annotations_service/update_annotations.py
```diff
@@ -20,10 +23,40 @@ def update_annotations(
         session: Database session for executing the operation.
         annotation_updates: List of objects containing updates for the annotations.
 
+    Info:
```
michal-lightly: 1) We do not use info section. We simply put text before args.
2) What happens if there are conflicting annotation updates?

lightly_studio/src/lightly_studio/services/annotations_service/update_annotations.py
```diff
@@ -20,10 +23,40 @@ def update_annotations(
...
+            # Overwriting the value reflects the last input update for that track
+            track_target_label_ids[result.object_track_id] = result.annotation_label_id
+
+    # Update the label for all annotations in the track if needed
+    for object_track_id, annotation_label_id in track_target_label_ids.items():
+        siblings = annotation_resolver.get_all_by_object_track_id(
+            session=session,
+            object_track_id=object_track_id,
+        )
+        for sibling in siblings:
```
michal-lightly, line 56: Wouldn't this be a performance issue? We would do a commit potentially for every video frame. Have you rested the performance?

lightly_studio/tests/resolvers/annotations/test_get_all_by_object_track_id.py
```diff
@@ -0,0 +1,67 @@
...
+        annotation for annotation in annotations if annotation.object_track_id == track_a
+    ]
+    assert len(track_a_annotations) == 2
+    result = annotation_resolver.get_all_by_object_track_id(
+        session=test_db, object_track_id=track_a
+    )
+    assert len(result) == 2
+    ids = {annotation.sample_id for annotation in result}
+    assert track_a_annotations[0].sample_id in ids
+    assert track_a_annotations[1].sample_id in ids
```
michal-lightly: Minor: The for loop is an unnecessary complication:

```suggestion
    result = annotation_resolver.get_all_by_object_track_id(
        session=test_db, object_track_id=track_a
    )
    assert len(result) == 2
    assert {annotation.sample_id for annotation in result} == {annotations[0].sample_id, annotations[1].sample_id}
```

lightly_studio/tests/services/annotations_service/test_update_annotations.py
```diff
@@ -0,0 +1,96 @@
...
+                object_track_id=track_b,
+            ),
+        ],
+    )
+    track_a_annotations = [
+        annotation for annotation in annotations if annotation.object_track_id == track_a
+    ]
+    track_b_annotations = [
+        annotation for annotation in annotations if annotation.object_track_id == track_b
+    ]
```
michal-lightly, line 71: Also in this test, this should not be needed - we know the ids exactly.

lightly_studio/src/lightly_studio/services/annotations_service/update_annotations.py
horatiualmasan: If multiple updates for annotations in the same track are provided, the last update in the list determines the final label for all annotations in that track. Updated the description.

lightly_studio/tests/resolvers/annotations/test_get_all_by_object_track_id.py
horatiualmasan: `track_a_annotations` is still needed. For me on windows at least annotations returned by `create_annotations` helper were not in the same order as the input (annotations[1] would have `object_track_id==track_b).`

michal-lightly: Where do we need the order? Don't we compare sets?

lightly_studio/src/lightly_studio/services/annotations_service/update_annotations.py
michal-lightly: Discussed offline, let's add a TODO

lightly_studio/tests/resolvers/annotations/test_get_all_by_object_track_id.py
horatiualmasan: Yes, but you do the assumption that `annotations[0].sample_id, annotations[1].sample_id` are the ones we need because the first 2 objects passed to `create_annotations` had `object_track_id=track_a`. This is not the case.

michal-lightly: I see, the problem is the implementation of `create_annotations` helper. That's a pity, we should fix it at some point (the underlying create_many implementation is fine).

Ok to keep as is.

michal-lightly (approved):

lightly_studio/tests/services/annotations_service/test_update_annotations.py
horatiualmasan: Same as above

lightly_studio/src/lightly_studio/services/annotations_service/update_annotations.py
horatiualmasan: added

PR #633, opened 2026-02-26
Use selectinload to avoid joining the same data many times

IgorSusmelj:
## What has changed and why?

- replaced `joinedload` with `selectinload` for tag relationships in image listing.
- eager loading for annotation sample tags to avoid per-annotation lazy loads
- according to [this blog](https://medium.com/@dresraceran/understanding-sqlalchemys-eager-loading-joinedload-selectinload-and-contains-eager-e12d98c8c8b0) we can use several queries with `selectinload` rather than 1 complex query with lots of joins with `joinedload` that results in lots of duplicate data.

## How has it been tested?

This should not be a functional change as we simply give a different hint to our ORM (sqlmodel) to write the SQL query differently. Was >3x faster for me locally (1.8s -> 300-400ms).

Tested manually on different datasets. All frontend queries worked fine.

- [X] Yes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Fixed right-click `Copy image` in grid and detail views to copy images from the GUI.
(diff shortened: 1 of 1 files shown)


MalteEbner (approved):
LGTM^2

PR #634, opened 2026-02-27
Context Plugin: update endpoint

JonasWurst:
## What has changed and why?

this change introduces the new context information to the anpiendpoint. it will be responsible to validate the provided context against the allowed scopes of a operator. Further, it will sanitze the provided filter/sample_id.

## How has it been tested?

tests are added.

- [ ] Yes

### DIFF
--- a/lightly_studio/src/lightly_studio/resolvers/group_resolver/__init__.py
+++ b/lightly_studio/src/lightly_studio/resolvers/group_resolver/__init__.py
@@ -3,6 +3,9 @@ from lightly_studio.resolvers.group_resolver.get_all import (
     get_all,
 )
 from lightly_studio.resolvers.group_resolver.get_by_id import get_by_id
+from lightly_studio.resolvers.group_resolver.get_collection_id_by_group_id import (
+    get_collection_id_by_group_id,
+)
 from lightly_studio.resolvers.group_resolver.get_group_component_with_type import (
     get_group_component_with_type,
 )
@@ -18,6 +21,7 @@ __all__ = [
     "create_many",
     "get_all",
     "get_by_id",
+    "get_collection_id_by_group_id",
     "get_group_component_with_type",
     "get_group_components_as_dict",
     "get_group_previews",
--- a/lightly_studio/src/lightly_studio/resolvers/group_resolver/get_collection_id_by_group_id.py
+++ b/lightly_studio/src/lightly_studio/resolvers/group_resolver/get_collection_id_by_group_id.py
@@ -1,0 +1,28 @@
+"""Get collection_id by group sample_id."""
+
+from __future__ import annotations
+
+from uuid import UUID
+
+from sqlmodel import Session, col, select
+
+from lightly_studio.models.group import GroupTable
+from lightly_studio.models.sample import SampleTable
+
+
+def get_collection_id_by_group_id(session: Session, group_id: UUID) -> UUID | None:
+    """Retrieve collection_id for a group by its group_id.
+
+    Args:
+        session: Database session for executing queries.
+        group_id: The group_id of the group.
+
+    Returns:
+        The collection_id of the group, or None if the group doesn't exist.
+    """
+    query = (
+        select(SampleTable.collection_id)
+        .join(GroupTable, col(GroupTable.sample_id) == col(SampleTable.sample_id))
+        .where(col(GroupTable.sample_id) == group_id)
+    )
+    return session.exec(query).one_or_none()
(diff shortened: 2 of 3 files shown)


lightly_studio/src/lightly_studio/api/routes/api/operator.py
```diff
@@ -70,11 +83,39 @@ def execute_operator(
             detail=f"Operator '{operator_id}' not found",
         )
 
-    context = ExecutionContext(collection_id=collection_id, filter=None)
+    context = request.context
+    if context is None:
+        context = OperatorContextRequest(collection_id=None, context_filter=None)
```
horatiualmasan: Optional: you could pass here collection_id and then use context.colletion_id below directly
```suggestion
        context = OperatorContextRequest(collection_id=colletion_id , context_filter=None)
```

lightly_studio/src/lightly_studio/api/routes/api/operator.py
```diff
@@ -8,20 +8,33 @@
...
+from lightly_studio.resolvers import collection_resolver
 
 operator_router = APIRouter(prefix="/operators", tags=["operators"])
 
 
+class OperatorContextRequest(BaseModel):
+    """Client-supplied execution context for scoped operator calls."""
+
+    collection_id: UUID | None = None
+    context_filter: AnyFilter = None
```
michal-lightly: Can you please add docstrings to explain the meaning? It should be in tripple-quotes under the defined term, so that it propagates to the api docs.

Some questions I have:
- Why are they both optional, and on top even `context` below can be optional? In which cases would we set each to None? Can we make some mandatory?
- `collection_id` is passed both here and as a param to `execute_operator`. What is the plan?

lightly_studio/tests/api/routes/api/test_operator.py
```diff
@@ -213,3 +299,61 @@ class EmptyParamsOperator(TestOperator):
...
+        return [
+            OperatorScope.ROOT,
+            OperatorScope.IMAGE,
+            OperatorScope.VIDEO,
+            OperatorScope.VIDEO_FRAME,
+        ]
+
+    def execute(
+        self,
+        *,
```
horatiualmasan: You can remove *

lightly_studio/src/lightly_studio/plugins/operator_context.py
```diff
@@ -4,15 +4,21 @@
...
 from lightly_studio.resolvers.annotations.annotations_filter import AnnotationsFilter
 from lightly_studio.resolvers.group_resolver.group_filter import GroupFilter
 from lightly_studio.resolvers.image_filter import ImageFilter
+from lightly_studio.resolvers.sample_resolver.sample_filter import SampleFilter
 from lightly_studio.resolvers.video_frame_resolver.video_frame_filter import VideoFrameFilter
 from lightly_studio.resolvers.video_resolver.video_filter import VideoFilter
 
+AnyFilter = Union[
+    ImageFilter, VideoFrameFilter, VideoFilter, AnnotationsFilter, GroupFilter, SampleFilter, None
+]
```
michal-lightly, line 20: Note that we already had a use for a union of filters for the sample adjacents endpoint. Wanted to raise it to both of you so you can see if it should be unified. Cc @horatiualmasan.

lightly_studio/src/lightly_studio/api/routes/api/operator.py
JonasWurst: The collection_id on the route will be removed in a follow up, making the collection_id in the context mandatory, hence also the collection will be mandatory.

lightly_studio/src/lightly_studio/api/routes/api/operator.py
```diff
@@ -85,15 +91,14 @@ def execute_operator(
 
     context = request.context
     if context is None:
-        context = OperatorContextRequest(collection_id=None, context_filter=None)
+        context = OperatorContextRequest(collection_id=collection_id, context_filter=None)
 
     # The context may specify a focused sub-collection; fall back to the route collection.
-    collection_id = context.collection_id or collection_id
     collection = collection_resolver.get_by_id(session=session, collection_id=collection_id)
```
horatiualmasan: You could already use here `context.collection_id`

lightly_studio/src/lightly_studio/plugins/operator_context.py
JonasWurst: There is no overlap anymore.

PR #650, opened 2026-03-02
feat: get information about collection_id by group_id

ikondrat:
## What has changed and why?

This PR introduces function to get information about collection_id for given group_id
(sample_id of the group)

We need this information to use in the function returning group view representation  by group_id
https://github.com/lightly-ai/lightly-studio/pull/651/changes#diff-247821752e632947650315aa4ba318ccc828f30e9170e906be7a4ddec212879c

## How has it been tested?

By unit test

- [ ] Yes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Settings dialog is now scrollable on small screens.
 - Added `✕` button to clear an active text search, matching the existing image search behaviour.
 - Fixed spurious "Annotation updated successfully" toast appearing when clicking a bounding box without moving or resizing it.
 
--- a/lightly_studio_view/src/lib/components/Settings/SettingsDialog.svelte
+++ b/lightly_studio_view/src/lib/components/Settings/SettingsDialog.svelte
@@ -137,7 +137,9 @@
 <Dialog.Root open={$isSettingsDialogOpen} onOpenChange={(isOpen) => setOpen(isOpen)}>
     <Dialog.Portal>
         <Dialog.Overlay />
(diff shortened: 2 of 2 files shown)


lightly_studio/tests/resolvers/group_resolver/test_get_collection_id_by_group_id.py
```diff
@@ -0,0 +1,105 @@
...
+    non_existent_id = uuid4()
+    collection_id = group_resolver.get_collection_id_by_group_id(
+        session=db_session, group_id=non_existent_id
+    )
+
+    # Verify it returns None
+    assert collection_id is None
+
+
+def test_get_collection_id_by_group__multiple_groups(db_session: Session) -> None:
```
horatiualmasan: This test is similar to the first one. We could append it there and keep only one.

lightly_studio/tests/resolvers/group_resolver/test_get_collection_id_by_group_id.py
```diff
@@ -0,0 +1,105 @@
+from uuid import uuid4
+
+from sqlmodel import Session
+
+from lightly_studio.models.collection import SampleType
+from lightly_studio.resolvers import collection_resolver, group_resolver
+from tests.helpers_resolvers import ImageStub, create_collection, create_images
+
+
+def test_get_collection_id_by_group(db_session: Session) -> None:
```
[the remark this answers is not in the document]
horatiualmasan: I would do this

lightly_studio/tests/resolvers/group_resolver/test_get_collection_id_by_group_id.py
ikondrat: I would keep it as more complicated test and keep the first one as a separate simple test. 
I think it is better if we document separately these cases

PR #655, opened 2026-03-02
feat: introduce GroupComponentView model

ikondrat:
## What has changed and why?

This PR introduces GroupComponentView to represent group component for the frontend

It will be part of the logic:
- fetching information about group components by the given group_id 
- fetching details about group components content by recieved sample ids from previous method (**this PR**)

## How has it been tested?

By unit test

- [ ] Yes

### DIFF
--- a/lightly_studio/src/lightly_studio/models/group.py
+++ b/lightly_studio/src/lightly_studio/models/group.py
@@ -78,8 +78,8 @@ class GroupComponentView(BaseModel):
     @classmethod
     def from_image_table(
         cls,
-        image: "ImageTable",
         collection: ComponentCollectionView,
+        image: "ImageTable",
         similarity_score: Optional[float] = None,
     ) -> "GroupComponentView":
         """Create a GroupComponentView from an ImageTable.
@@ -102,10 +102,10 @@ class GroupComponentView(BaseModel):
     @classmethod
     def from_video_table(
         cls,
-        video: "VideoTable",
         collection: ComponentCollectionView,
-        similarity_score: Optional[float] = None,
+        video: "VideoTable",
         frame: Optional["FrameView"] = None,
+        similarity_score: Optional[float] = None,
     ) -> "GroupComponentView":
         """Create a GroupComponentView from a VideoTable.
 


lightly_studio/src/lightly_studio/models/group.py
```diff
@@ -12,6 +12,10 @@
 from lightly_studio.models.sample import SampleTable, SampleView
 from lightly_studio.models.video import VideoView
 
+if TYPE_CHECKING:
```
michal-lightly: I'm afraid we have to remove the `TYPE_CHECKING` guard. Otherwise, the ImageTable and VideoTable model is not loaded during runtime, which means that the generated OpenAPI spec will miss the type.

Note that we have already removed all `TYPE_CHECKING` guards for all models, except for annotations.

lightly_studio/src/lightly_studio/models/group.py
```diff
@@ -52,3 +56,74 @@ class GroupViewsWithCount(BaseModel):
...
+
+class GroupComponentView(BaseModel):
+    """GroupComponentView representation.
+
+    Represents a group component with its name and associated media (image or video).
+    A component is always either an image or a video, never both.
+
+    A "GroupComponent" is a sample that has the following relationships:
+    - Collection relationship (samples.collection_id → collections.collection_id): The component
+      belongs to a collection/dataset.
```
michal-lightly:
```suggestion
    - Collection relationship (samples.collection_id → collections.collection_id): The component
      sample belongs to a component collection.
```

lightly_studio/src/lightly_studio/models/group.py
```diff
@@ -52,3 +56,74 @@ class GroupViewsWithCount(BaseModel):
...
+    - Group relationship (via SampleGroupLinkTable): The component is linked to a parent group
+      sample through the SampleGroupLinkTable join table, where the component is referenced
+      by sample_id and the parent group by parent_sample_id.
+    - Content relationship: Each sample's actual content (media file information) is stored in
+      either ImageTable or VideoTable, linked via sample_id as a foreign key. A sample_id exists
+      in SampleTable and exactly one of ImageTable/VideoTable - never both. ImageTable stores
+      image-specific data (file_name, file_path_abs, width, height), while VideoTable stores
+      video-specific data (file_name, file_path_abs, width, height, fps, duration_s).
```
michal-lightly: Minor: Unnecessary information

```suggestion
      in SampleTable and exactly one of ImageTable/VideoTable - never both. 
```

lightly_studio/src/lightly_studio/models/group.py
```diff
@@ -52,3 +56,74 @@ class GroupViewsWithCount(BaseModel):
...
+    """
+
+    component_name: str
+    image: Optional[ImageView] = None
+    video: Optional[VideoView] = None
+
+    @classmethod
+    def from_image_table(cls, image: "ImageTable", component_name: str) -> "GroupComponentView":
```
michal-lightly: We need to pass the optional parameters for `from_image_table` here. Also below for `from_video_table`

lightly_studio/src/lightly_studio/resolvers/group_resolver/get_group_component_details_by_ids.py
```diff
@@ -0,0 +1,65 @@
...
+from uuid import UUID
+
+from sqlmodel import Session, col, select
+
+from lightly_studio.models.collection import CollectionTable
+from lightly_studio.models.group import GroupComponentView
+from lightly_studio.resolvers import image_resolver, video_resolver
+
+
+def get_group_component_details_by_ids(
```
michal-lightly: Minor: Makes it clearer what is returned.
```suggestion
def get_group_component_views_by_ids(
```

lightly_studio/src/lightly_studio/resolvers/group_resolver/get_group_component_details_by_ids.py
```diff
@@ -0,0 +1,65 @@
...
+    session: Session, sample_ids: list[UUID]
+) -> list[GroupComponentView]:
+    """Get group component details for a list of sample IDs.
+
+    Args:
+        session: Database session.
+        sample_ids: List of GroupComponent IDs (primary keys from the samples table).
```
michal-lightly: This does not look correct. How is this function going to be used?

I think there are two possible ways it would make sense:
* Accept a list of GroupTable.sample_id and return all component views
* Accept a list of sample_ids of a components - then I would expect all to be of the same type, we can pass in collection_id and make the logic simpler.

lightly_studio/src/lightly_studio/models/group.py
```diff
@@ -52,3 +56,74 @@ class GroupViewsWithCount(BaseModel):
...
+      sample through the SampleGroupLinkTable join table, where the component is referenced
+      by sample_id and the parent group by parent_sample_id.
+    - Content relationship: Each sample's actual content (media file information) is stored in
+      either ImageTable or VideoTable, linked via sample_id as a foreign key. A sample_id exists
+      in SampleTable and exactly one of ImageTable/VideoTable - never both. ImageTable stores
+      image-specific data (file_name, file_path_abs, width, height), while VideoTable stores
+      video-specific data (file_name, file_path_abs, width, height, fps, duration_s).
+    """
+
+    component_name: str
```
michal-lightly: I am not sure about passing `str` here. I think there are two options that make more sense:
* Pass `collection_id` instead
* If we need the collection info "resolved", let's instead define a class. Sp that we can also pass information as e.g. `group_component_index`. My suggestion would be to add a class `GroupComponentCollectionView` to `models/collections.py`.

lightly_studio/src/lightly_studio/models/group.py
```diff
@@ -52,3 +56,74 @@ class GroupViewsWithCount(BaseModel):
...
+    image: Optional[ImageView] = None
+    video: Optional[VideoView] = None
```
michal-lightly: Shouldn't we rather use a Union? We do it already for filters.

lightly_studio/src/lightly_studio/models/group.py
ikondrat: ahh yes

lightly_studio/src/lightly_studio/resolvers/group_resolver/get_group_component_details_by_ids.py
ikondrat: Here is usage example

https://github.com/lightly-ai/lightly-studio/pull/656/changes
It will accept **a list of sample_ids of a components** 
We will get
group_resolver.get_group_components_by_group_id(session=session, group_id=group_id)

and fill it with details right in the endpoint

lightly_studio/src/lightly_studio/models/group.py
ikondrat: My motivation was - it would be easier for usage than union

to use union properly you would need to have typeguard e.g. 
```
isVideo(component.details)
```

When you have separate typed field it will require less logic resolve
```
if component.vidoe
     ... render video
```

I would avoid use union when it is possible and rather return optional filelds

lightly_studio/src/lightly_studio/models/group.py
michal-lightly: This was not addressed. Shouldn't we accept also `similarity_score`?

ikondrat: ohh missed this, thanks

updated
not sure we are going to use it right away but let's pass it

lightly_studio/src/lightly_studio/models/group.py
```diff
@@ -52,3 +53,85 @@ class GroupViewsWithCount(BaseModel):
...
+
+    collection: ComponentCollectionView
+    details: Union[ImageView, VideoView, None] = None
+
+    @classmethod
+    def from_image_table(
+        cls,
+        image: "ImageTable",
+        collection: ComponentCollectionView,
+        similarity_score: Optional[float] = None,
```
michal-lightly, line 83: Optional: It would be nice to match the order:
```py
        collection: ComponentCollectionView,
        image: "ImageTable",
        similarity_score: Optional[float] = None,
```

lightly_studio/src/lightly_studio/models/group.py
```diff
@@ -52,3 +53,85 @@ class GroupViewsWithCount(BaseModel):
...
+            details=ImageView.from_image_table(image=image, similarity_score=similarity_score),
+        )
+
+    @classmethod
+    def from_video_table(
+        cls,
+        video: "VideoTable",
+        collection: ComponentCollectionView,
+        similarity_score: Optional[float] = None,
+        frame: Optional["FrameView"] = None,
```
michal-lightly: Optional: It would be nice to match the order:
```py
        collection: ComponentCollectionView,
        video: "VideoTable",
        frame: Optional["FrameView"] = None,
        similarity_score: Optional[float] = None,
```

PR #656, opened 2026-03-02
feat: introduce endpoint to fetch group components by group id

ikondrat:
## What has changed and why?

To render the group components in the menu we need to have endpoint returning the list of the components for group details.
This PR introduces endpoint collecting the components belonging to the group and returning the details about this components. Including component name and image or video

It is based on changes in https://github.com/lightly-ai/lightly-studio/pull/672

## How has it been tested?

By unit tests

- [ ] Yes

### DIFF
--- a/lightly_studio/tests/api/routes/api/test_group.py
+++ b/lightly_studio/tests/api/routes/api/test_group.py
@@ -1,3 +1,5 @@
+import uuid
+
 from fastapi.testclient import TestClient
 from sqlmodel import Session
 
@@ -577,8 +579,6 @@ def test_get_group_components_by_group_id__nonexistent_group(
     test_client: TestClient,
 ) -> None:
     """Test GET group components endpoint with non-existent group ID."""
-    import uuid
-
     # Act
     non_existent_id = uuid.uuid4()
     response = test_client.get(f"/api/groups/{non_existent_id}/components")


lightly_studio/src/lightly_studio/api/routes/api/group.py
```diff
@@ -43,3 +44,16 @@ def get_all_groups(
...
     )
+
+
+@group_router.get("/groups/{group_id}/components", response_model=list[GroupComponentView])
+def get_group_components_by_group_id(
+    session: SessionDep,
+    group_id: Annotated[UUID, Path(title="Group Id")],
+) -> list[GroupComponentView]:
+    """Get all component views that belong to a group sample."""
+    samples = group_resolver.get_group_components_by_group_id(session=session, group_id=group_id)
```
horatiualmasan: For the embedding plot optimisations one big optimisation was not returning the sample objects and do a query just for the object IDs. It was visible in large datasets. If you think it could also happen here you can create a resolver function that will directly fetch the IDs.

ikondrat: I think here we rather need a sample

I think we shouldn't have a problem since it's a limited number of components per group

It should be on average 3 components

PR #659, opened 2026-03-02
Fix: Avoid unnecessary pop up for annotation deletion

IgorSusmelj:
## What has changed and why?

- Before this PR we always showed a pop up asking the user if they are sure whether they want to delete an annotation
- After this PR we only ask for views where we can't undo the deletion with the undo button or "ctrl-z"

Additionally:
- I made sure the undo feature also works for deletion of captions (was not possible before this PR)
- I modified E2E tests to cover the new functionality in all views

https://github.com/user-attachments/assets/fc928596-0383-4c5c-91a6-08c0555553cf

## How has it been tested?

- Updated unit and end2end tests
- Manual testing

- [X] Yes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,12 +11,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Restructured the navigation menu to follow the dataset structure. This makes the menu work with group datasets.
+
 ### Deprecated
 
 ### Removed
 
 ### Fixed
 
+- Settings dialog is now scrollable on small screens.
 - Added `✕` button to clear an active text search, matching the existing image search behaviour.
 - Fixed spurious "Annotation updated successfully" toast appearing when clicking a bounding box without moving or resizing it.
 - Improved deletion UX in sample details and captions list views: replaced confirmation popups with immediate deletion and Ctrl+Z undo support.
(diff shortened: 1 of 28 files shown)


lightly_studio_view/src/lib/components/SampleDetails/SampleDetailsCaptionsSegment/SampleDetailsCaptionSegment.svelte
```diff
@@ -27,9 +28,19 @@
...
+            addCaptionDeleteToUndoStack({
+                text: caption.text ?? '',
+                parentSampleId: sampleId,
+                addReversibleAction,
+                createCaption,
+                refetch
+            });
+            await deleteCaption(captionId);
             toast.success('Caption deleted successfully');
             refetch();
```
mihnea-necsulescu, line 45: this find -> add to stack -> delete -> toast -> refetch operation seems to be duplicated in `CaptionsItem`, too

maybe it would be worth extracting this to a hook?

IgorSusmelj: I'd not do this now. The logic for the undo is already extracted and it's just a few lines of code. We can look into a refactoring later down the road though!

PR #665, opened 2026-03-03
Operator Context: Add Example

JonasWurst:
## Description

An example has been added to demonstrate the use of various operators within a group dataset context. This includes operators for different collection types such as ROOT, IMAGE, VIDEO_FRAME, VIDEO, ANNOTATION, and GROUP.

## How has it been tested?

It will be required for follow up PRs and tested there.

- [ ] Yes

### DIFF
--- a/lightly_studio/src/lightly_studio/models/collection.py
+++ b/lightly_studio/src/lightly_studio/models/collection.py
@@ -51,6 +51,13 @@ class CollectionView(CollectionBase):
     children: list["CollectionView"] = []
 
 
+class ComponentCollectionView(CollectionBase):
+    """Collection view for group components."""
+
+    group_component_name: str
+    group_component_index: int
+
+
 class CollectionViewWithCount(CollectionView):
     """Collection view with total sample count."""
 
--- a/lightly_studio/src/lightly_studio/models/group.py
+++ b/lightly_studio/src/lightly_studio/models/group.py
@@ -8,9 +8,10 @@ from pydantic import Field as PydanticField
 from sqlalchemy.orm import Mapped
 from sqlmodel import Field, Relationship, SQLModel
 
-from lightly_studio.models.image import ImageView
+from lightly_studio.models.collection import ComponentCollectionView
+from lightly_studio.models.image import ImageTable, ImageView
 from lightly_studio.models.sample import SampleTable, SampleView
-from lightly_studio.models.video import VideoView
+from lightly_studio.models.video import FrameView, VideoTable, VideoView
 
 
 class GroupTable(SQLModel, table=True):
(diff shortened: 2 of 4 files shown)


lightly_studio/src/lightly_studio/examples/example_group_operators.py
```diff
@@ -0,0 +1,311 @@
...
+
+from environs import Env
+from sqlmodel import Session
+
+import lightly_studio as ls
+from lightly_studio import db_manager
+from lightly_studio.plugins.base_operator import BaseOperator, OperatorResult
+from lightly_studio.plugins.operator_context import ExecutionContext, OperatorScope
+from lightly_studio.plugins.operator_registry import operator_registry
+from lightly_studio.plugins.parameter import BaseParameter
```
michal-lightly, line 17: Food for thought: Should we make the imports easier by exposing some things from lightly_studio.plugins?

lightly_studio/src/lightly_studio/examples/example_group_operators.py
```diff
@@ -0,0 +1,311 @@
...
+from lightly_studio.resolvers import collection_resolver, sample_resolver
+from lightly_studio.resolvers.sample_resolver.sample_filter import SampleFilter
+
+
+@dataclass
+class DemoRootOperator(BaseOperator):
+    """Demo operator for the ROOT scope."""
```
michal-lightly, line 37: Minor: We should call the scope "DATASET", we are renaming the places that refer to "root" collection. WDYT?

lightly_studio/src/lightly_studio/examples/example_group_operators.py
```diff
@@ -0,0 +1,311 @@
...
+        return [OperatorScope.ROOT]
+
+    def execute(
+        self,
+        session: Session,
+        context: ExecutionContext,
+        parameters: dict[str, Any],
+    ) -> OperatorResult:
+        """Execute the operator."""
+        _ = parameters
```
michal-lightly: Delete

michal-lightly: Or what is the purpose of this? If it is for example, shouldn't we add a dummy string parameter ("Example str parameter")?

lightly_studio/src/lightly_studio/examples/example_group_operators.py
```diff
@@ -0,0 +1,311 @@
...
+        return OperatorResult(
+            success=True,
+            message=(
+                f"ANNOTATION executed on {context.collection_id} (type={sample_type}, n={count})."
+            ),
+        )
+
+
+@dataclass
+class DemoGroupOperator(BaseOperator):
```
michal-lightly, line 249: What is the point of adding an example for every type? Is it for testing?

lightly_studio/src/lightly_studio/examples/example_group_operators.py
```diff
@@ -0,0 +1,311 @@
...
+    ``context_filter`` may be deserialized as any filter subtype by Pydantic's
+    union matching (e.g. AnnotationsFilter instead of SampleFilter when both
+    carry ``sample_ids``).  Use ``getattr`` so the count works regardless of
+    which concrete type was selected.
+    """
+    collection = collection_resolver.get_by_id(
+        session=session, collection_id=context.collection_id
+    )
+    sample_type = collection.sample_type.value if collection else "unknown"
+    sample_ids = getattr(context.context_filter, "sample_ids", None)
```
michal-lightly: Do not use `getattr`. There are ways to tell the type system.

This implementation is very likely incorrect - the filter classes do not usually have `sample_ids` field.

What is the intent? The proper way is to define a protocol or an ABC that describes the common filter interface (i.e. "apply" and I'm not sure if there is more).

lightly_studio/src/lightly_studio/examples/example_group_operators.py
JonasWurst: Yes I agree, I will do it in a follow up.

lightly_studio/src/lightly_studio/examples/example_group_operators.py
JonasWurst: Yes, I will create a follow-up issue

lightly_studio/src/lightly_studio/examples/example_group_operators.py
JonasWurst: it is to silence mypy :)

PR #667, opened 2026-03-03
feat: introduce conversion function for VideoView

ikondrat:
## What has changed and why?

We need to have consistent view conversion for the groups. 
There is a new way to define conversion functions for views defined in https://github.com/lightly-ai/lightly-studio/pull/653

Let's make similar conversion for video to handle it simirlarly.

## How has it been tested?

By unit tests

- [ ] Yes


lightly_studio/src/lightly_studio/models/video.py
```diff
@@ -65,6 +65,35 @@ class VideoView(SQLModel):
     frame: Optional["FrameView"] = None
     similarity_score: Optional[float] = None
 
+    @classmethod
+    def from_video_table(
+        cls,
+        video: VideoTable,
+        frame: Optional["VideoFrameTable"] = None,
+        similarity_score: Optional[float] = None,
```
michal-lightly, line 73: Very optional: I would consider removing the default for these two. Then the caller must explicitly pass None, we avoid forgetting to pass them by mistake.

lightly_studio/src/lightly_studio/models/video.py
```diff
@@ -65,6 +65,35 @@ class VideoView(SQLModel):
...
+            )
+        return cls(
+            width=video.width,
+            height=video.height,
+            duration_s=video.duration_s,
+            fps=video.fps,
+            file_name=video.file_name,
+            file_path_abs=video.file_path_abs,
+            sample_id=video.sample_id,
+            sample=SampleView.model_validate(video.sample),
```
michal-lightly, line 92: This is suspicious, the function in `api/frame.py` used a helper `_build_sample_view` for conversion. Are you sure this will work? It seems so, but double-checking.

We could also add a todo and implement `SampleView.from_sample_table` later on.

lightly_studio/src/lightly_studio/api/routes/api/frame.py
michal-lightly, line 181: Please delete the `build_frame_view` function, it is not used anywhere anymore.

lightly_studio/tests/models/test_video.py
```diff
@@ -0,0 +1,101 @@
...
+        video_view = VideoView.from_video_table(
+            video=video_table, frame=None, similarity_score=similarity_score
+        )
+
+        assert video_view.similarity_score == 0.95
+        assert video_view.frame is None
+
+    def test_from_video_table__with_frame(self, db_session: Session) -> None:
+        """Test VideoView.from_video_table with VideoFrameTable conversion."""
+        from lightly_studio.resolvers import video_frame_resolver
```
michal-lightly, line 62: Imports must be at the top

lightly_studio/tests/models/test_video.py
```diff
@@ -0,0 +1,101 @@
...
+        video_table = video_resolver.get_by_id(
+            session=db_session, sample_id=video_with_frames.video_sample_id
+        )
+        assert video_table is not None
+
+        frame_table = video_frame_resolver.get_by_id(
+            session=db_session, sample_id=video_with_frames.frame_sample_ids[0]
+        )
+        assert frame_table is not None
+
```
michal-lightly, line 91: Remove. We are only using `frame_table.sample_id`. We already have it, it is `video_with_frames.frame_sample_ids[0]`.

michal-lightly (changes requested):

lightly_studio/src/lightly_studio/models/video.py
michal-lightly: Actually lets do this - it would prevent a bug it the next PR.

lightly_studio/src/lightly_studio/models/video.py
ikondrat: works according to tests - I was also surprised 🤔  

I will check as well visually

ikondrat: I think it requires a bit more solid approach to refactor all of these build functions properly 

With refactoring all other conversion functions as well

PR #669, opened 2026-03-03
Fix: Update readme and docs to use simplified VideoDataset import

IgorSusmelj:
## What has changed and why?

- Simplified the readme and docs example about using VideoDataset

## How has it been tested?

- No need for new tests

- [X] Yes


CHANGELOG.md
JonasWurst, line 1: I would not add all the mini docs changes to the changelog. It will be listed as part of the release anyway.

PR #675, opened 2026-03-05
Operator Context: Prepare Backend for GUI changes

JonasWurst:
## What has changed and why?

This PR registers the operator scope such that the frontend can use the type. also it improves the error ahndling for operator executed in the wrong scope. this is required for follow up prs

## How has it been tested?

tests are passing.

- [ ] Yes

### DIFF
--- a/lightly_studio/src/lightly_studio/api/routes/api/export.py
+++ b/lightly_studio/src/lightly_studio/api/routes/api/export.py
@@ -17,6 +17,7 @@ from lightly_studio.api.routes.api import collection as collection_api
 from lightly_studio.core.dataset_query.dataset_query import DatasetQuery
 from lightly_studio.db_manager import SessionDep
 from lightly_studio.export import export_dataset
+from lightly_studio.models.annotation.annotation_base import AnnotationType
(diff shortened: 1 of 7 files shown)


lightly_studio/tests/api/routes/api/test_operator.py
```diff
@@ -204,7 +203,10 @@ def test_execute_operator__scope_mismatch(
         json={"parameters": {}, "context": {"collection_id": str(video_collection.collection_id)}},
     )
 
-    assert response.status_code == HTTP_STATUS_UNPROCESSABLE_ENTITY
+    response_payload = response.json()
+    assert response.status_code == HTTP_STATUS_OK
+    assert response_payload["success"] is False
```
lukas-lightly, line 208: I always found HTTP endpoints returning 200 but then having `{success: false}` in the message confusing and harder to work with. Can you explain the reasoning?

JonasWurst: we use operatorResult to pass success and error to the GUI, adding proper messages

PR #678, opened 2026-03-05
Show group component names in the menu

michal-lightly:
## What has changed and why?

Show group component names in the menu.

## How has it been tested?

Tested manually. I plan to add a test in a follow-up, after a bug with isSelected is fixed (now we highlight based on sampleType, we should highlight based on collectionId).

Screenshot 2026-03-05 at 10 17 40

- [ ] Yes

### DIFF
--- a/lightly_studio_view/src/lib/components/NavigationMenu/utils.ts
+++ b/lightly_studio_view/src/lib/components/NavigationMenu/utils.ts
@@ -5,16 +5,17 @@ import { Image, WholeWord, Video, Frame, ComponentIcon, LayoutDashboard } from '
 import type { BreadcrumbLevel, NavigationMenuItem } from './types';
 
 export function getMenuItem(
-    sampleType: SampleType,
     pageId: string | null,
     datasetId: string,
-    collectionType: string,
-    collectionId: string
+    collectionId: string,
+    sampleType: SampleType,
+    groupComponentName?: string | null
 ): NavigationMenuItem {
+    const collectionType = sampleType.toLowerCase();
     switch (sampleType) {
         case SampleType.IMAGE:
             return {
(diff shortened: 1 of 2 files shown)


lightly_studio_view/src/lib/components/NavigationMenu/utils.ts
```diff
@@ -5,16 +5,17 @@ import { Image, WholeWord, Video, Frame, ComponentIcon, LayoutDashboard } from '
...
 
 export function getMenuItem(
-    sampleType: SampleType,
     pageId: string | null,
     datasetId: string,
-    collectionType: string,
-    collectionId: string
+    collectionId: string,
+    sampleType: SampleType,
+    groupComponentName?: string | null
```
LeonardoRosaa, line 12: Umm. I think we can simplify this to:

```ts
groupComponentName?: string
```

michal-lightly: Type check complains because `groupComponentName` can be null, and `?` allows just undefined.

lightly_studio_view/src/lib/components/NavigationMenu/utils.ts
```diff
@@ -5,16 +5,17 @@ import { Image, WholeWord, Video, Frame, ComponentIcon, LayoutDashboard } from '
...
 ): NavigationMenuItem {
+    const collectionType = sampleType.toLowerCase();
     switch (sampleType) {
         case SampleType.IMAGE:
             return {
-                title: 'Images',
+                title: groupComponentName || 'Images',
```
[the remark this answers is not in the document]
michal-lightly: We'll keep `||`, it is desired to rewrite an empty string

PR #680, opened 2026-03-05
Highlight only the current collection in the menu

michal-lightly:
## What has changed and why?

Highlight only the current collection in the menu. Passes `currentCollectionId` instead of `pageId` to `getMenuItem`.

## How has it been tested?

Added test. Created tests for getMenuItem function.

- [ ] Yes

### DIFF
--- a/lightly_studio/src/lightly_studio/api/routes/api/annotation.py
+++ b/lightly_studio/src/lightly_studio/api/routes/api/annotation.py
@@ -122,7 +122,7 @@ def read_annotations(
         filters=AnnotationsFilter(
             collection_ids=[collection_id],
             annotation_label_ids=annotation_label_ids,
-            annotation_tag_ids=tag_ids,
+            tag_ids=tag_ids,
         ),
     )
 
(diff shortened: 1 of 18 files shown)


lightly_studio_view/src/lib/components/NavigationMenu/utils.ts
```diff
@@ -1,66 +1,67 @@
...
 import { SampleType } from '$lib/api/lightly_studio_local';
-import { APP_ROUTES, routeHelpers } from '$lib/routes';
+import { routeHelpers } from '$lib/routes';
 import { Image, WholeWord, Video, Frame, ComponentIcon, LayoutDashboard } from '@lucide/svelte';
 import type { BreadcrumbLevel, NavigationMenuItem } from './types';
 
 export function getMenuItem(
-    pageId: string | null,
     datasetId: string,
+    currentCollectionId: string | undefined,
```
mihnea-necsulescu, line 9: it's a bit awkward that this is defined as `string | undefined`, but `groupComponentName` is defined as `string | null` few lines below - could this also use null?

michal-lightly: I believe the  type check would complain

PR #682, opened 2026-03-05
Operator Lifecycle: Enhance startup and shutdown management

JonasWurst:
## What has changed and why?

adding the operator lifecycle managment to the app lifespan.

## How has it been tested?

tests are added.

- [ ] Yes


lightly_studio/src/lightly_studio/api/app.py
```diff
@@ -60,9 +61,13 @@ async def lifespan(_: FastAPI) -> AsyncGenerator[None, None]:
         None when the application is ready.
     """
     try:
+        operator_registry.startup_all()
         yield
```
lukas-lightly, line 65: Does this work as a async generator, that can be cancelled, e.g. via a timeout, uncaught exception or a signal (TERM, KILL, etc.)?

JonasWurst: as discussed

PR #683, opened 2026-03-05
Operator lifecycle: add operator registry handling

JonasWurst:
## What has changed and why?

This adds the operator registry hadnling to startup/shutdown all registered operators.

## How has it been tested?

tests are added.

- [ ] Yes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,11 +10,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - `Dataset.export()` now supports instance segmentation export to COCO using `to_coco_instance_segmentations()`.
+- Added keyboard shortcuts to switch between the brush and eraser, as well as to adjust the brush size.
 
 ### Changed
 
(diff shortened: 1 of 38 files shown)


lightly_studio/src/lightly_studio/plugins/operator_registry.py
```diff
@@ -30,6 +33,42 @@ def register(self, operator: BaseOperator) -> None:
...
+                    "Operator '%s' (%s) failed to start.",
+                    operator.name,
+                    operator_id,
+                    exc_info=True,
+                )
+
+    def shutdown_all(self) -> None:
+        """Stop all registered operators sequentially."""
+        for operator_id, operator in self._operators.items():
+            operator.status = OperatorStatus.STOPPING
```
horatiualmasan, line 59: Should we check the status of the operator here? What would happen if we try to stop an operator with OperatorStatus.ERROR?

JonasWurst: Valid point. For now I would leave this to the operator. Even if there is an error, it may still require some shutdown.

PR #684, opened 2026-03-05
Operator Lifecycle: Return error to GUI when operator is not ready

JonasWurst:
## What has changed and why?

Return error to GUI when the operator to execute is not ready. 

Also adding changelog note about the new way for the lifecycle management of operators.

## How has it been tested?

test added.

- [x] Yes


lightly_studio/src/lightly_studio/api/routes/api/operator.py
```diff
@@ -79,6 +79,12 @@ def execute_operator(
             detail=f"Operator '{operator_id}' not found",
         )
 
+    if operator.status != OperatorStatus.READY:
+        return OperatorResult(
+            success=False,
+            message=f"Operator '{operator_id}' is not ready (status: {operator.status.value})",
```
michal-lightly: Is this meant for the user? It is not a very useful message, the user does not know if this is ok or not. It is better to tell the user what to do.

E.g. operator is starting, please wait.

michal-lightly (approved):

PR #695, opened 2026-03-06
Disable annotation text rendering by default

IgorSusmelj:
## What has changed and why?

- For a while now we always rendere the annotation class label text in the GUI. With many annotations this cane appear cluttered. This was always editable in the settings. Now with this PR we simply change the default.

Before this PR:
image

After this PR:
image

## How has it been tested?

Updated tests

- [X] Yes

### DIFF
--- a/lightly_studio_view/e2e/pages/video-frames-page.ts
+++ b/lightly_studio_view/e2e/pages/video-frames-page.ts
@@ -10,8 +10,9 @@ export class VideoFramesPage {
         await this.page.getByTestId('navigation-menu-frames').click();
 
         // Wait for frame grid items to be visible
+        // TODO(Horatiu 01/2026): 20s timeout might be too long, investigate if we can reduce it.
         await expect(this.getVideoFrames().first()).toBeVisible({
-            timeout: 10000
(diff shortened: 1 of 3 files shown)


lightly_studio_view/src/lib/components/Settings/SettingsDialog.svelte
```diff
@@ -47,7 +47,7 @@
                 keyToolbarEraser: $settingsStore.key_toolbar_eraser || 'x'
             };
             gridViewRendering = $settingsStore.grid_view_sample_rendering || 'contain';
-            showAnnotationTextLabels = Boolean($settingsStore.show_annotation_text_labels ?? true);
+            showAnnotationTextLabels = Boolean($settingsStore.show_annotation_text_labels ?? false);
```
LeonardoRosaa: This `?? false` condition is redundant, no?

IgorSusmelj: You're right. The Boolean part should be the same. I'll leave the ?? condition to match the other syntax.

PR #719, opened 2026-03-09
Add a navigation e2e test

michal-lightly:
## What has changed and why?

Add a navigation e2e test. It clicks on all the items in the menu and checks that at least some item was loaded.

## How has it been tested?

CI

- [ ] Yes

## Summary by CodeRabbit

* **Tests**
  * Added end-to-end navigation tests covering groups, collections, and view flows.
  * Removed a placeholder/dummy end-to-end test.
  * Set default action and navigation timeouts to improve test reliability.

* **Minor UI**
  * Added a test identifier to a group item element to aid UI testing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Add tooltips to features in annotation items in the right panel to improve usability.
+- Added instance segmentation export to COCO in the GUI.
+
 ### Changed
 
 ### Deprecated
--- a/lightly_studio/Makefile
+++ b/lightly_studio/Makefile
@@ -97,6 +97,11 @@ test-ci:
 	@echo "Running tests..."
 	uv run pytest
 
+.PHONY: test-postgres
+test-postgres:
+	@echo "Running tests against Postgres..."
+	uv run pytest --postgres
+
 .PHONY: check-uv-lock
 check-uv-lock:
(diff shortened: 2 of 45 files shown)


lightly_studio_view/e2e/groups/navigation.e2e-test.ts
```diff
@@ -0,0 +1,40 @@
+import { test, expect } from '@playwright/test';
+
+test.describe('groups-menu-navigation', () => {
+    test('navigate between collections', async ({ page }) => {
+        await page.goto('/');
+
+        // 1. Verify groups page loads with group items
+        await expect(page.getByTestId('group-grid-item').first()).toBeVisible({ timeout: 5000 });
```
mihnea-necsulescu: the explicit 5s timeouts in this test should not be needed - I think that's the default anyway

lightly_studio_view/e2e/groups/navigation.e2e-test.ts
```diff
@@ -0,0 +1,40 @@
+import { test, expect } from '@playwright/test';
+
+test.describe('groups-menu-navigation', () => {
+    test('navigate between collections', async ({ page }) => {
```
mihnea-necsulescu, line 4: we should probably register a groupsPage, too, but this should be sufficient for now

michal-lightly: Yes, we can do it later.

mihnea-necsulescu (approved):

PR #736, opened 2026-03-10
Run unit tests on Postgres

mihnea-necsulescu:
## What has changed and why?

Added the infrastructure to run the existing test suite against a real Postgres instance (managed by testcontainers):
- registered a `--postgres` CLI flag that can be used to run the tests on Postgres (e.g., `uv run pytest --postgres`)
- registered `duckdb_only` and `postgres_only` pytest markers (currently only used in a single place) based on which certain tests can be skipped 
- added setup for the session-scoped Postgres container
- updated some fixture to use the proper DB engine 
- some updates to tests and code
- added `test-postgres` target

## How has it been tested?

Manually: all tests pass on both DB backends. Running the tests with a single worker takes roughly the same amount of time between DuckDB and Postgres (01:13 for DuckDB; 01:14 for Postgres); there's a difference, though, when running them with multiple workers (21s for DuckDB; 31s for Postgres).

- [ ] Yes

## Summary by CodeRabbit

* **New Features**
  * Added PostgreSQL testing support via new `--postgres` flag for test execution.
  * Enhanced database compatibility detection for improved backend flexibility.

* **Tests**
  * Expanded test infrastructure with dual-database testing capabilities.
  * Improved test assertions for database operations.

### DIFF
--- a/lightly_studio/tests/conftest.py
+++ b/lightly_studio/tests/conftest.py
@@ -487,7 +487,7 @@ def assert_contains_properties(
 def patch_collection(
     mocker: MockerFixture,
     _db_engine: DatabaseEngine,
-) -> Generator[None, None, None]:
+) -> None:
     """Fixture to patch the collection resources.
 
     Patches get_engine() to return the per-test engine (in-memory DuckDB or
@@ -516,5 +516,3 @@ def patch_collection(
 
     # Create test-specific lightly_studio_active_features.
     mocker.patch.object(features, "lightly_studio_active_features", [])
-
-    yield


lightly_studio/src/lightly_studio/models/embedding_model.py
```diff
@@ -5,15 +5,15 @@
...
+from sqlmodel import VARCHAR, Column, Field, SQLModel
 
 
 class EmbeddingModelBase(SQLModel):
     """Base class for the EmbeddingModel."""
 
     name: str
     parameter_count_in_mb: int | None = None
-    embedding_model_hash: str = Field(default="", sa_column=Column(CHAR(128)))
+    embedding_model_hash: str = Field(default="", sa_column=Column(VARCHAR(128)))
```
mihnea-necsulescu, line 16: `CHAR` in Postgres is not working the same way:
`AssertionError: assert 'same_hash   ...             ' == 'same_hash'` ; 
it is padding with trailing spaces

this change should be fine, since in DuckDB, `CHAR` is just an alias for `VARCHAR` (https://duckdb.org/docs/stable/sql/data_types/text)

lightly_studio/src/lightly_studio/resolvers/annotations/annotations_filter.py
```diff
@@ -91,11 +91,7 @@ def _apply_annotation_filters(
 
         # Filter by tags
         if self.tag_ids:
-            query = (
-                query.join(annotation_sample.tags)
-                .where(annotation_sample.tags.any(col(TagTable.tag_id).in_(self.tag_ids)))
-                .distinct()
-            )
```
mihnea-necsulescu, line 98: this was not working for Postgres:
`sqlalchemy.exc.ProgrammingError: (psycopg.errors.InvalidColumnReference) for SELECT DISTINCT, ORDER BY expressions must appear in select list` ; Postgres requires all columns present in the order by to also be present in select distinct

this change should be fine; the now removed `join` and `distinct` were doing the same thing that was done with just `any` -> the `join` was creating duplicates that were removed by `distinct`, `.any()` already generates an `exists` subquery which would correctly filter without creating duplicates that have to be removed

for example, if called from [here](https://github.com/lightly-ai/lightly-studio/blob/main/lightly_studio/src/lightly_studio/resolvers/annotation_resolver/get_all.py#L66), the old version would produce something like:
```sql
select distinct ...
from annotation_base
  left outer join image on ...
  left outer join video_frame on ...
  left outer join  video on ...
  join sample on ...
  join samplelinktable on ...
  join tag on ...
where exists (
  select 1 from tag, samplelinktable
  where ... and tag.tag_id in (...)
order by ...
```

the new version would produce something like:
```sql
select ...
from annotation_base
  left outer join image on ...
  left outer join video_frame on ...
  left outer join  video on ...
  join sample on ...
where exists (
  select 1 from tag, samplelinktable 
  where ... and tag.tag_id in (...)
order by ...
```

lightly_studio/src/lightly_studio/db_manager.py
```diff
@@ -253,7 +253,7 @@ def _detect_backend_from_url(engine_url: str) -> DatabaseBackend:
     """
     if engine_url.startswith("duckdb://"):
         return DatabaseBackend.DUCKDB
-    if engine_url.startswith(("postgresql://", "postgres://")):
+    if engine_url.startswith("postgres"):
```
mihnea-necsulescu, line 256: testcontainers explicitly sets the driver (i.e., `"postgresql+psycopg://"`) and this would fail with an unsupported database url; just checking if the url starts with "postgres" should be fine

lightly_studio/tests/resolvers/group_resolver/test_create_many.py
```diff
@@ -102,7 +102,7 @@ def test_create_many__unique_component_samples(db_session: Session) -> None:
         groups=[{front_image.sample_id}],
     )
     # Assigning the sample to a second group should fail
-    with pytest.raises(IntegrityError, match='Constraint Error: Duplicate key "sample_id'):
+    with pytest.raises(IntegrityError, match="[Dd]uplicate key"):
```
mihnea-necsulescu, line 105: changed the regex so it would work for both DuckDB and Postgres (Postgres uses lowercase "d" in "duplicate")

lightly_studio/tests/resolvers/test_sample_embedding_resolver.py
```diff
@@ -36,7 +36,7 @@ def test_create_sample_embedding(db_session: Session) -> None:
         embedding_model_id=embedding_model_id,
     )
     assert sample_embedding.sample_id == sample_id
-    assert sample_embedding.embedding == [1.0, 2.0, 3.0]
+    assert list(sample_embedding.embedding) == [1.0, 2.0, 3.0]
```
mihnea-necsulescu, line 39: pgvector seems to return numpy arrays: `ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()`

lightly_studio/uv.lock
```diff
@@ -6,9 +6,12 @@ resolution-markers = [
     "python_full_version == '3.12.*'",
     "python_full_version == '3.11.*'",
     "python_full_version == '3.10.*'",
-    "python_full_version < '3.10' and platform_machine == 'arm64' and sys_platform == 'darwin'",
-    "python_full_version < '3.10' and platform_machine == 'aarch64' and sys_platform == 'linux'",
-    "(python_full_version < '3.10' and platform_machine != 'arm64' and sys_platform == 'darwin') or (python_full_version < '3.10' and platform_machine != 'aarch64' and sys_platform == 'linux') or (python_full_version < '3.10' and sys_platform != 'darwin' and sys_platform != 'linux')",
+    "python_full_version >= '3.9.2' and python_full_version < '3.10' and platform_machine == 'arm64' and sys_platform == 'darwin'",
```
mihnea-necsulescu, line 9: I'm not really sure why those were changed

lightly_studio/tests/conftest.py
```diff
@@ -44,12 +46,111 @@
...
+
+    if use_postgres:
+        assert _postgres_engine is not None
+        with _postgres_engine.session() as session:
+            yield session
+            _truncate_tables(session)
+    else:
+        test_manager = DatabaseEngine("duckdb:///:memory:", single_threaded=True)
+        with test_manager.session() as session:
+            yield session
```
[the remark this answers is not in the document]
mihnea-necsulescu: no test uses both fixtures at the same time

lightly_studio/src/lightly_studio/resolvers/annotations/annotations_filter.py
michal-lightly: The new version should be even more efficient, no?

lightly_studio/src/lightly_studio/db_manager.py
michal-lightly: Optional: It looks weirdly inconsistent with line 255. We could add a comment, or delete `://` from above.

lightly_studio/tests/resolvers/test_sample_embedding_resolver.py
```diff
@@ -107,16 +107,16 @@ def test_add_sample_embedding_to_sample(db_session: Session) -> None:
...
 
     assert len(image.sample.embeddings) == 1
-    assert sample_embedding.embedding == image.sample.embeddings[0].embedding
+    assert list(sample_embedding.embedding) == list(image.sample.embeddings[0].embedding)
```
michal-lightly, line 113: Optional: np array can't be compared with each other without casting?

lightly_studio/tests/conftest.py
```diff
@@ -44,12 +46,110 @@
...
+@pytest.fixture(scope="session")
+def _postgres_url(_postgres_container: PostgresContainer) -> str | None:
+    """Return the Postgres connection URL for the test container, or None for DuckDB."""
+    if _postgres_container is None:
+        return None
+    return _postgres_container.get_connection_url()  # type: ignore[no-any-return]
+
+
+@pytest.fixture(scope="session")
+def _postgres_engine(_postgres_url: str | None) -> Generator[DatabaseEngine | None, Any, None]:
```
michal-lightly: Minor: Avoid `Any` types:
```suggestion
@pytest.fixture(scope="session")
def _postgres_engine(_postgres_url: str | None) -> Generator[DatabaseEngine | None, None, None]:
```

lightly_studio/tests/conftest.py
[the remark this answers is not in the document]
michal-lightly: @mihnea-necsulescu Let's implement the original suggestion, it is good. We can have a fixture `_db_engine` that is test scoped.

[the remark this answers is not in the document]
michal-lightly: Optional: Also we can fix the test that it pointed out, it should use `dataset.session` instead of `db_session`. It currently does `dataset.session = db_session` which is hacky.

lightly_studio/tests/test_db_vector.py
```diff
@@ -8,81 +8,82 @@
...
 from sqlalchemy import ARRAY, Float
 from sqlalchemy.dialects import postgresql, sqlite
+from sqlmodel import Session
 
 from lightly_studio import db_vector
-from lightly_studio.db_manager import DatabaseEngine
 from lightly_studio.db_vector import VectorType
 
 
-class TestVectorType:
```
michal-lightly, line 17: I think this PR is getting long. Can you please comment out these tests, and change them in a follow-up PR?

lightly_studio/tests/conftest.py
```diff
@@ -44,12 +46,110 @@
...
+    pg_container = PostgresContainer(
+        image="pgvector/pgvector:0.8.1-pg18-bookworm", driver="psycopg"
+    )
+    pg_container.start()
+    yield pg_container
+    pg_container.stop()
+
+
+@pytest.fixture(scope="session")
+def _postgres_url(_postgres_container: PostgresContainer) -> str | None:
```
michal-lightly: Optional: We could merge with the fixture above - simply yield `pg_container.get_connection_url()`.

lightly_studio/tests/conftest.py
```diff
@@ -44,12 +46,110 @@
...
+        action="store_true",
+        default=False,
+        help="Run tests against a Postgres container instead of in-memory DuckDB.",
+    )
+
+
+def pytest_configure(config: pytest.Config) -> None:
+    """Register custom markers for DB backend selection."""
+    config.addinivalue_line("markers", "duckdb_only: mark test to run only with DuckDB backend")
+    config.addinivalue_line("markers", "postgres_only: mark test to run only with Postgres backend")
```
michal-lightly: Two comments:
1) Let's have both the "user marker" (currently `duckdb_only`) and internal marker (currently `skip_on_postgres`) the same semantics.
    * I think they should both have the "skip" semantics (will scale to more db backends in theory)
    * If they can both be `skip_on_postgress` -> great, let's use that
    * If not, the latter one can be `skip_on_postgress_internal`
2) I would suggest to declare the markers in `pyproject.toml`, that's the usual place: https://docs.pytest.org/en/stable/example/markers.html#registering-markers

lightly_studio/src/lightly_studio/resolvers/annotations/annotations_filter.py
mihnea-necsulescu: yes, it should

lightly_studio/tests/resolvers/test_sample_embedding_resolver.py
mihnea-necsulescu: not with `==` apparently :/

lightly_studio/tests/conftest.py
mihnea-necsulescu: I moved them to `pytest.ini` - I don't think pytest considers `pyproject.toml` if `pytest.ini` is also present (https://docs.pytest.org/en/stable/reference/customize.html#pytest-ini)

we should probably consolidate everything into `pyproject.toml` at some point

lightly_studio/tests/conftest.py
```diff
@@ -44,12 +46,107 @@
...
+    Table names are quoted to handle reserved words (e.g. "group").
+    """
+    table_names = [table.name for table in reversed(SQLModel.metadata.sorted_tables)]
+    for table_name in table_names:
+        session.execute(sqlalchemy.text(f'TRUNCATE TABLE "{table_name}" CASCADE'))
+    session.commit()
+
+
+@pytest.fixture(scope="session")
+def use_postgres(request: pytest.FixtureRequest) -> bool:
```
michal-lightly: Minor: This is a helper used only in this file:

```suggestion
def _use_postgres(request: pytest.FixtureRequest) -> bool:
```

lightly_studio/tests/conftest.py
```diff
@@ -44,12 +46,107 @@
...
+    use_postgres: bool,
+    _postgres_engine: DatabaseEngine | None,
+) -> Generator[DatabaseEngine, None, None]:
+    """Provide a single DatabaseEngine for each test.
+
+    DuckDB mode (default): creates a fresh in-memory engine per test.
+    Postgres mode (--postgres): reuses the session-scoped Postgres engine.
+    """
+    if use_postgres:
+        assert _postgres_engine is not None
```
michal-lightly, line 115: Minor: You can inline contents of `_postgres_engine` and delete `_postgres_engine` fixture. Actually it will be quite elegant.

lightly_studio/tests/conftest.py
```diff
@@ -44,12 +46,107 @@
...
+        yield _postgres_engine
```
michal-lightly, line 116: Optional: I would suggest `_truncate_tables(session)` to be rather here than in `db_session`. Seems more tied to the engine lifecycle rather than session lifecycle.

lightly_studio/tests/conftest.py
mihnea-necsulescu: I'd keep them separate - otherwise the Postgres engine will become test-scoped and it would be confusing, i.e., why is the Postgres container session-scoped, but the Postgres engine test-scoped? also, making it test-scoped can add some (maybe negligible) overhead - e.g., for Postgres, `SQLModel.metadata.create_all` doesn't make sense to be called per test

michal-lightly: Yes, good point

PR #740, opened 2026-03-10
Igor lig 8849 fix redundant requests to fetch videos from frontend

IgorSusmelj:
## What has changed and why?

- Fixed duplicate initial video list requests by deriving the first video query filter directly from the page state instead of waiting for the shared video filter store to initialize
- Make sure we don't retrigger loading of data when text_embedding changes between None and []
- Reuse a buildVideoFilter helper across the frontend code

## How has it been tested?

Manual testing

- [X] Yes

## Summary by CodeRabbit

* **Bug Fixes**
  * Reduced redundant initial video requests for faster load and smoother navigation.
  * Improved handling of text embeddings: searches now clear and apply embeddings reliably, avoiding stale or empty embeddings.

* **Refactor**
  * Simplified video filtering and query payloads for more consistent filtering and better synchronization when selecting samples.

### DIFF
--- a/lightly_studio/Makefile
+++ b/lightly_studio/Makefile
@@ -97,6 +97,11 @@ test-ci:
 	@echo "Running tests..."
 	uv run pytest
 
+.PHONY: test-postgres
+test-postgres:
+	@echo "Running tests against Postgres..."
+	uv run pytest --postgres
+
 .PHONY: check-uv-lock
 check-uv-lock:
 	@echo "Ensuring uv.lock is up to date..."
--- a/lightly_studio/pyproject.toml
+++ b/lightly_studio/pyproject.toml
@@ -84,6 +84,7 @@ dev = [
     "types-tqdm>=4.67.0.20241221",
     "types-PyYAML>=6.0.0",
     "pytest-xdist>=3.6.1",
+    "testcontainers[postgres]>=4.0.0",
 ]
 
 docs = [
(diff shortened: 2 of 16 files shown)


lightly_studio_view/src/lib/hooks/useVideos/useVideos.ts
```diff
@@ -9,12 +9,13 @@ export const useVideos = (
     filter: VideoFilter,
     text_embedding?: Array<number>
 ) => {
+    const normalizedTextEmbedding = text_embedding?.length ? text_embedding : undefined;
```
mihnea-necsulescu: I think it's better to directly update [setTextEmbedding](https://github.com/lightly-ai/lightly-studio/blob/main/lightly_studio_view/src/lib/hooks/useGlobalStorage.ts#L85-L87) to make it accept undefined, too

then change its call sites to directly pass `undefined` instead of empty arrays (e.g., [here](https://github.com/lightly-ai/lightly-studio/blob/main/lightly_studio_view/src/routes/datasets/%5Bdataset_id%5D/%5Bcollection_type%5D/%5Bcollection_id%5D/%2Blayout.svelte#L447))

that would improve the image text search, too + this const won't be needed anymore

IgorSusmelj: I looked into it. Good idea. I updated the code :)

PR #742, opened 2026-03-11
Update test_db_vector

mihnea-necsulescu:
## What has changed and why?

Split the test_db_vector updates to a new PR

## How has it been tested?

Test passes

- [ ] Yes

### DIFF
--- a/lightly_studio/tests/conftest.py
+++ b/lightly_studio/tests/conftest.py
@@ -69,15 +69,15 @@ def _truncate_tables(session: Session) -> None:
 
 
 @pytest.fixture(scope="session")
-def use_postgres(request: pytest.FixtureRequest) -> bool:
+def _use_postgres(request: pytest.FixtureRequest) -> bool:
     """Return True when the test suite is running against Postgres."""
     return bool(request.config.getoption("--postgres"))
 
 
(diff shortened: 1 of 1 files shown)


lightly_studio/tests/test_db_vector.py
```diff
@@ -8,81 +8,82 @@
...
-        """cosine_distance compiles to <=> with ::vector casts for PostgreSQL."""
-        expr = db_vector.cosine_distance(sqlalchemy.column("col1"), sqlalchemy.column("col2"))
+@pytest.mark.skip_on_postgres
+def test_load_dialect_impl__duckdb(db_session: Session) -> None:
+    """VectorType returns ARRAY(Float) for DuckDB dialect."""
+    assert db_session.bind is not None
+    dialect = db_session.bind.dialect
+    vector_type = VectorType()
+    result = vector_type.load_dialect_impl(dialect=dialect)
+    assert isinstance(result, ARRAY)
```
michal-lightly: Optional: I wonder if we could have
```py
if cond:
    assert isinstance(result, ARRAY)
    assert isinstance(result.item_type, Float)
else:
    assert isinstance(result, Vector)
```
Where `cond` is some condition on `dialect`.

PR #779, opened 2026-03-17
feat: introduce component to to provide information about currentFrame

ikondrat:
## What has changed and why?

This PR introduces component to show information about current frame.
It is using a binding pattern to provide information outside. We need to get the playback time for VideoPlayer

## How has it been tested?

By the unit test

- [ ] Yes

## Overview
Introduces a new `VideoFrameDetails` Svelte component that displays information about the current video frame.

## Changes
- **New Component**: `VideoFrameDetails.svelte` - Displays frame details including:
  - Frame number
  - Timestamp (formatted to 3 decimal places with "s" suffix)
  - Button to view the frame with provided URL
  
- **Test Coverage**: Comprehensive test suite validating:
  - Frame number rendering
  - Timestamp formatting precision
  - Button href configuration
  
- **Public Export**: Component exposed via the central components index barrel (`src/lib/components/index.ts`)

## Implementation Details
The component accepts two props:
- `frame` (FrameView type): Contains frame_number and frame_timestamp_s
- `frameURL` (string): Link destination for the view frame button

Renders as a Segment titled "Current Frame" with a full-width secondary button for navigation.

## Test Results
All tests passing, validating correct rendering and formatting behavior.


lightly_studio_view/src/lib/components/VideoFrameDetails/VideoFrameDetails.svelte
```diff
@@ -0,0 +1,95 @@
+<script lang="ts">
```
LeonardoRosaa, line 1: The component looks great. However, I’m not sure `useVideoFrames` should live here. It feels like too much responsibility for a child component. I’d expect to keep this logic in the parent component, since it will also be more readable there. WDTY?

ikondrat: hmm good point, maybe it makes sense to just consume hook results

I will change it a bit

ikondrat: updated

PR #783, opened 2026-03-18
Add FilterWithCollectionId

horatiualmasan:
## What has changed and why?

Add new class `FilterWithCollectionId` that will be used in cases where we need to keep filters and corresponding collection_id close together. 

## How has it been tested?

New test

- [ ] Yes

## Overview

Adds a new Pydantic model `FilterWithCollectionId` that couples a `collection_id` with an inner filter object and scopes queries to that collection before delegating to the inner filter.

## Changes

**New Module**: lightly_studio/src/lightly_studio/resolvers/filter_with_collection_id.py
- Added `FilterProtocol` (runtime-checkable structural Protocol) requiring `apply(self, query) -> QueryType`.
- Added generic `FilterWithCollectionId[T](BaseModel, Generic[T])` with fields:
  - `collection_id: UUID`
  - `filter: T`
- `FilterWithCollectionId.apply(query)`:
  - Scopes the provided query by adding a `where` predicate equating `SampleTable.collection_id` to the stored `collection_id`.
  - Delegates to the inner filter's `apply()` using the scoped query and returns the resulting query.
  - Configured to allow arbitrary inner filter types (ConfigDict(arbitrary_types_allowed=True)).

**New Test Module**: lightly_studio/tests/resolvers/test_filter_with_collection_id.py
- Test 1: Creates two collections and images, applies `FilterWithCollectionId` (scoped to one collection + width >= 500) and asserts a single expected row with correct `sample_id` and `width`.
- Test 2: Uses a random `collection_id` with an empty inner `ImageFilter` and asserts the result set is empty, verifying collection scoping.

## Impact

- Files added: 2 (implementation + tests)
- Lines changed: +45 / +58 (implementation + tests) — total +103
- Internal change only; CHANGELOG.md not updated

### DIFF
--- a/lightly_studio_view/src/lib/components/Typography/constants.ts
+++ b/lightly_studio_view/src/lib/components/Typography/constants.ts
@@ -15,6 +15,18 @@ export const TYPOGRAPHY_VARIANTS = [
 
 export type TypographyVariant = (typeof TYPOGRAPHY_VARIANTS)[number];
 
-export const COMPONENT_TYPES = ['h1', 'h2', 'h3', 'h4', 'h5', 'h6', 'p', 'span', 'div'] as const;
+export const COMPONENT_TYPES = [
+    'h1',
+    'h2',
+    'h3',
+    'h4',
+    'h5',
+    'h6',
+    'p',
+    'span',
+    'div',
+    'dt',
+    'dd'
+] as const;
 
 export type ComponentType = (typeof COMPONENT_TYPES)[number];
--- a/lightly_studio_view/src/lib/components/VideoSampleMetadata/VideoSampleMetadata.stories.svelte
+++ b/lightly_studio_view/src/lib/components/VideoSampleMetadata/VideoSampleMetadata.stories.svelte
@@ -1,0 +1,76 @@
+<script module>
+    import { defineMeta } from '@storybook/addon-svelte-csf';
+    import VideoSampleMetadata from './VideoSampleMetadata.svelte';
+
+    const { Story } = defineMeta({
+        title: 'Components/VideoSampleMetadata',
+        component: VideoSampleMetadata,
+        tags: ['autodocs']
+    });
+</script>
+
+<Story
+    name="Default"
+    args={{
+        video: {
+            file_name: 'sample_video.mp4',
+            width: 1920,
+            height: 1080,
+            duration_s: 45.5,
+            fps: 30.0
+        }
+    }}
+/>
+
+<Story
+    name="Long Filename"
+    args={{
+        video: {
+            file_name:
+                'very_long_filename_that_should_wrap_or_break_properly_in_the_metadata_panel_to_test_overflow_behavior.mp4',
+            width: 1920,
+            height: 1080,
+            duration_s: 45.5,
+            fps: 30.0
+        }
+    }}
+/>
+
+<Story
+    name="Zero Duration"
+    args={{
+        video: {
+            file_name: 'zero_duration.mp4',
+            width: 1920,
+            height: 1080,
+            duration_s: 0,
+            fps: 30.0
+        }
+    }}
+/>
+
+<Story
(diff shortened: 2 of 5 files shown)


lightly_studio/tests/resolvers/test_filter_with_collection_id.py
```diff
@@ -0,0 +1,58 @@
...
+        db_session=db_session,
+        collection_id=col_b.collection_id,
+        images=[ImageStub(path="b_small.png", width=100)],
+    )
+
+    query = select(ImageTable).join(ImageTable.sample)
+    filtered = FilterWithCollectionId[ImageFilter](
+        collection_id=col_a.collection_id,
+        filter=ImageFilter(width=FilterDimensions(min=500)),
+    ).apply(query)
```
ikondrat, line 41: I am not sure about this 🤔 

For me seems like writing directly
`query.where(col(sample_table.collection_id) == self.collection_id)` where we need would be shorter

I wouldn't hide this query under the additional abstraction

lightly_studio/src/lightly_studio/resolvers/filter_with_collection_id.py
```diff
@@ -0,0 +1,53 @@
...
+from uuid import UUID
+
+from pydantic import BaseModel, ConfigDict
+from sqlmodel import col
+from typing_extensions import Self
+
+from lightly_studio.models.sample import SampleTable
+
+
+class QueryWithWhereProtocol(Protocol):
```
michal-lightly: Minor: Shouldn't this one also be @runtime_checkable?

lightly_studio/src/lightly_studio/resolvers/filter_with_collection_id.py
```diff
@@ -0,0 +1,53 @@
...
+
+    collection_id: UUID
+    filter: T
+
+    def apply(
+        self,
+        query: Q,
+        sample_table: type[SampleTable] = SampleTable,
+    ) -> Q:
+        """Apply collection scope then the inner filter to the query."""
```
michal-lightly: Let's describe the arguments. It is non-trivial why we need `sample_table`.

lightly_studio/src/lightly_studio/resolvers/filter_with_collection_id.py
```diff
@@ -0,0 +1,53 @@
...
+        scoped = query.where(col(sample_table.collection_id) == self.collection_id)
+        return cast(Q, self.filter.apply(scoped))
```
michal-lightly: Why is the cast needed? Any way to avoid it?

michal-lightly: I think yes: In FilterProtocol, don't use `Any` type, make it generic instead.

lightly_studio/tests/resolvers/test_filter_with_collection_id.py
michal-lightly: This is okay. FilterWithCollectionId is a class that holds data, and can be potentially constructed in the frontend.

lightly_studio/src/lightly_studio/resolvers/filter_with_collection_id.py
```diff
@@ -0,0 +1,45 @@
...
+        ...
+
+
+T = TypeVar("T", bound=FilterProtocol)
+
+
+class FilterWithCollectionId(BaseModel, Generic[T]):
+    """Pairs a collection ID with a filter."""
+
+    model_config = ConfigDict(arbitrary_types_allowed=True)
```
michal-lightly, line 30: Btw. why is this needed?

horatiualmasan: It's needed to make Pydantic work with protocols. Without it I would get runtime errors.

PR #889, opened 2026-04-07
Add tag from details view

JonasWurst:
## What has changed and why?

Updated the tag management functionality by adding a form for tag creation and enhancing the existing tag display. This change improves user interaction with tags, allowing for easier addition and removal of tags associated with samples.

## How has it been tested?

Tested the tag addition and removal functionalities to ensure they work as expected. Verified that the UI updates correctly when tags are added or removed.

- [x] Yes

## Summary by CodeRabbit

* **New Features**
  * Create and add tags directly from sample and annotation detail views via a searchable popover; supports on-the-fly tag creation and case-insensitive matching.
  * Tag controls now consistently appear in detail panels even when no tags are present.

* **Bug Fixes / UX**
  * Unified tag add/remove flow with clearer success/error feedback and automatic refresh of tag lists.


lightly_studio_view/src/lib/components/SegmentTags/SegmentTags.svelte
```diff
@@ -16,7 +19,7 @@
         onRefetch?: () => void;
     }
 
-    let { tags, collectionId, sampleId, onRefetch }: Props = $props();
+    let { tags, collectionId, sampleId, onRefetch = () => {} }: Props = $props();
```
LeonardoRosaa, line 22: Does it need a default value? Can you make it optional and use `onRefetch?.()` instead?

lightly_studio_view/e2e/general/sample-details/tag-sample-detail.e2e-test.ts
```diff
@@ -0,0 +1,27 @@
+import { test, expect } from '../../utils';
+
+test.describe('sample-details-tag-flow', () => {
+    test('user can create and attach multiple new tags in sample details', async ({
```
LeonardoRosaa, line 5: I'd add tests for the following scenarios:

- Attempting to add a tag that is already assigned
- Deleting a tag created in this view
- Adding a tag to the current sample, then navigating to the next sample and adding the same tag again

lightly_studio_view/src/lib/hooks/useAddTagToSample/useAddTagToSample.ts
```diff
@@ -0,0 +1,70 @@
...
+        } catch {
+            toast.error('Failed to add tag. Please try again.');
+            return;
+        } finally {
+            _busy.set(false);
+        }
+        onRefetch();
+    }
+
+    async function createAndAdd(name: string) {
```
LeonardoRosaa, line 42: We could implement this with a single API call. I’d create a follow-up issue. WDTY?

LeonardoRosaa (changes requested):

lightly_studio_view/src/lib/components/SegmentTags/SegmentTags.svelte
JonasWurst: We would need to propagate it to the hook then as well. Hence I'd keep it like this if possible

lightly_studio_view/src/lib/hooks/useAddTagToSample/useAddTagToSample.ts
JonasWurst: I will add an issue.

LeonardoRosaa (approved):

PR #893, opened 2026-04-07
Improve interraction with Embedding plot selection

horatiualmasan:
## What has changed and why?

Added a filter entry for the embedding plot filter in the filter panel.  It will show the number of selected samples in the embedding plot allowing show/hide and clear.
Removed "Clear selection" button from the plot. It can now be made from the filter item.

## How has it been tested?
New tests and manual test

https://github.com/user-attachments/assets/7f92b053-84ae-4baf-8e47-cf427d9b9e86

- [x] Yes

## Release Notes

* **New Features**
  * Added Embedding Selection functionality with toggleable visibility and selection management controls in embedding plots.

* **UI/UX Updates**
  * Updated legend label terminology for embedding selections.


lightly_studio_view/src/routes/datasets/[dataset_id]/[collection_type]/[collection_id]/+layout.svelte
```diff
@@ -484,6 +492,21 @@
     const showLeftSidebar = $derived(
         isSamples || isAnnotations || isVideos || isVideoFrames || isGroups
     );
+
+    const embeddingFilter = useEmbeddingFilter({
+        collectionId: collectionIdStore,
+        isVideos: toStore(() => isVideos),
+        isSamples: toStore(() => isSamples),
+        setRangeSelectionForcollection
+    });
```
Comment (ikondrat) [inline lightly_studio_view/src/routes/datasets/[dataset_id]/[collection_type]/[collection_id]/+layout.svelte]: I would avoid cascading isVideos and isImages( let's name it correctly)

What is the problem now:
We cascading branching between images and videos 

Let's rather create separate:
useEmbeddingFilterForImages
useEmbeddingFilterForVideos

lightly_studio_view/src/lib/hooks/useEmbeddingFilter/useEmbeddingFilter.ts
```diff
@@ -0,0 +1,146 @@
...
+        if (!shouldShow) {
+            hideEmbeddingFilters();
+            return;
+        }
+        const alreadyApplied = get(activePlotFilterSampleIds).length > 0;
+        const hasHidden = get(hiddenEmbeddingFilterSampleIds).length > 0;
+        if (!alreadyApplied && hasHidden) showEmbeddingFilters();
+    }
+
+    function clearPlotFilter() {
```
ikondrat: I would avoid mixing to many logic in one place

let's not introduce anything related to plot in generic embedding filtering

ikondrat (changes requested):
I am struggling to understand what it is doing. It looks like Hook is not really following SOLID and the single responsibility principle, and it is hard to understand what it does. 

**Problems**
Multiple Responsibilities: The hook handles:
Active filter state derivation (video vs image)
Hidden filter state management
Filter visibility toggling logic
UI-related computed values (count, labels, context flags)
Coordination with video/image filters
Poor Abstraction: Mixing video/image filter logic with visibility toggle logic
Large Surface Area: Returns 9 different properties/functions
Tight Coupling: Direct dependency on useImageFilters and useVideoFilters

**Proposals from AI**
1. Extract Hidden Filter Store (useHiddenFilters.ts)
Responsibility: Manage hidden filters per collection
Single purpose: Store/retrieve/clear hidden filter state
// Returns: { hiddenFilters, setHidden, clearHidden, getForCollection }

2. Extract Active Filter Adapter (useActiveFilterAdapter.ts)
Responsibility: Abstract video vs image filter differences
Single purpose: Unified interface for getting/setting sample IDs
// Returns: { activeSampleIds, updateSampleIds, isApplied }

3. Extract Filter Visibility Manager (useFilterVisibility.ts)
Responsibility: Toggle between active and hidden states
Single purpose: Coordinate show/hide operations
// Returns: { setVisibility, clearAll }

4. Simplified Main Hook (useEmbeddingFilter.ts)
Responsibility: Compose the above hooks and provide UI-facing computed values
Returns only what consumers actually need


My personal proposal is to challenge is we need to keep branching
isImages and isVideos is static information we dont need to have it reactive 

so you can technically have
useEmbeddingFilterForVideos
useEmbeddingFilterForImages

and it would be easier to handle
 
one more point - please challenge reactivity usage 

From what I see, reactivity is wasted because it is only used 
onVisibilityChange={embeddingFilter.setEmbeddingFilterVisibility}
onClear={embeddingFilter.clearPlotFilter}

maybe reactivity is not required at all in this hook

lightly_studio_view/src/routes/datasets/[dataset_id]/[collection_type]/[collection_id]/+layout.svelte
```diff
@@ -506,6 +548,15 @@
                             </div>
                             <Segment title="Filters" icon={SlidersHorizontal}>
```
Comment (ikondrat) [inline lightly_studio_view/src/routes/datasets/[dataset_id]/[collection_type]/[collection_id]/+layout.svelte:513]: It looks like this useEmbeddingFilterForImages is required only for <Segment title="Filters" 

Can we move this out as a separate Component?

E.g. <LayoutFilters
I think you can pass filtersType = "image" | "video"

and Based  on type call either 
<LayoutFiltersVideos
<LayoutFiltersImages

this way we will be able to support more sample types just by introducing new <LayoutFiltersType + adding type without changing a lot

wdyt?

Comment (ikondrat) [reply lightly_studio_view/src/routes/datasets/[dataset_id]/[collection_type]/[collection_id]/+layout.svelte:513]: and you will not need to call bothe as well
const embeddingFilterForImages = useEmbeddingFilterForImages(
        collectionIdStore,
        setRangeSelectionForCollection
    );
    const embeddingFilterForVideos = useEmbeddingFilterForVideos(
        collectionIdStore,
        setRangeSelectionForCollection
    );

ikondrat (approved):
lgtm

PR #913, opened 2026-04-09
Tag Details: Add tag popover

JonasWurst:
## What has changed and why?

Adding a popover compnent to create tags from details views.

## How has it been tested?

will be tested in follow up pr

- [ ] Yes

## Summary by CodeRabbit

* **New Features**
  * Added a tag popover interface for adding tags to segments with support for selecting existing tags or creating new ones.
  * Includes case-insensitive duplicate prevention and input validation.

* **Tests**
  * Added comprehensive test coverage for the tag popover component.


lightly_studio_view/src/lib/components/SegmentTags/AddTagPopover.svelte
```diff
@@ -0,0 +1,76 @@
...
+        onSelect: (name: string) => void;
+    }
+
+    let { options, attachedTagNames, busy, onSelect }: Props = $props();
+
+    let open = $state(false);
+    let inputValue = $state('');
+
+    $effect(() => {
+        if (!open) inputValue = '';
```
ikondrat: if you would like to change inputValue on toggling the state rather make it in the callback explicetely using onOpenChange
https://bits-ui.com/docs/components/popover#api-reference
like 

```
onOpenChange = (isOpen: boolean) => {
   if (!isOpen) {
     inputValue = '';
   }
}
```

it would be more self-describing, otherwise the intention of code is less clear

ikondrat (approved):
Lgtm but I would reduce $effect usage if we have provided target callback for state changes

every $effect is additional state-track handling, we don't need it if we have aleady callback for it

PR #926, opened 2026-04-10
Tag workflow: Add tag input component

JonasWurst:
## What has changed and why?

Introdcuing new tag input componeent that will be used by the new tag menu.

## How has it been tested?

test are added

- [ ] Yes

## Summary by CodeRabbit

* **New Features**
  * Added a tag-assignment input with autocomplete dropdown: live search/filtering, select existing tags or create a new tag from the input.
  * Keyboard support: Enter to select or create (trimmed) tag, Escape to cancel and clear.
  * Respects a busy/disabled state to block interactions while busy.

* **Tests**
  * Added tests for typing, selection, creation, Enter behavior, and Escape cancelation.


lightly_studio_view/src/lib/components/TagsMenu/TagAssignInput.svelte
```diff
@@ -0,0 +1,91 @@
...
+        onSelect: (name: string) => void;
+    }
+
+    let { options, busy, onSelect }: Props = $props();
+
+    let searchQuery = $state('');
+    let showDropdown = $state(false);
+
+    const filteredOptions = $derived<TagView[]>(
+        searchQuery.trim()
```
ikondrat: I would make trim only once, otherwise I just make it several times: once for the condition and once for actual usage

lightly_studio_view/src/lib/components/TagsMenu/TagAssignInput.svelte
```diff
@@ -0,0 +1,91 @@
...
+            ? options.filter((t) => t.name.toLowerCase().includes(searchQuery.trim().toLowerCase()))
+            : options
+    );
+
+    const hasExactMatch = $derived(
+        options.some((t) => t.name.toLowerCase() === searchQuery.trim().toLowerCase())
```
ikondrat: plean define variable for searchQuery.trim() result to make once

lightly_studio_view/src/lib/components/TagsMenu/TagAssignInput.svelte
```diff
@@ -0,0 +1,91 @@
...
+    }
+</script>
+
+<div class="relative pt-2">
+    <Input
+        type="text"
+        placeholder="Assign tag to selection"
+        bind:value={searchQuery}
+        onkeydown={handleKeydown}
+        oninput={() => (showDropdown = true)}
```
ikondrat: this way you create function on every render


I would rather avoid it with oninput={handleOnInput}

ikondrat (approved):
lgtm

minor: It would be better to have a variable with the result of searchQuery.trim()  once
I think it is defined 3-4 times

ikondrat (approved):

PR #927, opened 2026-04-13
[DO NOT MERGE] Query prototype 

JonasWurst:
Query form GUI prototype


lukas-lightly (commented):
Regarding serialization (*wire): I'd strongly suggest using an existing library. E.g. in plain Python we can do
```
from dataclasses import dataclass, asdict
import json

@dataclass
class User:
    id: int
    name: str

user = User(id=42, name="Alex")

# Convert to dict first, then to JSON
json_data = json.dumps(asdict(user))
```

Which I think is more than enough in this case, but it needs to be checked. There's also `user.model_dump_json()` in `pydantic`.

lukas-lightly:
> Regarding serialization (*wire): I'd strongly suggest using an existing library. E.g. in plain Python we can do
> 
> ```
> from dataclasses import dataclass, asdict
> import json
> 
> @dataclass
> class User:
>     id: int
>     name: str
> 
> user = User(id=42, name="Alex")
> 
> # Convert to dict first, then to JSON
> json_data = json.dumps(asdict(user))
> ```
> 
> Which I think is more than enough in this case, but it needs to be checked. There's also `user.model_dump_json()` in `pydantic`.

Consider also a recursive example:

```
from pydantic import BaseModel
from typing import List, Union

class BoolLeaf(BaseModel):
    kind: str = "leaf"
    value: bool

class LogicalNode(BaseModel):
    kind: str = "node"
    operator: str
    # Self-reference works out of the box in modern Pydantic
    children: List[Union["LogicalNode", BoolLeaf]]

# Usage
tree = LogicalNode(
    operator="AND",
    children=[
        BoolLeaf(value=True),
        LogicalNode(operator="OR", children=[BoolLeaf(value=False)])
    ]
)

print(tree.model_dump_json(indent=2))
```

PR #930, opened 2026-04-13
Add AnnotationAccessor

lukas-lightly:
## What has changed and why?
It allows querying for annotations, so far only object detections are supported and only x, y, width, height and label.

## How has it been tested?
Added tests

- [ ] Yes

## Summary by CodeRabbit

* **New Features**
  * Query and filter samples by object-detection annotation properties (width, height, x/y coordinates, label) via a new annotation accessor in the public API.

* **Tests**
  * Added tests validating SQL generation and runtime filtering for object-detection annotation queries.
  * Updated tag-expression tests to use the accessor-based expression pattern.

### DIFF
--- a/lightly_studio/src/lightly_studio/core/dataset_query/annotation_expression.py
+++ b/lightly_studio/src/lightly_studio/core/dataset_query/annotation_expression.py
@@ -1,0 +1,131 @@
+"""Annotation field classes for building dataset queries on sample annotations."""
+
+from __future__ import annotations
+
+from dataclasses import dataclass
+from typing import Any
+
+from sqlalchemy import ColumnElement, and_
+from sqlalchemy.orm import Mapped
+from sqlmodel import col
+
+from lightly_studio.core.dataset_query.field import ComparableField, NumericalField
+from lightly_studio.core.dataset_query.match_expression import MatchExpression
+from lightly_studio.models.annotation.annotation_base import (
+    AnnotationBaseTable,
+    AnnotationType,
+)
+from lightly_studio.models.annotation.object_detection import ObjectDetectionAnnotationTable
+from lightly_studio.models.annotation_label import AnnotationLabelTable
+from lightly_studio.models.sample import SampleTable
+
+
+# Ignore PLW1641 because `==` and `!=` create query conditions here, so these
+# classes do not need normal hash behavior.
+class ObjectDetectionNumericalField:  # noqa: PLW1641
+    """Numerical field for object detection properties."""
+
+    def __init__(self, column: Mapped[int | float]) -> None:
+        """Initialize the object detection numerical field.
+
+        Args:
+            column: The database column this field represents.
+        """
+        self.field = NumericalField(column)
+
+    def __gt__(self, other: float | int) -> ObjectDetectionMatchExpression:
+        """Create a greater-than expression."""
+        return ObjectDetectionMatchExpression(self.field.__gt__(other))
+
+    def __lt__(self, other: float | int) -> ObjectDetectionMatchExpression:
+        """Create a less-than expression."""
+        return ObjectDetectionMatchExpression(self.field.__lt__(other))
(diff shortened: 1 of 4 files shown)


lightly_studio/src/lightly_studio/core/dataset_query/annotation_expression.py
```diff
@@ -0,0 +1,133 @@
...
+    # TODO(lukas, 4/2026): try using Mapped[T] as the type of relationship
+    relationship: str = "object_detection_details"
+
+    def get(self) -> ColumnElement[bool]:
+        """Get the object detection match expression."""
+        # Get the relationship property from AnnotationBaseTable
+        rel = getattr(AnnotationBaseTable, self.relationship)
+        condition = rel.has(self.criterion.get())
+
+        return SampleTable.annotations.any(
```
lukas-lightly, line 143: We only allow `SampleTable.annotations.any(...)` queries. I'd keep this in the initial PR but let's see if we need to make this more flexible in the future. I can add a TODO.

lightly_studio/src/lightly_studio/core/dataset_query/annotation_expression.py
```diff
@@ -0,0 +1,133 @@
...
+    """Expression for checking if a sample has an object detection matching a criterion."""
+
+    criterion: MatchExpression
+    # TODO(lukas, 4/2026): try using Mapped[T] as the type of relationship
+    relationship: str = "object_detection_details"
+
+    def get(self) -> ColumnElement[bool]:
+        """Get the object detection match expression."""
+        # Get the relationship property from AnnotationBaseTable
+        rel = getattr(AnnotationBaseTable, self.relationship)
```
lukas-lightly: This part is ugly and I tried using `Mapped` for `relationship`, but couldn't make it work in a few minutes.
In the interest of time let's continue with a TODO.

lightly_studio/tests/core/dataset_query/test_tags_expression.py
```diff
@@ -2,21 +2,21 @@
 
 from sqlmodel import Session, select
 
-from lightly_studio.core.dataset_query.image_sample_field import ImageSampleField
+from lightly_studio.core.dataset_query.tags_expression import TagsAccessor
```
lukas-lightly, line 5: I changed this file because it tests `ImageSampleField` instead of `TagsAccessor`

lightly_studio/src/lightly_studio/core/dataset_query/annotation_expression.py
lukas-lightly: Done

lightly_studio/src/lightly_studio/core/dataset_query/annotation_expression.py
```diff
@@ -0,0 +1,133 @@
...
+        condition = rel.has(self.criterion.get())
+
+        return SampleTable.annotations.any(
+            and_(
+                col(AnnotationBaseTable.annotation_type) == AnnotationType.OBJECT_DETECTION,
+                condition,
+            )
```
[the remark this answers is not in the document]
lukas-lightly: I added a test in 6b806c0 testing this behavior and the test is correct. Can you explain it?

lightly_studio/src/lightly_studio/core/dataset_query/annotation_expression.py
lukas-lightly: Codex has a good comment that this could allow two filter like "width > 100" and "height == 100" matching two different detections on the same sample instead of one detection satisfying both constraints.

However, I added a test for this and it worked as expected, although I can't explain it completely at this moment.

lukas-lightly:
I'm closing this PR, we need a new approach. The generated SQL expression is wrong and we probably need a whole new API for it.

PR #933, opened 2026-04-13
Rework tag workflow

JonasWurst:
## What has changed and why?

this pr moves the tagging worfklow form the dialog to the sidepanel. it simplifies the ux.

## How has it been tested?

tests are added

- [x] Yes

## Summary by CodeRabbit

* **New Features**
  * Create and assign tags from the side panel in all grid views (samples, annotations, video frames, videos); assignment respects current selection and disables input when nothing is selected.

* **Tests**
  * Added end-to-end tests for tag creation/assignment from the side panel across views and unit tests covering tag creation, assignment flows, and UI states.


lightly_studio_view/src/lib/components/TagsMenu/TagsMenu.svelte
```diff
@@ -3,32 +3,97 @@
...
+
+    const { getSelectedSampleIds, selectedSampleAnnotationCropIds } = useGlobalStorage();
+
+    const selectedSampleIds = $derived(getSelectedSampleIds(collection_id));
+    const hasSelection = $derived(
+        tagKind === 'annotation'
+            ? ($selectedSampleAnnotationCropIds[collection_id]?.size ?? 0) > 0
+            : $selectedSampleIds.size > 0
     );
+    const selectedIds = $derived(
```
ikondrat, line 30: what is selectedSampleAnnotationCropIds? 
I am not sure what is crop? 

I would expect here to see $selectedAnnotationIds by logic

lightly_studio_view/src/lib/components/TagsMenu/TagsMenu.svelte
```diff
@@ -3,32 +3,97 @@
...
+                path: { collection_id, tag_id: existingTag.tag_id },
+                body: { sample_ids: [...selectedIds] }
+            });
+            assignBusy = false;
+            if (response.error) {
+                toast.error('Failed to assign tag. Please try again.');
+                return;
+            }
+        } else {
+            const createResponse = await createTag({
```
ikondrat: please handle error properly 

async operations should be handled with try/catch

https://wesbos.com/javascript/12-advanced-flow-control/71-async-await-error-handling

lightly_studio_view/src/lib/components/TagsMenu/TagsMenu.test.ts
```diff
@@ -0,0 +1,170 @@
...
+            response: mockResponse
+        });
+    });
+
+    it('assigns an existing tag to the selected samples', async () => {
+        mocks.selectedSampleIdsByCollection = {
+            'collection-1': new Set(['sample-1', 'sample-2'])
+        };
+
+        renderComponent('samples');
```
ikondrat: please use rather render

```
render(TagsMenu, {
        props: {
            gridType: 'samples',
             ...defautlProps
        }
    }

```

The test will be more readable

We need to describe in the test what we changed in the component render

lightly_studio_view/src/lib/components/TagsMenu/TagsMenu.test.ts
```diff
@@ -0,0 +1,170 @@
...
+            },
+            body: {
+                sample_ids: ['annotation-1', 'annotation-2']
+            }
+        });
+        expect(mocks.loadTags).toHaveBeenCalled();
+    });
+
+    it('disables tag assignment when no samples are selected', () => {
+        renderComponent('samples');
```
ikondrat: I think using renderComponent does not really simplify the test, because it is not clear what ' renderComponent('samples')' means. I would rather use direct rendering by changing the gridType

ikondrat (changes requested):
Please handle errors with try/catch

The rest of the comments are optional

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added thumbnail quality setting in the Settings dialog. Enable "High Quality Thumbnails" to load compressed JPEG thumbnails in grid views, reducing bandwidth for large datasets.
 - Added a floating selection panel to grid pages.
 - Added ability to add and create tags directly from the sample and annotation detail view.
+- Added ability to add and create tags directly from the side panel in the grid view.
```
michal-lightly:
```suggestion
- You can add and create tags directly from the sample and annotation detail view.
- You can add and create tags directly from the side panel in the grid view.
```
or
```suggestion
- Tags can be created and assigned directly from the sample and annotation detail view.
- Tags can be created and assigned directly from the side panel in the grid view.
```

lightly_studio_view/src/lib/components/TagsMenu/TagsMenu.svelte
JonasWurst: Those are the selectedAnnotationIds, but the name is just not nice. It was done like this also in the dialog: https://github.com/lightly-ai/lightly-studio/blob/f59cbcb3b36761221342e1feeddd4be1db03e84c/lightly_studio_view/src/lib/components/TagCreateDialog/TagCreateDialog.svelte#L131

lightly_studio_view/e2e/pages/annotations-page.ts
```diff
@@ -106,12 +110,28 @@ export class AnnotationsPage {
     }
 
     async createTag(tagName: string) {
-        await this.page.getByTestId('tag-create-dialog-button').click();
-
-        await this.page.getByTestId('tag-create-dialog-input').fill(tagName);
+        const createTagResponsePromise = this.page.waitForResponse(
+            (response) =>
+                response.request().method() === 'POST' &&
+                /\/tags(?:\?|$)/.test(response.request().url()) &&
```
ikondrat: I would drop group capturing and have the similar check as below
```
response.url().includes('/tags')
```

ikondrat (changes requested):
I think we don't need group capturing in regexp

lightly_studio_view/src/lib/components/TagsMenu/TagsMenu.svelte
ikondrat: Shall we create task to reduce confusion and rename
selectedSampleAnnotationCropIds 
to
selectedAnnotationIds
?


because it will continue to be confusing

ikondrat (approved):
I think we don't need group capturing in a regexp

Otherwise lgtm

PR #946, opened 2026-04-14
Introduce createSourceMaskManager

LeonardoRosaa:
## What has changed and why?

This PR introduces the `createSourceMaskManager` used by the `useInstanceSegmentationPreview` hook.

## How has it been tested?

Unit tests.

- [ ] Yes

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive test suite validating mask-to-canvas conversion and canvas-to-mask operations with proper alpha channel handling.

* **New Features**
  * Implemented canvas-based mask manager for instance segmentation previews, enabling bidirectional conversion between binary masks and canvas rendering with automatic canvas recreation on dimension changes.


lightly_studio_view/src/lib/hooks/useInstanceSegmentationPreview/sourceMaskManager.ts
```diff
@@ -0,0 +1,88 @@
...
+    const drawMaskToSourceCanvas = (mask: Uint8Array | null, width: number, height: number) => {
+        const sourceContext = ensureSourceMaskContext(width, height);
+        if (!sourceContext || !sourceMaskCanvas) return;
+
+        sourceContext.clearRect(0, 0, sourceMaskCanvas.width, sourceMaskCanvas.height);
+        if (!mask) return;
+
+        // Encode binary mask as alpha-only pixels.
+        const imageData = sourceContext.createImageData(width, height);
+        const pixels = imageData.data;
```
ikondrat: Have you considered using Uint32Array?

Chatty says it will improve  performance 4x by processing 4 bytes at once instead of individual bytes

lightly_studio_view/src/lib/hooks/useInstanceSegmentationPreview/sourceMaskManager.ts
```diff
@@ -0,0 +1,88 @@
...
+
+        for (let i = 0; i < mask.length; i++) {
+            if (mask[i] !== 1) continue;
+            pixels[i * 4 + 3] = 255; // +3 selects alpha in RGBA.
+        }
+
+        sourceContext.putImageData(imageData, 0, 0);
+    };
+
+    const toBinaryMaskFromSourceCanvas = (width: number, height: number) => {
```
ikondrat: I think you should use passed width/height not only for creation of new sourceMaskCanvas

lightly_studio_view/src/lib/hooks/useInstanceSegmentationPreview/sourceMaskManager.ts
```diff
@@ -0,0 +1,88 @@
...
+        if (
+            !sourceMaskCanvas ||
+            sourceMaskCanvas.width !== width ||
+            sourceMaskCanvas.height !== height
+        ) {
+            // Recreate when image size changes.
+            sourceMaskCanvas = document.createElement('canvas');
+            sourceMaskCanvas.width = width;
+            sourceMaskCanvas.height = height;
+            sourceMaskContext = null;
```
ikondrat: It looks like you need to recreate context here

And all logic looks like 

```
const needToRecreate = !sourceMaskCanvas ||
            sourceMaskCanvas.width !== width ||
            sourceMaskCanvas.height !== height;

if (needToRecreate) {
    sourceMaskCanvas = document.createElement('canvas');
    sourceMaskContext = sourceMaskCanvas.getContext('2d');
}

return sourceMaskContext;
````

I would avoid to split logic to several ifs

File: lightly_studio_view/src/lib/hooks/useInstanceSegmentationPreview/sourceMaskManager.ts [hunk 4]
```diff
@@ -0,0 +1,88 @@
...
+        }
+
+        if (sourceMaskContext) return sourceMaskContext;
+
+        sourceMaskContext = sourceMaskCanvas.getContext('2d');
+        return sourceMaskContext;
+    };
+
+    const drawMaskToSourceCanvas = (mask: Uint8Array | null, width: number, height: number) => {
```
ikondrat: I would rather add a dedicated clearSourceCanvas

passing null to the function called drawMaskToSourceCanvas is rather confusing

lightly_studio_view/src/lib/hooks/useInstanceSegmentationPreview/sourceMaskManager.ts
```diff
@@ -0,0 +1,88 @@
...
+    toBinaryMaskFromSourceCanvas: (width: number, height: number) => Uint8Array | null;
+    getSourceMaskCanvas: () => HTMLCanvasElement | null;
+    destroy: () => void;
+}
+
+export function createSourceMaskManager(): SourceMaskManager {
+    let sourceMaskCanvas: HTMLCanvasElement | null = null;
+    let sourceMaskContext: CanvasRenderingContext2D | null = null;
+
+    const ensureSourceMaskContext = (width: number, height: number) => {
```
ikondrat: I think the naming of this function is not perfect, it doesn't reflect what it actually returns context

I think smth like 
getMaskContext
or
getOrCreateMaskContext

would make more sense

wdyt?

Would be much cleaner to have smth like 
```
export interface MaskManager {
    getMaskContext(width, height): ...
    drawMask(mask, width, height): ...
    clear(width, height): ...
    extractMask(width, height): ...
    getCanvas(): ...
    destroy(): ...
}
```

lightly_studio_view/src/lib/hooks/useInstanceSegmentationPreview/sourceMaskManager.ts
```diff
@@ -0,0 +1,88 @@
+interface SourceMaskManager {
```
ikondrat: I think "source" prefix is rather redundant here, can we just use canvas, context (locals) or maskCanvas, maskContext?

lightly_studio_view/src/lib/hooks/useInstanceSegmentationPreview/sourceMaskManager.ts
LeonardoRosaa: Let's keep it as is since we only need to assign the alpha channel.

ikondrat (approved):
the comments are rather nitpicks and optional

lightly_studio_view/src/lib/hooks/useInstanceSegmentationPreview/sourceMaskManager.ts
LeonardoRosaa: We should cover the case where the mask can be null. I'd not abstract this since it is clear now. WDTY?

PR #952, opened 2026-04-15
Faster preview rendering while erasing

LeonardoRosaa:
## What has changed and why?

Uses `useInstanceSegmentationPreview` to improve preview performance rendering while erasing, which renders segmentation previews directly on a canvas and coalesces updates with requestAnimationFrame, reducing unnecessary work during drawing.

## How has it been tested?

Unit tests + manual tests.

- [ ] Yes

## Release Notes

* **Refactor**
  * Enhanced eraser tool preview rendering mechanism for more responsive and immediate visual feedback during erasing operations.


ikondrat:
I’m still not getting immediate feedback when using the eraser.

Here’s a video showing the behavior:

https://github.com/user-attachments/assets/b26ac08c-26c8-4839-b64f-09157735e28a



Current issues:

- When I draw a segmentation, it briefly appears and then disappears (blinks).
- The stroke starts with one color, but changes to a different color once I finish drawing.
- Erasing itself feels smooth (actually better than on the current main branch), but once I finish, the erased area briefly reappears and flickers. As a result, the feedback doesn’t feel immediate.

ikondrat (changes requested):
The code looks solid, but if the goal of this task was to provide immediate feedback when using the eraser, that part still isn’t really achieved.

Even though the erasing isn’t as discrete as it is on the main branch, the current behavior—with the erased result flickering—creates a confusing visual effect. It doesn’t feel like immediate feedback to the user.

ikondrat (approved):

PR #962, opened 2026-04-16
Show saving indicator while annotation edits are being persisted

LeonardoRosaa:
## What has changed and why?

This PR adds a save-pending UI state for annotation editing actions so users get clear feedback while async save operations are in flight.

Key changes:

- Added a `Saving` indicator (spinner + label) in `SampleDetailsImageContainer`, rendered next to the zoom panel controls.

## How has it been tested?

Unit tests + manual tests.

loading

- [x] Yes

## Release Notes

* **Bug Fixes**
  * Annotation editing in sample details now displays a Saving indicator while changes are being persisted, providing clearer feedback during erasing, brush strokes, and bounding box creation operations.


lightly_studio_view/src/lib/components/SampleDetails/savePendingChange.ts
```diff
@@ -0,0 +1,6 @@
+export type SavePendingChange =
+    | boolean
+    | {
+          token: string;
+          isPending: boolean;
+      };
```
ikondrat: I think I need an explanation why you decided to make union based type instead of 2 different types

For me, it seems that it would be just harder to handle

lightly_studio_view/src/lib/components/SampleDetails/SampleDetailsImageContainer/SampleDetailsImageContainer.svelte
```diff
@@ -180,6 +185,27 @@
 
         return true;
     });
+
+    const handleSavePendingChange = (pendingChange: SavePendingChange) => {
```
ikondrat: let's have consistent type, I think you can always have pendingChange.isPending

lightly_studio_view/src/lib/components/SampleDetails/savePendingChange.ts
LeonardoRosaa: The Boolean union is no longer required. I’ve removed it and simplified the design.

lightly_studio_view/src/lib/components/SampleDetails/savePendingChange.ts
```diff
@@ -0,0 +1,6 @@
+export type SavePendingChange =
+    | boolean
+    | {
+          token: string;
```
ikondrat: I am wondering if "operation" would suit better here

lightly_studio_view/src/lib/components/ZoomPanel/ZoomPanel.svelte
```diff
@@ -11,7 +11,8 @@
         isZoominDisabled,
         isZoomoutDisabled,
         isResetDisabled,
-        content
+        content,
+        rightContent
```
ikondrat: rightContent kind of conflicts with naming pattern here in props

see 
onZoom{smth}
isReset{smth}

I think calling it like 

content{smth} would follow pattern better

e.g.
contentMain
contentRight
contentLeft 
etc

lightly_studio_view/src/lib/components/SampleDetails/SampleDetailsImageContainer/SampleDetailsImageContainer.svelte
ikondrat: I am pretty sure you would need to do it somewhere else

Maybe it does make sense to move it out of the component

to have smth like
```
const pendingOps = usePendingOperations();
const { isPending, pendingCount } = pendingOps;

async function saveData() {
  const operationId = `save-${Date.now()}`;
  pendingOps.startOperation(operationId);
 
 try {
      await api.saveData();
 } finally {
     pendingOps.completeOperation(operationId);
}

<button disabled={$isPending} onclick={save}>
   Save
</button>

```
wdyt?

lightly_studio_view/src/lib/components/SampleDetails/SampleInstanceSegmentationRect/SampleInstanceSegmentationRect.svelte
```diff
@@ -201,14 +234,22 @@
...
 
-        brushApi.finishBrush(
-            updatedMask,
-            targetAnnotation,
-            $labels.data ?? [],
-            updateAnnotation,
-            annotationLabelContext.lockedAnnotationIds
-        );
-        setIsDrawing(false);
+        const pendingToken = startFinishBrushPending();
```
ikondrat: I would recommend to consolidate pending operation functions into the hook

I gave example above, here you technically need only the operation name

ikondrat (changes requested):
let's consolidate pending state handling into the hook and add jsdoc comments about it. 
To describe functionality better

I would recommend naming it rather **operation** instead of "token" because it is actually state handling for operation

lightly_studio_view/src/lib/hooks/usePendingSaveTokens/usePendingSaveTokens.ts
```diff
@@ -0,0 +1,51 @@
...
+    token: string;
+    isPending: boolean;
+};
+
+type UsePendingSaveTokensProps = {
+    tokenPrefix: string;
+    onPendingChange?: (pendingOperation: PendingSaveOperation) => void;
+};
+
+export const usePendingSaveTokens = ({
```
ikondrat: Please add proper jsdocs describing what it is doing

lightly_studio_view/src/lib/hooks/usePendingSaveTokens/usePendingSaveTokens.ts
```diff
@@ -0,0 +1,51 @@
...
+    let pendingSaveTokenCounter = 0;
+
+    const setPending = (pendingOperation: PendingSaveOperation) => {
+        onPendingChange?.(pendingOperation);
+    };
+
+    // Manage pending save operations with deterministic token lifecycle events.
+    const startPending = () => {
+        pendingSaveTokenCounter += 1;
+        const token = `${tokenPrefix}-${pendingSaveTokenCounter}`;
```
ikondrat: I am still struggling to understand why you call a token

It seems like you need the operation name and the counter. Correct me if I'm wrong

I would then use the operation name
and the counter would put to setPending({ token, isPending: true, counter: currentCounter + 1 });

lightly_studio_view/src/lib/components/SampleDetails/usePendingSaveState.ts
```diff
@@ -0,0 +1,27 @@
+import { derived, writable } from 'svelte/store';
+import type { SavePendingChange } from './savePendingChange';
+
+export const usePendingSaveState = () => {
+    const pendingSaveTokens = writable<string[]>([]);
+    const isSavePending = derived(pendingSaveTokens, ($tokens) => $tokens.length > 0);
+
+    const handleSavePendingChange = (pendingChange: SavePendingChange) => {
```
ikondrat: It seems handlePendingChange would be better

not sure what "Save" means here

lightly_studio_view/src/lib/hooks/usePendingSaveTokens/usePendingSaveTokens.ts
ikondrat: I would also call it a more general way

smth like 
usePendingOperations 

not sure why we need "Save" in the naming?

LeonardoRosaa: Initially, it was implemented to solve a specific save request, so it seems we should make it a bit more generic now.

ikondrat (approved):
great job

PR #965, opened 2026-04-16
Fix: Rename of tags is done properly

IgorSusmelj:
## What has changed and why?

- Change rename behaviour in backend. Before we were deleting + inserting a new tag. This would break the links between the tables. Now we do a proper rename. 

## How has it been tested?

- Additional tests

- [ ] Yes

### DIFF
--- a/lightly_studio/src/lightly_studio/resolvers/tag_resolver.py
+++ b/lightly_studio/src/lightly_studio/resolvers/tag_resolver.py
@@ -67,18 +67,22 @@ def update(session: Session, tag_id: UUID, tag_data: TagUpdate) -> TagTable | No
     if conflicting_tag and conflicting_tag.tag_id != tag_id and conflicting_tag.kind == tag.kind:
         raise IntegrityError(statement=None, params=None, orig=Exception("Tag already exists"))
 
-    sample_ids = [sample.sample_id for sample in tag.samples]
+    sample_ids = [
+        sample_id
+        for sample_id in session.exec(
+            select(SampleTagLinkTable.sample_id).where(col(SampleTagLinkTable.tag_id) == tag_id)
+        ).all()
+        if sample_id is not None
+    ]
 
     session.exec(
         sqlmodel.delete(SampleTagLinkTable).where(col(SampleTagLinkTable.tag_id) == tag_id)
     )
     session.commit()
 
-    # due to duckdb/OLAP optimisations, update operations effecting unique
-    # constraints (e.g colums) will lead to a unique constraint violation.
-    # This is due to a update is implemented as delete+insert. The error
-    # happens only within the same session.
-    # To fix it, we can delete, commit + insert a new tag.
+    # DuckDB checks unique constraints too eagerly on updates, so we recreate
+    # the tag row. We must temporarily delete the link rows first, otherwise
+    # DuckDB rejects deleting the tag row because of the foreign key.
     # https://duckdb.org/docs/sql/indexes#over-eager-unique-constraint-checking
     session.delete(tag)
     session.commit()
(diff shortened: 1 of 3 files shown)


lightly_studio/src/lightly_studio/resolvers/tag_resolver.py
```diff
@@ -69,14 +83,23 @@ def update(session: Session, tag_id: UUID, tag_data: TagUpdate) -> TagTable | No
...
-    tag_updated.updated_at = datetime.now(timezone.utc)
+    tag_updated = TagTable(
+        tag_id=tag.tag_id,
+        collection_id=tag.collection_id,
+        name=tag_data.name,
+        description=tag_data.description,
+        kind=tag.kind,
+        created_at=tag.created_at,
+        updated_at=datetime.now(timezone.utc),
+    )
```
JonasWurst, line 98: why is this change required?

lightly_studio/src/lightly_studio/resolvers/tag_resolver.py
```diff
@@ -69,14 +83,23 @@ def update(session: Session, tag_id: UUID, tag_data: TagUpdate) -> TagTable | No
...
 
     session.add(tag_updated)
     session.commit()
+
+    for sample_id in sample_ids:
+        session.merge(SampleTagLinkTable(sample_id=sample_id, tag_id=tag_id))
```
JonasWurst: Can we do it in bulk? This might lead to very slow renaming for large payloads

lightly_studio/tests/api/routes/api/test_dataset_tag.py
```diff
@@ -57,3 +57,30 @@ def test_delete_tag__deletes_tag_with_sample_links(
...
+        },
+    )
+
+    assert response.status_code == HTTP_STATUS_OK
+    assert response.json()["name"] == "renamed_tag"
+
+    updated_tag = tag_resolver.get_by_id(session=db_session, tag_id=tag.tag_id)
+    assert updated_tag is not None
+    assert updated_tag.name == "renamed_tag"
+    assert [sample.sample_id for sample in updated_tag.samples] == [image.sample.sample_id]
```
JonasWurst: we only have one image. can just directly access

lightly_studio/tests/resolvers/test_tag_resolver.py
```diff
@@ -109,6 +109,33 @@ def test_update_tag(db_session: Session) -> None:
     assert tag_updated.name == data_update.name
 
 
+def test_update_tag__preserves_sample_links(db_session: Session) -> None:
```
JonasWurst, line 112: Lets add a test for the error-out opt if a name is already used by another tag.

JonasWurst (changes requested):

lightly_studio/tests/resolvers/test_tag_resolver.py
IgorSusmelj: You mean `test_update_tag__unique_tag_name`? There is a test already

lightly_studio/src/lightly_studio/resolvers/tag_resolver.py
IgorSusmelj: Still the same issue as before (see comment, I updated it). 
DuckDB currently fails this rename as a normal update because of how it enforces unique constraints internally. The workaround is to recreate the tag row instead, and to preserve the existing sample-tag assignments we temporarily remove and then restore the link rows.

lightly_studio/tests/resolvers/test_tag_resolver.py
JonasWurst: I see. I only saw that you added the check.

lightly_studio/src/lightly_studio/resolvers/tag_resolver.py
ikondrat: yes deleting is a mess in duckdb

ikondrat (approved):

PR #998, opened 2026-04-22
feat: add button to get value

ikondrat:
## What has changed and why?

Introduce button to get value from editor + test for component

## How has it been tested?

By test

- [ ] Yes

## Summary by CodeRabbit

* **New Features**
  * Optional "Save" button for the query editor; shown when a save handler is provided and disabled in read-only mode.

* **Documentation**
  * Storybook story added to expose and trigger the Save action for previewing save behavior.

* **Tests**
  * New tests confirm Save button visibility, click handling (handler invoked once with current value), and disabled state when read-only.

### DIFF
--- a/lightly_studio_view/src/lib/components/QueryEditor/QueryEditor.svelte
+++ b/lightly_studio_view/src/lib/components/QueryEditor/QueryEditor.svelte
@@ -3,7 +3,7 @@
     import { Button } from '$lib/components/ui/button';
     import { LIGHTLY_QUERY_DEFAULT_VALUE } from './monaco-lightly-query.js';
     import { useLightlyQueryEditor } from './useLightlyQueryEditor.js';
-    import Typography from '../Typography/Typography.svelte';
+    import Typography from '$lib/components/Typography/Typography.svelte';
 
     interface QueryEditorProps {
         value?: string;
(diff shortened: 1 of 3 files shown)


lightly_studio_view/src/lib/components/QueryEditor/QueryEditor.svelte
```diff
@@ -32,13 +35,56 @@
...
 <style>
     .query-editor {
         width: 100%;
         border: 1px solid rgb(60 60 60);
         border-radius: 0.5rem;
         overflow: hidden;
+        background: #1e1e1e;
+    }
+
+    .query-editor__toolbar {
```
michal-lightly: Don't we have already styled components for toolbars and buttons?

lightly_studio_view/src/lib/components/QueryEditor/QueryEditor.test.ts
```diff
@@ -0,0 +1,47 @@
...
+    LIGHTLY_QUERY_LANGUAGE_ID: 'lightly-query',
+    LIGHTLY_QUERY_THEME_ID: 'lightly-query-theme',
+    registerLightlyQueryMonacoLanguage: vi.fn()
+}));
+
+describe('QueryEditor', () => {
+    it('does not render the toolbar when onSave is not provided', () => {
+        render(QueryEditor, { props: { value: 'query' } });
+
+        expect(screen.queryByRole('button', { name: 'Save' })).not.toBeInTheDocument();
```
michal-lightly, line 21: We will likely rename the button. Wouldn't it be better to use a test id?

ikondrat: I think it is fine to update the test if we rename the button

lightly_studio_view/src/lib/components/QueryEditor/QueryEditor.svelte
ikondrat: I already use the strandart button, there is no custom styles for the button in the file

About toolbar, we don't have anything shared and suitable, and I doubt what we need it to be honest

michal-lightly (approved):

PR #1005, opened 2026-04-23
feat: introduce hook to control query filter

ikondrat:
## What has changed and why?

Introduce query filter to control filter query editing state

## How has it been tested?

By unit tests

- [ ] Yes

## Release Notes

* **New Features**
  * Added query filter functionality with toggle capability for editing mode, controlled via feature flag configuration.

* **Tests**
  * Comprehensive test coverage validates query filter behavior, including toggle state management and feature flag integration.

### DIFF
--- a/lightly_studio_view/src/lib/hooks/useQueryFilter/useQueryFilter.ts
+++ b/lightly_studio_view/src/lib/hooks/useQueryFilter/useQueryFilter.ts
@@ -1,0 +1,29 @@
+import { derived, get, readonly, writable, type Readable } from 'svelte/store';
+import { useFeatureFlags } from '$lib/hooks/useFeatureFlags/useFeatureFlags';
+
+export const QUERY_FILTER_FEATURE_FLAG = 'query_filter';
+
+interface UseQueryFilterReturn {
+    isEnabled: Readable<boolean>;
+    isEditing: Readable<boolean>;
+    toggleEditing: () => void;
+}
+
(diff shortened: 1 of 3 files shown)


lightly_studio_view/src/lib/hooks/useQueryFilter/useQueryFilter.ts
```diff
@@ -0,0 +1,29 @@
...
+
+interface UseQueryFilterReturn {
+    isEnabled: Readable<boolean>;
+    isEditing: Readable<boolean>;
+    toggleEditing: () => void;
+}
+
+export function useQueryFilter(): UseQueryFilterReturn {
+    const { featureFlags } = useFeatureFlags();
+    const isEnabled = derived(featureFlags, ($flags) => $flags.includes(QUERY_FILTER_FEATURE_FLAG));
```
michal-lightly, line 14: The feature flag should not be here. It should only toggle visibility of the "Query" section - the visibility of the button that opens the query editor.

michal-lightly (changes requested):

lightly_studio_view/src/lib/hooks/useQueryFilter/useQueryFilter.ts
ikondrat: hmm in the second though if we would move out feature flag control outside we don't need hook

I will close PR for now

PR #1043, opened 2026-04-29
Add: Backend endpoint and resolvers for only retrieving sample ids based on filters

IgorSusmelj:
## What has changed and why?

- Add resolver support to only get sample ids for images, video frames and annotations
- Expose the new resolvers in new endpoints so we can use them by the frontend
- There will be a follow up issue updating the frontend. We will use the new endpoints to select all samples based on the current active filters. Important, the big difference is that this new endpoints now only return sample ids and nothing else.

## How has it been tested?

- New unit tests for happy path. We don't need extensive tests as we already cover the query mechanism in other tests.

- [X] Yes

## Summary by CodeRabbit

* **New Features**
  * POST endpoints to retrieve sample IDs for images, video frames, and annotations; support optional filters and select-all workflows.

* **Tests**
  * Integration and resolver tests added for sample-ID retrieval with and without filters across images, video frames, and annotations.

* **Documentation**
  * Unreleased changelog entry noting the new sample-ID endpoints.


lightly_studio/src/lightly_studio/api/routes/api/annotation.py
```diff
@@ -85,6 +85,28 @@ def read_annotations(
...
 
 
+class ReadAnnotationSampleIdsRequest(BaseModel):
+    """Request body for reading matching annotation sample ids."""
+
+    filters: AnnotationsFilter | None = Field(None, description="Filter parameters for annotations")
+
+
+@annotations_router.post("/annotations/sample_ids", response_model=list[UUID])
+def get_annotation_sample_ids(
```
horatiualmasan, line 95: Please add tests also for all the new api functions.

horatiualmasan (approved):
LGTM
Please add tests for the api functions

lightly_studio/src/lightly_studio/api/routes/api/annotation.py
IgorSusmelj: Ok

PR #1049, opened 2026-04-30
Temporarily disable E2E tests that are timing out

lukas-lightly:
## What has changed and why?
These tests are timing out since the evening of April 29.

## How has it been tested?
Just by eye-balling the CI runs

## Did you update [CHANGELOG.md](../CHANGELOG.md)?

- [ ] Yes
- [x] Not needed (internal change)

## Release Notes

* **Tests**
  * Disabled E2E test for brush tool keyboard navigation functionality across samples while creating masks as the feature undergoes refinement
  * Disabled E2E test for selecting all filtered images using keyboard shortcut functionality, pending further development
  * All test code and assertions remain intact for future re-enablement

PR #1055, opened 2026-04-30
Fix: Make E2E tests more stable

IgorSusmelj:
## What has changed and why?

- Two tests were skipped in #1049 because they timed out in CI. This PR fixes the underlying causes and re-enables both.                                                                       
- The other changes we implemented to address E2E issues:
  - select all with label filter: Was listening for a /images/list response that arrived before the label click event was bound. Fixed by using waitForRequestsToSettle and proper visibility assertions.
  - brush tool keyboard navigation: The route mock was matching too broadly on /annotations, intercepting unrelated requests like /annotations/count. Fixed by using a precise path regex for POST /annotations.
  - annotation details performance test: Was calling annotationsPage.goto() inside the measurement loop, which loads /, waits for the samples grid, and only then navigates to annotations. This measured unrelated bootstrap/navigation work 5 times and exceeded the 30s timeout. Fixed by resolving the annotations route once during setup, navigating directly to that route in each measured iteration, and reducing iterations to 3.
  - CI diagnostics: Server log is now uploaded as an artifact on failure, with traceback extraction and a backend-alive check, so the next mystery failure has context.

What is still open to be done in a separate PR:
- Split up E2E tests to make them run faster (currently takes over 9min)

## How has it been tested?

- Ran E2E tests locally 3 times
- Ran E2E tests in CI 4 times (4 out of 4 were a pass)

- [ ] Yes

## Summary by CodeRabbit

* **Tests**
  * Enabled multiple E2E tests (keyboard navigation, selection) and tightened request matching and waits
  * Improved reliability with explicit response timeouts, page-object construction from page, removed in-measurement reloads, and added UI readiness checks before interactions

* **Chores**
  * Improved CI failure diagnostics with structured log dumps, traceback extraction, process-alive checks, and log artifact upload on failure (5-day retention)
  * Improved query-editor expression handling for object-detection fields


lightly_studio_view/e2e/general/images-performance.e2e-test.ts
```diff
@@ -96,20 +98,17 @@ test('sample details renders within 5 seconds', async ({ page, samplesPage }) =>
...
+    const sampleDetailsPage = new SampleDetailsPage(page);
 
     await setNetworkThrottling(page, 'Fast4G');
 
     const result = await measureRenderAndMemory(async () => {
         await samplesPage.goto();
         await samplesPage.doubleClickFirstSample();
 
-        await page.getByText('Sample 1 of 128').waitFor();
+        await page.getByText('Sample 1 of 128').waitFor({ state: 'visible', timeout: 10000 });
```
ikondrat: now sure why we need it? I think it was working before without flackiness

lightly_studio_view/e2e/general/images-performance.e2e-test.ts
```diff
@@ -134,19 +133,16 @@ test('sample details renders next image within 5 seconds', async ({
...
+    const sampleDetailsPage = new SampleDetailsPage(page);
 
     await setNetworkThrottling(page, 'Fast4G');
 
     const result = await measureRenderAndMemory(async () => {
         await samplesPage.goto();
-        await samplesPage.getSampleByIndex(1).dblclick();
-        await page.getByText('Sample 2 of 128').waitFor({ state: 'visible' });
+        await samplesPage.doubleClickNthSample(1);
+        await page.getByText('Sample 2 of 128').waitFor({ state: 'visible', timeout: 10000 });
```
ikondrat: It is a bit conflicts with the name of the test, "sample details renders prev image within 5 seconds"

lightly_studio_view/e2e/general/images-performance.e2e-test.ts
```diff
@@ -134,19 +133,16 @@ test('sample details renders next image within 5 seconds', async ({
...
 });
 
-test('sample details renders prev image within 5 seconds', async ({
-    page,
-    samplesPage,
-    sampleDetailsPage
-}) => {
-    await samplesPage.goto();
+test('sample details renders prev image within 5 seconds', async ({ page }) => {
+    const samplesPage = new SamplesPage(page);
```
ikondrat: It looks like you don't need it, you will make await samplesPage.goto(); in the test anyway

lightly_studio_view/e2e/general/images-performance.e2e-test.ts
```diff
@@ -96,20 +98,17 @@ test('sample details renders within 5 seconds', async ({ page, samplesPage }) =>
...
 });
 
-test('sample details renders next image within 5 seconds', async ({
-    page,
-    samplesPage,
-    sampleDetailsPage
-}) => {
-    await samplesPage.goto();
+test('sample details renders next image within 5 seconds', async ({ page }) => {
+    const samplesPage = new SamplesPage(page);
```
ikondrat:
```suggestion

```

lightly_studio_view/e2e/general/images-performance.e2e-test.ts
```diff
@@ -96,20 +98,17 @@ test('sample details renders within 5 seconds', async ({ page, samplesPage }) =>
...
+    const sampleDetailsPage = new SampleDetailsPage(page);
```
ikondrat:
```suggestion

```

ikondrat (changes requested):
It looks like we can drop loading pages for perf tests
const annotationsPage = new AnnotationsPage(page);
const annotationDetailsPage = new AnnotationDetailsPage(page);

because we have loading pages inside the tests anyway

timeout 10000 with a combination of "in 5 sec" seems off
Shall we update the test name then?

ikondrat (approved):

PR #1080, opened 2026-05-07
Add endpoint read_annotation_collections

MalteEbner:
## What has changed and why?

Adds `GET /api/collections/{collection_id}/annotation_collections`, which lists every direct child collection of the given parent whose `sample_type == ANNOTATION`. The frontend menu uses it to discover available annotation collections (id + name) under the user's current collection.

The implementation mirrors the existing `read_annotation_labels` pattern but scopes the lookup by `parent_collection_id` rather than `dataset_id`, and reuses the existing `CollectionView` response model so no new model is introduced.

## How has it been tested?

new unittests

- [ ] Yes

## Summary by CodeRabbit

* **New Features**
  * New API endpoint available to retrieve annotation collections under a specified parent collection, with results ordered by creation date.

* **Tests**
  * Added test coverage for the annotation collections endpoint, including validation of collection filtering and retrieval of direct children in collection hierarchies.

### DIFF
--- a/lightly_studio/src/lightly_studio/api/routes/api/annotation.py
+++ b/lightly_studio/src/lightly_studio/api/routes/api/annotation.py
@@ -10,6 +10,7 @@ from fastapi.params import Query
 from pydantic import BaseModel, Field
 
 from lightly_studio.api.routes.api import annotations as annotations_module
+from lightly_studio.api.routes.api.collection import get_and_validate_collection_id
 from lightly_studio.api.routes.api.status import HTTP_STATUS_NOT_FOUND
 from lightly_studio.api.routes.api.validators import Paginated, PaginatedWithCursor
 from lightly_studio.db_manager import SessionDep
@@ -20,7 +21,8 @@ from lightly_studio.models.annotation.annotation_base import (
     AnnotationViewsWithCount,
     AnnotationWithPayloadAndCountView,
 )
-from lightly_studio.resolvers import annotation_resolver
+from lightly_studio.models.collection import CollectionTable, CollectionView
+from lightly_studio.resolvers import annotation_resolver, collection_resolver
 from lightly_studio.resolvers.annotation_resolver.get_all import (
     GetAllAnnotationsResult,
 )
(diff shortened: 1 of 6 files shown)


lightly_studio/src/lightly_studio/api/routes/api/annotation.py
```diff
@@ -37,6 +39,25 @@
...
 
 
+@annotations_router.get(
+    "/annotation_collections",
+    response_model=list[CollectionView],
+)
+def read_annotation_collections(
+    session: SessionDep,
+    collection: Annotated[
+        CollectionTable,
```
ikondrat, line 49: I would rather define a dedicated type(CollectionView is too generic to work with) to return to avoid returning unused props 

I would call it AnnotationCollectionView
with 
collection_id
name

and we don't have name of the collection in CollectionView and we need it

ikondrat (changes requested):
We should also include the name in the return value.

I’d avoid using overly abstract types like CollectionView here. Using something more specific, such as AnnotationCollectionView, makes the API clearer and helps avoid issues with unions or ambiguity when determining what the frontend actually receives.

lightly_studio/src/lightly_studio/api/routes/api/annotation.py
MalteEbner: Added a separate `AnnotationCollectionView`.

Note that the `CollectionView` contained the `name`, by inheriting from `CollectionBase`

ikondrat (approved):

lightly_studio/src/lightly_studio/api/routes/api/annotation.py
```diff
@@ -37,6 +39,25 @@
...
+    "/annotation_collections",
+    response_model=list[AnnotationCollectionView],
+)
+def read_annotation_collections(
+    session: SessionDep,
+    collection: Annotated[
+        CollectionTable,
+        Path(title="collection Id"),
+        Depends(get_and_validate_collection_id),
+    ],
```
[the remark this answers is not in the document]
MalteEbner: Skipping — `Path(...) + Depends(...)` on the same parameter is the established pattern in `collection.py` (`read_collection`, `update_collection`, `delete_collection`, `has_embeddings`, `deep_copy`, `delete_dataset`), all paired with `get_and_validate_collection_id` and all with passing tests. FastAPI binds the path variable via the dependency function's own `collection_id: UUID` parameter; the route-level `Path(...)` is redundant OpenAPI metadata, not a broken wiring. The new route's tests confirm it works.

lightly_studio/src/lightly_studio/api/routes/api/annotation.py
```diff
@@ -37,6 +39,25 @@
...
+) -> list[CollectionTable]:
```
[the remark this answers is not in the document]
MalteEbner: Skipping — the codebase convention is to annotate the actual Python return type, not the `response_model`. See `read_annotations` directly above (line 72): `-> GetAllAnnotationsResult` while `response_model=AnnotationViewsWithCount`. The function returns `CollectionTable` rows from the resolver; FastAPI handles the conversion to `AnnotationCollectionView` via `response_model`. Changing the annotation to `list[AnnotationCollectionView]` would misrepresent what the function actually returns at runtime.

lightly_studio/src/lightly_studio/api/routes/api/annotation.py
ikondrat: aah true

PR #1089, opened 2026-05-11
Extract shortcut settings config and ShortcutSettingRow component

michal-lightly:
## What has changed and why?

Part 2 of the settings refactor.

Extracts the repeated shortcut row markup in `SettingsDialog.svelte` into a typed config array and a reusable component, as part of the settings dialog refactor (step 2).

- Create `settingsDialogConfig.ts` with typed `shortcutSettings` and `staticShortcuts` arrays, giving every shortcut a unique stable HTML ID
- Extract `ShortcutSettingRow.svelte` with explicit props (`id`, `label`, `value`, `isRecording`, `disabled`, `onStartRecording`)
- Replace ~180 lines of duplicated shortcut markup with `{#each}` loops over the config arrays
- Fix 5 shortcut rows that previously shared `id="toggle-edit-mode"` — all 10 rows now have distinct IDs

## How has it been tested?

- Unit tests

- [ ] Yes

## Summary by CodeRabbit

* **Refactor**
  * Settings dialog keyboard shortcut UI rewritten to be configuration-driven and simplified for maintainability.
  * Shortcut recording now syncs and persists normalized key values with the form state.

* **New Features**
  * Added modular shortcut row component and a config-driven list separating editable and static/display-only shortcuts.

* **Tests**
  * Added/updated tests to cover shortcut rows, recording interactions, unique element ids, save flow, and dialog closing.

[Review Change Stack](https://app.coderabbit.ai/change-stack/lightly-ai/lightly-studio/pull/1089)

### DIFF
--- a/lightly_studio_view/src/lib/components/Settings/SettingsDialog.svelte
+++ b/lightly_studio_view/src/lib/components/Settings/SettingsDialog.svelte
@@ -26,7 +26,7 @@
     let isSaving = $state(false);
 
     const initialFormState = createSettingsDialogFormState($settingsStore);
-    let shortcutState = $state(initialFormState.shortcutSettings);
+    let shortcuts = $state(initialFormState.shortcutSettings);
     let gridViewRendering: RenderingMode = $state(initialFormState.gridViewRendering);
     let gridViewThumbnailQuality: ThumbnailQualityMode = $state(
         initialFormState.gridViewThumbnailQuality
@@ -43,7 +43,7 @@
     $effect(() => {
         if ($settingsStore && $isLoadedStore && !initialized) {
             const formState = createSettingsDialogFormState($settingsStore);
-            shortcutState = formState.shortcutSettings;
+            shortcuts = formState.shortcutSettings;
             gridViewRendering = formState.gridViewRendering;
             gridViewThumbnailQuality = formState.gridViewThumbnailQuality;
             showAnnotationTextLabels = formState.showAnnotationTextLabels;
@@ -85,7 +85,7 @@
         try {
             await saveSettings(
                 createSettingsSavePayload({
-                    shortcutSettings: shortcutState,
+                    shortcutSettings: shortcuts,
                     gridViewRendering,
                     gridViewThumbnailQuality,
                     showAnnotationTextLabels,
@@ -115,7 +115,7 @@
         event.preventDefault();
         event.stopPropagation();
 
-        shortcutState[recordingShortcut] = normalizeShortcutKey(event);
+        shortcuts[recordingShortcut] = normalizeShortcutKey(event);
 
         // Stop recording
         recordingShortcut = null;
@@ -146,7 +146,7 @@
                             <ShortcutSettingRow
                                 id={setting.id}
                                 label={setting.label}
-                                value={shortcutState[setting.key]}
+                                value={shortcuts[setting.key]}
                                 isRecording={recordingShortcut === setting.key}
                                 onStartRecording={() => startRecording(setting.key)}
                             />
--- a/lightly_studio_view/src/lib/components/Settings/SettingsDialog.svelte
+++ b/lightly_studio_view/src/lib/components/Settings/SettingsDialog.svelte
@@ -11,9 +11,11 @@
         createSettingsSavePayload,
         normalizeShortcutKey
     } from './settingsDialogState';
+    import { shortcutSettings, staticShortcuts } from './settingsDialogConfig';
+    import type { ShortcutSettingKey } from './settingsDialogConfig';
+    import ShortcutSettingRow from './ShortcutSettingRow/ShortcutSettingRow.svelte';
 
     type SettingsDialogFormState = ReturnType<typeof createSettingsDialogFormState>;
-    type ShortcutSettingKey = keyof SettingsDialogFormState['shortcutSettings'];
     type RenderingMode = SettingsDialogFormState['gridViewRendering'];
     type ThumbnailQualityMode = SettingsDialogFormState['gridViewThumbnailQuality'];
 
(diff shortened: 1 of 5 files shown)


lightly_studio_view/src/lib/components/Settings/SettingsDialog.svelte
```diff
@@ -24,7 +26,7 @@
     let isSaving = $state(false);
 
     const initialFormState = createSettingsDialogFormState($settingsStore);
-    let shortcutSettings = $state(initialFormState.shortcutSettings);
+    let shortcutState = $state(initialFormState.shortcutSettings);
```
michal-lightly: Renamed from `shortcutSettings` to `shortcutState` to avoid collision with the imported `shortcutSettings` config array.

lightly_studio_view/src/lib/components/Settings/SettingsDialog.svelte
```diff
@@ -11,9 +11,11 @@
         createSettingsSavePayload,
         normalizeShortcutKey
     } from './settingsDialogState';
+    import { shortcutSettings, staticShortcuts } from './settingsDialogConfig';
+    import type { ShortcutSettingKey } from './settingsDialogConfig';
```
michal-lightly, line 15: `ShortcutSettingKey` is now exported from `settingsDialogConfig.ts` (the canonical source of shortcut definitions) and imported here, replacing the previous inline derivation from form state.

lightly_studio_view/src/lib/components/Settings/settingsDialogConfig.ts
```diff
@@ -0,0 +1,56 @@
...
+
+export interface StaticShortcutConfig {
+    id: string;
+    label: string;
+    /** Static display value (not recordable). */
+    displayValue: string;
+}
+
+/** Recordable shortcut rows shown in the Keyboard Shortcuts section. */
+export const shortcutSettings: readonly ShortcutSettingConfig[] = [
```
michal-lightly, line 37: IDs match the ones previously hardcoded in the template. The 5 rows that used to share `id="toggle-edit-mode"` now have distinct IDs (`toolbar-selection`, `toolbar-drag`, `toolbar-bounding-box`, etc.).

lightly_studio_view/src/lib/components/Settings/settingsDialogConfig.ts
```diff
@@ -0,0 +1,56 @@
...
+    {
+        id: 'toolbar-segmentation-mask',
+        label: 'Toolbar segmentation mask',
+        key: 'keyToolbarSegmentationMask'
+    },
+    { id: 'toolbar-brush-mode', label: 'Toolbar brush mode', key: 'keyToolbarBrush' },
+    { id: 'toolbar-eraser-mode', label: 'Toolbar eraser mode', key: 'keyToolbarEraser' }
+] as const;
+
+/** Non-recordable shortcut rows (static display only). */
```
michal-lightly, line 53: The "Change brush size" row is separated into its own config array since it is not recordable — it uses a static `displayValue` instead of a `key` into shortcut state.

michal-lightly (commented):
Review comments to help navigate the diff.

lightly_studio_view/src/lib/components/Settings/ShortcutSettingRow/ShortcutSettingRow.svelte
```diff
@@ -0,0 +1,36 @@
...
+<div class="grid grid-cols-2 items-center gap-4">
+    <Label for={id} class="text-right text-foreground">
+        {label}
+    </Label>
+    <button
+        {id}
+        type="button"
+        {disabled}
+        class="rounded-md border border-input bg-background p-2 text-left text-foreground"
+        onclick={(e) => {
```
ikondrat: Why you need to prevent default here? 🤔 

I would name it better and reflect what it does

it looks like it toogles the state 

onclick={onClickToggle} makes more sense to me

lightly_studio_view/src/lib/components/Settings/ShortcutSettingRow/ShortcutSettingRow.svelte
```diff
@@ -0,0 +1,36 @@
...
+            e.preventDefault();
+            onStartRecording?.();
+        }}
+    >
+        {#if isRecording}
```
ikondrat, line 27: It looks like you using value only when you pass isRecording 

I would simplify this then and drop isRecording
{#if value}
   <span>{value}</span>
{:else}
<span class="italic opacity-70">Press a key...</span>
{/if}

lightly_studio_view/src/lib/components/Settings/SettingsDialog.svelte
ikondrat:
```suggestion
    let shortcuts = $state(initialFormState.shortcutSettings);
    
    shortcuts[recordingShortcut]
```

I would avoid meaningless parts in the naming

I think "state" in naming doesn't give any information

ikondrat (approved):
lgtm

just few minor comments

lightly_studio_view/src/lib/components/Settings/ShortcutSettingRow/ShortcutSettingRow.svelte
michal-lightly: Also the style changes when the keyboard shortcut is being recorded (italics, opacity).

michal-lightly: And we need to keep `isRecording`, since if the user aborts recording, the previous value is used.

lightly_studio_view/src/lib/components/Settings/ShortcutSettingRow/ShortcutSettingRow.svelte
michal-lightly: Removed `preventDefault`, it was not needed.

Kept `onStartRecording`, it is descriptive - it means the user is recording a new shortcut.

(We can debate is to rename `recording` to something clearer everywhere, but I'd say that's out of scope of this PR.)

PR #1111, opened 2026-05-13
Make the query editor horizontally resizable

lukas-lightly:
## What has changed and why?
The query editor can take too much space, the user should be able to resize it.

## How has it been tested?
Manually

- [ ] Yes

## Summary by CodeRabbit

* **New Features**
  * Split-pane editor: when actively editing query filters, the interface shows a right-side editor for better screen use.
  * Search and embeddings controls now appear consistently in both split and single-column layouts.

* **Refactor**
  * Unified header, selection display, and resizer behaviors for smoother layout switching and consistent UI across modes.

[Review Change Stack](https://app.coderabbit.ai/change-stack/lightly-ai/lightly-studio/pull/1111)


lightly_studio_view/src/routes/datasets/[dataset_id]/[collection_type]/[collection_id]/+layout.svelte
```diff
@@ -417,60 +417,148 @@
...
-                {#if isQueryFilterEnabled && isQueryFilterEditing}
-                    <QueryEditorPanel />
+                {:else}
+                    <PaneGroup direction="horizontal" class="flex-1">
+                        <Pane defaultSize={65} minSize={35} class="flex">
+                            <div
+                                class="relative flex flex-1 flex-col space-y-4 rounded-[1vw] bg-card p-4 pb-2"
+                            >
+                                {#if isImages || isAnnotations || isVideos || isVideoFrames || isGroups}
+                                    <GridHeader>
```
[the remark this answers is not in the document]
Comment (lukas-lightly) [reply lightly_studio_view/src/routes/datasets/[dataset_id]/[collection_type]/[collection_id]/+layout.svelte]: It's improved now, can you check again?

lukas-lightly:
Notice the six dots in the middle divider, you can move it left and right:
image

lightly_studio_view/src/routes/datasets/[dataset_id]/[collection_type]/[collection_id]/+layout.svelte
```diff
@@ -416,62 +468,33 @@
...
+                        </div>
+                    </Pane>
+
+                    <PaneResizer
+                        class="relative mx-2 flex w-1 cursor-col-resize items-center justify-center"
+                    >
+                        <div class="bg-brand z-10 flex h-7 min-w-5 items-center justify-center">
+                            <GripVertical class="text-diffuse-foreground" />
+                        </div>
+                    </PaneResizer>
```
Comment (mihnea-necsulescu) [inline lightly_studio_view/src/routes/datasets/[dataset_id]/[collection_type]/[collection_id]/+layout.svelte]: this can also be extracted to a snippet

mihnea-necsulescu (commented):
the query filter makes the main content lose its rounded corners:

Image

vs 

Image

also, this may be out of scope, but noting it here - the behavior when opening the query filter and the embedding plot is not consistent:
1. if I click on the "Show Embeddings" button when the query filter is open, the embedding plot will open on top of the query filter
2. if I click on the query filter button when the embedding plot is open, the query filter opens "in the background", i.e., it's open but behind the embedding plot; if I close the embedding plot, the query filter will then be visible 

I think the best UX would be that if I open one of those components when the other one is already open, then that one should be replaced - this may be a bigger effort, though, and I'm not even sure if it's worth doing now, as it seems like an edge case

lightly_studio_view/src/routes/datasets/[dataset_id]/[collection_type]/[collection_id]/+layout.svelte
Comment (lukas-lightly) [reply lightly_studio_view/src/routes/datasets/[dataset_id]/[collection_type]/[collection_id]/+layout.svelte]: Done

lukas-lightly:
> the query filter makes the main content lose its rounded corners:
> 
> Image
> vs
> 
> Image
> also, this may be out of scope, but noting it here - the behavior when opening the query filter and the embedding plot is not consistent:
> 
> 1. if I click on the "Show Embeddings" button when the query filter is open, the embedding plot will open on top of the query filter
> 2. if I click on the query filter button when the embedding plot is open, the query filter opens "in the background", i.e., it's open but behind the embedding plot; if I close the embedding plot, the query filter will then be visible
> 
> I think the best UX would be that if I open one of those components when the other one is already open, then that one should be replaced - this may be a bigger effort, though, and I'm not even sure if it's worth doing now, as it seems like an edge case

Yes, this is out of scope for this task but we added a task yesterday about it to this sprint.

mihnea-necsulescu (approved):

PR #1212, opened 2026-05-26
Fix query editor Apply button disabled on first open

michal-lightly:
## What has changed and why?

The Apply button in the query editor was disabled when the panel first opened, even though the query had never been applied. This happened because the disable logic compared the editor draft against the initial prop value — which are identical on mount.

Replaced the check with a `lastAppliedValue` tracker (initially `null`) so the button is enabled until the user applies, then disables until the text changes again.

## How has it been tested?

Manual testing: opened the query editor panel, confirmed Apply is enabled on first open, disabled after clicking Apply, and re-enabled after editing.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Query Editor change tracking so the Apply button is enabled only when there are unapplied draft changes and remains disabled in read‑only mode.

* **Documentation**
  * Clarified panel helper text to explain when completion hints appear while typing.

* **Tests**
  * Expanded tests to cover Apply-button enable/disable across edit/apply cycles and reopening the editor.

[Review Change Stack](https://app.coderabbit.ai/change-stack/lightly-ai/lightly-studio/pull/1212?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


ikondrat (changes requested):
I don't understand why we need it, let's rather remove disabled state

If we don't need to track disabled state, let's just remove it

horatiualmasan (approved):
LGTM
Can you add a test that checks that Apply is enabled by default?

ikondrat (approved):

PR #1213, opened 2026-05-26
Fix GUI query not refetching after TanStack v6 upgrade

michal-lightly:
## What has changed and why?

Bugfix: Query from GUI application was not working.

This was caused by the TanStack Query v5 → v6 upgrade. The `query_expr` parameter was used in the request body but was missing from the TanStack query key. V6 is stricter about only refetching when the query key changes, so the cached result was returned instead.

Fixed by including `query_expr` in the query key.

## How has it been tested?

- Existing unit tests
- Manual: apply a query expression → verify new results are fetched

## Summary by CodeRabbit

* **Refactor**
  * Improved query parameter handling to ensure distinct search expressions are tracked and cached independently, preventing overlap between different queries and enhancing search accuracy.

[Review Change Stack](https://app.coderabbit.ai/change-stack/lightly-ai/lightly-studio/pull/1213?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


horatiualmasan (approved):
LGTM

PR #1216, opened 2026-05-26
Factor out filter chip, use it for query filter

michal-lightly:
## What has changed and why?

Use the same UI element - dubbed `Filter Chip` - for query and embedding plot filter.

* Factored out the embedding plot design to a custom element
* Applied it to query filter
* Changed the color - no longer orange
* The query chip no longer closes the panel - for that we have the X button

## How has it been tested?

Manually.

Screenshot 2026-05-26 at 22 25 53

## Summary by CodeRabbit

* **New Features**
  * Added a reusable "Filter Chip" UI for filters: checkbox, title, optional subtitle, and clear action.
  * Storybook examples demonstrating the chip (default and with subtitle).

* **Refactor**
  * Replaced previous embedding and query filter UIs with the unified Filter Chip for consistent behavior and appearance.
  * Query filter now opens the editor via the chip and shows a truncated subtitle for query text or item counts.

[Review Change Stack](https://app.coderabbit.ai/change-stack/lightly-ai/lightly-studio/pull/1216?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/lightly_studio_view/src/lib/components/FilterChip/FilterChip.svelte
+++ b/lightly_studio_view/src/lib/components/FilterChip/FilterChip.svelte
@@ -26,7 +26,7 @@
     }: Props = $props();
 </script>
 
-<div class="rounded-md border border-amber-500/35 bg-amber-500/10 px-2 py-1.5" data-testid={testId}>
+<div class="rounded-md border border-[#3c3c3c] bg-muted px-2 py-1.5" data-testid={testId}>
     <div class="flex items-center gap-2">
         <Checkbox {checked} aria-label={checkboxLabel} {onCheckedChange} />
         {#if onclick}
--- a/lightly_studio_view/src/lib/components/FilterChip/FilterChip.svelte
+++ b/lightly_studio_view/src/lib/components/FilterChip/FilterChip.svelte
@@ -1,0 +1,52 @@
+<script lang="ts">
+    import { Checkbox } from '$lib/components/ui/checkbox';
+    import { X } from '@lucide/svelte';
+    import type { Snippet } from 'svelte';
+
+    interface Props {
+        checked: boolean;
+        title: string;
+        checkboxLabel: string;
+        onCheckedChange: (checked: boolean | 'indeterminate') => void;
+        onClear: () => void;
+        onclick?: () => void;
+        subtitle?: Snippet;
+        testId?: string;
+    }
+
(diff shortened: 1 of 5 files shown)


lightly_studio_view/src/lib/components/FilterChip/FilterChip.svelte
```diff
@@ -0,0 +1,52 @@
+<script lang="ts">
```
ikondrat, line 1: Add test for component + story for storybook to test it

lightly_studio_view/src/lib/components/FilterChip/FilterChip.svelte
```diff
@@ -0,0 +1,52 @@
...
+        testId?: string;
+    }
+
+    let {
+        checked,
+        title,
+        checkboxLabel,
+        onCheckedChange,
+        onClear,
+        onclick,
```
ikondrat, line 23: combination of onclick + onCheckedChange seems a bit confusing

I would keep only onCheckedChange

I think the problem comes from non-proper checkbox usage

For checkbox click on label should trigger onCheckedChange
here is an example https://www.shadcn-svelte.com/docs/components/checkbox

lightly_studio_view/src/lib/components/FilterChip/FilterChip.svelte
```diff
@@ -0,0 +1,52 @@
...
+        subtitle,
+        testId
+    }: Props = $props();
+</script>
+
+<div class="rounded-md border border-[#3c3c3c] bg-muted px-2 py-1.5" data-testid={testId}>
```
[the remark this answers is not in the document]
michal-lightly: We want the specified color, not border-border. We do not support light mode. The change is not worth it at this point.

ikondrat (commented):
Refactoring wise it is good

But UX wise we have weird behavior, click on label should also trigger onCheckedChange

ikondrat (approved):

lightly_studio_view/src/lib/components/FilterChip/FilterChip.svelte
michal-lightly: We cannot have only onCheckedChange, clicking the element opens the query panel. We can do it once the way the query panel opens it changed.

PR #1231, opened 2026-05-27
fix: remove DEFAULT class when creating new annotations

ikondrat:
## How has it been tested?
Previously, when drawing a bounding box or segmentation mask without a class selected, the code would fall back to finding or creating a label named 'DEFAULT'. This caused confusing UX where annotations were silently assigned to a DEFAULT class.

Now:
- If no class is selected when drawing, an error toast is shown: 'Please select a class before creating an annotation'
- If a class is selected but doesn't exist in the dataset yet, it is created with the user's chosen name (not 'DEFAULT')

Fixes: LIG-9632

## Testing Steps

1. Run the app
2. Go to image details and click **Edit annotations**
3. Try to add a new **Bounding Box** without selecting a class
4. A dialog should appear with a list of available classes to select
5. Select a class from the dialog and confirm
6. The annotation should be created with the selected class name
7. Try to add another **Bounding Box** — the previously selected class should be used automatically (no dialog)

- [ ] Yes

## Summary by CodeRabbit

* **New Features**
  * Added a select-class dialog and flow so brush/rectangle tools can prompt for or request a class before creating annotations.

* **Bug Fixes**
  * Creating annotations now requires an explicit class selection; shows an error and aborts if none is chosen.
  * Removed implicit fallback to a default label; new labels are created using the selected class name.

* **Tests**
  * Updated unit and e2e tests to cover early-abort behavior and label-creation using the chosen class; e2e reload ensures selected label is loaded from session storage.

[Review Change Stack](https://app.coderabbit.ai/change-stack/lightly-ai/lightly-studio/pull/1231?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


lightly_studio_view/src/lib/components/SampleDetails/SampleSegmentationMaskRect/SampleSegmentationMaskRect.svelte
```diff
@@ -73,6 +74,37 @@
...
+    const settleRequestLabel = (label: string | null) => {
+        showSelectClassDialog = false;
+        resolveRequestLabel?.(label);
+        resolveRequestLabel = null;
+        pendingLabelRequest = null;
+    };
+
+    const handleClassSelected = (label: string) => settleRequestLabel(label);
+
+    const handleClassDialogCancel = () => settleRequestLabel(null);
```
mihnea-necsulescu, line 105: could these be extracted to a hook? they are also defined in `SampleObjectDetectionRect.svelte`

mihnea-necsulescu (approved):

lightly_studio_view/src/lib/components/SampleDetails/SampleSegmentationMaskRect/SampleSegmentationMaskRect.svelte
ikondrat: yes, I will do it in the follow up though

PR #1262, opened 2026-05-29
Send all color categories per sample for embedding plot (backend)

michal-lightly:
## What has changed and why?

Send all color categories per sample for embedding plot. The motivation is to enable category toggling in the frontend. This simplifies the backend logic (no need to pick the winning category).

Adds a new `color_categories` table column to the data returned to the frontend. The old field is preserved
so that frontend does not need an update yet. It will be removed in a later PR.

The `color_categories` table column is of type `list[int]` - Arrow allows a list to be a column type.

Note: `fulfils_filter` has been removed from the backend functions. The plan is to handle the rendering by the frontend.

## How has it been tested?

- `make static-checks` and `make test` in `lightly_studio` (targeted: `test_embeddings2d.py` + `embedding_coloring/`).
- Extended endpoint tests assert the multi-category sample returns all categories in priority order while the scalar stays the primary.

- [ ] Yes

## Release Notes

* **New Features**
  * Samples in 2D embedding visualizations now support multiple simultaneous color categories, allowing richer representation of multi-label data in the coloring system.

* **Refactor**
  * Simplified and restructured internal color categorization logic to enable per-sample multi-valued color assignments with improved legend handling.


LeonardoRosaa (approved):
LGTM

PR #1265, opened 2026-05-29
fix: batch bulk DB operations to stay under Postgres' 65,535-parameter limit (1/4)

MalteEbner:
### Error description
PostgreSQL caps a single statement at 65,535 bind parameters, so bulk operations whose parameter count grows with the dataset size fail on large datasets:

```
psycopg.OperationalError: number of parameters must be between 0 and 65535
```

This made large datasets unusable. For example, a dataset with more than ~65,535 samples could be created but then failed to delete (which also blocked iterating on it), and the same limit affects bulk inserts.

### Changes
The fix splits oversized id lists and rows into sub-limit chunks. Behavior is identical on DuckDB and PostgreSQL.

This is the first of several PRs. It adds the shared batching utility and applies it to a first set of bulk operations:

- **New `utils/batching.py`** — a `batched()` helper.
- **`delete_dataset`** uses batching now
- **`sample_resolver.create_many`** uses batching now
- **`annotation_collection_coverage_resolver.add_many`** uses batching now

Follow-up PRs apply the same batching to the remaining bulk inserts, the id-list select/delete paths, and deep_copy.

## How has it been tested?

new unittests

## Did you update [CHANGELOG.md](../CHANGELOG.md)?

- [x] Yes

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed PostgreSQL failures when deleting very large datasets (more than ~65,535 items) by implementing batched delete operations.

* **Tests**
  * Added test coverage for batch processing and large dataset deletion operations.

[Review Change Stack](https://app.coderabbit.ai/change-stack/lightly-ai/lightly-studio/pull/1265?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

PR #1302, opened 2026-06-02
Fix: Make more queries work with more than 65k samples using Postgres

MalteEbner:
## What has changed and why?

Follow-up to PRs #1265 and #1291. 

A few `WHERE col.in_(python_list)` sites can't be batched like #1291 for one of these reasons:
- they are *composable* (the shared `SampleFilter`)
- they are*paginated* (the image/video grid list queries)
- they are *ordered by a joined column* (the annotation parent-sample lookups). 

On PostgreSQL their expanding `IN` still exceeds the 65,535 bind-parameter cap when a user selects/filters a large set.

This binds the list as a **single array parameter** instead. Uses new dialect-aware `in_array(column, values)` construct (`db_array.py`, mirroring `db_vector.py`'s `@compiles` pattern) compiles to:
- PostgreSQL: `column = ANY(:array)` — one bound `UUID[]` parameter.
- DuckDB: ordinary expanding `column.in_(values)` (no comparable cap).

## Covered parts

Covered is almost everything. Exceptions are:
- everything that is batched is left as-is
- everything that is using a subquery is left as-is
- everything in `deep_copy.py` will be done in a follow-up PR
image

## How has it been tested?

New unittests for some resolvers. However, there are far too many resolves, did not add new unittests for all of them.

## Next steps / PRs
- Move all the db_***.py files from the root into a subdir.
- Also apply these fixes to deep copy

- [x] Not needed — covered by the existing "operations on large datasets … `number of parameters must be between 0 and 65535`" entry.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed edge-case behavior when provided ID lists are empty to avoid invalid or unnecessary queries.

* **Refactor**
  * Introduced a database-level membership helper to improve filtering by ID and handle very large ID lists reliably across supported databases, improving performance and avoiding bind-parameter limits.

* **Tests**
  * Added tests validating membership behavior across dialects and large-ID scenarios.


lightly_studio/src/lightly_studio/resolvers/sample_resolver/sample_filter.py
```diff
@@ -46,7 +47,7 @@ def apply(self, query: QueryType) -> QueryType:
 
     def _apply_sample_ids_filter(self, query: QueryType) -> QueryType:
         if self.sample_ids:
-            return query.where(col(SampleTable.sample_id).in_(self.sample_ids))
+            return query.where(in_array(col(SampleTable.sample_id), self.sample_ids))
```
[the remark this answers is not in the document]
MalteEbner: Done — `in_array` is now called with keyword arguments (`column=`, `values=`) at all call sites, matching the project convention.

lightly_studio/src/lightly_studio/resolvers/annotation_resolver/get_all_by_parent_sample_ids.py
```diff
@@ -28,7 +31,7 @@ def get_all_by_parent_sample_ids(
             col(VideoFrameTable.sample_id) == col(AnnotationBaseTable.parent_sample_id),
         )
         .outerjoin(VideoTable, col(VideoTable.sample_id) == col(VideoFrameTable.parent_sample_id))
-        .where(col(AnnotationBaseTable.parent_sample_id).in_(parent_sample_ids))
+        .where(in_array(col(AnnotationBaseTable.parent_sample_id), parent_sample_ids))
```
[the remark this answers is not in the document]
MalteEbner: Done — `in_array` is now called with keyword arguments (`column=`, `values=`) at all call sites, matching the project convention.

lightly_studio/src/lightly_studio/resolvers/image_resolver/get_all_by_collection_id.py
```diff
@@ -182,8 +183,8 @@ def _get_all_with_similarity(  # noqa: PLR0913
 
     # TODO(Michal, 06/2025): Consider adding sample_ids to the filters.
     if sample_ids:
-        samples_query = samples_query.where(col(ImageTable.sample_id).in_(sample_ids))
-        total_count_query = total_count_query.where(col(ImageTable.sample_id).in_(sample_ids))
+        samples_query = samples_query.where(in_array(col(ImageTable.sample_id), sample_ids))
+        total_count_query = total_count_query.where(in_array(col(ImageTable.sample_id), sample_ids))
```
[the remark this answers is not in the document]
MalteEbner: Done — `in_array` is now called with keyword arguments (`column=`, `values=`) at all call sites, matching the project convention.

lightly_studio/src/lightly_studio/resolvers/video_resolver/get_all_by_collection_id.py
```diff
@@ -159,8 +160,8 @@ def _get_all_with_similarity(  # noqa: PLR0913
     )
 
     if sample_ids:
-        samples_query = samples_query.where(col(VideoTable.sample_id).in_(sample_ids))
-        total_count_query = total_count_query.where(col(VideoTable.sample_id).in_(sample_ids))
+        samples_query = samples_query.where(in_array(col(VideoTable.sample_id), sample_ids))
+        total_count_query = total_count_query.where(in_array(col(VideoTable.sample_id), sample_ids))
```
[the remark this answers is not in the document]
MalteEbner: Done — `in_array` is now called with keyword arguments (`column=`, `values=`) at all call sites, matching the project convention.

lightly_studio/src/lightly_studio/db_array.py
```diff
@@ -0,0 +1,66 @@
...
+from sqlalchemy.ext.compiler import compiles
+from sqlalchemy.orm import Mapped
+from sqlalchemy.sql.compiler import SQLCompiler
+from sqlalchemy.sql.elements import ColumnElement
+
+
+class in_array(ColumnElement[bool]):  # noqa: N801
+    """Membership test over a UUID column, compiled per dialect.
+
+    Use like ``column.in_(values)`` inside ``.where(...)``. On PostgreSQL it binds
```
michal-lightly: The comment is not clear to me, I would expect something like this
```suggestion
    Use instead of ``column.in_(values)`` inside ``.where(...)``. On PostgreSQL it binds
```

michal-lightly (approved):

lightly_studio/src/lightly_studio/db_array.py
MalteEbner: Reworked the docstring completely

mihnea-necsulescu:
there are still a couple of places where `.in_(list)` is used that I think could cause problems on Postgres, e.g.:
- https://github.com/lightly-ai/lightly-studio/blob/main/lightly_studio/src/lightly_studio/resolvers/collection_resolver/export.py#L202
- https://github.com/lightly-ai/lightly-studio/blob/main/lightly_studio/src/lightly_studio/resolvers/collection_resolver/export.py#L221
- https://github.com/lightly-ai/lightly-studio/blob/main/lightly_studio/src/lightly_studio/resolvers/video_frame_resolver/get_all_by_video_ids.py#L26

shouldn't these also be addressed in this pr or have they been skipped intentionally?

mihnea-necsulescu (commented):
my bad, left a comment directly on the pr which did not trigger the bot; see -> https://github.com/lightly-ai/lightly-studio/pull/1302#issuecomment-4611054094

lightly_studio/src/lightly_studio/resolvers/annotation_label_resolver/get_by_ids.py
```diff
@@ -15,11 +16,18 @@ def get_by_ids(session: Session, ids: Sequence[UUID]) -> list[AnnotationLabelTab
 
     Output order matches the input order.
     """
+    ids_list = list(ids)
+    if not ids_list:
```
mihnea-necsulescu:
```suggestion
    if not ids:
```

lightly_studio/src/lightly_studio/db_array.py
```diff
@@ -0,0 +1,70 @@
+"""Dialect-aware ``in_array`` membership test for UUID columns.
+
+Compiles to ``column = ANY(:array)`` (one bound array parameter) on PostgreSQL to dodge the
+65,535 expanding-IN bind cap, and to expanding ``IN`` on DuckDB. Callers must guard against
+empty ``values`` (``= ANY(ARRAY[])`` errors on PostgreSQL).
```
mihnea-necsulescu: I'm not sure the "callers must guard..." part is accurate - I think the related PG error would be "cannot determine type of empty array", which won't be the case, as the array would always be typed as uuid[]

the guard still makes sense to avoid an unnecessary DB hit, but I'd update the reasoning

mihnea-necsulescu (approved):

lightly_studio/src/lightly_studio/db_array.py
MalteEbner: I deleted that sentence

PR #1325, opened 2026-06-03
Fix deep copy to work with more than 65k samples using Postgres

MalteEbner:
## What has changed and why?

Fix deep copy to work with more than 65k samples using Postgres. Same as https://github.com/lightly-ai/lightly-studio/pull/1302 

## How has it been tested?

Ran `test_deep_copy__exceeds_postgres_param_limit` manually with both duckdb and Postgres. Then deactivated it, cause it takes too long.

## Next steps

Last PR: Move all the `db_***.py` files into a subdirectory.

Already covered:

```
Fixed operations on large datasets (more than ~65,535 samples) failing on PostgreSQL with
  `number of parameters must be between 0 and 65535`.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Dataset deep-copy now handles very large collections without hitting database parameter limits.

* **Refactor**
  * Annotation detail copying reworked to perform bulk copying of supported detail tables after remapping base annotation rows.
  * Filtering switched to array-based operations to improve large-batch handling.

* **Tests**
  * Added integration test validating deep-copy with extremely large sample and annotation counts.


lightly_studio/src/lightly_studio/resolvers/dataset_resolver/deep_copy.py
```diff
@@ -443,37 +461,21 @@ def _copy_annotations(
...
-    annotation_type: AnnotationType,
-) -> None:
-    """Copy annotation detail table based on type."""
-    if annotation_type == AnnotationType.OBJECT_DETECTION:
-        old_obj_det = session.get(ObjectDetectionAnnotationTable, old_sample_id)
-        if old_obj_det:
-            new_obj_det = _copy_with_updates(
-                old_obj_det,
-                {"sample_id": new_sample_id},
+    # Go over detail annotation tables to copy bboxes and segmentation mask.
```
MalteEbner, line 464: Unrelated speedup. Instead of calling `session.get(ObjectDetectionAnnotationTable, old_sample_id)` for every annotation separately (one DB call per annotation), call it once for all annotations in batch.

Made the test much faster

lightly_studio/tests/resolvers/dataset_resolver/test_deep_copy.py
```diff
@@ -626,6 +629,61 @@ def test_deep_copy__with_annotations(db_session: Session) -> None:
     assert is_detail.height == 8
     assert is_detail.segmentation_mask == [1, 0, 0, 1]
 
+#@pytest.mark.skip(reason="On an M4 Pro, it takes 47s for duckdb and 38s for postgres.")
```
MalteEbner: Deactivated

ikondrat (approved):

michal-lightly (approved):

PR #1335, opened 2026-06-04
Make it clear what we recommend 3.10 as default version

ikondrat:
## What has changed and why?

Make it clear which version we recommend 

## How has it been tested?

no need to test, just doc changes

- [ ] Yes

## Summary by CodeRabbit

* **Documentation**
  * Updated installation and contribution guidelines to recommend Python 3.10 for improved compatibility with SAM autolabeling and other plugins, while continuing to support Python 3.9–3.14.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -101,7 +101,7 @@ written in Markdown (MyST flavor). For more information regarding formatting, se
 ## Development Environment Setup
 
 ### Requirements
-- Python **3.9+**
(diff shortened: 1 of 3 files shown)


MalteEbner (approved):
Please use a better PR name:
- leave away the P0
- include the term "docs"
- make it clear that the version refers to the python version and not a version of something else

E.g. "Docs: Include recommendation to use python version 3.10"

PR #1392, opened 2026-06-11
Add NullableNumericalField for `confidence` field

lukas-lightly:
## What has changed and why?
Note that we can't use the existing NumericalField, as it's only for float or int but not for `float | int | None`, i.e. an optional number.

## How has it been tested?
Added tests.

- [x] Yes

## Summary by CodeRabbit

* **New Features**
  * Query editor now supports filtering by annotation confidence for classification, object detection, and segmentation masks (including nullable/optional confidence).

* **Tests**
  * Added unit and integration database tests validating confidence-based query translation and runtime filtering, including handling of null confidence values.

* **Documentation**
  * Updated changelog to document confidence field support in the query editor.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Show plugin description and a link to documentation when no plugins are available.
 - When sorting the image grid by a numeric field, the field value is now shown as an overlay on each sample.
-- Query editor now supports annotation `source` field in queries.
+- Query editor now supports annotation `source` and `confidence` fields in queries.
 
 ### Changed
 
--- a/lightly_studio/src/lightly_studio/core/dataset_query/field.py
+++ b/lightly_studio/src/lightly_studio/core/dataset_query/field.py
@@ -78,7 +78,44 @@ class OrdinalField(Field, Generic[T]):  # noqa: PLW1641
         return OrdinalFieldExpression(field=self, operator="!=", value=other)
 
 
+class NullableOrdinalField(Field, Generic[T]):  # noqa: PLW1641
+    """Generic field for nullable ordinal values that support comparison operations."""
+
+    def __init__(self, column: Mapped[T | None]) -> None:
+        """Initialize the nullable ordinal field with a database column."""
+        self._column = column
+
+    def get_sqlmodel_field(self) -> Mapped[T | None]:
+        """Get the nullable ordinal database column or property."""
+        return self._column
+
+    def __gt__(self, other: T) -> OrdinalFieldExpression[T]:
+        """Create a greater-than expression."""
+        return OrdinalFieldExpression(field=self, operator=">", value=other)
+
+    def __lt__(self, other: T) -> OrdinalFieldExpression[T]:
+        """Create a less-than expression."""
+        return OrdinalFieldExpression(field=self, operator="<", value=other)
+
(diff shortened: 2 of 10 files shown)


lightly_studio/src/lightly_studio/core/dataset_query/field.py
```diff
@@ -78,7 +78,44 @@ def __ne__(self, other: T) -> OrdinalFieldExpression[T]:  # type: ignore[overrid
         return OrdinalFieldExpression(field=self, operator="!=", value=other)
 
 
+class NullableOrdinalField(Field, Generic[T]):  # noqa: PLW1641
```
lukas-lightly, line 81: Note: the code is identical with `OrdinalField`, we could theoretically just have one type, but I wouldn't do it.

Nullable is a non-standard behavior and maybe these classes will eventually diverge. It also makes the code more readable IMO.

lightly_studio/src/lightly_studio/core/dataset_query/field_expression.py
```diff
@@ -36,7 +37,7 @@
 class OrdinalFieldExpression(MatchExpression, Generic[T]):
     """Generic expression for ordinal field comparisons."""
 
-    field: OrdinalField[T]
+    field: OrdinalField[T] | NullableOrdinalField[T]
```
lukas-lightly, line 40: We've just added another type here, meaning that the nullable confidence can only be compared numerically. You can't say `confidence IS NULL`, at least not yet.

CHANGELOG.md
```diff
@@ -11,7 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Show plugin description and a link to documentation when no plugins are available.
 - When sorting the image grid by a numeric field, the field value is now shown as an overlay on each sample.
-- Query editor now supports annotation `source` field in queries.
+- Query editor now supports annotation `source` and `confidence` field in queries.
```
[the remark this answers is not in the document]
lukas-lightly: Done

lightly_studio/src/lightly_studio/core/dataset_query/field.py
```diff
@@ -78,7 +78,44 @@ def __ne__(self, other: T) -> OrdinalFieldExpression[T]:  # type: ignore[overrid
...
+    """Generic field for nullable ordinal values that support comparison operations."""
```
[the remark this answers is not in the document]
lukas-lightly: We might add additional functionality for Nullable*, e.g. the `IS NULL` predicate. Let's keep it for now, it's extensible in the future.

lightly_studio/tests/core/dataset_query/test_field_expression.py
```diff
@@ -130,6 +137,50 @@ def test_apply__less_than_or_equal(self) -> None:
         assert "where image.created_at <= '2024-06-15 10:30:00+00:00'" in sql
 
 
+class TestNullableNumericalFieldExpression:
```
michal-lightly: The AI does not understand our test naming scheme, it is wrong in this file. Only tests for classes should be in classes. There is no `NullableNumericalFieldExpression` class. This tests etither `NullableNumericalField` or `OrdinalFieldExpression`.

michal-lightly (approved):

lightly_studio/tests/core/dataset_query/test_field_expression.py
lukas-lightly: Done

PR #1395, opened 2026-06-11
Fix unreadable red hint text in strategy tooltip

LeonardoRosaa:
## What has changed and why?

The disabled reason text in the `AddStrategyButton` tooltip was hard to read in dark mode because the red color was too dark against the background.

It was changed to use lighter red text directly in the component, making it easier to read in dark mode by introducing a dedicated `text-destructive` CSS token.

Screenshot 2026-06-15 at 13 06 56

## How has it been tested?

Manual test

- [ ] Yes

## Summary by CodeRabbit

* **Style**
  * Improved the destructive theme by adding a dedicated destructive text color for clearer contrast in dark mode.
  * Applied the updated destructive text styling across critical UI elements, including alerts, destructive/hover states, validation “required” indicators, and error/warning messages in operator and sampling dialogs.


lightly_studio_view/src/lib/components/Sampling/AddStrategyButton.svelte
```diff
@@ -106,7 +106,7 @@
         >
             <p>{hoveredStrategy.description}</p>
             {#if hoveredDisabledReason}
-                <p class="mt-1.5 border-t border-border pt-1.5 text-destructive">
+                <p class="mt-1.5 border-t border-border pt-1.5 text-red-600 dark:text-red-400">
```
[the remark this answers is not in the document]
LeonardoRosaa: There is no need to change it throughout the application.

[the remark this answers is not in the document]
michal-lightly: Why is there no need? In other places the text is readable? My instinct would be to change `text-destructive`.

lightly_studio_view/src/app.css
```diff
@@ -65,6 +65,10 @@
     }
 }
 
+.dark .text-destructive {
+    color: hsl(0 84% 65%);
+}
```
michal-lightly: The color looks good!

I checked with AI; this way of solving the problem is slightly hacky. There is an clean solution:

>Reusing --destructive for both fills and standalone text mixes two different roles. --destructive and --destructive-foreground are designed to work together: a red background with near-white text on top, so the text contrasts against the fill. text-destructive instead puts that red directly on the page background, where it has to provide the contrast itself.
>
>The .dark .text-destructive { color: hsl(0 84% 65%) } override in app.css shows this: the dark --destructive works behind white text but is too dark to use as text on --background, so it was patched separately.
>
>Suggest adding a dedicated --destructive-text token (with a text-destructive-text utility in tailwind.config.ts), set to a contrast-checked red in both :root and .dark — e.g. roughly hsl(0 72% 50%) for light and hsl(0 84% 65%) for dark. Keep --destructive/--destructive-foreground for fills only, switch error/hint text over to the new utility, and drop the .dark .text-destructive override.

Can we do it that way or would there be a problem?

LeonardoRosaa: As discussed, the initial implementation was hacky just to test it. The proper solution is the one AI mentioned.

lightly_studio_view/src/app.css
```diff
@@ -33,6 +33,7 @@
         --accent-foreground: 240 5.9% 10%;
         --destructive: 0 72.22% 50.59%;
         --destructive-foreground: 0 0% 98%;
+        --destructive-text: 0 72% 50%;
```
LeonardoRosaa, line 36: I think we can remove this since the light theme is not used.

michal-lightly: That's out of scope of this PR, let's keep it for now.

michal-lightly (approved):

PR #1396, opened 2026-06-11
Limit diversity strategy to one per sampling combination

LeonardoRosaa:
## What has changed and why?

Disables the **Diversity** option in `AddStrategyButton` once a diversity strategy has already been added to the combination. Only one embedding space is available per collection, so adding multiple diversity strategies would be redundant and confusing.

## How has it been tested?

Unit tests.

- [ ] Yes

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevent adding more than one diversity strategy — the add menu item is disabled with a displayed reason; duplicate controls on diversity strategy cards are also disabled.
  * Add action now requires a selected value and clears the selection after adding, avoiding stale selections.

* **Tests**
  * Added unit tests verifying the add-menu disables the diversity option when one exists and that the duplicate button is disabled for diversity strategy cards.

### DIFF
--- a/lightly_studio_view/src/lib/components/Sampling/SamplingCombinationDialog.svelte
+++ b/lightly_studio_view/src/lib/components/Sampling/SamplingCombinationDialog.svelte
@@ -106,6 +106,7 @@
                                         annotationLabels={strategyOptions.annotationLabels}
                                         annotationSourceOptions={strategyOptions.annotationSourceOptions}
                                         metadataFieldNames={strategyOptions.metadataFieldNames}
+                                        isDuplicateDisabled={instance.type === 'diversity'}
                                         onRemove={() => removeStrategy(instance.id)}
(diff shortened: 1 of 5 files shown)


lightly_studio_view/src/lib/components/Sampling/SamplingCombinationDialog.svelte
```diff
@@ -33,6 +33,10 @@
 
     const strategyOptions = useStrategyOptions(() => collectionId);
 
+    // TODO: Update once there are multiple embedding models - currently only one diversity
```
MalteEbner: Not sure about the frontend code style, but we have a specific format for backend TODOs: https://app.notion.com/p/getlightly/Python-code-style-290e575bc2134c3eb4c67d28e350d954?source=copy_link#f4bc0107775a43f881c0414ebb771c96

MalteEbner (approved):
LGTM, I tested it myself

PR #1402, opened 2026-06-12
Skip deep copy tests on DuckDB

mihnea-necsulescu:
## What has changed and why?

Reintroduces the `postgres_only` Pytest marker and skips the deep copy tests on DuckDB.

## How has it been tested?

Tests pass

- [ ] Yes

### DIFF
--- a/lightly_studio/tests/conftest.py
+++ b/lightly_studio/tests/conftest.py
@@ -64,16 +64,11 @@ def pytest_addoption(parser: pytest.Parser) -> None:
     )
 
 
-def pytest_configure(config: pytest.Config) -> None:
-    """Register custom markers for database backend selection."""
-    config.addinivalue_line(
-        "markers",
-        "postgres_only: run the test only on the Postgres backend (skipped under DuckDB)",
-    )
-
-
 def pytest_collection_modifyitems(config: pytest.Config, items: list[pytest.Item]) -> None:
-    """Skip postgres_only tests unless the suite runs with --postgres."""
+    """Skip postgres_only tests unless the suite runs with --postgres.
+
+    The ``postgres_only`` marker is registered in ``pytest.ini``.
+    """
     if config.getoption("--postgres"):
         return
     skip_marker = pytest.mark.skip(reason="postgres_only: skipped under DuckDB")
--- a/lightly_studio/tests/conftest.py
+++ b/lightly_studio/tests/conftest.py
@@ -20,10 +20,7 @@ from lightly_studio.api import features
 from lightly_studio.api.app import app
 from lightly_studio.dataset import embedding_manager
 from lightly_studio.dataset.embedding_generator import RandomEmbeddingGenerator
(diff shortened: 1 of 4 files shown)


lightly_studio/tests/conftest.py
```diff
@@ -57,6 +64,24 @@ def pytest_addoption(parser: pytest.Parser) -> None:
     )
 
 
+def pytest_configure(config: pytest.Config) -> None:
+    """Register custom markers for database backend selection."""
+    config.addinivalue_line(
+        "markers",
+        "postgres_only: run the test only on the Postgres backend (skipped under DuckDB)",
+    )
```
MalteEbner: According to Claude, this is duplicate to the changes in the pytest.ini

lightly_studio/tests/conftest.py
```diff
@@ -31,7 +34,11 @@
     AnnotationLabelCreate,
     AnnotationLabelTable,
 )
-from lightly_studio.models.collection import CollectionCreate, CollectionTable, SampleType
+from lightly_studio.models.collection import (
+    CollectionCreate,
+    CollectionTable,
+    SampleType,
+)
```
MalteEbner: Unnecessary change

MalteEbner (commented):
This looks quite a lot of code to me given what we need.

One alternative idea: Don't have a marker, but just make the tests skip directly if Postgres is not available. E.g. using pytest.mark.skipif for individual tests. For the small number of tests, this is fine, I don't think the marker is worth the overhead.

mihnea-necsulescu:
> This looks quite a lot of code to me given what we need.
> For the small number of tests, this is fine, I don't think the marker is worth the overhead.

I disagree - there are currently 13 deep copy tests + we'll also want to skip the delete dataset tests as well (12 more tests) - that would translate to at least 25 lines of code if we skip tests individually (it would probably be more, as it would also require a helper most likely); the PR currently sits at 21 lines of new code after addressing your other comments 

also, there may be other tests in the future that we'd want to skip for DuckDB 

I'd keep the marker

MalteEbner:
> > This looks quite a lot of code to me given what we need.
> > For the small number of tests, this is fine, I don't think the marker is worth the overhead.
> 
> I disagree - there are currently 13 deep copy tests + we'll also want to skip the delete dataset tests as well (12 more tests) - that would translate to at least 25 lines of code if we skip tests individually (it would probably be more, as it would also require a helper most likely); the PR currently sits at 21 lines of new code after addressing your other comments
> 
> also, there may be other tests in the future that we'd want to skip for DuckDB
> 
> I'd keep the marker

There is the option to skip the whole file on condition
```python
if should_skip_this_file():
    pytest.skip("reason this file is skipped", allow_module_level=True)
```

MalteEbner (approved):
PR is small, thus approving.

Removing the marker is optional. It would still make it easier to understand it my eyes.

PR #1404, opened 2026-06-12
Prevent negative strength values for diversity and class-balancing strategies

LeonardoRosaa:
## What has changed and why?

Added an optional `min` prop to `StrengthField`. The `min={0}` is passed for diversity and class balancing strategies, preventing negative strength values from being applied.

## How has it been tested?

Unit tests.

- [ ] Yes

## Summary by CodeRabbit

* **Bug Fixes**
  * Strength inputs for diversity and class‑balancing sampling now enforce a minimum of 0 to prevent negative values.

* **New Features**
  * Strength field accepts an optional minimum constraint and updates its help text to omit the “negative inverts scoring” note when non‑negative only.

* **Tests**
  * Added tests verifying input validation and tooltip content when a minimum is set.

### DIFF
--- a/lightly_studio_view/src/lib/components/Sampling/StrategyCard/StrategyCard.svelte
+++ b/lightly_studio_view/src/lib/components/Sampling/StrategyCard/StrategyCard.svelte
@@ -104,6 +104,7 @@
                             strength={instance.params.strength}
                             id={`diversity-strength-${instance.id}`}
                             testid={`strategy-diversity-strength-input-${instance.id}`}
+                            min={0}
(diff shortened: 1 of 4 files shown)


MalteEbner (commented):
Generally approving the PR. Good catch with the tooltip for negative values only showing conditionally.

The warning that negative values are not allowed only happens when clicking the sampling button. I expect the warning to happen already when inserting, at least when then clicking enter or changing the field

Image

MalteEbner (approved):
LGTM

PR #1405, opened 2026-06-12
Run unit tests conditionally on directory changes

michal-lightly:
## What has changed and why?

Skips test jobs that a change can't affect.

This PR introduces dependency on the 3rd party `dorny/paths-filter` action for action detection. It is a popular action with an MIT license. The check works specifically also for merge queues.

Note that the Docs check before was omitted from required checks in GitHub settings (by mistake), this PR will make it required.

Note: A change of required checks from multiple jobs to the single `Unit Success Check` is needed before merging.

## How has it been tested?

* This PR's own CI run.
* For merge queue, we'll need to test live

- [ ] Yes

## Summary by CodeRabbit

* **Chores**
  * Enhanced CI/CD pipeline efficiency by running checks only when relevant parts of the codebase change.
  * Added an always-on merge validation gate that ensures required checks either succeed or are skipped before allowing merges.

### DIFF
--- a/.github/workflows/test_and_release.yml
+++ b/.github/workflows/test_and_release.yml
@@ -37,10 +37,14 @@ jobs:
         uses: dorny/paths-filter@fbd0ab8f3e69293af611ebaee6363fc25e6d187d # v4.0.1
         with:
           filters: |
+            # static-checks runs `mypy .` / `ruff check` over the whole tree,
+            # so every Python dir it lints belongs here, not just src/tests.
             backend:
               - '.github/workflows/test_and_release.yml'
               - 'lightly_studio/src/**'
               - 'lightly_studio/tests/**'
+              - 'lightly_studio/e2e-tests/**'
+              - 'lightly_studio/dataset_examples/**'
               - 'lightly_studio/pyproject.toml'
               - 'lightly_studio/uv.lock'
               - 'lightly_studio/ruff.toml'
--- a/.github/workflows/test_and_release.yml
+++ b/.github/workflows/test_and_release.yml
@@ -16,8 +16,111 @@ concurrency:
 
   
 jobs:
+  # Detects which areas of the repo changed so downstream jobs can skip work
+  # that the change can't affect (e.g. a docs-only PR skips the test jobs).
+  # The required status check is `Unit Success Check`, not these jobs, so the
+  # test jobs are free to be skipped without stranding the merge queue.
+  detect:
+    name: Detect changes
+    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/test_and_release.yml
```diff
@@ -102,6 +148,9 @@ jobs:
 
   test-unit:
     name: Unit (${{ matrix.python }}, ${{ matrix.database }})
+    needs: detect
+    # Matrix runs Python unit tests (backend) and view tests (frontend).
+    if: needs.detect.outputs.backend == 'true' || needs.detect.outputs.frontend == 'true'
```
MalteEbner, line 153: Should we split these two tests, e.g. in a follow-up PR?

.github/workflows/test_and_release.yml
```diff
@@ -177,6 +226,9 @@ jobs:
 
   check-docs:
     name: Build Docs
+    needs: detect
+    # API reference is generated from backend docstrings.
```
MalteEbner, line 230: Can the `make docs` command fail if the backend docstrings are wrong, e.g. wrongly formatted?
If not, then we can remove the dependency on the backend changes

.github/workflows/test_and_release.yml
```diff
@@ -211,3 +263,54 @@ jobs:
       - name: Build docs
         working-directory: lightly_studio/docs
         run: make docs
+
+  # The single required status check: always runs, so gated jobs can be skipped
+  # without stranding the merge queue. Require this in the branch ruleset.
+  unit-success:
+    name: Unit Success Check
```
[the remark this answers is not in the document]
lukas-lightly: Agreed, please make the name more descriptive

lukas-lightly (approved):
This is awesome, thanks!

MalteEbner (approved):
This PR LGTM, but not enough to close the issue.

~~The E2E tests still run unconditionally.~~
They run based on a file hash and are thus cached: https://github.com/lightly-ai/lightly-studio/blob/d276cd2865cfb35916ce4b898556f4a629f02079/.github/workflows/end2end_test.yml#L42-L55

.github/workflows/test_and_release.yml
michal-lightly: Yes, we shall do that as a part of the other issue I have (CI speedups).

.github/workflows/test_and_release.yml
michal-lightly: Yes, it might invalidate links to the API reference

PR #1408, opened 2026-06-15
Rework delete dataset

mihnea-necsulescu:
## What has changed and why?

Similar to deep copy - improved the dataset deletion performance by running deletions on the DB layer only by using subqueries:
- all delete helpers issue a single statement -> IDs never leave the DB (this is done together with setting `synchronize_session` to False - check the docstring for details)
- deletes are now done as part of a single transaction (unlike on DuckDB)
- delete dataset tests are skipped for DuckDB

Note: Also considered a simple cascading delete, but that would transform into a maintenance burden, as `ON DELETE CASCADE` is not valid DDL on DuckDB - this means that we couldn't have updated all SQLModels to set cascade, which would require manual Postgres migrations to be maintained.

## How has it been tested?

Manual + tests still pass

- [ ] Yes

## Summary by CodeRabbit

* **Refactor**
  * Improved dataset deletion performance by switching to database-side, hierarchy-scoped bulk removals.
  * Deletion now uses a more consistent cleanup order and a single transaction for more reliable results, including related evaluation data.
  * Dataset deletion continues to require the PostgreSQL backend.

* **Tests**
  * Restricted dataset deletion tests to PostgreSQL-only environments.
  * Added an end-to-end round-trip test to verify deep-copy followed by deletion fully removes data for the copied dataset while leaving others unchanged.


lightly_studio/src/lightly_studio/resolvers/dataset_resolver/delete_dataset.py
```diff
@@ -43,218 +53,260 @@ def delete_dataset(
...
 
 
-def _delete_sample_group_links(session: Session, sample_ids: list[UUID]) -> None:
-    """Delete sample-group links for the given samples."""
-    for batch in batching.batched(items=sample_ids):
-        session.exec(
-            delete(SampleGroupLinkTable).where(col(SampleGroupLinkTable.sample_id).in_(batch))
-        )
+def _execute_delete(session: Session, statement: Delete) -> None:
+    """Run a server-side bulk DELETE without ORM session synchronization.
```
MalteEbner: This function adds quite a lot of indirection and extra lines of code. An alternative that gets rid of the extra function, while keeping the options at one place:

```python

# Comment on why this option makes sense.
DELETE_EXECUTION_OPTIONS={"synchronize_session": False}

def _delete_sample_tag_links(session: Session, dataset_id: UUID) -> None:
    """Delete sample-tag links for the dataset's samples."""
    session.exec(
        delete(SampleTagLinkTable).where(
            col(SampleTagLinkTable.sample_id).in_(_sample_ids_subquery(dataset_id))
        ),
        execution_options=DELETE_EXECUTION_OPTIONS
    )
```

MalteEbner (approved):

PR #1415, opened 2026-06-15
Index FKs

mihnea-necsulescu:
## What has changed and why?

#1408 was not enough on its own - the delete operation was still running after 15 minutes for ~250k samples. Unlike DuckDB, Postgres does not automatically index foreign keys. Deleting a parent row makes Postgres verify no orphan remains in every referencing table. Without indexes, this means that each parent row deletion will translate into a sequential scan over referencing tables.

This PR adds explicit indexes for every declared foreign key + the migration file for Postgres.

The same delete dropped from >15 min to ~330s (still pretty long, but it's expected). joins and filters should also be faster on enterprise now.

## How has it been tested?

Manual (with #1416)  + tests still pass

- [ ] Yes

## Summary by CodeRabbit

* **Chores**
  * Improved database performance by optimizing query lookups across annotations, samples, collections, embedding models, evaluation runs, and related entities, resulting in faster application responsiveness.


mihnea-necsulescu:
@coderabbitai review

MalteEbner (approved):

PR #1422, opened 2026-06-16
Split general e2e tests into two shards

michal-lightly:
## What has changed and why?

The general e2e suite was the CI bottleneck. Both the `General` (DuckDB) and `General Postgres` matrix jobs ran the full suite serially at ~10-11 min each, and test execution accounted for ~77% of that time.

Each is now split into two parallel shards using Playwright's `--shard` flag, dropping the critical path to ~6.5 min. The `shard` field is required on every matrix entry; unsharded suites (Videos, Groups) use `1/1`.

Note:
- Sharding is whole-file (config has `workers: 1`, no `fullyParallel`), so files are never split mid-way.
- Balance is by file count, not duration, so the split is roughly 55/45 rather than exactly even.

## How has it been tested?

Manually checked the CI runs on this PR.

- [ ] Yes

## Summary by CodeRabbit

* **Tests**
  * Improved test execution efficiency with optimized shard configuration
  * Enhanced reliability of annotation element detection in tests
  * Adjusted performance metrics collection for sharded test environments

### DIFF
--- a/.github/workflows/end2end_test.yml
+++ b/.github/workflows/end2end_test.yml
@@ -407,8 +407,7 @@ jobs:
         with:
           name: performance-metrics-${{ matrix.name }}
           path: lightly_studio_view/performance-metrics.json
-          # Only the shard that happens to run the performance test produces this file;
-          # the other general/videos shards legitimately have nothing to upload.
+          # Only the shard running the performance test produces this file.
           if-no-files-found: ignore
 
       - name: Create e2e success marker
--- a/.github/workflows/end2end_test.yml
+++ b/.github/workflows/end2end_test.yml
@@ -159,16 +159,26 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - name: General
+          - name: General 1/2
             index_script: index_general.py
             playwright_project: general
-            success_marker: .e2e-success-general
-            success_key: e2e-success-general
-            artifact_name: Playwright Report (General)
+            shard: "1/2"
+            success_marker: .e2e-success-general-1
+            success_key: e2e-success-general-1
+            artifact_name: Playwright Report (General 1)
+            postgres: false
+          - name: General 2/2
+            index_script: index_general.py
+            playwright_project: general
+            shard: "2/2"
+            success_marker: .e2e-success-general-2
+            success_key: e2e-success-general-2
+            artifact_name: Playwright Report (General 2)
             postgres: false
           - name: Videos
             index_script: index_video.py
             playwright_project: videos
+            shard: "1/1"
             success_marker: .e2e-success-videos
             success_key: e2e-success-videos
             artifact_name: Playwright Report (Videos)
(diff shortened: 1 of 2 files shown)


lightly_studio_view/e2e/pages/sample-details-page.ts
```diff
@@ -67,7 +67,11 @@ export class SampleDetailsPage {
     }
 
     getAnnotationBoxByLabel(label: string) {
-        return this.page.locator(`[data-testid="selectable-svg-group"] [data-label="${label}"]`);
+        // Match `data-annotation-label` (always rendered), not the `data-label` text overlay that
+        // only appears when the "show annotation text labels" setting is on.
+        return this.page.locator(
+            `[data-testid="selectable-svg-group"] [data-annotation-label="${label}"] [data-testid="annotation_box"]`
+        );
```
michal-lightly, line 74: This test depended on some state from a test from a different shard. This change fixes it.

lightly_studio_view/e2e/pages/sample-details-page.ts
```diff
@@ -67,7 +67,11 @@ export class SampleDetailsPage {
...
     }
```
MalteEbner, line 75: Is this an unrelated change?

.github/workflows/end2end_test.yml
```diff
@@ -159,37 +159,58 @@ jobs:
...
+            playwright_project: general
+            shard: "2/2"
+            success_marker: .e2e-success-general-2
+            success_key: e2e-success-general-2
+            artifact_name: Playwright Report (General 2)
             postgres: false
           - name: Videos
             index_script: index_video.py
             playwright_project: videos
+            shard: "1/1"
```
MalteEbner, line 181: Is this needed or is it the default?

MalteEbner (approved):
LGTM

lightly_studio_view/e2e/pages/sample-details-page.ts
michal-lightly: This fixes a related bug, added a comment.

.github/workflows/end2end_test.yml
michal-lightly: This is needed, otherwise `--shard=` would be passed to the command (i.e. without a parameter). An option would be to test for emptiness below, but I find this solution cleaner.

.github/workflows/end2end_test.yml
```diff
@@ -345,8 +366,13 @@ jobs:
...
         working-directory: lightly_studio_view
-        run: npx playwright test --reporter=html --trace=on --project=${{ matrix.playwright_project }}
-        timeout-minutes: 10
+        run: >-
+          npx playwright test
+          --reporter=html
+          --trace=on
+          --project=${{ matrix.playwright_project }}
+          --shard=${{ matrix.shard }}
+        timeout-minutes: 6
```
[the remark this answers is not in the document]
michal-lightly: Tested that the new timeout suffices. The other tests were not a bottleneck.

lightly_studio_view/e2e/pages/sample-details-page.ts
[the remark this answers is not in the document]
michal-lightly: Not worth changing. label here is hardcoded COCO fixture data ("cell phone",
  "apple"), not untrusted input, so there's no injection surface and no special
  characters to escape. The same unescaped pattern is already the convention in
  samples-page.ts ([data-annotation-label="${label}"]); escaping only this line
  would be an inconsistent one-off. Leaving as-is.

PR #1428, opened 2026-06-16
Create only one DB test fixture per worker

michal-lightly:
## What has changed and why?

Speeds up DuckDB python tests.

Before, every test got its own in-memory DuckDB, rebuilding the full schema (~62 ms) every time. Now each pytest worker gets its own DuckDB, resetting only the table data between tests.

Changes:
- **One engine per worker** also for DuckDB (already previously for Postgres)
- **Backend-aware table truncation.** DuckDB needs special handling because of the strict FK check. The self-referential `collections` table has to be deleted step-by-step from collection leaves to the root.
- **File-based DuckDB instead of `:memory:`.** A reused in-memory connection results in issues with cached query plans — this broke 4 tests on `Backend (3.14, DuckDB)`. The file backend invalidates these plans correctly, and moreover is closer to production setup. Full investigation in Linear.

## How has it been tested?

* CI passes, the time decreases
    * py3.9: from 190s to 145s (-45s)
    * py3.14: from 175s to 138s (-37s)

- [ ] Yes

## Summary by CodeRabbit

* **Tests**
  * Strengthened the test database setup to distinguish DuckDB file-based mode from PostgreSQL and to select the appropriate backend for test runs.
  * Improved per-test teardown to fully reset database state using backend-specific strategies, including careful handling of DuckDB foreign key constraints for consistent test isolation and reliability.


MalteEbner (approved):
LGTM

lukas-lightly (approved):

PR #1430, opened 2026-06-17
Add EvaluationMetricField class

lukas-lightly:
## What has changed and why?
This will be useful for querying evaluation metrics later

## How has it been tested?
Added tests

- [ ] Yes

## Summary by CodeRabbit

* **New Features**
  * Added dataset query support for filtering samples by persisted evaluation metric values for a specific evaluation run, with comparisons using `>`, `<`, `>=`, `<=`, `==`, and `!=`.

* **Public API**
  * Exposed `EvaluationMetricField` as part of the dataset query public API for simpler imports.

* **Tests**
  * Added coverage to validate generated SQL for the supported operators and to confirm end-to-end filtering returns only matching samples.


lightly_studio/src/lightly_studio/core/dataset_query/evaluation_metric_expression.py
```diff
@@ -0,0 +1,90 @@
...
+class EvaluationMetricField:  # noqa: PLW1641
+    """Queryable per-sample metric field from an evaluation run.
+
+    Example:
+        ``EvaluationMetricField(evaluation_run_id=run.id, metric_name="miou") < 0.3``
+    """
+
+    def __init__(self, evaluation_run_id: UUID, metric_name: str) -> None:
+        """Initialize a queryable metric reference."""
+        self.evaluation_run_id = evaluation_run_id
```
lukas-lightly, line 27: We use the run ID for the select, Codex review complained when I used the name, as the name doesn't have to be unique.

In the Query we'd like to take in run name but we'd need to convert it first to UUID.

lukas-lightly:
Note: the DB class we are querying here is called `EvaluationSampleMetricTable`, so this could also be called `EvaluationSampleMetricField`, but "sample context" is the default context, so I'd omit that for brevity.

For annotations we will have `EvaluationAnnotationMetricField` or similar.

lightly_studio/src/lightly_studio/core/dataset_query/evaluation_metric_expression.py
horatiualmasan: I think we could also make the name unique if it will simplify the query in the GUI. I don't recall any specific reason against it.

lightly_studio/tests/core/dataset_query/test_evaluation_metric_expression.py
```diff
@@ -0,0 +1,132 @@
...
+                sample_id=image_b.sample_id,
+                metric_name="score",
+                value=0.9,
+            ),
+        ],
+    )
+
+    results = (
+        DatasetQuery(dataset=dataset, session=db_session)
+        .match(EvaluationMetricField(evaluation_run_id=run.id, metric_name="score") > 0.5)
```
horatiualmasan, line 128: How would equality work for float values (for classification and semantic segmentation the metrics would be float values)? Should we consider this or leave it to the user to not use it?

lukas-lightly: Equality only makes sense with an epsilon, maybe we should add a `1e-6` tolerance? WDYT @JonasWurst ?

lightly_studio/src/lightly_studio/core/dataset_query/evaluation_metric_expression.py
lukas-lightly: I'd:
* Keep this PR as is
* Implement this issue fully, including the query in the GUI. See how it works with the name.
* Adjust the code here, if necessary.

Adding a TODO.

lightly_studio/tests/core/dataset_query/test_evaluation_metric_expression.py
JonasWurst: I Do not see many use cases for it. In case someone needs to filter like that two filters <> might be best.

lukas-lightly: Should I drop `==` then? I'd keep it though, it's weird to have all the other operators but not equality. Just don't use it :)

horatiualmasan (approved):
LGTM

PR #1441, opened 2026-06-18
Filter-based tagging: INSERT … SELECT tag-assignment helper (PR1)

michal-lightly:
## What has changed and why?

First backend PR towards filter-based tagging for "Select all" (LIG-9942). The "Select all → assign tag" path currently ships every matching sample ID over the wire, which fails with HTTP 413 at scale (~10M IDs ≈ 390 MB body).

This PR adds the DB-side primitives that let a later endpoint tag samples without materializing IDs on the client:

- `insert_from_select_ignoring_conflicts` — a single server-side `INSERT … SELECT` with dialect-aware conflict handling (Postgres `ON CONFLICT DO NOTHING`, DuckDB/SQLite `OR IGNORE`); no client-side rows, no bind-parameter cap, no batching.
- `add_samples_to_tag_from_query` — tags every sample a query returns, appending the constant `tag_id` column so it matches the link table. Idempotent via the composite primary key.

No endpoint or behavior change yet; these are wired up in follow-up PRs.

## How has it been tested?

New unit tests for both helpers, passing on DuckDB and PostgreSQL. Covers link creation, subset queries, idempotent re-runs, already-tagged skipping, and empty-query no-ops.

- [ ] Yes

## Release Notes

* **New Features**
  * Added server-side conflict handling for batch insertions, improving performance for large data imports.
  * New resolver function enables adding samples to tags based on custom query results.
  * Expanded database documentation on bulk-insert patterns.

* **Tests**
  * Comprehensive test coverage added for new bulk-insert and query-based sample-linking functionality.


lightly_studio/tests/database/test_db_insert.py
```diff
@@ -23,10 +30,11 @@ def test_insert_ignoring_conflicts__inserts_rows(db_session: Session) -> None:
...
     tag = create_tag(session=db_session, collection_id=collection_id)
-    samples = [
-        create_image(session=db_session, collection_id=collection_id, file_path_abs=f"/p/{i}.png")
-        for i in range(3)
-    ]
+    samples = create_images(
+        db_session=db_session,
+        collection_id=collection_id,
+        images=[ImageStub(path="/p/0.png"), ImageStub(path="/p/1.png")],
+    )
```
michal-lightly, line 37: Modernised the olds tests - used `create_images`

MalteEbner (approved):

PR #1444, opened 2026-06-19
Centralize effective annotation coloring mode

JonasWurst:
## Summary

This PR introduces a single, reusable decision function — `resolveEffectiveColorBySource` — as the canonical path for resolving whether annotations should be colored by source or by class. When **Enforce Coloring by Class** is enabled, the function always returns false regardless of how many sources are visible, overriding the existing multi-source heuristic. When the setting is off, existing behavior is preserved.

The function is applied to three places in the grid view:

- **Grid overlay coloring** (`SampleAnnotations`) — bounding boxes and segmentation masks on image thumbnails now color by class when enforcement is on, instead of by source.
- **Annotation Sources filter** (`AnnotationCollectionsMenu`) — source color markers in the filter panel are hidden when enforcement is on, since the source colors are no longer semantically meaningful.
- **Annotation Classes filter** (`LabelsMenu`) — class color legends, which were previously hidden whenever multiple sources were selected, now reappear when enforcement is on, reflecting the active coloring mode.

## Still pending (covered in follow-up PRs)

The following rendering paths still contain their own inline source-coloring decisions and have not yet been wired to `resolveEffectiveColorBySource`:

- **Image detail view overlays** — object-detection boxes, segmentation masks, and annotation text labels rendered in the full-size image detail view.
- **Side panel annotation rows** — color markers on individual annotation rows in the detail-view side panel.
- **Side panel source group headers** — color markers on source group headers in the detail-view side panel.
- **Classification pills and classification rows** — source vs. class coloring in the classification UI.
- **Video frame annotation overlays** — object-detection and segmentation overlays during video playback.

## Test plan

- [x] Toggle **Enforce Coloring by Class** in Settings with two annotation sources active in the grid view and confirm bounding box colors on thumbnails switch between source-based and class-based.
- [x] Confirm source color dots in the Annotation Sources filter panel disappear when enforcement is on.
- [x] Confirm class color swatches in the Annotation Classes filter panel reappear when enforcement is on.
- [x] `resolveEffectiveColorBySource` unit tests cover all four input combinations.
- [ ] `SampleAnnotations`, `AnnotationCollectionsMenu`, and `LabelsMenu` component tests cover both enforce-on and enforce-off cases with multiple sources.
- [ ] Full test suite passes (`224 test files, 1683 tests`).

## Summary by CodeRabbit

* **New Features**
  * Added an “enforce coloring by class” setting to control how annotation colors are assigned (class-based when enabled; source-based only when multiple sources are shown and enforcement is off).
* **UI Updates**
  * Updated annotation color markers and the label color legend visibility/behavior to reflect the new setting and multi-source selection rules.
* **Bug Fixes**
  * Corrected sample annotation coloring so label-to-color mapping matches the setting and current source selection.
* **Tests**
  * Expanded coverage for the setting-driven rendering and color behavior, including new utility tests.

### DIFF
--- a/lightly_studio_view/src/lib/components/LabelsMenu/LabelsMenu.svelte
+++ b/lightly_studio_view/src/lib/components/LabelsMenu/LabelsMenu.svelte
@@ -13,10 +13,10 @@
     const { enforceColoringByClassStore } = useSettings();
 
     const showClassColorLegend = $derived(
-        !resolveEffectiveColorBySource(
-            $selectedCollectionIds.length > 1,
-            $enforceColoringByClassStore
-        )
+        !resolveEffectiveColorBySource({
+            multipleSourcesVisible: $selectedCollectionIds.length > 1,
+            enforceColoringByClass: $enforceColoringByClassStore
+        })
     );
 
     let {
@@ -59,13 +59,12 @@
                         class="flex min-w-0 flex-1 cursor-pointer items-center space-x-2 text-nowrap peer-disabled:cursor-not-allowed peer-disabled:opacity-70"
                     >
                         {#if showClassColorLegend}
-                            <div data-testid="label-color-legend">
-                                <AnnotationColorLegend
-                                    labelName={label_name}
-                                    className="h-3 w-3"
-                                    {selected}
-                                />
-                            </div>
+                            <AnnotationColorLegend
+                                labelName={label_name}
+                                className="h-3 w-3"
+                                {selected}
+                                testId="label-color-legend"
+                            />
                         {/if}
                         <p
                             class="flex-1 truncate text-base font-normal"
--- a/lightly_studio_view/src/lib/components/LabelsMenu/LabelsMenu.svelte
+++ b/lightly_studio_view/src/lib/components/LabelsMenu/LabelsMenu.svelte
@@ -3,12 +3,14 @@
     import { Checkbox } from '$lib/components/ui/checkbox/index.js';
     import { Label } from '$lib/components/ui/label/index.js';
     import type { Annotation } from '$lib/types';
-    import { formatInteger } from '$lib/utils';
+    import { formatInteger, resolveEffectiveColorBySource } from '$lib/utils';
     import { type Readable } from 'svelte/store';
     import AnnotationColorLegend from '../AnnotationColorLegend/AnnotationColorLegend.svelte';
     import { useAnnotationCollectionsFilter } from '$lib/hooks/useAnnotationCollectionsFilter/useAnnotationCollectionsFilter';
+    import { useSettings } from '$lib/hooks/useSettings';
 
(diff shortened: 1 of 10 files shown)


lightly_studio_view/src/lib/components/LabelsMenu/LabelsMenu.svelte
```diff
@@ -49,12 +58,14 @@
...
                         class="flex min-w-0 flex-1 cursor-pointer items-center space-x-2 text-nowrap peer-disabled:cursor-not-allowed peer-disabled:opacity-70"
                     >
-                        {#if $selectedCollectionIds.length < 2}
-                            <AnnotationColorLegend
-                                labelName={label_name}
-                                className="h-3 w-3"
-                                {selected}
-                            />
+                        {#if showClassColorLegend}
+                            <div data-testid="label-color-legend">
```
LeonardoRosaa: Minor: you could pass this `data-testid` via `AnnotationColorLegend` props, or just define it in `AnnotationColorLegend` itself.

lightly_studio_view/src/lib/utils/resolveEffectiveColorBySource.ts
```diff
@@ -0,0 +1,13 @@
+/**
+ * Resolves whether annotations should be colored by source.
+ *
+ * When enforceColoringByClass is true, class colors always win regardless of how many
+ * sources are visible.
+ */
+export const resolveEffectiveColorBySource = (
```
LeonardoRosaa: I'd suggest using an inline object type since it's easy to make a mistake with two booleans. Example:

```ts
export const resolveEffectiveColorBySource = ({
  multipleSourcesVisible,
  enforceColoringByClass,
}: {
  multipleSourcesVisible: boolean;
  enforceColoringByClass: boolean;
}) => {
  // ...
};
```

lightly_studio_view/src/lib/utils/resolveEffectiveColorBySource.ts
```diff
@@ -0,0 +1,13 @@
...
+    multipleSourcesVisible: boolean,
+    enforceColoringByClass: boolean
+): boolean => {
+    if (enforceColoringByClass) return false;
```
LeonardoRosaa, line 14: We could simplify this condition, no? Example:

```ts
multipleSourcesVisible && !enforceColoringByClass
```

LeonardoRosaa (approved):

lightly_studio_view/src/lib/utils/resolveEffectiveColorBySource.ts
JonasWurst: I actually prefer the current version. It is most obvious that `enforceColoringByClass` is overruling everything when true.

PR #1461, opened 2026-06-22
Add image crop embedding infrastructure

horatiualmasan:
## What has changed and why?

Introduce ImageCrop and embed_image_crops() on the image embedding generators (MobileCLIP, Perception Encoder, Random) so object-level crops can be embedded with the same models as full images. Each source image is loaded once and crops are batched. Foundation for annotation-level embeddings.

## How has it been tested?

New tests

- [ ] Yes

## Release Notes

* **New Features**
  * Added a shared image-crop definition to describe regions to embed.
  * Introduced batched crop embedding with optional progress reporting, returning embeddings in the same order as the input crops.
  * Extended embedding generators to support embedding custom image crops (including MobileCLIP and Perception Encoder), with correct results even for empty inputs.

* **Tests**
  * Added coverage for empty-input handling, batching limits, and embedding order across multiple images.
  * Added checks that full-image crops match standard full-image embeddings.

### DIFF
--- a/lightly_studio/src/lightly_studio/dataset/image_crop_embedding.py
+++ b/lightly_studio/src/lightly_studio/dataset/image_crop_embedding.py
@@ -1,0 +1,114 @@
+"""Shared helper for batched image-crop embedding."""
+
+from __future__ import annotations
+
+from collections.abc import Callable
+from typing import Any
+
+import fsspec
+import numpy as np
+import torch
+from numpy.typing import NDArray
+from PIL import Image
+from tqdm import tqdm
+
+from lightly_studio.dataset.embedding_generator import ImageCrop
+
+
+def embed_image_crops_batched(  # noqa: PLR0913
+    image_crops: list[ImageCrop],
+    embedding_dimension: int,
+    max_batch_size: int,
+    device: torch.device,
+    preprocess: Callable[[Image.Image], torch.Tensor],
+    encode_batch: Callable[[torch.Tensor], NDArray[np.float32]],
+    show_progress: bool,
+) -> NDArray[np.float32]:
+    """Embed image crops, opening each source file once and preserving input order.
+
+    Args:
+        image_crops: Crop definitions to embed.
+        embedding_dimension: Output embedding width.
+        max_batch_size: Maximum crops encoded per model forward pass.
+        device: Torch device for model inference.
+        preprocess: Callable that converts a PIL crop to a model input tensor.
+        encode_batch: Callable that encodes a batch tensor and returns embeddings.
+        show_progress: Whether to show a tqdm progress bar.
+
+    Returns:
+        Float32 array of shape ``(len(image_crops), embedding_dimension)``.
+    """
+    total_crops = len(image_crops)
+    if not total_crops:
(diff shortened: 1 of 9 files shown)


lightly_studio/src/lightly_studio/dataset/image_crop_embedding.py
```diff
@@ -0,0 +1,115 @@
...
+    device: torch.device,
+    preprocess: Callable[[Image.Image], torch.Tensor],
+    encode_batch: Callable[[torch.Tensor], NDArray[np.float32]],
+    show_progress: bool,
+) -> NDArray[np.float32]:
+    """Embed image crops, opening each source file once and preserving input order.
+
+    Args:
+        image_crops: Crop definitions to embed.
+        embedding_dimension: Output embedding width.
```
JonasWurst:
```suggestion
        embedding_dimension: Output embedding dimension.
```

lightly_studio/src/lightly_studio/dataset/image_crop_embedding.py
```diff
@@ -0,0 +1,115 @@
...
+from tqdm import tqdm
+
+from lightly_studio.dataset.embedding_generator import ImageCrop
+
+
+def embed_image_crops_batched(  # noqa: PLR0913
+    image_crops: list[ImageCrop],
+    embedding_dimension: int,
+    max_batch_size: int,
+    device: torch.device,
```
JonasWurst: I would wrap them into a data class to prevent the PLR0913.

`EmbeddingContext`

I would also add the `tensor_size` as arg -> we then can avoid appending the preprocessed images but directly initialize a tensor. should speed up the method

lightly_studio/src/lightly_studio/dataset/image_crop_embedding.py
```diff
@@ -0,0 +1,115 @@
...
+    total_crops = len(image_crops)
+    if not total_crops:
+        return np.empty((0, embedding_dimension), dtype=np.float32)
+
+    crops_by_filepath: dict[str, list[tuple[int, ImageCrop]]] = {}
+    for index, image_crop in enumerate(image_crops):
+        crops_by_filepath.setdefault(image_crop.filepath, []).append((index, image_crop))
+
+    embeddings = np.empty((total_crops, embedding_dimension), dtype=np.float32)
+    batch_tensors: list[torch.Tensor] = []
```
JonasWurst: This I would change, see comment above

lightly_studio/tests/dataset/test_image_crop_embedding.py
JonasWurst, line 1: This test needs rework after my suggested change

JonasWurst (changes requested):
See my comments on the input tensor handling

lightly_studio/tests/dataset/test_image_crop_embedding.py
horatiualmasan: The tests is fine. It will test that the order of the embeddings is the same as the input crop order ( as they will be order by file path before processing)

JonasWurst (approved):
LGTM

PR #1469, opened 2026-06-23
Cache embedding generators across collections in EmbeddingManager

horatiualmasan:
## What has changed and why?

Cache loaded embedding generators by sample type so image and annotation collections share one in-memory model instance.
Each collection still gets its own embedding-model DB record and ID.

## How has it been tested?

new tests

- [ ] Yes

## Summary by CodeRabbit

* **Refactor**
  * Improved embedding model loading by normalizing collection sample types so compatible collections (e.g., image and annotation) reuse the same underlying generator instead of loading weights repeatedly.
  * For unsupported sample types, default embedding model loading now cleanly returns no model rather than proceeding.

* **Tests**
  * Added a unit test ensuring the generator is loaded only once, shared across the resulting embedding model instances, and that persisted model record IDs are created distinctly.

### DIFF
--- a/lightly_studio/src/lightly_studio/dataset/embedding_manager.py
+++ b/lightly_studio/src/lightly_studio/dataset/embedding_manager.py
@@ -323,7 +323,7 @@ class EmbeddingManager:
             return None
 
         if cache_key not in self._generators:
-            generator = _load_embedding_generator_from_env(sample_type=sample_type)
+            generator = _load_embedding_generator_from_env(sample_type=cache_key)
             if generator is None:
                 return None
             self._generators[cache_key] = generator
--- a/lightly_studio/src/lightly_studio/dataset/embedding_manager.py
+++ b/lightly_studio/src/lightly_studio/dataset/embedding_manager.py
@@ -35,6 +35,13 @@ logger = logging.getLogger(__name__)
 # round-trips but higher peak memory. 1024 balances the two.
 EMBEDDING_INSERTION_BATCH_SIZE = 1024
 
+# Mapping of sample types to the generator type used for embedding generation.
+_GENERATOR_SAMPLE_TYPE: dict[SampleType, SampleType] = {
+    SampleType.IMAGE: SampleType.IMAGE,
+    SampleType.ANNOTATION: SampleType.IMAGE,
+    SampleType.VIDEO: SampleType.VIDEO,
+}
+
 
 class EmbeddingManagerProvider:
     """Provider for the EmbeddingManager singleton instance."""
@@ -71,9 +78,7 @@ class EmbeddingManager:
         """Initialize the embedding manager."""
         self._models: dict[UUID, EmbeddingGenerator] = {}
         self._collection_id_to_default_model_id: dict[UUID, UUID] = {}
-        # Loaded generators cache. A single physical model is held
-        # in memory once and shared across collections.
-        self._generators: dict[SampleType, EmbeddingGenerator] = {}
+        self._sample_type_to_model_id: dict[SampleType, UUID] = {}
 
     def register_embedding_model(
         self,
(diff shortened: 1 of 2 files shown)


lightly_studio/src/lightly_studio/dataset/embedding_manager.py
```diff
@@ -357,6 +384,19 @@ def _store_embeddings(
...
+    """Map a sample type to the key under which its generator is cached.
+
+    Image and annotation samples share the same image generator, so both map to
+    SampleType.IMAGE and reuse a single loaded instance.
+    """
+    if sample_type in (SampleType.IMAGE, SampleType.ANNOTATION):
+        return SampleType.IMAGE
+    if sample_type == SampleType.VIDEO:
+        return SampleType.VIDEO
+    return None
```
JonasWurst: I'm not sure if that will scale. At the moment we only support OD, not annotations in general

MalteEbner: This could be a dict instead, which makes it easier to understand:

```
_GENERATOR_SAMPLE_TYPE: dict[SampleType, SampleType] = {
    SampleType.IMAGE: SampleType.IMAGE,
    SampleType.ANNOTATION: SampleType.IMAGE,
    SampleType.VIDEO: SampleType.VIDEO,
}
```

lightly_studio/src/lightly_studio/dataset/embedding_manager.py
```diff
@@ -303,6 +306,30 @@ def load_or_get_default_model(
 
         return embedding_model.embedding_model_id
 
+    def _get_or_load_generator(self, sample_type: SampleType) -> EmbeddingGenerator | None:
```
MalteEbner: Instead of implementing a cache ourselves, we could also just use [functools.cache](https://docs.python.org/3/library/functools.html#functools.cache). Much less LOC, no logic by us. It caches by sample_type automatically.

```python
@functools.cache
def _load_cached_embedding_generator_from_env(
    sample_type: SampleType,
) -> EmbeddingGenerator | None:
    return _load_embedding_generator_from_env(sample_type=sample_type)
```

MalteEbner (commented):
I think this can be made easier by replacing `_generator_cache_key` with a `_GENERATOR_SAMPLE_TYPE` dict and the caching with `functools.cache`. See my other comments. Then there would be no need for an extra function and the change in `load_or_get_default_model` would only be

```python
generator_sample_type = _GENERATOR_SAMPLE_TYPE.get(dataset.sample_type)
if generator_sample_type is None:
    return None

embedding_generator = _load_cached_embedding_generator_from_env(
    sample_type=generator_sample_type
)
```

lightly_studio/src/lightly_studio/dataset/embedding_manager.py
```diff
@@ -71,6 +71,9 @@ def __init__(self) -> None:
         """Initialize the embedding manager."""
         self._models: dict[UUID, EmbeddingGenerator] = {}
         self._collection_id_to_default_model_id: dict[UUID, UUID] = {}
+        # Loaded generators cache. A single physical model is held
+        # in memory once and shared across collections.
+        self._generators: dict[SampleType, EmbeddingGenerator] = {}
```
JonasWurst: I think this change is not fully integrated:

If I need an embedding model for a certain colleciton, this will now either instantiate a new one or get one from `_generators`, then this model will be passed to the registration, So we will register the same model twice, is that intended?

Instead what i would assume:
- we check for a collection if we already have a registered model mapped - use it
- if not, we check if any of the registered models can be reused (Here the type should be sufficient eg ImageEmbeddingGenerator etc.),
- If not only then instantiate new model an register it.

Hence we have the 
self._models: dict[UUID, EmbeddingGenerator]
self._collection_id_to_model_id:dict[UUID, UUID]

At some point we might even want to extend this interface to 
self._collection_id_to_model_id:dict[UUID, list[UUID]]

JonasWurst (changes requested):

lightly_studio/src/lightly_studio/dataset/embedding_manager.py
MalteEbner: This assumes that we support different models for different collections for the same sample type. We don't at the moment. And it is out of scope for this PR. 

Thus I would keep the scope of this PR as it is: Just make sure that the image and annotation embedding model are the same.

lightly_studio/src/lightly_studio/dataset/embedding_manager.py
MalteEbner: I think it is fine that we map all annotation kinds to that model. We can abstract once we actually have different models for different annotation types, but no need to do that yet.

JonasWurst (approved):

PR #1472, opened 2026-06-24
Speed up sample_embedding_resolver.get_by_sample_ids more than 10x on Postgres

MalteEbner:
## What has changed and why?

`sample_embedding_resolver.get_by_sample_ids` is a scaling bottleneck when loading embeddings for many samples (2D view, sampling, few-shot classifier).

- **Part 1** (tiny). Before this, it used to call `.where(cast(SampleEmbeddingTable.sample_id, String).in_(string_ids))`, which made the primary-key index unusable and forced a full sequential scan per batch (≈ O(N²)). It now compares UUIDs directly so the index is used. This is a speedup only at 1M/10M samples. At 100k the table is small enough that it is not yet visible.
- **Part 2** (tiny). The result is now a NamedTuple `SampleEmbeddingRow` (`sample_id` + `embedding`) instead of a full `SampleEmbeddingTable` object. Faster in general. Required for Part 3.
- **Part 3**. Split the function in a Postgres and DuckDB part. On Postgres create`SampleEmbeddingRow` directly as binary floats. Previously it was floats -> strings like "1.342" -> floats again.  This change also makes part 1 not apply to Postgres anymore.

## Benchmark 
on 100k samples, 512-dim embeddings, loading speed

| backend | part 1 only | all parts |
|---|---|---|
| local Postgres | 27.19s | 1.74s |
| DuckDB | 16.50s | 14.44s |

Did not test before part 1, cause I had committed both together. Does not give speedups on 100k samples anyway.

## How has it been tested?

Unit tests, plus a new benchmark file.

- [ ] Yes

## Summary by CodeRabbit

* **New Features**
  * Improved sample embedding lookups to return results faster and with lower memory use.
  * Sample embeddings now preserve the requested order and skip missing items cleanly.
  * Added support for more efficient reads across different database backends.

* **Bug Fixes**
  * Empty embedding requests now return immediately instead of running a query.


lightly_studio/tests/benchmarks/sample_embedding_lookup_benchmark.py
```diff
@@ -0,0 +1,306 @@
+"""Sample-embedding lookup micro-benchmark.
```
MalteEbner, line 1: Do NOT review this benchmark

lightly_studio/src/lightly_studio/resolvers/sample_embedding_resolver.py
```diff
@@ -61,21 +75,81 @@ def get_by_sample_ids(
...
+    """Load embeddings via the SQLAlchemy query path (DuckDB).
+
+    DuckDB returns the vectors as arrays natively, so selecting just ``sample_id`` and
+    ``embedding`` avoids creating full ``SampleEmbeddingTable`` objects. IDs are batched
+    to stay under the database's parameter limit.
+    """
+    rows: list[SampleEmbeddingRow] = []
+    for batch in batching.batched(items=sample_ids):
+        results = session.exec(
+            select(SampleEmbeddingTable.sample_id, col(SampleEmbeddingTable.embedding))
```
MalteEbner, line 134: This is part 1, replacing the previous `.where(cast(SampleEmbeddingTable.sample_id, String).in_(string_ids))`

lightly_studio/src/lightly_studio/resolvers/sample_embedding_resolver.py
```diff
@@ -48,7 +62,7 @@ def get_by_sample_ids(
     session: Session,
     sample_ids: list[UUID],
     embedding_model_id: UUID,
-) -> list[SampleEmbeddingTable]:
+) -> list[SampleEmbeddingRow]:
```
MalteEbner, line 65: Part 2, it gives another speedup and allows more flexibility in the actual creation -> relevant for part 3

lightly_studio/src/lightly_studio/resolvers/sample_embedding_resolver.py
```diff
@@ -61,21 +75,81 @@ def get_by_sample_ids(
...
+        # Register pgvector once per connection so psycopg decodes ``vector`` columns
+        # from binary instead of parsing them from text.
+        pgvector.psycopg.register_vector(driver_connection)
+        connection.info["pgvector_registered"] = True
+    query = (
+        "SELECT sample_id, embedding FROM sample_embedding "
+        "WHERE embedding_model_id = %s AND sample_id = ANY(%s)"
+    )
+    with driver_connection.cursor(binary=True) as cursor:
+        cursor.execute(query, (embedding_model_id, sample_ids))
```
MalteEbner, line 113: Part 3, transmit the vector in binary format instead of floats -> strings like "0.0003" -> float again

lightly_studio/src/lightly_studio/resolvers/sample_embedding_resolver.py
```diff
@@ -3,12 +3,15 @@
 from __future__ import annotations
 
 import hashlib
+from typing import NamedTuple
 from uuid import UUID
 
-from sqlalchemy import String, cast, func
+import pgvector.psycopg
```
mihnea-necsulescu: ~should we move this into `_get_by_sample_ids_postgres` as a local import?~

won't be needed at all after https://github.com/lightly-ai/lightly-studio/pull/1472/changes#r3466613169

lightly_studio/src/lightly_studio/resolvers/sample_embedding_resolver.py
```diff
@@ -61,21 +75,81 @@ def get_by_sample_ids(
...
-        string_ids = [str(id_) for id_ in batch]
-        results.extend(
-            session.exec(
-                select(SampleEmbeddingTable)
-                .where(cast(SampleEmbeddingTable.sample_id, String).in_(string_ids))
-                .where(SampleEmbeddingTable.embedding_model_id == embedding_model_id)
-            ).all()
+    if not sample_ids:
+        return []
+    if session.get_bind().dialect.name == "postgresql":
```
mihnea-necsulescu: can we directly compare against the Postgres enum?
```suggestion
    if session.get_bind().dialect.name == DatabaseBackend.POSTGRESQL.value:
```

lightly_studio/src/lightly_studio/resolvers/sample_embedding_resolver.py
```diff
@@ -61,21 +75,81 @@ def get_by_sample_ids(
...
     embedding_map = {embedding.sample_id: embedding for embedding in results}
     return [embedding_map[id_] for id_ in sample_ids if id_ in embedding_map]
 
 
+def _get_by_sample_ids_postgres(
+    session: Session,
+    sample_ids: list[UUID],
+    embedding_model_id: UUID,
+) -> list[SampleEmbeddingRow]:
+    """Load embeddings with a binary psycopg cursor (PostgreSQL only).
```
mihnea-necsulescu, line 98: this basically fully bypasses SQLAlchemy, no? I think it's worth mentioning that, too

lightly_studio/src/lightly_studio/resolvers/sample_embedding_resolver.py
```diff
@@ -61,21 +75,81 @@ def get_by_sample_ids(
...
+    session.flush()
+    connection = session.connection().connection
+    driver_connection = connection.driver_connection
+    if driver_connection is None:  # pragma: no cover - a live session always has one
+        raise RuntimeError("PostgreSQL session has no underlying psycopg connection.")
+    if not connection.info.get("pgvector_registered"):
+        # Register pgvector once per connection so psycopg decodes ``vector`` columns
+        # from binary instead of parsing them from text.
+        pgvector.psycopg.register_vector(driver_connection)
+        connection.info["pgvector_registered"] = True
```
mihnea-necsulescu: this is a lot of pgvector specific stuff that's leaking into the resolver - I think it makes sense to extract this into a helper function, something like `get_pgvector_connection`, maybe in `db_vector.py`

mihnea-necsulescu (changes requested):

lightly_studio/src/lightly_studio/resolvers/sample_embedding_resolver.py
MalteEbner: Yes, done

lightly_studio/src/lightly_studio/resolvers/sample_embedding_resolver.py
MalteEbner: Yes, done

lightly_studio/src/lightly_studio/resolvers/sample_embedding_resolver.py
MalteEbner: Yes, Done

mihnea-necsulescu (approved):

PR #1474, opened 2026-06-24
Add select-all filter snapshot stores to useGlobalStorage

michal-lightly:
## What has changed and why?

First PR to allow tagging by filter in the frontend.

Introduces stores to capture the current filter state when select-all is toggled.

Any manual selection mutation invalidates the matching snapshot, so a non-null snapshot means "still an unmodified select-all." Behavior-neutral — nothing reads the snapshot yet.

## How has it been tested?

New tests.

- [ ] Yes

## Summary by CodeRabbit

* **New Features**
  * Added cached “select all” snapshot memory separately for sample and annotation grids, preserved per collection to improve consistency when switching actions.

* **Bug Fixes**
  * “Select all” snapshots now invalidate correctly when individual samples/annotations are toggled or when selections are cleared.
  * Unrelated actions (such as range filtering and bulk “set all selected” updates) no longer reset these snapshots unexpectedly.
  * Sample and annotation “select all” states remain independent from each other.

### DIFF
--- a/lightly_studio_view/src/lib/hooks/useGlobalStorage.test.ts
+++ b/lightly_studio_view/src/lib/hooks/useGlobalStorage.test.ts
@@ -2,9 +2,7 @@ import { get } from 'svelte/store';
 import { beforeEach, describe, expect, it, vi } from 'vitest';
 import { useGlobalStorage } from './useGlobalStorage';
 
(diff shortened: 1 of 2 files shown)


michal-lightly:
@coderabbit review

ikondrat (approved):

PR #1476, opened 2026-06-24
Drop SampleEvaluationQuery.match()

lukas-lightly:
## What has changed and why?
This is a subtle change but I think it makes the API much nicer to use:
- Use `SampleEvaluationQuery.__init__` instead of `SampleEvaluationQuery.match()`
- Drop `SampleEvaluationMatchExpression`, instead make `SampleEvaluationQuery` inherit from `MatchExpression`.
- Rename the file after the name change

## How has it been tested?
Existing tests

- [ ] Yes

## Summary by CodeRabbit

* **Bug Fixes**
  * Simplified sample evaluation query usage so queries can be created directly with a run name and criteria, improving consistency in filtering workflows.
  * Updated query examples to match the current syntax for building evaluation filters.
* **Tests**
  * Refreshed query-related tests to cover the updated construction style and filtering behavior.


lightly_studio/tests/core/dataset_query/test_sample_evaluation_query.py
```diff
@@ -66,7 +66,7 @@ def test_sample_evaluation_query__filters_matching_samples(db_session: Session)
 
     results = (
         DatasetQuery(dataset=dataset, session=db_session)
-        .match(SampleEvaluationQuery.match("run1", EvaluationMetricField("score") > 0.5))
+        .match(SampleEvaluationQuery("run1", EvaluationMetricField("score") > 0.5))
```
lukas-lightly, line 69: Note: we don't strictly follow the rule to use named arguments everywhere, it would make this too verbose:
`SampleEvaluationQuery(run_name="run1", criterion=EvaluationMetricField("score") > 0.5)`

michal-lightly (approved):

PR #1481, opened 2026-06-24
evaluations(pr1): evaluate an explicit sample set in ImageDatasetEvaluate

ikondrat:
## What has changed and why?

Take an explicit set of sample IDs (instead of ImageSample objects) so callers, including the HTTP trigger, can drive evaluation from sample IDs.

## How has it been tested?

By the test

- [ ] Yes

## Summary by CodeRabbit

* **New Features**
  * Evaluation results now include an **evaluation run ID** for easier tracking and reference.

* **Bug Fixes**
  * Image dataset evaluation now bases result selection directly on the **selected sample IDs**, improving consistency.
  * An explicit error is raised if the same source is used for both **ground truth** and **predictions**, preventing accidental misconfiguration.


horatiualmasan (approved):
LGTM

PR #1484, opened 2026-06-25
feat: refactor evaluation routes into a barrel package

ikondrat:
## What has changed and why?

Pure structural refactor — no endpoint or behavior change. The single `evaluation.py` route file is split into an `evaluation/` package where each route handler lives in its own module with its own `APIRouter`, and a barrel `__init__.py` aggregates them into the `evaluation_router` the app mounts:

- `evaluation/get_runs.py`
- `evaluation/get_sample_metrics_info.py`
- `evaluation/get_confusion_matrix.py`
- `evaluation/__init__.py` (barrel)

This matches the per-function + barrel convention used elsewhere (resolvers, services) and keeps each handler small and independently reviewable.

The route functions move modules, so the tests' `mocker.patch` targets are repointed accordingly (e.g. `…api.evaluation.evaluation_run_resolver` → `…api.evaluation.get_runs.evaluation_run_resolver`). No test logic changes here — the file split comes in the next PR.

## How has it been tested?

```bash
cd lightly_studio && make static-checks && make test
```

The existing `test_evaluation.py` continues to pass against the same endpoints (only its patch paths changed).

- [ ] Yes

## Summary by CodeRabbit

* **New Features**
  * Added endpoints to list evaluation runs, retrieve sample metric info, and fetch confusion matrices for supported evaluation runs.
  * Evaluation responses now include clearer source names for ground-truth and prediction annotations.
* **Bug Fixes**
  * Improved handling for missing or unsupported evaluation run types, returning appropriate errors instead of failing unexpectedly.
* **Chores**
  * Reorganized evaluation routes to a new structure with no expected change in user-facing behavior.


lightly_studio/src/lightly_studio/evaluation/image_dataset_evaluate.py
```diff
@@ -229,7 +233,7 @@ def _prepare_evaluation_data(
             config_json=config_json,
         )
 
-        selected_sample_ids = {sample.sample_id for sample in self.samples}
+        selected_sample_ids = set(self.sample_ids)
```
JonasWurst: why again set?

lightly_studio/src/lightly_studio/models/evaluation_run.py
```diff
@@ -61,5 +61,5 @@ class EvaluationRunView(BaseModel):
     name: str
     evaluation_run_configuration: dict[str, Any]
     created_at: datetime
-    gt_annotation_source: str
-    pred_annotation_source: str
+    gt_annotation_source: str | None
+    pred_annotation_source: str | None
```
JonasWurst: why would we want this? such a run does not make sense. Can you motivate this?

JonasWurst (commented):
Can you clarify the open questions

lightly_studio/src/lightly_studio/models/evaluation_run.py
ikondrat: fixed

lightly_studio/src/lightly_studio/evaluation/image_dataset_evaluate.py
ikondrat: added comment below it was proposal from AI

```
A set is required for the intersection with the covered-sample sets below; it
also dedups the input. Ordering is irrelevant since annotations are regrouped
by parent sample ID afterwards.
```

JonasWurst (approved):

PR #1487, opened 2026-06-25
Add routing function to trigger model evaluation from gui

ikondrat:
## What has changed and why?

Adds the ability to start an evaluation run over HTTP, with the orchestration extracted into a service so the route stays thin and the logic is unit-testable without the HTTP layer.

- **`services/evaluation_service/run_evaluation.py`** — `run_evaluation(...)`: resolves the samples to evaluate (optional active-view `filter`), builds the `ImageDatasetEvaluate` facade, and dispatches to the task method. The task is selected by the **config type**, so the function takes only primitives — no dependency on the route's request schemas. Raises `ValueError` (→ 400) for a non-image collection, the same source used for GT and predictions, an unknown source, or a source whose annotations don't match the task type.
- **`evaluation/create_run.py`** — thin `POST /datasets/{dataset_id}/evaluation/runs` route: validates the collection exists in the dataset (404), then delegates to `run_evaluation`. Registered in the barrel `__init__`.

Request body is discriminated on `task_type` (`object_detection` | `classification` | `semantic_segmentation`) with shared fields (`gt_annotation_source`, `pred_annotation_source`, `collection_id`, optional `filter`/`name`, per-task `config`); returns `201` with `EvaluationResult`.

**Scope:** synchronous (blocking) by design; the GUI sends no filter (full dataset) for now; no frontend in this PR.

## How has it been tested?

`tests/.../evaluation/test_create_run.py`: object-detection and classification triggers, name/default-config generation, `filter` respected, and the error cases — `400` for same source, wrong annotation type, and unknown source; `404` for collection not found; `400` for a non-image collection.

```bash
cd lightly_studio && make static-checks && make test
```

- [ ] Yes

## Summary by CodeRabbit

* **New Features**
  * Added an evaluation run API so evaluations can be triggered directly from the dataset context.
  * Supports object detection, classification, and semantic segmentation run creation.
  * Runs can be filtered to specific image subsets and use custom (or auto-generated) run names.

* **Bug Fixes**
  * Added validation to ensure evaluations only run for valid image collections under the selected dataset.
  * Improved handling for missing collections/annotation sources and invalid annotation-source setups.

* **Tests**
  * Added route and service tests covering success paths, defaults, filtering, name uniqueness, and error cases.


lightly_studio/src/lightly_studio/services/evaluation_service/run_evaluation.py
```diff
@@ -0,0 +1,113 @@
...
+    ClassificationEvaluationConfig,
+    SemanticSegmentationEvaluationConfig,
+]
+
+
+def run_evaluation(  # noqa: PLR0913
+    session: Session,
+    *,
+    collection: CollectionTable,
+    task_type: EvaluationTaskType,
```
horatiualmasan, line 39: We could remove the task_type parameter and create the name after the config type is resolved. This would also remove any possible mismatches between task_type and config type.

lightly_studio/src/lightly_studio/services/evaluation_service/run_evaluation.py
```diff
@@ -0,0 +1,113 @@
...
+    config: EvaluationConfig,
+    filters: ImageFilter | None = None,
+    name: str | None = None,
+) -> EvaluationResult:
+    """Run an evaluation for ``collection`` and persist it.
+
+    The task is selected by the concrete type of ``config`` (which also carries
+    the task-specific options). ``filters`` scopes the evaluated samples; when
+    omitted the whole collection is evaluated. ``name`` defaults to
+    ``"<task> <UTC timestamp>"`` (with microsecond precision so concurrent
```
horatiualmasan:
```suggestion
    ``"<task_type> <UTC timestamp>"`` (with microsecond precision so concurrent
```

lightly_studio/tests/services/evaluation_service/test_run_evaluation.py
```diff
@@ -0,0 +1,241 @@
...
+from lightly_studio.services import evaluation_service
+from tests.helpers_resolvers import (
+    create_annotation,
+    create_annotation_label,
+    create_collection,
+    create_image,
+)
+
+
+def _create_dataset_with_annotations(
```
horatiualmasan: It is also implemented in api/routes/api/evaluation/helpers.py, we can use it from there or move it to tests\helpers_resolvers.py

horatiualmasan (approved):
Left small comments, otherwise LGTM

lightly_studio/src/lightly_studio/services/evaluation_service/run_evaluation.py
ikondrat: I think it would be overcomplicated, would be more straightforward to just pass type here

PR #1488, opened 2026-06-25
Expose annotation types per source

ikondrat:
### What
Surfaces the distinct annotation types contained in each annotation collection
on the `/annotation_collections` response, so the trigger-evaluation dialog can
filter source dropdowns to only the sources compatible with the chosen task
(classification / object detection / semantic segmentation).

### Why
This is a UX guardrail, not a correctness requirement — the backend already
rejects mismatched or missing sources with a 400 (`run_evaluation` validates the
source's annotation type against the task). Pre-filtering the dropdowns just
stops the user from picking an incompatible source in the first place, avoiding a
confusing failed round-trip. The field is consumed by the frontend source-filter
work that stacks on top of this one.

### Changes
- New resolver `get_annotation_types_by_collection_ids` — batched query returning
  `dict[collection_id, sorted list[str]]` of distinct annotation types (joins
  annotations to their sample's `collection_id`; avoids an N+1 over collections).
- New `annotation_types: list[str]` field on `AnnotationCollectionView` (typed as
  plain strings to avoid a circular import with `annotation_base`).
- `read_annotation_collections` populates the field for each returned collection.

### Tests
- Resolver unit test: distinct + alphabetically sorted, empty input returns `{}`,
  collections with no annotations are omitted.
- Route test: `test_read_annotation_collections` now asserts `annotation_types` is
  returned by the endpoint.

## Summary by CodeRabbit

* **New Features**
  * Annotation collections now return `annotation_types` for each collection.
  * Enables displaying/using the available annotation types when working with selected collections.

* **Bug Fixes**
  * `annotation_types` are deduplicated and returned in alphabetical order.
  * Collections with no annotations are handled gracefully, and empty requests return empty results.

* **Tests**
  * Updated API tests for the expanded response.
  * Added resolver tests for distinctness, sorting, and empty/missing-annotation cases.


lightly_studio/src/lightly_studio/resolvers/collection_resolver/get_annotation_types_by_collection_ids.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+def get_annotation_types_by_collection_ids(
+    session: Session,
+    collection_ids: Sequence[UUID],
+) -> dict[UUID, list[str]]:
+    """Return the distinct annotation types present in each annotation collection.
+
+    Annotations belong to a collection through their sample's ``collection_id``
+    (mirrors the join used by the evaluation validators).
```
horatiualmasan: [optional] We could remove this.

horatiualmasan (approved):
LGTM

PR #1503, opened 2026-06-26
Resolve confusion-matrix FP/FN buckets in the sample filter

JonasWurst:
## What

Part 1 of 2 (backend). Extends the `confusion_cell` sample filter to the confusion matrix's synthetic margin buckets, so a caller can select **false positives** (a prediction with no matching ground truth) and **false negatives** (a ground truth the model missed), not just confusions between two real classes.

## Changes

- `ConfusionCell.gt_label` / `pred_label` are now optional. A `null` label selects a margin bucket:
  - `gt_label = null` → false-positive bucket (no gt pairing)
  - `pred_label = null` → false-negative bucket (no pred pairing)
  - a model validator rejects the meaningless `gt_label = null && pred_label = null` corner.
- `_build_confusion_cell_subquery` now handles each side independently: a real label inner-joins its annotation/label and matches by name; a `null` label instead requires that side's annotation id to be `NULL`.

## Tests

`test_sample_filter.py` gains coverage for the FP bucket (`gt_label = null`), the FN bucket (`pred_label = null`), and the rejected null/null combination. The filter is task-agnostic, so it works for both classification and object-detection runs.

## Notes for reviewer

- The OpenAPI schema / generated TS client are gitignored build artifacts; the frontend PR (part 2) regenerates them and consumes the now-nullable labels.

## Summary by CodeRabbit

* **New Features**
  * Added support for additional confusion-matrix “margin” buckets when either ground-truth or prediction is missing.
  * Updated sample filtering so results match these false-positive/false-negative bucket definitions.
* **Bug Fixes**
  * Strengthened validation to reject confusion-matrix cells where both ground-truth and prediction are missing.
* **Tests**
  * Added coverage for false-positive and false-negative scenarios to ensure the correct sample is returned.

### DIFF
--- a/lightly_studio/src/lightly_studio/models/evaluation_confusion_matrix.py
+++ b/lightly_studio/src/lightly_studio/models/evaluation_confusion_matrix.py
@@ -4,7 +4,7 @@ from __future__ import annotations
 
 from uuid import UUID
 
-from pydantic import BaseModel, Field
+from pydantic import BaseModel, Field, model_validator
 
 # Row label for predictions with no matching ground truth (false positives).
 NO_GROUND_TRUTH_ROW_LABEL = "(no ground truth)"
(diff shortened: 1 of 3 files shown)


horatiualmasan (approved):
LGTM

PR #1507, opened 2026-06-29
make the evaluation side panel reachable for image collections for zero evaluation runs

ikondrat:
## What has changed and why?

Part of LIG-9843 (trigger model evaluation from the GUI). This PR makes the
evaluation side panel reachable for image collections **even when there are zero
evaluation runs**, so users can trigger their first run from the GUI instead of
the panel only appearing after a run already exists.

Key changes:
- Gate the evaluation tab/panel on `supportsEvaluation` (currently `isImages`)
  rather than on whether evaluation runs already exist. The panel and its tab now
  render with no runs present.
- Rename the `SidePanelTabs` prop `hasEvaluationRuns` → `supportsEvaluation` and
  simplify its render condition (`isImages && hasEvaluationRuns` → `supportsEvaluation`).
- Drop the now-unused `hasEvaluationRuns` derivation in the collection layout and
  update `panelIsVisible` / panel mount / tab-visibility conditions accordingly.

(Backend pieces of LIG-9843 — the POST endpoint to trigger evaluation runs and
exposing annotation types per source — land in the base branch.)

## How has it been tested?

- Updated `SidePanelTabs.test.ts` to cover the new `supportsEvaluation` prop,
  including that the Eval tab renders when `supportsEvaluation` is true and is
  hidden otherwise.
- Updated `SidePanelTabs.stories.svelte` stories ("Images With Evaluation", etc.)
  to the new prop.
- Manually verified in the GUI that an image collection with no evaluation runs
  shows the evaluation tab and that the panel opens.

- [ ] Yes

## Summary by CodeRabbit

* **New Features**
  * Evaluation tools are now available for image collections even when no evaluation results exist yet.
  * The side panel can now open the Evaluation tab based on collection type, improving access to evaluation-related actions.

* **Bug Fixes**
  * Updated tab visibility so the Evaluation option appears consistently in the appropriate image collection views.
  * Improved related app behavior and tests to match the new evaluation tab availability.


lightly_studio/src/lightly_studio/resolvers/collection_resolver/get_annotation_types_by_collection_ids.py
mihnea-necsulescu: I'd either remove this or also alias `AnnotationBaseTable` - it's a bit confusing to have both

the statement is simple enough, I think `SampleTable` can be inlined

mihnea-necsulescu (approved):

PR #1515, opened 2026-06-29
script to synthesize gt/pred annotation sources for object detection, classification, and semantic segmentation

ikondrat:
## What has changed and why?

Part of LIG-9843. Adds a developer/QA indexing script,
`lightly_studio/e2e-tests/index_evaluations_sources.py`, to make it easy to
exercise the GUI evaluation trigger across all supported task types.

The script creates a single dataset and synthesizes `gt`/`pred` annotation
sources for object detection, classification, and semantic segmentation, then
launches the UI so you can open the Evaluation panel and start runs for any task
type. Annotations are generated synthetically (boxes and segmentation masks
sized to each image), so the only required input is an images folder — set via
`EXAMPLES_IMAGES_PATH` (defaults to the COCO subset used by `index_evaluations.py`).
GT and prediction geometries are deliberately offset so they partially overlap,
and predictions carry random confidences, giving non-trivial metrics.

By default it resets the local DuckDB (`EXAMPLES_CLEANUP_DB=true`) so re-runs work,
since DuckDB has no single-dataset delete; set `EXAMPLES_CLEANUP_DB=false` to keep
existing data (the dataset name must then not already exist).

## How has it been tested?

- Ran the script locally against the default COCO subset, confirmed the dataset
  and all six sources are created, and verified each evaluation type can be
  triggered from the GUI panel.

`uv run e2e-tests/index_evaluations_sources.py`

- [ ] Yes

## Summary by CodeRabbit

* **New Features**
  * Added a new evaluation setup flow for image collections, including choosing task type, sources, and run settings.
  * Evaluation panels can now appear even before any runs exist, making it easier to start the first one.
  * Annotation collections now show which evaluation types they support, improving source selection.

* **Bug Fixes**
  * Improved evaluation panel labels and visibility rules for clearer, more consistent navigation.
  * Added broader test coverage for evaluation dialogs, source filtering, and panel behavior.


lightly_studio/e2e-tests/index_evaluations_sources.py
```diff
@@ -0,0 +1,217 @@
...
+    """Return a binary mask (height x width) with a filled rectangle."""
+    x, y, w, h = _box(width, height, left_fraction, top_fraction)
+    mask = np.zeros((height, width), dtype=np.int_)
+    mask[y : y + h, x : x + w] = 1
+    return mask
+
+
+def main() -> None:
+    """Create the dataset, synthesize sources, and launch the UI."""
+    db_manager.connect(db_file="lightly_studio.db", cleanup_existing=CLEANUP_DB)
```
horatiualmasan: Do we need to run it with cleanup_existing=False or provide the database? 
We could simply  do `db_manager.connect(cleanup_existing=True)`

horatiualmasan (approved):
LGTM

PR #1516, opened 2026-06-29
Add YOLO Export Format

MalteEbner:
## What has changed and why?

Add python functions to export in YOLO format.
Add GUI dropdown option of exporting in YOLO format as zip.

Implementation copies the one for COCO if applicable. For exporting a directory instead of single file, copies from PascalVoc instead, which also exports a directory. Note that the exported directory/zip does NOT include the original images, same as for COCO and PascalVOC

Also renamed many "export object detection" variable names to "coco export" variable names, so that using "yolo export" variable names fits better.

image
image

## How has it been tested?

New unittests and e2e tests
Manually: Open the GUI, download the zip, inspect it. 

- [x] Yes

## Summary by CodeRabbit

* **New Features**
  * Added YOLO object detection export alongside existing COCO export options.
  * Export dialogs now let you choose a specific annotation source when multiple collections are available.
  * Downloaded YOLO exports now include the expected ZIP contents and filenames.
* **Bug Fixes**
  * Annotation exports now respect the selected annotation collection across object detection and segmentation formats.
  * Export links and filenames now match the selected format more consistently.

### DIFF
--- a/lightly_studio/src/lightly_studio/api/routes/api/export.py
+++ b/lightly_studio/src/lightly_studio/api/routes/api/export.py
@@ -8,6 +8,7 @@ from datetime import datetime, timezone
 from pathlib import Path as PathlibPath
 from tempfile import TemporaryDirectory
 from typing import Annotated
+from uuid import UUID
 
 from fastapi import APIRouter, Depends, HTTPException, Path
 from fastapi.responses import PlainTextResponse, StreamingResponse
@@ -37,6 +38,7 @@ def export_collection_annotations(
     ],
     session: SessionDep,
     export_format: ExportFormat = ExportFormat.OBJECT_DETECTION_COCO,
+    annotation_collection_id: UUID | None = None,
 ) -> StreamingResponse:
     """Export collection annotations in the selected export format."""
     # Query to export - all samples in the collection.
@@ -54,6 +56,7 @@ def export_collection_annotations(
                 dataset_id=collection.dataset_id,
                 samples=dataset_query,
                 output_json=output_path,
+                annotation_collection_id=annotation_collection_id,
             )
         except Exception:
             temp_dir.cleanup()
@@ -68,6 +71,7 @@ def export_collection_annotations(
                 dataset_id=collection.dataset_id,
                 samples=dataset_query,
                 output_json=output_path,
+                annotation_collection_id=annotation_collection_id,
             )
         except Exception:
             temp_dir.cleanup()
@@ -82,6 +86,7 @@ def export_collection_annotations(
                 dataset_id=collection.dataset_id,
                 samples=dataset_query,
                 output_folder=output_path,
+                annotation_collection_id=annotation_collection_id,
             )
         except Exception:
             temp_dir.cleanup()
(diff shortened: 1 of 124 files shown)


lightly_studio/src/lightly_studio/api/routes/api/export.py
```diff
@@ -62,6 +62,35 @@ def export_collection_annotations(
...
+                output_folder=output_path,
+                annotation_collection_id=annotation_collection_id,
+            )
+        except Exception:
+            temp_dir.cleanup()
+            # Reraise.
+            raise
+
+        # For YOLO export, the exporter produces a directory (data.yaml + labels/),
+        # so this route streams the folder as a .zip instead of streaming a single file.
```
MalteEbner, line 82: This is different to COCO, but the same as for Pascal VOC, see below:

```python
# For Pascal VOC export, the exporter produces a directory,
# so this route should stream the folder as a .zip instead of streaming a single file.
```

lightly_studio/tests/api/routes/api/test_export.py
```diff
@@ -35,7 +36,7 @@
 from tests.resolvers.video.helpers import VideoStub, create_video_with_frames
 
 
-def test_export_collection_annotations(
+def test_export_collection_coco(
```
MalteEbner, line 39: One of the renamings of a generic "annotations" to the more specific "coco"

MalteEbner:
Thanks @chatgpt-codex-connector for the review. Decisions on the two points:

**P1 — make `annotation_collection_id` optional again (route):** Keeping it required, intentionally. This parameter and its required-ness come from the annotation-source-selection work this branch builds on, not from the YOLO change — it only appears in this diff because that branch is merged in here. The export dialog always sends an `annotation_collection_id` (it defaults to the first available source), so the 422-on-missing path is not reachable from the app. Leaving it consistent with that PR's design rather than diverging here.

**P2 — YOLO label filename uniqueness:** Keeping `sample.file_name`, intentionally. The YOLO exporter deliberately mirrors the existing Pascal VOC exporter, which derives its per-file output names from `sample.file_name` the same way. Same-basename collisions across directories are a pre-existing limitation shared by all file-per-image exporters (ingestion stores only `Path(...).name`); changing it for YOLO alone would diverge from Pascal VOC, and a real fix belongs in a separate, broader change.

The three documentation/changelog/style points (export docs, changelog entry, `import uuid`) are addressed in 82683c9e.

lightly_studio/src/lightly_studio/export/lightly_studio_label_input.py
```diff
@@ -88,6 +88,33 @@ def get_labels(self) -> Iterable[ImageObjectDetection]:
...
 
 
+class LightlyStudioYOLOObjectDetectionInput(LightlyStudioObjectDetectionInput):
+    """Labelformat adapter for YOLO object detection export.
+
+    Uses relative filenames so the YOLO writer places label files inside the output
+    ``labels`` directory. The base adapter uses absolute paths, which are fine for COCO
+    (stored verbatim as strings) but would break YOLO: the writer joins the filename
+    with the output directory to build each label file path, and an absolute path would
+    escape that directory.
```
MalteEbner, line 98: Same as for Pascal VOC

horatiualmasan (approved):
LGTM

PR #1517, opened 2026-06-29
Add stored annotation embedding lookup endpoint

horatiualmasan:
## What has changed and why?

GET /collections/{id}/annotations/{sample_id}/embedding returns the stored embedding vector for an annotation sample. Drag-to-search uses this to query with an annotation's own embedding so self-similarity scores ~1.0 instead of re-embedding a UI crop. 

## How has it been tested?

New tests

- [ ] Yes

## Summary by CodeRabbit

* **New Features**
  * Added a new endpoint to retrieve an annotation’s stored embedding vector.
  * Successfully returning an annotation embedding now provides the vector as a list of floating-point values.

* **Bug Fixes**
  * Requests for annotations without a stored embedding now return a clear 404 response.
  * Requests for collections without available embedding data also return 404.

* **Tests**
  * Added coverage for successful embedding retrieval and missing-embedding handling.


lightly_studio/src/lightly_studio/api/routes/api/annotation.py
```diff
@@ -160,6 +166,47 @@ def read_annotations_with_payload(
...
+    """Return the stored embedding vector for a single annotation sample.
+
+    Used for drag-to-search self-similarity: searching with an annotation's own
+    stored embedding.
+    """
+    embedding_model_id = session.exec(
+        select(EmbeddingModelTable.embedding_model_id)
+        .where(EmbeddingModelTable.collection_id == collection_id)
+        .limit(1)
+    ).first()
```
JonasWurst: I think this should be in a resolver

lightly_studio/src/lightly_studio/api/routes/api/annotation.py
```diff
@@ -160,6 +166,47 @@ def read_annotations_with_payload(
...
 
+@annotations_router.get(
+    "/annotations/{sample_id}/embedding",
+    response_model=list[float],
+)
+def read_annotation_embedding(
+    collection_id: Annotated[
+        UUID, Path(title="collection Id", description="The ID of the annotation collection")
+    ],
+    sample_id: Annotated[UUID, Path(title="sample Id", description="The annotation sample ID")],
```
JonasWurst, line 176: Shouldn't this be multiple?

lightly_studio/src/lightly_studio/api/routes/api/annotation.py
```diff
@@ -160,6 +166,47 @@ def read_annotations_with_payload(
...
+        sample_ids=[sample_id],
+        embedding_model_id=embedding_model_id,
+    )
+    if not embeddings:
+        raise HTTPException(
+            status_code=HTTP_STATUS_NOT_FOUND,
+            detail="No stored embedding found for this annotation.",
+        )
+
+    return [float(value) for value in embeddings[0].embedding]
```
JonasWurst, line 205: are we returning the actual embeddings (not the 2d) here? what is the purpose

JonasWurst (changes requested):

lightly_studio/src/lightly_studio/api/routes/api/annotation.py
horatiualmasan: used existing resolver function for it

lightly_studio/src/lightly_studio/api/routes/api/annotation.py
horatiualmasan: It will be used when searching by image in the annotation grid using drag&drop. It will only use one annotation item.

lightly_studio/src/lightly_studio/api/routes/api/annotation.py
horatiualmasan: The embedding will be used when searching using drag&drop of a annotation sample in the grid. It is used instead of recalculating it from the crop to avoid mismatches in the size and get an exact comparison.

lightly_studio/src/lightly_studio/api/routes/api/annotation.py
JonasWurst: so we get an embedding vector and pass it back in to another endpoint? Would it nicer to just have a new endpoint doing both?

JonasWurst: clarified

JonasWurst (approved):

PR #1544, opened 2026-07-02
Add limit support to fsspec file listing

IgorSusmelj:
## What has changed and why?

 - Adds limit support to the core file-listing logic
 - This is the core building block for a follow-up PR that exposes limit on the ImageDataset/VideoDataset import methods

## How has it been tested?

- Unit tests

- [ ] Yes

## Summary by CodeRabbit

* **New Features**
  * Added an optional `limit` parameter to file listing, stopping after discovering up to N files (in discovery order) and returning only those results.
  * Added validation so `limit` must be positive when provided, otherwise an error is raised.

* **Tests**
  * Added coverage for `limit` behavior (capped results, full results with large limits) and for limit validation, including error messages for invalid values.

### DIFF
--- a/lightly_studio/tests/dataset/test_fsspec_lister.py
+++ b/lightly_studio/tests/dataset/test_fsspec_lister.py
@@ -224,6 +224,19 @@ class TestFsspecLister:
         result = list(fsspec_lister.iter_files_from_path("s3://test/images/empty_dir"))
         assert len(result) == 0
 
+    def test_get_file_list_from_s3__limit(self, s3_bucket: str, mock_fsspec_s3fs: None) -> None:  # noqa: ARG002
+        # The directory holds 5 files; limit caps how many are yielded.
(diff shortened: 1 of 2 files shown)


lightly_studio/tests/dataset/test_fsspec_lister.py
```diff
@@ -235,3 +248,15 @@ def test_get_file_list_from_s3__single_file_read(
         with fsspec.open(result[0], "rb") as file:
             content = file.read()
             assert content == b"fake_bmp_content"
+
+
+class TestValidateLimit:
```
michal-lightly: Style: Do not wrap in `TestValidateLimit` class. You can point the AI at the python guidelines in our repo to resolve.

michal-lightly: Consider this optional. I see that the tests in this file are written using a wrong convention already for TestFsspecLister. There should be no class.

michal-lightly (approved):

PR #1545, opened 2026-07-02
Add: New argument to limit how many samples to index

IgorSusmelj:
## What has changed and why?

This PR adds a new argument limit that allows to set a `limit` of the number of samples to be used for indexing. This is helpful for debugging or testing on smaller subsets before indexing large datasets.

## How has it been tested?

- Manual testing
- Updated unit tests

- [X] Yes

## Summary by CodeRabbit

* **New Features**
  * Added an optional `limit` setting when importing images and videos, so you can cap how many items are loaded at once.
  * Extended this limit support across multiple dataset import paths, including label formats, COCO, YOLO, Pascal VOC, Lightly, and video sources.
* **Bug Fixes**
  * Improved handling when the limit excludes some media items, preventing related annotations from causing import failures.
  * Added validation so invalid limit values now return a clear error.

### DIFF
--- a/lightly_studio/src/lightly_studio/core/image/add_images.py
+++ b/lightly_studio/src/lightly_studio/core/image/add_images.py
@@ -149,8 +149,7 @@ def load_into_dataset_from_labelformat(  # noqa: PLR0913
         input_labels: The labelformat input containing images and annotations.
         images_path: The path to the directory containing the images.
         collection_name: Optional name for the annotation collection.
-        limit: If set, load at most this many samples (the first ``limit`` in the order
-            the input yields them). If None, all samples are loaded.
+        limit: Maximum number of samples to load. By default, all samples are loaded.
 
     Returns:
         A list of UUIDs of the created samples.
@@ -248,8 +247,7 @@ def load_into_dataset_from_coco_captions(
         root_collection_id: Identifier of the root collection that receives the samples.
         annotations_json: Path to the COCO captions annotations file.
         images_path: Directory containing the referenced images.
-        limit: If set, load at most this many samples (the first ``limit`` images listed
-            in the annotations file). If None, all samples are loaded.
+        limit: Maximum number of samples to load. By default, all samples are loaded.
 
     Returns:
         The list of newly created sample identifiers.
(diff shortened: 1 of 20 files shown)


lightly_studio/src/lightly_studio/core/image/add_images.py
LeonardoRosaa, line 263: Minor: It is still iterating over all annotations without considering the images restricted by the `limit` argument. This might be an issue if the limit is low and the dataset is large, with many annotations.

lightly_studio/src/lightly_studio/core/image/add_images.py
```diff
@@ -147,6 +149,8 @@ def load_into_dataset_from_labelformat(
         input_labels: The labelformat input containing images and annotations.
         images_path: The path to the directory containing the images.
         collection_name: Optional name for the annotation collection.
+        limit: If set, load at most this many samples (the first ``limit`` in the order
```
LeonardoRosaa: Suggestion: maximum number of samples to load per dataset. By default, all samples are loaded.

lightly_studio/src/lightly_studio/core/image/image_dataset.py
```diff
@@ -144,18 +145,23 @@ def add_images_from_path(
                 - `tag_depth=0` (default): No automatic tagging is performed.
                 - `tag_depth=1`: Automatically creates a tag for each
                   image based on its parent directory's name.
+            limit: If set, load at most this many images (the first ``limit`` in discovery
```
LeonardoRosaa: Could we keep the same docstring throughout the API?

lightly_studio/tests/core/image/test_image_dataset__coco.py
```diff
@@ -130,6 +130,26 @@ def test_add_samples_from_coco__valid_insseg(
         )
         assert len(list(dataset)) == 2
 
+    def test_add_samples_from_coco__limit(
```
LeonardoRosaa, line 133: Should we also test with negative `limit` argument?

lightly_studio/src/lightly_studio/core/image/add_images.py
IgorSusmelj: We can't change that. The annotations are not ordered by image. So to make sure we capture all we need to scan through. But this should not be slow.

LeonardoRosaa (approved):

PR #1551, opened 2026-07-02
Fix: Improve UX for embedding plot and its legend

IgorSusmelj:
## What has changed and why?

- Makes scatter plot legend fixed height to be less space consuming
- Hide warning if WebGPU is not enabled

Before this change:
image

After this change:

https://github.com/user-attachments/assets/838995c7-14ab-466c-a110-21d718f99f03

## How has it been tested?

Manual testing

- [X] Yes

## Summary by CodeRabbit

* **New Features**
  * Improved the embedding plot legend to be more compact and clearly scrollable, with fade cues that indicate when additional items are available.
* **Bug Fixes**
  * Removed the WebGPU fallback notice from the plot area to reduce unnecessary on-screen messaging.
* **Tests**
  * Added coverage to verify the legend’s edge-fade cue behavior across multiple scroll positions.


lightly_studio_view/src/lib/components/PlotPanel/PlotPanel.svelte
```diff
@@ -451,4 +451,11 @@
         height: 25px !important;
         line-height: 25px !important;
     }
+    /* Hide the library's status message slot (e.g. "WebGPU is unavailable. Falling back
+       to WebGL.") while keeping the selection tools, scale, and point count visible. */
+    :global(
```
LeonardoRosaa, line 456: This is quite dangerous

lightly_studio_view/src/lib/components/PlotPanel/PlotPanelLegend.svelte
```diff
@@ -36,6 +36,30 @@
         onDoubleClickCategory
     }: Props = $props();
 
+    // Fade the clipped edge(s) of the list so users see the legend is scrollable.
+    let scrollContainer: HTMLDivElement | null = $state(null);
```
LeonardoRosaa, line 40: I'd write unit tests

LeonardoRosaa (approved):

LeonardoRosaa:
Rename the PR title

lightly_studio_view/src/lib/components/PlotPanel/PlotPanel.svelte
IgorSusmelj: Do you remember how we did it for the other css rules we wanted to change? (FYI, this only affects the rendered text and not any console logging). Are you concerned that we also overwrite other rules?

LeonardoRosaa: We did something similar and overwrote the other rules. See: https://github.com/lightly-ai/lightly-studio/blob/7281ae341f4b9cc2658a82126511f9c6f4fe6f97/lightly_studio_view/src/lib/components/PlotPanel/PlotPanel.svelte#L441

These rules only change sizes, which is not very risky. I’m okay with your implementation, and I don’t think we have a better solution for now.

PR #1558, opened 2026-07-03
Image folder ingest: record missing/broken outcomes instead of silent skip

JonasWurst:
## What

Wires the image **folder** ingest path (`load_into_dataset_from_paths`) into the per-run `FileOutcomeReport` so missing/broken files are recorded instead of silently dropped.

## Why

The path was already routed through `report.track(...)`, but the loop still `continue`d on a failed header read. Because `continue` exits the tracked block *cleanly*, the report counted missing/broken files as **added** — the silent-skip bug this slice fixes.

## How

At the I/O boundary inside `report.track(path)`:
- **missing** — detected proactively via `fs.exists()` → `MissingInputFileError` (`FileNotFoundError` is unreliable across fsspec backends and is a subclass of `OSError`);
- **broken** — failed header read (`PIL.Image.open` + `.size`) → `BrokenInputFileError(...) from e`;
- **already-present** — unchanged, from the existing path-dedup;
- **any other exception propagates** — a bug or infra failure is never recorded as "broken".

The run closes with `report.log_summary()` then `report.raise_if_all_failed()`, so a folder where every attempted file fails raises loudly instead of returning silently.

## Behaviour change to call out

`load_into_dataset_from_paths` (and therefore `add_images_from_path` / `CreateImage.create_in_collection`) now raises `AllInputFilesFailedError` when **every** attempted file is missing/broken, where it previously returned an empty list. Empty folders (0 attempted / all already-present) still do **not** raise. `Raises:` docstrings updated accordingly.

## Tests

- New test: a folder mixing good / missing / broken / already-present asserts one outcome per file (`added=1, already_present=1, missing=1, broken=1`) and that only the good file is added.
- Updated the corrupt-file test to expect the new loud `AllInputFilesFailedError`.
- `make static-checks` clean on changed files; affected image/group test modules pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved bulk image imports by differentiating missing input paths from broken/unreadable images with dedicated error cases.
  * Bulk loading now raises an error when all discovered images are missing or broken, rather than returning results without adding usable images.
* **Tests**
  * Added coverage for mixed imports (valid, missing, broken, already present) and verified per-outcome logging summary.
  * Updated corrupted-file test to expect failure when no valid images are available.
* **Documentation**
  * Updated API documentation to include the additional “all inputs failed” error scenarios.


lightly_studio/src/lightly_studio/core/image/add_images.py
```diff
@@ -89,13 +91,21 @@ def load_into_dataset_from_paths(
             if normalized_path in seen_or_existing_paths:
                 raise AlreadyPresentInputFileError()
 
+            # Detect a missing path proactively: FileNotFoundError is unreliable across
+            # fsspec backends and is a subclass of OSError, which we treat as broken.
+            fs, fs_path = fsspec.core.url_to_fs(normalized_path)
```
JonasWurst, line 100: we could discuss adding this also for `load_into_dataset_from_labelformat, ``load_into_dataset_from_coco_captions`

JonasWurst: -> will be done in a follow up

lightly_studio/tests/core/image/test_add_images.py
```diff
@@ -88,6 +88,55 @@ def test_load_into_collection_from_paths(db_session: Session, tmp_path: Path) ->
...
+    samples = image_resolver.get_all_by_collection_id(
+        session=db_session, collection_id=collection.collection_id
+    ).samples
+    assert {sample.file_name for sample in samples} == {"already_present.jpg", "good.jpg"}
+
+    # Assert: the end-of-run summary records one outcome per file.
+    assert "added=1" in caplog.text
+    assert "already_present=1" in caplog.text
+    assert "missing=1" in caplog.text
+    assert "broken=1" in caplog.text
```
MalteEbner: This assertion would not catch confusing two outcomes. Would rather have different numbers each

lightly_studio/src/lightly_studio/core/image/add_images.py
MalteEbner: It seems to me that there is a lot of duplicate code among `load_into_dataset_from_paths`, `load_into_dataset_from_labelformat` and `load_into_dataset_from_coco_captions`.
Could we refactor this? E.g. the two latter ones calling the first one and then adding annotations later?

MalteEbner (approved):

PR #1559, opened 2026-07-03
Add hover-point and thumbnail-URL resolution for the embedding plot (LIG-10089, 1/3)

IgorSusmelj:
Part 1/3 of the embedding-plot hover preview ([LIG-10089](https://linear.app/lightly/issue/LIG-10089)). Stack: this PR → #1560 (preview card) → #1561 (PlotPanel wiring).

## What

Pure logic, no UI change (exports are consumed in parts 2 and 3):

- `createQuerySelection` — nearest-visible-point hit test over the embedding arrays (10px hover radius, skips hidden-category points), returning the sample id as the tooltip `identifier`. The array-based `EmbeddingView` emits **no hover tooltips unless this callback is provided** — that's why it exists.
- `createThumbnailUrlResolver` — maps a hovered sample id to a thumbnail URL per route: images directly, videos via their poster frame (`getVideoById`), annotations via the parent image (`getAnnotation`); per-sample promise cache so repeat hovers don't refetch.

## Test plan

- 6 unit tests cover nearest-point selection, hover radius, hidden points, all three route resolvers, lookup caching, and API-failure fallback.
- `make static-checks` and the PlotPanel test suite pass.

## Summary by CodeRabbit

* **New Features**
  * Added nearest-point selection for plot hover interactions.
  * Added thumbnail previews for image, video, and annotation samples.
  * Added caching for thumbnail lookups, with retries after failed requests.
  * Excluded hidden and filtered-out points from hover selection.

* **Bug Fixes**
  * Improved handling when hover data or selectable points are unavailable.


lightly_studio_view/src/lib/components/PlotPanel/PlotHoverPreview/tooltipThumbnail.ts
```diff
@@ -0,0 +1,142 @@
+import { getAnnotation, getVideoById } from '$lib/api/lightly_studio_local';
```
ikondrat: I would rather split this file to smaller files and add barrel export 

It would be easier to maintain and test wise as well (test for dedicated funcitons would be in dedicated files)

lightly_studio_view/src/lib/components/PlotPanel/PlotHoverPreview/tooltipThumbnail.test.ts
```diff
@@ -0,0 +1,87 @@
...
+
+describe('createQuerySelection', () => {
+    const params = {
+        x: new Float32Array([0, 10]),
+        y: new Float32Array([0, 10]),
+        sampleIds: ['sample-a', 'sample-b'],
+        category: new Uint8Array([INCLUDED_BY_FILTERS_CATEGORY, INCLUDED_BY_FILTERS_CATEGORY])
+    };
+
+    test('returns the nearest point with its sample ID as identifier', async () => {
```
ikondrat: We rather is "it" than test
```suggestion
    it('returns the nearest point with its sample ID as identifier', async () => {
```

I would try to be consistent

ikondrat (approved):

PR #1574, opened 2026-07-06
Add batched embedding helper for in-memory PIL images.

horatiualmasan:
## What has changed and why?

Introduce `embed_pil_images_batched` for embedding decoded frames and other in-memory PIL images .

## How has it been tested?

new tests

- [ ] Yes

## Summary by CodeRabbit

* **New Features**
  * Added support for embedding PIL images directly from memory.
  * Introduced batched PIL embedding with optional progress display while preserving input order.
* **Tests**
  * Added tests to verify empty-input behavior returns an appropriately shaped empty array.
  * Added tests to confirm embeddings are returned in the same order as the input images.

### DIFF
--- a/lightly_studio/src/lightly_studio/dataset/image_embedding.py
+++ b/lightly_studio/src/lightly_studio/dataset/image_embedding.py
@@ -95,44 +95,13 @@ def embed_image_files_batched(
     Returns:
         Float32 array of shape ``(len(filepaths), embedding_dimension)``.
     """
-    total_images = len(filepaths)
-    if not total_images:
-        return np.empty((0, context.embedding_dimension), dtype=np.float32)
-
-    if context.max_batch_size <= 0:
-        raise ValueError("max_batch_size must be positive.")
-
-    dataset = _ImageFileDataset(filepaths, context.preprocess)
-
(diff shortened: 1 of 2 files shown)


lightly_studio/src/lightly_studio/dataset/image_embedding.py
```diff
@@ -77,15 +103,53 @@ def embed_image_files_batched(
...
+
+    Returns:
+        Float32 array of shape ``(len(images), embedding_dimension)``.
+    """
+    return _embed_dataset_batched(
+        _PILImageDataset(images, context.preprocess),
+        len(images),
+        context,
+        show_progress,
+        _EmbeddingProgress(desc="Generating frame embeddings", unit=" frames"),
```
lukas-lightly, line 135: Is this only used for video frames? I'm asking because of the `frame` word, if it's used also for non video frame images, it might be confusing.

lukas-lightly (approved):

lightly_studio/src/lightly_studio/dataset/image_embedding.py
horatiualmasan: Currently it will only be used for video frames, I would leave it as is for now and update it if we use it also for different images

PR #1575, opened 2026-07-06
Embed video frames during extraction when embed_frames is enabled.

horatiualmasan:
## What has changed and why?

Add embedding manager support for VIDEO_FRAME collections and wire frame embedding into the video ingest path via add_videos.

## How has it been tested?

New tests

- [ ] Yes

## Summary by CodeRabbit

* **New Features**
  * Video frame import and annotation-based loading can now optionally generate and persist embeddings while decoding frames (with configurable embedding model selection).
  * Added in-memory PIL-image embedding support across embedding generators and manager, enabling embed-and-store workflows for frame images.
* **Bug Fixes**
  * If embedding is enabled but a compatible embedding model isn’t available, embedding is skipped gracefully and a warning is logged.
  * Updated embedding routing so video-frame embeddings use the image embedding path.
* **Tests**
  * Added coverage for embedding-enabled frame extraction to verify embeddings are created for each frame sample.


lightly_studio/src/lightly_studio/core/video/add_videos.py
JonasWurst: this might be too large now considering we also perform embedding

lightly_studio/src/lightly_studio/dataset/embedding_generator.py
```diff
@@ -94,6 +95,21 @@ def embed_image_crops(
         """
         ...
 
+    def embed_pil_images(
```
JonasWurst, line 98: i think it is quite convoluted now. Especially if you imagine users that need to implement this interface. we can go ahead for now, but should consider how to improve this. I.e. i would expect only one embed_images, not two.

lightly_studio/tests/dataset/test_embedding_manager.py
JonasWurst, line 1: should we also test mobileclip and perception encoder?

horatiualmasan: added tests

lightly_studio/src/lightly_studio/core/video/add_videos.py
horatiualmasan: lowered to 32

lightly_studio/src/lightly_studio/dataset/embedding_generator.py
horatiualmasan: maybe we can address this after we refactor to move the embedding at index time, we could pass directly the in memory image and won't need the variant with file paths

JonasWurst: sgtm make sure to create an issue

JonasWurst (approved):

PR #1584, opened 2026-07-06
Add embedding-region model and point-in-polygon core

ikondrat:
### Why

When a user lassos points in the embedding plot, the frontend currently sends the
**full list of selected sample ids** in the request body. That list grows with the
selection and doesn't scale — a large lasso produces a huge request.

The fix is to send the **selection geometry** (a handful of polygon
vertices, a few KB) instead, and reproduce the exact selection server-side by running
point-in-polygon over the cached 2D projection.

This PR lands the **shared foundation** used by both the images and annotations
filtering that follow in the stack. It adds no wiring into any resolver yet, so it's
inert on its own.

### What

- **`models/embedding_region.py`** — `Point2D` and `EmbeddingRegion` Pydantic models.
  Rectangle selections are normalized to four corner vertices on the frontend, so both
  lasso and rectangle arrive as a polygon. Validates ≥ 3 vertices.
- **`resolvers/embedding_region_resolver.py`** — the geometry core:
  - `get_sample_ids_in_region(...)` — loads the collection's cached 2D embedding and
    returns the sample ids inside the region.
  - `_points_in_polygon(...)` — vectorized even-odd ray-casting test, deliberately
    mirroring the frontend's `isPointInPolygon` so the server selects exactly the points
    the user saw highlighted.
- Tests covering the polygon math (convex + concave) and `get_sample_ids_in_region`
  (inside points, empty region, missing embedding model).

### Notes for reviewers

- No public route or resolver calls this code yet — the image/annotation resolve
  functions and filter wiring come in the next PRs in the stack (this file gains two more
  functions there).
- The embedding-model selection reuses the "first embedding model" convention from the
  `embeddings2d` endpoint; a `TODO(Kondrat, 07/2026)` marks making it an API parameter.

## Summary

* **New Features**
  * Added region-based sample selection on embedding plots using polygon geometry, returning the samples that fall inside the drawn region.

* **Bug Fixes**
  * Enforced minimum region polygon size (at least 3 vertices) for valid selections.
  * Improved empty-state behavior by returning no results when an embedding model or 2D projections are unavailable.

* **Tests**
  * Added automated coverage for point-in-polygon behavior (convex/concave) and region-to-sample selection edge cases.

* **Other**
  * Updated the embeddings 2D endpoint to use centralized default embedding model selection.

### DIFF
--- a/lightly_studio/src/lightly_studio/resolvers/embedding_region_resolver.py
+++ b/lightly_studio/src/lightly_studio/resolvers/embedding_region_resolver.py
@@ -13,11 +13,10 @@ from uuid import UUID
 
 import numpy as np
 from numpy.typing import NDArray
-from sqlmodel import Session, select
+from sqlmodel import Session
 
-from lightly_studio.models.embedding_model import EmbeddingModelTable
 from lightly_studio.models.embedding_region import EmbeddingRegion
-from lightly_studio.resolvers import twodim_embedding_resolver
+from lightly_studio.resolvers import embedding_model_resolver, twodim_embedding_resolver
 
 
 def get_sample_ids_in_region(
@@ -26,14 +25,14 @@ def get_sample_ids_in_region(
     region: EmbeddingRegion,
 ) -> list[UUID]:
     """Return the sample ids whose cached 2D coordinates fall inside ``region``."""
-    # Mirror the embeddings2d endpoint: use the collection's first embedding model.
+    # Resolve against the same deterministic default model as embeddings2d.get_2d_embeddings,
+    # so the region is tested against the exact projection the user lassoed over.
     # TODO(Kondrat, 07/2026): Select the embedding model via API parameter once supported,
     # matching embeddings2d.get_2d_embeddings.
-    embedding_model = session.exec(
-        select(EmbeddingModelTable)
-        .where(EmbeddingModelTable.collection_id == collection_id)
-        .limit(1)
-    ).first()
+    embedding_model = embedding_model_resolver.get_default_by_collection_id(
+        session=session,
+        collection_id=collection_id,
+    )
     if embedding_model is None:
         return []
 
--- a/lightly_studio/src/lightly_studio/resolvers/embedding_region_resolver.py
+++ b/lightly_studio/src/lightly_studio/resolvers/embedding_region_resolver.py
@@ -2,7 +2,7 @@
 
 The frontend sends the lasso/rectangle geometry (a handful of vertices) instead of the full
 list of selected sample ids, keeping the request body constant-size regardless of selection
-size (see LIG-9903). Here we reproduce the exact client-side selection server-side: load the
+size. Here we reproduce the exact client-side selection server-side: load the
 cached, deterministic 2D projection the user lassoed over and run a vectorized point-in-polygon
 test, then feed the resulting ids into the existing ``= ANY(...)`` filter path.
 """
--- a/lightly_studio/src/lightly_studio/resolvers/embedding_region_resolver.py
+++ b/lightly_studio/src/lightly_studio/resolvers/embedding_region_resolver.py
@@ -1,10 +1,9 @@
-"""Resolve an embedding-plot region selection to the sample ids it encloses.
+"""Resolve an embedding-plot region (a polygon of 2D vertices) to the sample ids it encloses.
 
-The frontend sends the lasso/rectangle geometry (a handful of vertices) instead of the full
-list of selected sample ids, keeping the request body constant-size regardless of selection
-size. Here we reproduce the exact client-side selection server-side: load the
-cached, deterministic 2D projection the user lassoed over and run a vectorized point-in-polygon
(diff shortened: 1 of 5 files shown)


lightly_studio/src/lightly_studio/resolvers/embedding_region_resolver.py
```diff
@@ -0,0 +1,85 @@
+"""Resolve an embedding-plot region selection to the sample ids it encloses.
+
+The frontend sends the lasso/rectangle geometry (a handful of vertices) instead of the full
+list of selected sample ids, keeping the request body constant-size regardless of selection
+size. Here we reproduce the exact client-side selection server-side: load the
+cached, deterministic 2D projection the user lassoed over and run a vectorized point-in-polygon
+test, then feed the resulting ids into the existing ``= ANY(...)`` filter path.
```
MalteEbner: This describes functionality outside of this file, which can become untrue. Rather tell its purpose independently. If the logic mirrors some frontend logic, best add a GitHub link to the exact LOC in the frontend.

lightly_studio/src/lightly_studio/resolvers/embedding_region_resolver.py
```diff
@@ -0,0 +1,85 @@
...
+
+
+def _points_in_polygon(
+    x: NDArray[np.float32],
+    y: NDArray[np.float32],
+    region: EmbeddingRegion,
+) -> NDArray[np.bool_]:
+    """Vectorized even-odd ray-casting point-in-polygon test.
+
+    Matches the frontend's ``isPointInPolygon`` (ray cast to the right, edges counted when
```
MalteEbner: Add a GitHub link to the  `isPointInPolygon` function

lightly_studio/src/lightly_studio/resolvers/embedding_region_resolver.py
```diff
@@ -0,0 +1,85 @@
...
+    exactly one endpoint is strictly above the point's y) so the server selects exactly the
+    points the user saw highlighted.
+    """
+    vertices_x = np.asarray([vertex.x for vertex in region.polygon], dtype=np.float64)
+    vertices_y = np.asarray([vertex.y for vertex in region.polygon], dtype=np.float64)
+
+    px = np.asarray(x, dtype=np.float64)
+    py = np.asarray(y, dtype=np.float64)
```
MalteEbner: I am quite sure that float32 is enough and faster

lightly_studio/tests/resolvers/test_embedding_region_resolver.py
```diff
@@ -0,0 +1,154 @@
...
+from sqlmodel import Session
+
+from lightly_studio.models.embedding_region import EmbeddingRegion, Point2D
+from lightly_studio.models.two_dim_embedding import TwoDimEmbeddingTable
+from lightly_studio.resolvers import embedding_region_resolver, sample_embedding_resolver
+from tests import helpers_resolvers
+from tests.helpers_resolvers import ImageStub
+
+
+def _square(x_min: float, y_min: float, x_max: float, y_max: float) -> EmbeddingRegion:
```
MalteEbner, line 89: Helper functions belong to the bottom, see our code style guidelines.

Furthermore, the naming in this file is wrong in general, see https://app.notion.com/p/getlightly/Python-testing-78e2dc49baf340178fa44ce20031c52e?source=copy_link#58c714f100b8413a8e9eb8b93fc1f62d

lightly_studio/tests/resolvers/test_embedding_region_resolver.py
```diff
@@ -0,0 +1,154 @@
...
+            coordinates=[(1.0, 1.0), (5.0, 5.0), (100.0, 100.0)],
+        )
+
+        selected = embedding_region_resolver.get_sample_ids_in_region(
+            session=db_session,
+            collection_id=collection_id,
+            region=_square(x_min=0, y_min=0, x_max=10, y_max=10),
+        )
+
+        assert set(selected) == {sample_ids[0], sample_ids[1]}
```
MalteEbner: From  Codex, makes sense to me:

[P2] lightly_studio/tests/resolvers/test_embedding_region_resolver.py:129 has a flaky expectation. _setup_collection_with_coordinates() assigns coordinates in creation order, but returns ordered_sample_ids from get_hash_by_collection_id(), which orders by sample_id UUID. The assertion assumes the first two returned IDs are the two inside points, which only holds when the outside sample happens to sort last by UUID. Return the expected inside IDs explicitly from the helper, or build the coordinate mapping in the same ordered ID space used by the assertion.

MalteEbner (commented):
The approach itself LGTM.

Some details should be improved to get this into a good shape.

lightly_studio/tests/resolvers/test_embedding_region_resolver.py
ikondrat: fixed

MalteEbner (approved):

PR #1585, opened 2026-07-06
Filter images by embedding-plot region selection

ikondrat:
## Filter images by embedding-plot region selection (LIG-9903)

Stacked on #1584 (shared region model + point-in-polygon resolver).

### Why
When you lasso/rectangle-select points in the embedding plot, the frontend currently POSTs
the full list of selected `sample_id` UUIDs. The body grows linearly with the selection
(~47k ids ≈ 1.9 MB → HTTP 413), so it can't scale toward the 10M-sample target. The fix is
to send the selection **geometry** and resolve it to ids server-side. This PR wires that
resolution into the image query path. (The frontend switch to sending geometry is a later PR
in the stack.)

### What
- `SampleFilter` gains an optional `embedding_region` field and applies it as a
  `sample_id = ANY(...)` predicate — the same path the raw id list already used, so the SQL
  side is unchanged.
- Adds `resolve_embedding_region(...)`, an image-filter wrapper over the shared
  `get_sample_ids_in_region` from #1584: it loads the cached 2D projection, runs the
  point-in-polygon test, and stores the resulting ids on the filter before the query runs
  (the query builder has no DB session, so resolution happens up front).
- Wires resolution into the three image entry points that share `ImageFilter`:
  `get_all_by_collection_id` (`/images/list`), `get_sample_ids`
  (`/images/sample_ids` + the `/embeddings2d` fulfils-filter mask), and
  `count_image_annotations_by_collection` (`/images/annotations/count`). One field change
  covers all three.

### Behavior notes
- An empty region (nothing enclosed) matches **no** rows (`WHERE false`), not everything.
- Applying an unresolved region raises — a guard against forgetting to resolve before querying.

### Validation
- Unit tests for the image wrapper (`test_embedding_region_resolver.py`) and for
  `SampleFilter` applying the region, including the empty-region and unresolved-region cases
  (`test_sample_filter.py`).

### Scope
- Backend only; no endpoint request-shape change is exercised until the frontend PR sends
  `embedding_region`. Reviewable on its own via the new tests.

## Summary by CodeRabbit

* **New Features**
  * Added support for filtering images and samples by a selected embedding region, so results now reflect the area chosen in the embedding view.
  * Region-based filters now work together with existing sample filters for more precise results.

* **Bug Fixes**
  * Improved filtering behavior so empty regions return no matches, and region selection is consistently applied across image and sample queries.

* **Tests**
  * Expanded automated coverage for region-based filtering and combined filter scenarios.

### DIFF
--- a/lightly_studio/src/lightly_studio/resolvers/image_resolver/count_image_annotations_by_collection.py
+++ b/lightly_studio/src/lightly_studio/resolvers/image_resolver/count_image_annotations_by_collection.py
@@ -10,6 +10,7 @@ from lightly_studio.models.annotation.annotation_base import AnnotationBaseTable
 from lightly_studio.models.annotation_label import AnnotationLabelTable
 from lightly_studio.models.image import ImageTable
 from lightly_studio.models.sample import SampleTable
+from lightly_studio.resolvers import embedding_region_resolver
 from lightly_studio.resolvers.image_filter import ImageFilter
 
 
@@ -24,6 +25,9 @@ def count_image_annotations_by_collection(
     label name and counted for total and filtered.
     Returns a list of (label_name, current_count, total_count) tuples.
     """
+    embedding_region_resolver.resolve_embedding_region(
+        session=session, collection_id=collection_id, filters=image_filter
+    )
     total_counts = _get_total_counts(session=session, collection_id=collection_id)
     current_counts = _get_current_counts(
         session=session,
--- a/lightly_studio/src/lightly_studio/resolvers/image_resolver/count_image_annotations_by_collection.py
+++ b/lightly_studio/src/lightly_studio/resolvers/image_resolver/count_image_annotations_by_collection.py
@@ -25,15 +25,20 @@ def count_image_annotations_by_collection(
     label name and counted for total and filtered.
     Returns a list of (label_name, current_count, total_count) tuples.
     """
-    region_sample_ids = embedding_region_resolver.get_region_sample_ids(
-        session=session, collection_id=collection_id, filters=image_filter
-    )
+    # Resolve any embedding-plot region selection to concrete sample ids on the filter before the
+    # query is built (the point-in-polygon test needs the session, which `apply` lacks).
+    sample_filter = image_filter.sample_filter if image_filter is not None else None
+    if sample_filter is not None and sample_filter.embedding_region is not None:
+        sample_filter.region_sample_ids = embedding_region_resolver.get_sample_ids_in_region(
+            session=session,
+            collection_id=collection_id,
+            region=sample_filter.embedding_region,
+        )
     total_counts = _get_total_counts(session=session, collection_id=collection_id)
     current_counts = _get_current_counts(
         session=session,
         collection_id=collection_id,
         image_filter=image_filter,
-        region_sample_ids=region_sample_ids,
     )
 
     return [
@@ -74,7 +79,6 @@ def _get_current_counts(
     session: Session,
     collection_id: UUID,
     image_filter: ImageFilter | None,
-    region_sample_ids: list[UUID] | None,
 ) -> dict[str, int]:
     """Returns filtered annotation counts per label for the collection."""
     filtered_query = (
@@ -100,10 +104,6 @@ def _get_current_counts(
 
     if image_filter is not None:
         filtered_query = image_filter.apply(filtered_query)
-
-    filtered_query = embedding_region_resolver.apply_region_sample_ids(
-        filtered_query, region_sample_ids
-    )
 
     # Group by label name and sort
     filtered_query = filtered_query.group_by(AnnotationLabelTable.annotation_label_name).order_by(
--- a/lightly_studio/src/lightly_studio/resolvers/image_resolver/count_image_annotations_by_collection.py
+++ b/lightly_studio/src/lightly_studio/resolvers/image_resolver/count_image_annotations_by_collection.py
@@ -25,7 +25,7 @@ def count_image_annotations_by_collection(
     label name and counted for total and filtered.
     Returns a list of (label_name, current_count, total_count) tuples.
     """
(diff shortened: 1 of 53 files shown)


lightly_studio/src/lightly_studio/resolvers/sample_resolver/sample_filter.py
```diff
@@ -44,9 +56,18 @@ class SampleFilter(BaseModel):
     # before applying this filter.
     query_expr: Optional[QueryExpr] = None
 
+    def set_resolved_region_sample_ids(self, sample_ids: list[UUID]) -> None:
+        """Store the sample ids that ``embedding_region`` resolves to.
+
+        Called by the region resolver before the filter is applied, since the point-in-polygon
+        test needs a database session that ``apply`` does not have access to.
+        """
+        self._resolved_region_sample_ids = sample_ids
```
MalteEbner: The design of making `_resolved_region_sample_ids` an attribute, set it via `resolve_embedding_region` and `set_resolved_region_sample_ids` and then reading it looks like quite a lot of boilerplate to me. Can't this be made easier?

Why not just compute the region it within the `_apply_embedding_region_filter`? This would reduce the LOC massively.  


```python

class SampleFilter(BaseModel):
    ...
    embedding_region: Optional[EmbeddingRegion] = None

    def apply(self, query: QueryType, session: Session, collection_id: UUID) -> QueryType:
        """Apply the filters to the given query."""
        query = self._apply_sample_ids_filter(query)
        query = self._apply_embedding_region_filter(query, session, collection_id)
        ...

    def _apply_embedding_region_filter(
        self, query: QueryType, session: Session, collection_id: UUID
    ) -> QueryType:
        if self.embedding_region is None:
            return query
        sample_ids = embedding_region_resolver.get_sample_ids_in_region(
            session=session, collection_id=collection_id, region=self.embedding_region
        )
        # An empty region encloses no points and must match nothing (not everything).
        if not sample_ids:
            return query.where(sqlalchemy.false())
        return query.where(
            db_array.in_array(column=col(SampleTable.sample_id), values=sample_ids)
        )
```

ikondrat: I think the main problem is what we need session which we don't have for filter builders

I will check what we can do

ikondrat: We can't really do what you propose. Because we need to have session to calculate sample_ids

I also considered to precalculate ids and then reused them as "sample filtering" of existing filtering setup but it is different comparing to embedding region filtering

Logic before:
1. Calculate ids
2. Store ids on filter level to use it in the apply (with setters/getters) 

Pros
- use the same filtering logic as rest of the filters with apply

Cons
- we store ids in the filter
- slightly more logic
- we have store with set/get in filters

Logic now:
1. Calculated ids on resolver level to use for filtering
2. Pass ids as region_sample_ids
3. Apply region_sample_ids outside of the geenral filter defenition

Pros
- less code and more clear logic

Cons
- embedding region filtering is handled outside of "filter class"

lightly_studio/src/lightly_studio/resolvers/sample_resolver/sample_filter.py
```diff
@@ -34,6 +36,13 @@ class SampleFilter(BaseModel):
     has_captions: Optional[bool] = None
     annotations_filter: Optional[AnnotationsFilter] = None
     confusion_cell: Optional[ConfusionCell] = None
+    embedding_region: Optional[EmbeddingRegion] = None
+
+    # Sample ids enclosed by ``embedding_region``, resolved server-side before the query is built
+    # (the point-in-polygon test needs a session that ``apply`` cannot access). ``None`` means no
+    # region restriction; an empty list matches nothing, since an empty region encloses no points.
+    # Populated by ``embedding_region_resolver.resolve_region_sample_ids``.
+    region_sample_ids: Optional[list[UUID]] = None
```
MalteEbner, line 45: This makes the `region_sample_ids` an attribute of the API endpoint. Would make it internal and only set it by a method.

lightly_studio/src/lightly_studio/resolvers/image_resolver/count_image_annotations_by_collection.py
```diff
@@ -24,6 +25,15 @@ def count_image_annotations_by_collection(
     label name and counted for total and filtered.
     Returns a list of (label_name, current_count, total_count) tuples.
     """
+    # Resolve any embedding-plot region selection to concrete sample ids on the filter before the
+    # query is built (the point-in-polygon test needs the session, which `apply` lacks).
+    sample_filter = image_filter.sample_filter if image_filter is not None else None
+    if sample_filter is not None and sample_filter.embedding_region is not None:
+        sample_filter.region_sample_ids = embedding_region_resolver.get_sample_ids_in_region(
```
MalteEbner, line 32: The same logic is now repeated three times:
```python
sample_filter = image_filter.sample_filter if image_filter is not None else None
    if sample_filter is not None and sample_filter.embedding_region is not None:
        sample_filter.region_sample_ids = embedding_region_resolver.get_sample_ids_in_region(
```


Would rather have it in one method of the sample_filter

MalteEbner (commented):
WDYT about the approach in https://github.com/lightly-ai/lightly-studio/pull/1600/changes?

It keeps the `SampleFilter._region_sample_ids` fully internal with both setting and using them only internally.

lightly_studio/src/lightly_studio/resolvers/image_resolver/count_image_annotations_by_collection.py
MalteEbner: Furthermore, if this resolution block of code is forgotten, then the filter is silently ignored, there is no error message for that case.

MalteEbner (approved):
As discussed personally, there is no good solution. Let's take this one

PR #1586, opened 2026-07-06
backend part to filter annotations by embedding-plot region selection

ikondrat:
## What has changed and why?

- Created mixin class to use it in annotations and images filtering to filter samples by the region
- Refactored the test to not double test the logic for region filtering in sample and annotation filtering

## How has it been tested?

- [ ] Yes

## Summary by CodeRabbit

* **New Features**
  * Added optional embedding-region filtering when reading annotations with payload.
  * Region selection now constrains annotation results to samples within the selected region.

* **Bug Fixes**
  * Empty region selection now returns no results (no fallback to unfiltered data).
  * Unset/unknown region selection leaves results unchanged.

* **Tests**
  * Added and updated coverage for embedding-region filtering across annotation and sample filtering paths.

### DIFF
--- a/lightly_studio/src/lightly_studio/resolvers/annotations/annotations_filter.py
+++ b/lightly_studio/src/lightly_studio/resolvers/annotations/annotations_filter.py
@@ -36,15 +36,7 @@ class AnnotationsFilter(GridFilterBase):
     sample_ids: list[UUID] | None = Field(
         default=None, description="List of annotation sample UUIDs to restrict to"
     )
-    # Lasso/rectangle selection from the embedding plot, sent as geometry (a few KB) instead
-    # of the full list of selected annotation sample ids (see LIG-9903). It is resolved to
-    # concrete sample ids server-side via point-in-polygon over the cached 2D projection.
     embedding_region: EmbeddingRegion | None = None
-
-    # Annotation sample ids enclosed by ``embedding_region``, resolved server-side before the query
-    # is built (the point-in-polygon test needs a session that ``apply`` cannot access). ``None``
-    # means no region restriction; an empty list matches nothing, since an empty region encloses no
-    # points. Populated by ``embedding_region_resolver.resolve_annotation_embedding_region``.
     region_sample_ids: list[UUID] | None = None
 
     def apply(
--- a/lightly_studio/src/lightly_studio/resolvers/annotations/annotations_filter.py
+++ b/lightly_studio/src/lightly_studio/resolvers/annotations/annotations_filter.py
@@ -142,22 +142,7 @@ class AnnotationsFilter(GridFilterBase):
             )
 
         # Filter by embedding-plot region selection, resolved server-side to sample ids.
-        if self.embedding_region is not None:
-            if self._resolved_region_sample_ids is None:
-                raise RuntimeError(
-                    "embedding_region must be resolved with set_resolved_region_sample_ids() "
-                    "before the filter is applied."
-                )
-            if not self._resolved_region_sample_ids:
-                # An empty region encloses no points and must match nothing (not everything).
-                query = query.where(sqlalchemy.false())
-            else:
-                query = query.where(
-                    db_array.in_array(
-                        column=col(annotation_sample.sample_id),
-                        values=self._resolved_region_sample_ids,
-                    )
-                )
+        query = self._apply_embedding_region_filter(query, annotation_sample=annotation_sample)
 
         # Filter by annotation label
         if self.annotation_label_ids:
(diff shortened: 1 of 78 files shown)


lightly_studio/tests/resolvers/annotations/test_annotations_filters.py
```diff
@@ -204,6 +205,74 @@ def test_combined_filters(
...
+
+    region_filter = AnnotationsFilter(embedding_region=_square_region())
+    filtered_annotations = annotations_resolver.get_all(
+        session=db_session, filters=region_filter
+    ).annotations
+
+    assert len(filtered_annotations) == len(all_annotations)
+
+
+def _square_region() -> EmbeddingRegion:
```
MalteEbner: Isn't this duplicate to the same function for images?

lightly_studio/src/lightly_studio/resolvers/annotations/annotations_filter.py
```diff
@@ -144,5 +152,23 @@ def _apply_annotation_filters(
 
         return query
 
+    def _apply_region_sample_ids_filter(
+        self,
+        query: QueryType,
+        annotation_sample: type[SampleTable],
+    ) -> QueryType:
```
MalteEbner: As this is an almost-identical copy as the same for SampleFilter, why not have it in a shared parent class?

lightly_studio/src/lightly_studio/resolvers/annotations/annotations_filter.py
```diff
@@ -34,6 +36,8 @@ class AnnotationsFilter(GridFilterBase):
     sample_ids: list[UUID] | None = Field(
         default=None, description="List of annotation sample UUIDs to restrict to"
     )
+    embedding_region: EmbeddingRegion | None = None
+    region_sample_ids: list[UUID] | None = None
```
MalteEbner: This has much less comments that the same attributes for the SampleFilter. Either have them at both or none. Also check correctness of the comments again.

And why not have it centrally in a parent class?

MalteEbner (commented):
Instead of duplicating all the logic, I would put it into a mixin. E.g.

```python
class EmbeddingRegionFilterMixin(BaseModel):
    # comments
    embedding_region: EmbeddingRegion | None = None
    region_sample_ids: list[UUID] | None = None

    def _apply_region_sample_ids_filter(self, query: QueryType, column: ...) -> QueryType:
        if self.region_sample_ids is None:
            return query
        if not self.region_sample_ids:
            return query.where(sqlalchemy.false())
        return query.where(db_array.in_array(column=column, values=self.region_sample_ids))
```

Then let both `SampleFilter` and `AnnotationFilter` inherit from it. SampleFilter passes col(SampleTable.sample_id); AnnotationsFilter passes col(annotation_sample.sample_id).

lightly_studio/src/lightly_studio/resolvers/annotations/annotations_filter.py
ikondrat: I also thought the same but we have differenent return types as well, so it is not fully identical

let's try to make it with generic

lightly_studio/src/lightly_studio/resolvers/region_sample_ids_filter.py
```diff
@@ -0,0 +1,47 @@
...
+    restriction; an empty list matches nothing, since an empty region encloses no
+    points.
+    """
+
+    embedding_region: EmbeddingRegion | None = None
+
+    # Sample ids enclosed by ``embedding_region``, resolved server-side before the query is built
+    # (the point-in-polygon test needs a session that ``apply`` cannot access). ``None`` means no
+    # region restriction; an empty list matches nothing, since an empty region encloses no points.
+    # Populated by ``embedding_region_resolver.resolve_region_sample_ids``.
```
MalteEbner: The comment was wrong previously already.
```suggestion
    # Populated by ``embedding_region_resolver.get_sample_ids_in_region``.
```

MalteEbner: The two comments tell almost the same, condense them.

MalteEbner (approved):
Mostly LGTM, but many details could be better.

Please ask for a thorough review by another AI first. E.g. coding with Claude, review with Codex.

What Claude found for me regarding the code:

Image

And the tests:

Image

lightly_studio/src/lightly_studio/resolvers/region_sample_ids_filter.py
ikondrat: I wouldn't agree about spy tests with AI here, it checks what we can rely on mixin logic

ikondrat: I cleaned it up, we don't need this info in the mixin

PR #1587, opened 2026-07-06
Send image embedding-plot selection as geometry

ikondrat:
## What has changed and why?

Image embedding-plot lasso/rectangle selections are now stored and sent to the backend as **region geometry** (`embedding_region`) instead of a resolved list of `sample_ids`. The backend resolves the polygon to sample ids server-side, so the request body stays small and constant-size regardless of how many items are selected.

Key pieces:

- **New shared plumbing** — `useEmbeddingPlotSelection` (a per-collection selection-count store: `setPlotSelectionCount` / `clearPlotSelectionCount` / `getPlotSelectionCount`) and `useRegionFilterVisibility`, which drives the sidebar chip's count/visibility from the plot-propagated count rather than a resolved id list. Clearing the filter removes the region entirely (geometry has no separate "hidden"/restore state).
- **`useImageFilters`** — adds `updateEmbeddingRegion(...)` to set/clear only the region in the existing filter params, and threads `embedding_region` into the derived `imageFilter` sent to the API.
- **Request/type wiring** — `embedding_region` is added to `NormalModeFilters`, `buildRequestBody`, and the external-filter sync (`syncFilterParams`) so the region is excluded from base-param comparison and carried forward across collections alongside `sample_ids`.
- **`useEmbeddingFilterForImages`** — switched from the `sample_ids`-based `useFilterVisibility` to the new `useRegionFilterVisibility`.

## How has it been tested?

Unit tests updated/added:

- `useEmbeddingPlotSelection.test.ts` — per-collection count isolation, defaults, set/clear.
- `useEmbeddingFilter.test.ts` — chip visibility/count now driven by propagated plot-selection count; `setVisibility(false)` / `clearFilter` clear the region and call `setRangeSelection`.
- `useImagesInfinite.test.ts` — `embedding_region` geometry is propagated to `sample_filter`, and omitted when no region is selected.
- `syncFilterParams.test.ts` — `embedding_region` treated as an external filter and carried forward across collections.

- [ ] Yes

## Summary by CodeRabbit

* **New Features**
  * Lasso and region selections now work consistently across images and annotations, with selection state preserved as highlighted regions.
  * Selection counts are now tracked per collection, improving visibility of active filters.

* **Bug Fixes**
  * Fixed selection handling so clearing or switching collections resets the correct region state.
  * Improved filter syncing so region-based selections remain accurate when navigating or reloading.
  * Select-all behavior now respects annotation regions instead of only item lists.

### DIFF
--- a/lightly_studio_view/src/lib/hooks/useEmbeddingFilter/useEmbeddingPlotSelection.ts
+++ b/lightly_studio_view/src/lib/hooks/useEmbeddingFilter/useEmbeddingPlotSelection.ts
@@ -1,0 +1,26 @@
+import { derived, writable, type Readable } from 'svelte/store';
+
+// Per-collection count of items currently selected by the embedding-plot lasso/rectangle.
+const selectionCountByCollection = writable<Record<string, number>>({});
+
+export function setPlotSelectionCount(collectionId: string, count: number): void {
+    selectionCountByCollection.update((counts) => ({ ...counts, [collectionId]: count }));
+}
+
+export function clearPlotSelectionCount(collectionId: string): void {
+    selectionCountByCollection.update((counts) => {
+        if (!(collectionId in counts)) {
+            return counts;
+        }
+        const next = { ...counts };
+        delete next[collectionId];
+        return next;
+    });
+}
+
+export function getPlotSelectionCount(collectionId: Readable<string>): Readable<number> {
+    return derived(
+        [selectionCountByCollection, collectionId],
+        ([$counts, $collectionId]) => $counts[$collectionId] ?? 0
+    );
+}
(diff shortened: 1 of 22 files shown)


lightly_studio_view/src/lib/hooks/useEmbeddingFilter/useEmbeddingPlotSelection.ts
```diff
@@ -0,0 +1,31 @@
+import { derived, writable, type Readable } from 'svelte/store';
+
+// Per-collection count of items currently selected by the embedding-plot lasso/rectangle.
```
LeonardoRosaa, line 3: Can you take a look at this comment format? It seems broken.

ikondrat: It works correctly

Image

LeonardoRosaa (approved):

PR #1591, opened 2026-07-06
Video ingest: record missing/broken/already-present in the per-run report (LIG-10032)

JonasWurst:
## What

Wires video ingest into the central `FileOutcomeReport`, so a folder of videos records one outcome per file — **missing** vs **broken** distinguished, **already-present** from path-dedup, **added** for the rest — feeding the same end-of-run summary as image ingest.

### `load_into_collection_from_paths`
- **Missing** detected proactively via `fs.exists()` → `MissingInputFileError` (not `FileNotFoundError`, which is unreliable across fsspec backends and a subclass of `OSError`).
- **Broken** translated at the I/O boundary: a failed `container.open` / header read → `BrokenInputFileError(...) from e`.
- Removed the catch-all `except (FileNotFoundError, OSError, IndexError, FFmpegError)` block that logged-and-skipped, so bad input files are now *recorded* instead of silently dropped and any unexpected exception propagates (a bug/infra failure is not a broken file).
- Closes the run with `report.log_summary()` then `report.raise_if_all_failed()`, matching image ingest.

### `load_video_annotations_from_labelformat`
- The annotation-matching loop raised `ValueError` when a referenced video had no created sample. A video is not created when it is beyond the limit or when the report already recorded it as missing/broken/already-present — so the run recorded the outcome and then crashed anyway. It now logs a warning and skips those annotations, in line with tolerate-don't-crash.

## Out of scope / follow-up

Adding annotations to *already-present* videos (rather than skipping) is deliberately left as-is: it would expose annotation/object-track duplication on rerun (annotation ingest appends with no dedup), which is annotation-rerun semantics best handled under LIG-9908 / a dedicated annotation slice.

## Tests

- `test_load_into_collection_from_paths__records_missing_broken_already_present_outcomes` — good / missing / broken / already-present mix, each with a distinct count.
- `test_load_video_annotations_from_labelformat__skips_annotations_for_broken_video` — a broken video's annotations are skipped while the good video's annotations still process.

Full backend suite passes (pre-existing unrelated failures in `test_server`/`test_enterprise`/`test_fsspec_lister` only); mypy, ruff check, and format-check clean on the changed files.

## Summary by CodeRabbit

* **Bug Fixes**
  * Video imports now proactively distinguish missing vs unreadable files and report per-outcome results (including clearer end-of-run failure signals).
  * The import run now fails if all selected videos cannot be loaded.
  * Annotation imports now skip annotations that can’t be matched to loaded videos, rather than erroring or using previous fallback behavior.
  * Improved cleanup of opened video resources during loading.
* **Tests**
  * Added coverage for missing, broken, and already-present video import outcomes.
  * Added coverage ensuring annotations for broken/unreadable videos are skipped.

### DIFF
--- a/lightly_studio/src/lightly_studio/core/video/add_videos.py
+++ b/lightly_studio/src/lightly_studio/core/video/add_videos.py
@@ -148,7 +148,7 @@ def load_into_collection_from_paths(  # noqa: PLR0913
             # fsspec backends and is a subclass of OSError, which we treat as broken.
             fs, fs_path = fsspec.core.url_to_fs(url=video_path)
             if not fs.exists(fs_path):
-                raise MissingInputFileError(video_path)
+                raise MissingInputFileError()
 
             video_file = fs.open(path=fs_path, mode="rb")
             try:
@@ -168,7 +168,7 @@ def load_into_collection_from_paths(  # noqa: PLR0913
                     else:
                         video_duration = None
                 except (OSError, IndexError, FFmpegError) as e:
-                    raise BrokenInputFileError(video_path) from e
+                    raise BrokenInputFileError() from e
 
                 # Create video sample
                 video_sample_ids = video_resolver.create_many(
--- a/lightly_studio/src/lightly_studio/core/video/add_videos.py
+++ b/lightly_studio/src/lightly_studio/core/video/add_videos.py
@@ -291,12 +291,14 @@ def load_video_annotations_from_labelformat(  # noqa: PLR0913
         video_path_without_suffix = str((root_path / video_annotation_filename).with_suffix(""))
         video_sample_id = video_path_without_suffix_to_sample_id.get(video_path_without_suffix)
         if video_sample_id is None:
-            if limit is not None:
-                # The video was not loaded because it is beyond the limit.
-                continue
-            raise ValueError(
-                f"No matching video ({video_annotation_filename}) for annotations found"
+            # The video was not created: it is beyond the limit, or the per-run report
+            # already recorded it as missing/broken/already-present. Skip its annotations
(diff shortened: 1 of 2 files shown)


lightly_studio/src/lightly_studio/core/video/add_videos.py
```diff
@@ -142,10 +144,16 @@ def load_into_collection_from_paths(  # noqa: PLR0913
                 raise AlreadyPresentInputFileError()
             seen_or_existing_paths.add(video_path)
 
+            # Detect a missing path proactively: FileNotFoundError is unreliable across
+            # fsspec backends and is a subclass of OSError, which we treat as broken.
+            fs, fs_path = fsspec.core.url_to_fs(url=video_path)
+            if not fs.exists(fs_path):
+                raise MissingInputFileError(video_path)
```
JonasWurst:
```suggestion
                raise MissingInputFileError()
```

lightly_studio/src/lightly_studio/core/video/add_videos.py
```diff
@@ -159,55 +167,52 @@ def load_into_collection_from_paths(  # noqa: PLR0913
...
-                            )
-                        ],
-                    )
-
-                    if len(video_sample_ids) != 1:
-                        video_container.close()
-                        raise RuntimeError(
-                            f"There was an error adding {video_path} to the dataset."
+                except (OSError, IndexError, FFmpegError) as e:
+                    raise BrokenInputFileError(video_path) from e
```
JonasWurst:
```suggestion
                    raise BrokenInputFileError() from e
```

MalteEbner (approved):

PR #1603, opened 2026-07-08
use annotation_type

ikondrat:
## What has changed and why?

Adds an optional `annotation_type` filter to the two annotation count resolvers.

**Motivation.** The count resolvers (`count_image_annotations_by_collection` and `count_video_frame_annotations_by_video_collection`) group annotations by label and return per-label totals, but they never distinguish between annotation types. A single sample can carry both a CLASSIFICATION annotation and OBJECT_DETECTION annotations, so the counts silently merge the two. For the upcoming class-distribution panel (frontend prototype in #1542), that mix is wrong: on a dataset with both types, a classification class distribution would be polluted by detection labels (and vice versa). To render a correct per-type distribution, the frontend needs to request counts scoped to one annotation type — and that scoping has to happen in the query, since the type dimension isn't present in the response payload and can't be reconstructed client-side.

**Change.** Both resolvers now accept an optional `annotation_type` parameter. When provided, it applies a `WHERE annotation_base.annotation_type == <type>` to **both** the total and the filtered/current count queries, so a requested type is isolated across the whole result (total column included). When omitted, the queries are unchanged and existing behaviour is fully preserved. The parameter is surfaced through the REST layer on the existing count endpoints (`ReadCountImageAnnotationsRequest` / `ReadVideoCountAnnotationsRequest`), accepting the standard `AnnotationType` enum values (e.g. `classification`, `object_detection`).

This is a small, backend-only change (~90 logic lines + tests) that unblocks the class-distribution work; the frontend wiring is tracked separately in LIG-9581.

Files touched: image/video count resolvers, the two REST route bodies, and a new test in each resolver's test file.

## How has it been tested?

- Added a unit test to each resolver covering: no filter (both types counted), `CLASSIFICATION` (isolates the classification label), and `OBJECT_DETECTION` (isolates the detection label), asserting both total and current counts.
- Existing resolver tests still pass unchanged, confirming behaviour is identical when `annotation_type` is not passed.
- Ran `make static-checks` (ruff lint + format + type-check) locally.

Reproduce: `cd lightly_studio && make static-checks && make test`

- [ ] Yes

## Summary by CodeRabbit

* **New Features**
  * Added optional annotation-type filtering for image annotation count results.
  * Added optional annotation-type filtering for video frame annotation count results.
  * When an annotation type is provided, both “current” and “total” counts are restricted to that type; otherwise results include all types.

* **Tests**
  * Added coverage to verify image count filtering by annotation type.
  * Added coverage to verify video frame count filtering by annotation type.

### DIFF
--- a/lightly_studio/src/lightly_studio/resolvers/image_resolver/count_image_annotations_by_collection.py
+++ b/lightly_studio/src/lightly_studio/resolvers/image_resolver/count_image_annotations_by_collection.py
@@ -6,7 +6,10 @@ from uuid import UUID
 
 from sqlmodel import Session, col, func, select
 
-from lightly_studio.models.annotation.annotation_base import AnnotationBaseTable
+from lightly_studio.models.annotation.annotation_base import (
+    AnnotationBaseTable,
+    AnnotationType,
+)
 from lightly_studio.models.annotation_label import AnnotationLabelTable
 from lightly_studio.models.image import ImageTable
 from lightly_studio.models.sample import SampleTable
@@ -18,12 +21,18 @@ def count_image_annotations_by_collection(
     session: Session,
     collection_id: UUID,
     image_filter: ImageFilter | None = None,
+    annotation_type: AnnotationType | None = None,
 ) -> list[tuple[str, int, int]]:
     """Count annotations for a specific image collection.
 
     Annotations for a specific collection are grouped by annotation
     label name and counted for total and filtered.
     Returns a list of (label_name, current_count, total_count) tuples.
+
+    When ``annotation_type`` is provided, both the total and filtered counts are
+    restricted to annotations of that type (e.g. only CLASSIFICATION or only
+    OBJECT_DETECTION). When it is not provided, counts include all types and the
+    behaviour is unchanged.
     """
     # Resolve any embedding-plot region selection to concrete sample ids on the filter before the
     # query is built (the point-in-polygon test needs the session, which `apply` lacks).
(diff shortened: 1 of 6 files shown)


lightly_studio/src/lightly_studio/resolvers/image_resolver/count_image_annotations_by_collection.py
```diff
@@ -18,12 +21,18 @@ def count_image_annotations_by_collection(
...
     """Count annotations for a specific image collection.
 
     Annotations for a specific collection are grouped by annotation
     label name and counted for total and filtered.
     Returns a list of (label_name, current_count, total_count) tuples.
+
+    When ``annotation_type`` is provided, both the total and filtered counts are
+    restricted to annotations of that type (e.g. only CLASSIFICATION or only
+    OBJECT_DETECTION). When it is not provided, counts include all types and the
+    behaviour is unchanged.
```
horatiualmasan: We can drop " and the behaviour is unchanged"

lightly_studio/src/lightly_studio/resolvers/video_resolver/count_video_frame_annotations_by_collection.py
```diff
@@ -23,16 +26,31 @@ class CountAnnotationsView(BaseModel):
...
+    filters: Optional[VideoFilter] = None,
+    annotation_type: Optional[AnnotationType] = None,
 ) -> list[CountAnnotationsView]:
-    """Count the annotations by video frames."""
+    """Count the annotations by video frames.
+
+    When ``annotation_type`` is provided, both the total and filtered counts are
+    restricted to annotations of that type (e.g. only CLASSIFICATION or only
+    OBJECT_DETECTION). When it is not provided, counts include all types and the
+    behaviour is unchanged.
```
horatiualmasan: Same here

horatiualmasan (approved):
LGTM

PR #1606, opened 2026-07-08
Delete Annotation embedding when deleting a annotation

horatiualmasan:
## What has changed and why?

Delete sample embeddings before deleting the annotation sample to prevent duckdb AssertionError

## How has it been tested?

New unit test and fixed existing e2e test

- [ ] Yes

## Summary

* **Bug Fixes**
  * Improved annotation deletion to also remove associated stored embedding data and the related sample record, ensuring full cleanup.

* **Testing**
  * Added a unit test covering embedding deletion when an annotation is removed.
  * Updated the end-to-end annotation deletion flow to use a dedicated delete helper, wait for navigation, and verify the backend deletion succeeds.


IgorSusmelj (approved):
LGTM

PR #1609, opened 2026-07-08
Preserve user-entered percentage

LeonardoRosaa:
## What has changed and why?

Change `percentageToSelect` from `derived` to `writable`. Its value is now set explicitly only when the user changes an input.

Additionally, clearing either input now leaves it empty instead of resetting to 0.

## How has it been tested?

Unit test + Manual test

- [ ] Yes

## Summary by CodeRabbit

* **Bug Fixes**
  * Empty sample-count and percentage fields now clear correctly (no invalid numeric display).
  * Validation now treats cleared inputs as invalid, updating the create-button tooltip accordingly.
  * User-entered percentage is preserved during background sample-count changes.
  * Form reset and submit now align with what’s shown in the inputs, including proper handling of cleared/empty values.


IgorSusmelj (approved):
LGTM

PR #1615, opened 2026-07-09
Add: python API method to add custom embeddings

IgorSusmelj:
## What has changed and why?

- Add sample.set_embedding(...) so users can plug in embeddings from their own model instead of the built-in one.
- Add an example script (example_custom_embeddings.py) and a new docs section showing how to use it, including its current limitations (no text search, image/video samples only).
- Add tests covering the new method and the interaction with the embedding plot and search UI.

## How has it been tested?

- New unit tests

- [X] Yes

### DIFF
--- a/lightly_studio/src/lightly_studio/database/db_manager.py
+++ b/lightly_studio/src/lightly_studio/database/db_manager.py
@@ -120,6 +120,7 @@ class DatabaseEngine:
             )
         else:
             SQLModel.metadata.create_all(self._engine)
+            _migrate_duckdb_schema(self._engine)
 
     @contextmanager
     def session(self) -> Generator[Session, None, None]:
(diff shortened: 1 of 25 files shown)


lightly_studio/src/lightly_studio/database/db_manager.py
```diff
@@ -310,6 +311,26 @@ def _initialize_postgres_schema(
     db_migrations.run_migrations(engine=engine, engine_url=engine_url)
 
 
+def _migrate_duckdb_schema(engine: Engine) -> None:
```
IgorSusmelj, line 314: This we could also leave away

lightly_studio/src/lightly_studio/dataset/embedding_utils.py
```diff
@@ -2,36 +2,63 @@
...
-from lightly_studio.dataset.embedding_manager import EmbeddingManagerProvider
 from lightly_studio.resolvers import (
+    embedding_model_resolver,
     sample_embedding_resolver,
 )
 
 
-def collection_has_embeddings(session: Session, collection_id: UUID) -> bool:
-    """Check if there are any embeddings available for the given collection.
+class CollectionEmbeddingsStatus(SQLModel):
```
IgorSusmelj, line 13: This was refactored. The new status tells us whether we have embeddings and whether we have text embeddings for search. We can later allow the user to hook up their own text encoder model to do also text based search with custom models.

PR #1624, opened 2026-07-09
Backend: numeric metadata histogram bins 

ikondrat:
### Motivation

The `/metadata/info` endpoint previously only exposed `min` and `max` for numeric metadata fields. That's enough to render a range, but not enough to show *how* values are distributed. To build distribution charts in the UI (e.g. a histogram slider for filtering numeric metadata), the frontend needs the actual value distribution, not just the bounds. This PR provides that data server-side so the client can render histograms directly.

### Changes

Adds value-distribution histograms for numeric metadata fields to the `/metadata/info` response for `integer` and `float` fields.

- **New `HistogramView` model** (`bin_edges`, `counts`) added to `models/metadata.py`, plus an optional `histogram` field on `MetadataInfoView`. Bin `i` covers the half-open interval `[bin_edges[i], bin_edges[i+1])`, with the last bin including its right edge; `bin_edges` has length `len(counts) + 1`.
- **Resolver reworked** (`get_metadata_info.py`): instead of computing min/max with SQL `func.min`/`func.max`, it now fetches the raw numeric values (`_get_metadata_numeric_values`) and computes min, max, and a 20-bin histogram in Python via a new `_compute_histogram` helper using `numpy.histogram`. Binning is done with numpy so it is independent of the database backend (no dialect-specific `width_bucket()`). Constant-value fields produce a single degenerate range, which is kept as-is.

## How has it been tested?

- Updated the API route test (`test_metadata.py`) to assert histogram bin edges and counts are serialized correctly, and that non-numeric fields return `histogram: null`.
- Updated resolver tests (`test_get_metadata_info.py`) to assert histograms are present for numeric fields and absent for string/boolean fields.
- Added two new resolver tests: one verifying bin-edge semantics (edges span min..max, `len(edges) == len(counts) + 1`, counts sum to N) and one for the constant-value edge case.

- [ ] Yes

## Summary by CodeRabbit

* **New Features**
  * Added histogram data to numerical metadata insights, including bin ranges and sample counts.
  * Metadata summaries now report histograms for integer and floating-point values.
  * Non-numerical metadata remains available without histogram data.

* **Bug Fixes**
  * Improved numerical metadata aggregation, including minimum, maximum, and total sample counts.
  * Added handling for constant-value datasets and consistent histogram boundaries.


horatiualmasan (approved):
LGTM

PR #1625, opened 2026-07-10
Allow registering a custom embedding model to override the env default

JonasWurst:
## What

Adds a public API for customers to register their own embedding model at runtime, bypassing the model selected by the `LIGHTLY_STUDIO_EMBEDDINGS_MODEL_TYPE` env var.

```python
import lightly_studio as ls

class MyGenerator:  # implements ls.ImageEmbeddingGenerator / ls.VideoEmbeddingGenerator
    ...

ls.set_default_embedding_model(MyGenerator())   # call BEFORE load_or_create
ds = ls.ImageDataset.create()
ds.add_images_from_path(path=...)
```

## How

- `EmbeddingManager` keeps a per-sample-type override map (`_override_generators`). The `IMAGE`/`VIDEO` slot(s) a generator fills are inferred from the `@runtime_checkable` `ImageEmbeddingGenerator` / `VideoEmbeddingGenerator` protocols it implements (raises `TypeError` if neither).
- `load_or_get_default_model` consults the override **before** falling back to `_load_embedding_generator_from_env`. DB rows are still created lazily per collection, so no session/collection is needed at register time.
- The scope is global (applies to every collection); an unregistered slot (e.g. video) cleanly falls back to the env default.
- Exports the `EmbeddingGenerator` protocols and `ImageCrop` from the top-level package so customers can type their implementations.
- Adds `example_custom_embedding_model.py` (mimics the MobileCLIP generator) that registers a custom model and **confirms via the DB** that it was the model used to embed the images.

## Known limitations (v1)

- Registration is in-process only. Re-launching the GUI via `lightly-studio gui` without re-running the registration script leaves stored embeddings intact but makes text search fall back to the env-var default model (different vector space). Documented in the API docstring.
- Register-time validation is minimal (protocol/duck-typed); a malformed generator surfaces errors at first embed rather than at registration.

## Testing

- 4 new unit tests in `tests/dataset/test_embedding_manager.py`: override beats env, both slots filled, per-slot env fallback, `TypeError` on incompatible generator.
- `make static-checks` (ruff, ruff format, mypy) and the embedding-manager test suite pass.

## Summary by CodeRabbit

* **New Features**
  * Added support for registering a custom default embedding generator for future image/video ingestion.
  * Overridden embedding models take precedence over environment-based defaults for subsequent collection loads (override isn’t persisted across GUI relaunches).
  * Exposed embedding-related utilities at the main package interface.
* **Examples**
  * Added an example showing how to set up a custom MobileCLIP-style image embedding generator before dataset creation.
* **Tests**
  * Added unit tests covering generator override behavior, protocol support, and error handling.


lightly_studio/src/lightly_studio/examples/example_custom_embedding_model.py
```diff
@@ -0,0 +1,161 @@
...
+# images were embedded with from the database and print its name.
+used_model = embedding_model_resolver.get_default_by_collection_id(
+    session=dataset.session,
+    collection_id=dataset.collection_id,
+)
+assert used_model is not None, "No embedding model was registered for the collection."
+print(f"Images were embedded with: {used_model.name} (dim={used_model.embedding_dimension})")
+assert used_model.name == "Custom Embedding Model", (
+    f"Expected the custom model to be used, but got '{used_model.name}'."
+)
```
JonasWurst: Happy to drop this.

horatiualmasan (approved):
LGTM

lightly_studio/src/lightly_studio/examples/example_custom_embedding_model.py
horatiualmasan: You can drop it

PR #1627, opened 2026-07-10
Add collection_filter support to image export endpoints

LeonardoRosaa:
## What has changed and why?

Adds an optional `collection_filter` field to the image `/export/` and `/export/stats` endpoints. When provided, it is applied as an intersection on top of the existing include/exclude filters. So, only samples that satisfy both the include/exclude selection and the active view filters are exported.

## How has it been tested?

Unit tests.

- [ ] Yes

## Summary by CodeRabbit

- **New Features**
  - Collection exports now accept an optional view filter to further constrain which samples are included.
  - Exported files and stats are computed as the intersection of export criteria (e.g., tags or explicit sample lists) and the active view filter, including embedding region–based filtering.
- **Tests**
  - Added API and resolver tests covering intersections between tag/sample filters and image-based view filters.
  - Added tests validating embedding region filtering for both exported content and sample counts.


horatiualmasan (approved):
LGTM

PR #1635, opened 2026-07-13
Consider also own annotations for video filter

horatiualmasan:
## What has changed and why?

Previously, filtering and counting video annotations only considered annotations attached to a video's frames. Videos can also carry annotations directly (e.g. ActivityNet-style event/classification labels on the whole video). This PR makes both the video filter and the annotation counter consider these direct video annotations in addition to frame annotations.

Changes:

`video_filter.py` — `_apply_annotation_filter` now matches a video if either one of its frames or the video itself has an annotation matching the filter (OR of the frame-parent and video-direct subqueries).
`count_video_frame_annotations_by_collection.py` — replaced the frame-only base query with `_build_label_video_pairs_subquery`, a UNION ALL of `(label_id, video_id)` pairs sourced from both frame annotations and direct video annotations. Distinct video counts per label are computed over the union, so a video labeled via both a frame and a direct annotation is still counted once.

## How has it been tested?

New tests

- [ ] Yes

## Summary by CodeRabbit

* **Bug Fixes**
  * Video annotation counts now include labels attached directly to videos in addition to frame-derived annotations.
  * Annotation filtering now matches both frame-derived and direct video-level annotations when constraining video results.
  * Filtered and total counts are reported consistently for both annotation sources.
* **Tests**
  * Added coverage for counting annotations by video collection with direct video annotations.
  * Added coverage ensuring annotation filters correctly restrict queries to directly annotated videos.


lightly_studio/src/lightly_studio/resolvers/video_resolver/count_video_frame_annotations_by_collection.py
```diff
@@ -105,8 +123,20 @@ def _build_base_query(
...
+        .join(VideoTable, col(VideoTable.sample_id) == col(SampleTable.sample_id))
+        .where(col(SampleTable.collection_id) == collection_id)
+    )
     if annotation_type is not None:
-        query = query.where(col(AnnotationBaseTable.annotation_type) == annotation_type)
-
-    return query
+        frame_pairs = frame_pairs.where(col(AnnotationBaseTable.annotation_type) == annotation_type)
+        video_pairs = video_pairs.where(col(AnnotationBaseTable.annotation_type) == annotation_type)
+    return union_all(frame_pairs, video_pairs).subquery("label_video_pairs")
```
LeonardoRosaa, line 142: Should we use `union` rather than `union_all` since it expects distinct values?

LeonardoRosaa (commented):
I'd expect a test with video and video frame annotations.

lightly_studio/src/lightly_studio/resolvers/video_resolver/count_video_frame_annotations_by_collection.py
horatiualmasan: I would leave union_all, results are deduplicated later on, I removed the distinct mention from the docstring so it is not expected.

LeonardoRosaa: It's your call. I'd apply it since it is a quick win 😄

LeonardoRosaa (approved):

PR #1640, opened 2026-07-14
Support exporting video frames (Python API)

MalteEbner:
## What has changed and why

Add `VideoFrameDatasetExport` mirroring `ImageDatasetExport` fully:
- using `video_frame_to_image` instead of `image_sample_to_image`
- same tests

## How was it tested
new unittests, they mirror the ones for images

### DIFF
--- a/lightly_studio/src/lightly_studio/export/video_frame_dataset_export.py
+++ b/lightly_studio/src/lightly_studio/export/video_frame_dataset_export.py
@@ -1,0 +1,63 @@
+"""Exports video-frame datasets from Lightly Studio into various formats."""
+
+from __future__ import annotations
+
+from collections.abc import Iterable
+from typing import cast
+from uuid import UUID
+
+from labelformat.model.image import Image
+from sqlmodel import Session
+
+from lightly_studio.core.sample import Sample
+from lightly_studio.core.video.video_frame_sample import VideoFrameSample
+from lightly_studio.export.dataset_export import DatasetExport
+
+
+class VideoFrameDatasetExport(DatasetExport):
+    """Provides methods to export a video-frame dataset or a subset of it.
+
+    This class is typically not instantiated directly but returned by
+    `VideoFrameDataset.export()`. It allows exporting data in various formats.
+    """
+
+    def __init__(
+        self,
+        session: Session,
+        dataset_id: UUID,
+        samples: Iterable[VideoFrameSample],
+    ):
+        """Initializes the VideoFrameDatasetExport object.
+
+        Args:
+            session: The database session.
+            dataset_id: The dataset ID for label retrieval.
+            samples: Samples to export.
+        """
+        super().__init__(
+            session=session,
+            dataset_id=dataset_id,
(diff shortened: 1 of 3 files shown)


lightly_studio/src/lightly_studio/export/video_frame_dataset_export.py
```diff
@@ -0,0 +1,63 @@
...
+    A frame has no file of its own, so the file name is synthesized from the parent video and
+    the frame number (``<video>/<frame_number>.jpg``) and the dimensions are the parent video's.
+    COCO stores the absolute video path verbatim; YOLO and Pascal VOC need a relative video name.
+    """
+    frame_sample = cast(VideoFrameSample, sample)
+    video = frame_sample.parent_video
+    video_reference = video.file_name if use_relative_filename else video.file_path_abs
+    return Image(
+        id=image_id,
+        filename=f"{video_reference}/{frame_sample.frame_number:09d}.jpg",
```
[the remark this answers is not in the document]
MalteEbner: Valid in principle, but this is a pre-existing, format-level limitation that frame export intentionally inherits from image export rather than a new issue here.

Image export does the same thing — `image_sample_to_image` uses the basename (`file_name`) for relative (YOLO/Pascal) exports, so two same-basename images in different dirs (`/a/img.jpg`, `/b/img.jpg`) collide identically:
https://github.com/lightly-ai/lightly-studio/blob/2519aab2bc9c8ab02bf79495221f535cb9e94e7b/lightly_studio/src/lightly_studio/export/image_dataset_export.py#L54

It is really inherent to YOLO/Pascal VOC: those writers key off a flat, relative filename (`labels/<stem>.txt`, `SegmentationClass/<stem>.png`) and an absolute path cannot be used (the writer joins it with the output dir, so it would escape). Same-basename inputs are fundamentally ambiguous there. Using the basename also matches the existing YouTube-VIS convention (`video_001.mp4/00000.jpg`).

Disambiguating only for frames would make frame export diverge from image export and the existing convention for a corner case (same-basename videos in different dirs) the image path already accepts. If we want to address it, it should be done uniformly for images and frames in a separate change. Resolving here to keep image/frame behavior consistent.

lukas-lightly, line 60: Is this a standardized filename for frames? Should we keep it as a class method, so that it can be reused rather than reimplemented later (possibly with bugs)? I.e. using the DRY principle.

lukas-lightly (approved):

lightly_studio/src/lightly_studio/export/video_frame_dataset_export.py
MalteEbner: I am not sure where else it we would use this. Thus I don't know where to put it instead. 
Would keep it here for now and standardise later.

There is not really a filename, cause there is no file. The filename is only used inside the .yaml or .txt files for coco/yolo/voc as an internal identifier. It also allows extracting the frame from the video later via external programs.

PR #1644, opened 2026-07-15
Add POST /collections/{collection_id}/metadata/histograms endpoint

ikondrat:
## What has changed and why?

Computes value-distribution histograms for all numeric metadata keys in one request.

Bin edges always span the full (unfiltered) value range of each key, so the chart's x-axis stays stable while counts change with the active filters.
Accepts an optional ImageFilter; each key's own metadata filter is excluded from its histogram (faceted-search behavior).
Optional bin_count (default 20), consumed by a later PR in the stack.
Constant-valued keys return a single degenerate bin with a filtered count.

## How has it been tested?

lightly_studio/tests/metadata/test_get_metadata_histograms.py covers shapes, filtering, self-filter exclusion, bin_count, and the constant-value edge case.

Part 1/6 of the numeric-metadata-histogram stack (LIG-9587).

- [ ] Yes

## Summary by CodeRabbit

- **New Features**
  - Added `POST /collections/{collection_id}/metadata/histograms` to generate numeric metadata histograms.
  - Supports optional filters and a validated `bin_count` (default 20).
  - Histogram bin ranges remain consistent across filter changes, with counts reflecting active filtering.
  - Faceting: a field’s own filter won’t restrict its histogram; non-numeric metadata keys are omitted.
  - Updated metadata info retrieval to use the centralized schema resolver.

- **Tests**
  - Added tests for unfiltered vs filtered totals, faceting behavior, constant-value handling, custom bin counts, and skipping non-numeric keys.

- **Chores**
  - Updated ignore rules to exclude local git configuration files.


lightly_studio/src/lightly_studio/api/routes/api/metadata.py
```diff
@@ -38,7 +39,45 @@ def get_metadata_info(
...
+        request: Optional request body carrying the active sample filters.
+
+    Returns:
+        Mapping of metadata key to its histogram.
+    """
+    return metadata_info_resolver.get_metadata_histograms(
+        session=session,
+        collection_id=collection_id,
+        filters=request.filters if request else None,
+        bin_count=request.bin_count if request else 20,
```
horatiualmasan: You can replace 20 with a constant

horatiualmasan (approved):
LGTM

PR #1648, opened 2026-07-15
Scopes the Annotation Classes panel counts to the annotation source

horatiualmasan:
## What has changed and why?

Previously the class "total" count summed a label's annotations across *all* annotation
sources. So when viewing a single source, a class that also exists in another source showed
e.g. `2 of 4` even though only 2 of those annotations belong to the source you're looking at.
The current (filtered) count was already source-scoped, so the two numbers were inconsistent.

Annotation grid view now also scopes the count to the current collection.

Now both counts respect the selected source(s):
- **Backend** (`count_image_annotations_by_collection`): the annotation-source restriction
  (`annotations_filter.collection_ids`) is applied to the *total* count as well as the
  filtered count. The shared join was extracted into a `_restrict_to_annotation_sources`
  helper used by both queries. A label that exists only in unselected sources now drops out
  of the results entirely rather than lingering with a `0` count.
- **Frontend** (`+layout.svelte`): on the annotations / annotation-details routes, counts are
  now scoped to the active annotation collection (`[collectionId]`) instead of being left
  unscoped.
- **Frontend** (`useAnnotationsFilter`): added `pruneInvalidSelections()`, called after each
  counts refresh, which deselects any class filter whose label no longer has matching rows so
  the active filter can't point at a label that isn't shown.

## How has it been tested?

new tests

- [ ] Yes

## Summary by CodeRabbit

* **Bug Fixes**
  * Annotation counts and totals now respect only the selected annotation source(s), excluding unselected sources.
  * Label selections are automatically pruned when refreshed count data no longer includes them.
  * Annotation views compute counts using the correct collection scope, improving consistency when switching sources.
* **Tests**
  * Updated resolver tests to verify source-scoped totals and correct label filtering behavior.
  * Added unit tests for pruning invalid label selections in the annotations filter hook.

### DIFF
--- a/lightly_studio/tests/resolvers/image_resolver/test_count_image_annotations_by_collection.py
+++ b/lightly_studio/tests/resolvers/image_resolver/test_count_image_annotations_by_collection.py
@@ -191,8 +191,9 @@ def test_count_image_annotations_by_collection_filters_by_annotation_source(
     )
     assert {label: current for label, current, _ in counts} == {"dog": 1, "cat": 1}
 
-    # Restricting to source A counts only its "dog"; totals stay full and "cat"
-    # drops to a current count of 0.
+    # Restricting to source A scopes both the current and total counts to that
+    # source: only "dog" is counted, and "cat" (which lives in source B) drops out
+    # of the results entirely rather than lingering with a zero current count.
     counts = image_resolver.count_image_annotations_by_collection(
         session=db_session,
         collection_id=collection_id,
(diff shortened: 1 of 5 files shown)


lightly_studio/tests/resolvers/image_resolver/test_count_image_annotations_by_collection.py
```diff
@@ -217,8 +217,68 @@ def test_count_image_annotations_by_collection_filters_by_annotation_source(
...
+                    )
+                ],
+                collection_name=source_name,
+            )
+            collection_ids.append(created[0].annotation_collection_id)
+        # All annotations added under the same source share one collection id.
+        assert len(set(collection_ids)) == 1
+        return collection_ids[0]
+
+    source_a_collection_id = _annotate_images(count=2, source_name="source_a")
```
LeonardoRosaa: Could we make the tests DAMP? Also, do we really need two entries, or is one per source enough?

LeonardoRosaa (approved):

lightly_studio/tests/resolvers/image_resolver/test_count_image_annotations_by_collection.py
horatiualmasan: updated

PR #1654, opened 2026-07-16
introduce function to build histogram plot

ikondrat:
## What has changed and why?

This PR introduces function building options for Histogram plot. It is accompanied by test and part of the [Numeric metadata values representation](https://linear.app/lightly/issue/LIG-9579/view-numeric-metadata-distribution-2ed). This function will be used by Histogram component coming with follow-up PR.

## How has it been tested?

It is covered by unit test - herefore jsut passing CI check should be enough

- [ ] Yes

## Summary by CodeRabbit

* **New Features**
  * Added histogram option builder with interactive range highlighting, pixel-snapped bars with consistent 1px gaps, and accent/dimmed coloring for selected vs. unselected bins.
  * Tooltips now show bin interval, count, and percentage; empty tooltip states for out-of-range bins.
  * Added reusable `HistogramData` and `HistogramRange` types, including support for constant-value (single-bin) datasets.
  * Added `showAxes` option with controlled axis/tick/label visibility and hover “snap” behavior.
* **Tests**
  * Added Vitest coverage for bin overlap semantics (half-open and zero-width cases), geometry rendering, tooltip content/emptiness, axis behavior, and hover resolution.

### DIFF
--- a/lightly_studio_view/src/lib/components/Histogram/buildHistogramOption.ts
+++ b/lightly_studio_view/src/lib/components/Histogram/buildHistogramOption.ts
@@ -27,21 +27,29 @@ export interface HistogramOptionOptions {
     showAxes?: boolean;
 }
 
+/** A single bar: the half-open value interval `[start, end)` and its count. */
 interface HistogramBin {
     start: number;
     end: number;
     count: number;
 }
 
+/** Inputs the x-axis needs to map integer bin indices back to domain values. */
 interface HistogramAxisOptions {
+    /** Number of bins; also the x-axis max (indices run `0..binCount`). */
     binCount: number;
+    /** Left edge of the first bin (domain minimum). */
     domainMin: number;
+    /** Right edge of the last bin (domain maximum). */
     domainMax: number;
+    /** Whether axis chrome is rendered (see `HistogramOptionOptions.showAxes`). */
     showAxes: boolean;
 }
 
+/** Inputs the bar series needs: the bins and the optional highlight range. */
 interface HistogramSeriesOptions {
     bins: HistogramBin[];
+    /** Selected value range; bins outside it render dimmed. Omit to highlight all. */
     range?: HistogramRange;
 }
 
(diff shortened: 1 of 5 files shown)


lightly_studio_view/src/lib/components/Histogram/buildHistogramOption.ts
```diff
@@ -0,0 +1,222 @@
...
+ * the gaps between them consistent regardless of chart width.
+ *
+ * The x-axis is a value axis over bin indices, so bin `i` spans `[i, i + 1]`
+ * and `api.coord` returns its left edge.
+ */
+export function renderHistogramBin(
+    params: RenderItemParams,
+    api: RenderItemApi
+): Record<string, unknown> {
+    const index = params.dataIndex;
```
mihnea-necsulescu: shouldn't it be `api.value(0)`?

lightly_studio_view/src/lib/components/Histogram/buildHistogramOption.ts
```diff
@@ -0,0 +1,222 @@
...
+}
+
+function buildSeries(options: HistogramSeriesOptions): Record<string, unknown>[] {
+    return [
+        {
+            type: 'custom',
+            renderItem: renderHistogramBin,
+            encode: { x: 0, y: 1 },
+            data: options.bins.map((bin, index) => ({
+                value: [index, bin.count],
```
mihnea-necsulescu: we're passing the index, which would be the left-most point, no? shouldn't we pass the center point, i.e., `index + 0.5` ? 

for example, if we were inside a bin and we were to hover towards the right most part of it, I think the next bin's tooltip would be used as the cursor would be closer to that one, if the data point is the left most one

Claude wrote a test to confirm this:
```ts
import { describe, expect, it } from 'vitest';
import * as echarts from 'echarts';
import { buildHistogramOption } from './buildHistogramOption';
import { normal } from './fixtures';

/**
 * SCRATCH — review verification only, delete after running.
 *
 * Bin i is DRAWN spanning x-axis values [i, i+1]. Hovering anywhere inside that
 * span should resolve the axis tooltip to dataIndex i.
 */
describe('axis tooltip snapping', () => {
    it('resolves the hovered bin', () => {
        const dom = document.createElement('div');
        document.body.appendChild(dom);
        const chart = echarts.init(dom, null, { renderer: 'svg', width: 800, height: 300 });
        chart.setOption(buildHistogramOption(normal));

        // eslint-disable-next-line @typescript-eslint/no-explicit-any
        const series = (chart as any).getModel().getSeriesByIndex(0);
  
        for (const axisValue of [0.2, 0.7, 3.2, 3.7, 19.2, 19.7]) {
            // null maxDistance == what axisTrigger.js passes for a value axis.
            const indices = series.indicesOfNearest('x', 'x', axisValue, null);
            const drawnBin = Math.floor(axisValue);
            // eslint-disable-next-line no-console
            console.log(
                `hover x=${axisValue} -> drawn bin ${drawnBin}, tooltip resolves to bin ${indices[0]}` +
                    (indices[0] === drawnBin ? '   OK' : '   <-- MISMATCH')
            );
        }
        expect(true).toBe(true);
    });
 });
```

maybe a test for this scenario would be worth it too

mihnea-necsulescu (approved):
left some comments, but they could also be validated/addressed while testing with the actual component, thus approving

lightly_studio_view/src/lib/components/Histogram/buildHistogramOption.ts
ikondrat: good catch, updated it

PR #1659, opened 2026-07-16
Record YOLO/lazy-scan folder broken images as BROKEN instead of crashing

JonasWurst:
## Context

Depends on the upstream `labelformat` `on_error` hook ([PR #88](https://github.com/lightly-ai/labelformat/pull/88), now merged).

**Scope note:** This PR was split per review feedback. It now covers only the **YOLO / lazily-scanned folder formats** (YOLO, KITTI, Lightly, mask pairs) plus the shared infrastructure. Pascal VOC handling is stacked on top in #1664.

## Problem

Lazily-scanned folder `labelformat` formats open every image via `get_image_dimensions` to read its dimensions during the `get_images()`/`get_labels()` scans. A broken/undecodable file raised a raw exception mid-scan that aborted the whole ingest and **bypassed the `BROKEN` outcome reporting** entirely.

## Changes

**Sample ingest** (`load_into_dataset_from_labelformat`): route broken images through one `_BrokenImageCollector` set as `input_labels.on_error` — records each once (deduped across the double `get_images()`/`get_labels()` scan) and re-raises anything that is not an `ImageDimensionError`.

**Annotation ingest** (`add_annotations_from_labelformat`): set a skip+log `on_error` hook at the shared chokepoint so a broken image is skipped (it was already recorded `BROKEN` at its own ingest) instead of aborting. A caller that already set a hook keeps its collector.

**Dependency**: pin `labelformat` to the merged PR #88 commit (`07b00a0`) until a release includes it; `uv.lock` regenerated to match.

## Testing

- End-to-end 3×3 matrix over all combinations of a healthy/broken/missing **image** with a healthy/broken/missing **annotation** for YOLO (`test_add_samples_from_yolo__image_x_annotation_matrix`), asserting sample creation, annotation counts, and `BROKEN` reporting.
- YOLO broken-image sample + annotation regression tests.
- `load_into_dataset_from_labelformat` broken-scan + non-`ImageDimensionError` re-raise via a synthetic folder-scanning input.

## Follow-up

- #1664 (stacked): Pascal VOC, whose construction-time `from_dirs` scan needs separate wiring.

## Summary by CodeRabbit

* **Bug Fixes**
  * Image imports now skip unreadable or corrupted files instead of stopping the entire ingestion process.
  * Broken images are recorded in import results and reported with warnings.
  * Unexpected errors continue to be surfaced normally.
  * Annotation imports now handle unreadable images consistently across supported formats.

* **Tests**
  * Added coverage for broken, missing, and valid image and annotation combinations.
  * Added regression tests verifying warnings, outcome reporting, and error propagation.


lightly_studio/src/lightly_studio/core/image/image_dataset.py
```diff
@@ -568,10 +574,24 @@ def add_samples_from_pascal_voc_segmentations(  # noqa: PLR0913
...
+            # bug or infra failure and must propagate.
+            if not isinstance(error, ImageDimensionError):
+                raise error
+            broken_image_paths.add(str(path))
+
         label_input = PascalVOCSemanticSegmentationInput.from_dirs(
             images_dir=images_path,
             masks_dir=masks_path,
             class_id_to_name=class_id_to_name,
+            on_error=_collect_broken_image,
```
MalteEbner: I am not a fan of this approach, as it basically uses the on_error function for an extra output argument. This feels like code smell to me.

lightly_studio/src/lightly_studio/core/image/add_images.py
```diff
@@ -440,6 +463,35 @@ def _group_captions_by_image_id(
     return captions_by_image_id
 
 
+class _BrokenImageCollector:
```
MalteEbner: Order: Classes, also those with _, should come before methods.

MalteEbner (commented):
Two wishes:

1. Please split the PR with one part for YOLO and one for PascalVOC, otherwise this is hard to review.

3. I am missing an E2E test or at least one manual test. It should cover all 3x3 combinations of healthy/broken/missing image files with healthy/broken/missing annotation files. According to Codex, this can cause problems:

> The collector installed at [add_images.py:203](https://github.com/lightly-ai/lightly-studio/blob/119ca2f89a3c1d0998c796c0c08980730313446b/lightly_studio/src/lightly_studio/core/image/add_images.py#L203) makes get_images() skip the broken file. However, Studio constructs LightlyObjectDetectionInput with skip_labels_without_image=False at [image_dataset.py:646](https://github.com/lightly-ai/lightly-studio/blob/119ca2f89a3c1d0998c796c0c08980730313446b/lightly_studio/src/lightly_studio/core/image/image_dataset.py#L646). Its get_labels() subsequently sees the broken image’s JSON without a corresponding scanned image and raises LabelWithoutImageError ([upstream reader](https://github.com/lightly-ai/labelformat/blob/07b00a0edf54e2dc7930a3abc756c7914c068829/src/labelformat/formats/lightly.py#L65-L80)). Thus this format still crashes—the exact behavior the PR aims to fix. Add an end-to-end Lightly regression test and distinguish broken skipped images from genuinely missing images.

JonasWurst:
@MalteEbner both wishes addressed:

1. **Split.** This PR is now scoped to YOLO / lazily-scanned folder formats + the shared infrastructure. Pascal VOC is carved out into a stacked PR #1664 (based on this branch), so each is small and reviewable on its own. The `broken_image_paths` / `on_error`-as-output-argument mechanism you flagged as a code smell now lives entirely in #1664, so that design discussion is isolated there rather than blocking this base PR.

2. **3×3 matrix test.** Added `test_add_samples_from_yolo__image_x_annotation_matrix` — a parametrized end-to-end test over all 9 combinations of healthy/broken/missing image × healthy/broken/missing annotation, asserting sample creation, annotation counts, and BROKEN reporting. #1664 adds the Pascal-VOC image-axis matrix.

On the Codex `LabelWithoutImageError` concern: that failure mode is specific to the **Lightly** format's `get_labels()` with `skip_labels_without_image=False`, not YOLO — YOLO folder-scans and simply skips solitary labels (verified by the `("missing", ...)` matrix rows), so it handles all 9 combos without crashing. The Lightly-format regression is a separate follow-up; happy to open a ticket/PR for it if you'd like.

JonasWurst:
Moved `_BrokenImageCollector` above the module-level functions (classes before methods) in `add_images.py` — commit `ccdb2b5`. The stacked #1664 was rebased on top so it inherits the new ordering.

lightly_studio/tests/core/video/test_add_annotations_activitynet.py
JonasWurst, line 1: This is required due to the labelformat bump

lightly_studio/tests/core/image/test_add_images.py
```diff
@@ -63,6 +66,48 @@ def get_labels(self) -> Iterable[ImageObjectDetection]:
...
 
 
+class FolderScanningLabelInput(ObjectDetectionInput):
+    """Minimal folder-scanning input mirroring labelformat folder formats.
+
+    Both ``get_images`` and ``get_labels`` open every file via
+    ``get_image_dimensions`` and honor the settable ``on_error`` hook, so a
+    broken image is skipped instead of aborting iteration. This mirrors the
+    yolov8/kitti/lightly/pascalvoc/maskpair readers.
+    """
```
MalteEbner: Why do we need to mirror it instead of just taking the original?

MalteEbner (approved):
LGTM

lightly_studio/tests/core/image/test_add_images.py
JonasWurst: The mirror was doing two separable things, and only one needs a fake:

- **Broken-folder-scan / dedup-across-double-scan**: already covered end-to-end by the real YOLO reader in `test_add_samples_from_yolo__records_broken_image`. It runs a genuine `YOLOv8ObjectDetectionInput` through the same `load_into_dataset_from_labelformat` → `BrokenImageCollector`, with the same double `get_images()`/`get_labels()` scan. Its `broken=1` assertion for a single broken file *is* the dedup check: the `on_error` hook fires twice for that file, so without the `_recorded_paths` dedup it would report `broken=2`. So I dropped the synthetic version — it only re-implemented labelformat's scan (drift risk) without adding coverage.
- **Non-`ImageDimensionError` propagation**: no real reader emits a non-dimension error through the hook, so a stub is the only way to hit that path. I reduced the fake to a minimal `ExplodingLabelInput` that just invokes the infra-installed `on_error` with a `ValueError` — it no longer mirrors the scan.

Done in 936b601a.

PR #1664, opened 2026-07-16
Pascal VOC: record folder-scan broken images as BROKEN (stacked on #1659)

JonasWurst:
## Context

Stacked on #1659 (YOLO / lazy-scan formats), which already introduces the public `BrokenImageCollector`. Split out per review feedback that #1659 was hard to review as one PR covering both YOLO and Pascal VOC. **Review #1659 first** — this PR's diff is only the Pascal-VOC-specific delta on top of it.

## Problem

Pascal VOC's `from_dirs` folder scan opens every image to read its dimensions at **construction time** (not lazily like YOLO), so a broken/undecodable image raised mid-scan and aborted the whole ingest, bypassing the `BROKEN` outcome reporting.

## Changes

This PR does **not** modify `BrokenImageCollector` (it lands in its final public form in #1659). It only wires Pascal VOC into it:

- `add_samples_from_pascal_voc_segmentations`: build a `BrokenImageCollector`, pass it as `from_dirs(on_error=...)`, and hand the same instance to `load_into_dataset_from_labelformat` via the new optional `broken_image_collector` parameter. The hook records `BROKEN` directly into the shared report — identical to the YOLO/lazy-scan path.
- `add_annotations_from_pascal_voc_segmentations`: skip a broken image with a warning via `skip_and_warn_unreadable_image` (it was already recorded `BROKEN` at its own ingest).
- Add the optional `broken_image_collector` parameter to `load_into_dataset_from_labelformat` (used only by the Pascal VOC construction-time-scan case).

## Testing

- Pascal-VOC image-axis matrix (healthy/broken/missing image with a valid mask).
- Broken-image sample + annotation regression tests.

The mask axis is intentionally out of scope: a missing mask raises at `from_dirs` construction and a broken mask raises during `get_labels`, both by design and unchanged by this PR.

## Summary by CodeRabbit

* **Bug Fixes**
  * Pascal VOC annotation and sample imports now skip unreadable/broken images instead of aborting the scan.
  * Import runs continue successfully for valid image/mask pairs even when other files are broken.
  * Broken images are consistently skipped and surfaced via warnings and import summary counts.
* **Tests**
  * Added coverage to verify broken-image skipping, warning messages, and recorded broken-image totals for Pascal VOC ingest paths.


JonasWurst:
Addressed the `on_error`-as-output-argument code smell (@MalteEbner's note on #1659).

Instead of a closure that appends broken paths to a `set` passed back via `broken_image_paths`, Pascal VOC now uses the **same `BrokenImageCollector`** the lazy formats use: build one instance, pass it as `from_dirs(on_error=...)`, and inject that same instance into `load_into_dataset_from_labelformat`. The hook records `BROKEN` directly into the shared report — identical to the YOLO path — so `on_error` is a real collaborator, not a hidden output channel. The collector owns its `FileOutcomeReport` so the construction-time scan and the later ingest share one report.

Diff: `98a14f2`.

lightly_studio/src/lightly_studio/core/image/add_images.py
MalteEbner, line 50: Why do we need this class again? Isn't it already in the previous PR?

lightly_studio/src/lightly_studio/core/image/image_dataset.py
```diff
@@ -568,10 +573,17 @@ def add_samples_from_pascal_voc_segmentations(  # noqa: PLR0913
         images_path = _normalize_input_path(path=images_path)
         masks_path = _normalize_input_path(path=masks_path)
 
+        # Pascal VOC opens every image to read its dimensions during the from_dirs folder scan,
+        # which happens here at construction (not lazily in get_images/get_labels). Record broken
+        # images through the same collector that load_into_dataset_from_labelformat uses for the
+        # lazy formats, so the scan does not abort and the broken images land in its report.
+        broken_image_collector = add_images.BrokenImageCollector()
```
MalteEbner: Wrong import style, import BrokenImageCollector directly

lightly_studio/tests/core/image/test_image_dataset__add_annotations.py
MalteEbner, line 270: Why do we add a yolo test for Pascal VOC?

MalteEbner (commented):
Are you sure that the stacking is correct? Looks as if it is not.

lightly_studio/src/lightly_studio/core/image/add_images.py
JonasWurst: You're right — it's from #1659. The branch had been stacked on the #1659 *branch* (via merge commits) and never updated after #1659 was squash-merged into `main`, so the whole #1659 diff (including this class) was re-appearing here. Merged latest `main` in (`4f3c22dd`); the PR now shows only the Pascal VOC delta and this class is no longer part of it.

lightly_studio/src/lightly_studio/core/image/image_dataset.py
JonasWurst: Fixed in `aa3f913c` — `BrokenImageCollector` is now imported directly (`from lightly_studio.core.image.add_images import BrokenImageCollector`) per the import guideline for classes.

lightly_studio/tests/core/image/test_image_dataset__add_annotations.py
JonasWurst: Same stacking issue — that YOLO test comes from #1659 and is already in `main`. It only showed up in this PR's diff because the branch hadn't been updated after #1659's squash-merge. After merging `main` (`4f3c22dd`) the PR diff no longer includes it; the only test this PR adds is the Pascal VOC one.

MalteEbner (approved):

PR #1666, opened 2026-07-16
Introduce hook to change bin count from the Distribution Panel

ikondrat:
## What has changed and why?

Part of the LIG-9587 numeric-metadata-histogram stack. Adds the data layer for metadata histograms; nothing renders yet — the panel UI (`.p2`) and the collection-view wiring (`.p3`) build on top of this.

- `useNumericMetadataDistribution`: tanstack query for `POST /collections/{collection_id}/metadata/histograms` (added in #1644). Takes the same `ImageFilter` that drives the grid, so the histograms refetch and track the active view; `placeholderData` keeps the previous bars on screen while a filter change refetches.
- `selectDistributions`: maps the response (`metadata key → HistogramView`, snake_case `bin_edges`) to the `HistogramData` shape the `Histogram` component consumes.
- Exported from `$lib/hooks`.

## How has it been tested?

- `useNumericMetadataDistribution.test.ts` covers the response mapping, including empty/pending states.

- [ ] Yes

## Summary by CodeRabbit

* **New Features**
  * Added numeric metadata distribution (histogram) data retrieval for collections.
  * Histogram results now refresh when image filters change, while keeping the prior results visible during updates.
* **Bug Fixes**
  * Improved behavior when histogram data is missing or empty to avoid rendering issues.
* **Tests**
  * Added tests to verify histogram data transformation and correct handling of undefined/empty inputs.


lightly_studio_view/src/lib/hooks/useNumericMetadataDistribution/useNumericMetadataDistribution.ts
```diff
@@ -0,0 +1,68 @@
...
+    collectionId: string;
+    filter?: ImageFilter;
+}) =>
+    createQuery(() => {
+        // Computed inside the reactive function so a change to collectionId or
+        // filter updates the query key and triggers a refetch.
+        const requestOptions = {
+            path: { collection_id: collectionId },
+            ...(filter ? { body: { filters: filter } } : {})
+        } as const;
```
LeonardoRosaa: Is this type cast necessary?

LeonardoRosaa (approved):

PR #1676, opened 2026-07-17
Add metadata filter chips to the left sidebar

ikondrat:
## What has changed and why?

Adds metadata filter chips to the left sidebar, making active numeric metadata filters visible and easier to manage.

Each narrowed metadata field now has a chip that:

- Displays the selected range
- Can temporarily disable the filter
- Restores the previous range when re-enabled
- Resets the filter and removes the chip when cleared

The chips are available for image, video, and video-frame collections. This PR also renames the fallback distribution label from “Class labels” to “Annotation labels”.

Part of LIG-9587.

https://github.com/user-attachments/assets/04a35b63-7301-4c3c-9448-b42627af7737

## How has it been tested?

Added component tests covering:

- Rendering chips only for narrowed metadata filters
- Displaying the selected range
- Disabling a filter while remembering its range
- Restoring the range when re-enabled
- Clearing a filter and removing its chip

- [x] Yes
- [] Not needed (covered by the parent feature PR)

## Summary by CodeRabbit

- **New Features**
  - Added metadata filter chips to the left sidebar for image/video frame collections, including min–max range subtitles.
  - Chips support enabling/disabling, clearing, and restoring previously narrowed ranges.
- **Bug Fixes**
  - Updated the default annotation class source label to **“Annotation classes”** when sources aren’t provided.
- **Tests**
  - Added component and hook tests covering rendering, toggle, clear, and state restoration behavior.
- **Documentation**
  - Updated the unreleased changelog to mention metadata filter chips in the left sidebar.


lightly_studio_view/src/lib/components/MetadataFilterChips/MetadataFilterChips.svelte
```diff
@@ -0,0 +1,43 @@
...
+
+    interface Props {
+        collectionId?: string;
+    }
+
+    const { collectionId }: Props = $props();
+
+    const hook = useMetadataFilterChips(collectionId);
+
+    const prettify = (key: string): string => key.replace(/_/g, ' ');
```
LeonardoRosaa: I'm not sure we should prettify the metadata key, since it was entered by the user. If the metadata key is `is_bool`, I wouldn't expect to see `is bool`.

lightly_studio_view/src/lib/components/MetadataFilterChips/useMetadataFilterChips.svelte.ts
```diff
@@ -0,0 +1,96 @@
...
+        return [...keys]
+            .filter((key) => boundsStore.current[key])
+            .map((key) => {
+                const active = isNarrowed(key);
+                const range: Range | undefined = active
+                    ? valuesStore.current[key]
+                    : lastRanges[key];
+                return { key, active, range };
+            })
+            .filter((chip): chip is MetadataFilterChip => !!chip.range);
```
LeonardoRosaa, line 57: I think this `is` clause is redundant since `derived.by` already makes the return type explicit.

LeonardoRosaa (approved):

lightly_studio_view/src/lib/components/MetadataFilterChips/useMetadataFilterChips.svelte.ts
ikondrat: The chip is MetadataFilterChip is a TypeScript type predicate — it narrows the type so TypeScript knows the filtered items have a defined range. Without it, the return type would still include range?: Range instead of range: Range.

PR #1684, opened 2026-07-20
Fast Track: derive current internal PR targets

michal-lightly:
## What has changed and why?

Second PR in the Fast Track bot series (follows #1681). Adds the bot's targeting layer: given the trusted `workflow_run` head SHA, resolve which pull request the verdict actually applies to, rather than trusting artifact-supplied routing.

`deriveTarget` accepts a PR only when exactly one open, non-draft, same-repository PR matches the head SHA; stale, draft, closed, fork, and ambiguous targets fail closed as a clean no-op. `refreshTarget` re-reads mutable PR state right before a write to close the check-then-act race window.

This is pure additive code under `src/bot/` and is not yet wired into any workflow — later PRs in the series consume it.

## How has it been tested?

`make -C fast_track static-checks` and `make -C fast_track test` both pass (16 files, 143 tests, including the new targeting suite).

- [ ] Yes

## Summary by CodeRabbit

* **New Features**
  * Added automated, safety-checked pull request targeting for workflow runs, deriving a valid target from the trusted commit.
  * Ensures the chosen pull request is open, non-draft, matches the expected commit, and is not an invalid/forked source.
  * Refreshes target information at processing time and invalidates targets if the pull request no longer matches expectations.

* **Tests**
  * Added Vitest coverage for target derivation and refresh behavior, including pagination, labels, and rejection of superseded/closed/draft or mismatched targets.

### DIFF
--- a/fast_track/src/bot/derive-target.test.ts
+++ b/fast_track/src/bot/derive-target.test.ts
@@ -43,21 +43,28 @@ function derive(pullRequests: unknown[]) {
 
 describe('deriveTarget', () => {
     it('returns the internal open PR bound to the trusted head', async () => {
-        await expect(derive([pullRequest()])).resolves.toEqual({
+        const pr = pullRequest({
+            number: 7,
+            base: { ref: 'release', sha: 'base456', repo: { id: 1 } },
+            labels: [{ name: 'fast-track' }]
+        });
+        await expect(derive([pr])).resolves.toEqual({
             prNumber: 7,
             headSha: HEAD_SHA,
-            baseRef: 'main',
-            baseSha: 'base123',
-            labels: []
+            baseRef: 'release',
+            baseSha: 'base456',
+            labels: ['fast-track']
         });
     });
 
     it('refuses to guess when multiple internal PRs share the trusted head', async () => {
-        await expect(derive([pullRequest(), pullRequest({ number: 8 })])).resolves.toBeNull();
+        await expect(
+            derive([pullRequest({ number: 7 }), pullRequest({ number: 8 })])
+        ).resolves.toBeNull();
     });
 
     it('walks every page so a second candidate cannot hide behind pagination', async () => {
-        const octokit = fakeOctokit([pullRequest(), pullRequest({ number: 8 })]);
+        const octokit = fakeOctokit([pullRequest({ number: 7 }), pullRequest({ number: 8 })]);
         await expect(
             deriveTarget({
                 octokit,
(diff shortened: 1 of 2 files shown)


fast_track/src/bot/derive-target.ts
```diff
@@ -0,0 +1,98 @@
...
+
+/** Reload mutable PR state before the bot writes an approval or comment. */
+export async function refreshTarget(
+    params: Omit<DeriveTargetParams, 'trustedHeadSha'> & { target: BotTarget }
+): Promise<BotTarget | null> {
+    const { data: pullRequest } = await params.octokit.rest.pulls.get({
+        owner: params.owner,
+        repo: params.repo,
+        pull_number: params.target.prNumber
+    });
```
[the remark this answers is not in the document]
michal-lightly: Writes are bound to the derived `prNumber` and cross-checked against the verdict's `pr_number`/`head_sha`/`base_ref`/`base_sha`, so a second PR at the same head is never the one approved. `refreshTarget` also isn't called yet. A write-time uniqueness re-check belongs in the orchestrator in a follow-up (`refreshTarget` → `deriveTarget`), not here.

fast_track/src/bot/derive-target.test.ts
```diff
@@ -0,0 +1,121 @@
...
+        owner: 'lightly-ai',
+        repo: 'lightly-studio',
+        trustedHeadSha: HEAD_SHA
+    });
+}
+
+describe('deriveTarget', () => {
+    it('returns the internal open PR bound to the trusted head', async () => {
+        await expect(derive([pullRequest()])).resolves.toEqual({
+            prNumber: 7,
```
lukas-lightly, line 52: The values you are asserting shouldn't be implicit, but explicit, see the Google blog post on DRY vs. DAMP tests: https://testing.googleblog.com/2019/12/testing-on-toilet-tests-too-dry-make.html

lukas-lightly: E.g. you could make it explicit with `pullRequest({number: 7})` and then also expect number 7. Expecting HEAD_SHA is sort of fine, as it's a defined constant. But then `baseRef` and `baseSha` are implicit and you shouldn't assert them.

fast_track/src/bot/derive-target.test.ts
```diff
@@ -0,0 +1,121 @@
...
+            headSha: HEAD_SHA,
+            baseRef: 'main',
+            baseSha: 'base123',
+            labels: []
+        });
+    });
+
+    it('refuses to guess when multiple internal PRs share the trusted head', async () => {
+        await expect(derive([pullRequest(), pullRequest({ number: 8 })])).resolves.toBeNull();
```
lukas-lightly: It would be more clear like this:
```suggestion
        await expect(derive([pullRequest({ number: 7} ), pullRequest({ number: 8 })])).resolves.toBeNull();
```

It's immediately clear that those are two different numbers.

lukas-lightly (approved):

PR #1691, opened 2026-07-21
Tolerate broken files per-item in image embedding

JonasWurst:
## Summary

Route the image-embedding file-loading path through the centralized `FileOutcomeReport` so an unreadable/undecodable file is **skipped per-item** instead of aborting the whole embedding run.

- `embed_image_files_batched` catches `(OSError, UnidentifiedImageError)` per file, drops the item **before batching** (so batches stay full-sized and no empty batch reaches `torch.stack`), and returns an `ImageEmbeddingResult` carrying `kept_indices` so callers can realign parallel per-file data. Outcomes are recorded in a `FileOutcomeReport`; when every attempted file is broken it raises `AllInputFilesFailedError`.
- `ImageEmbeddingResult` is propagated through the `ImageEmbeddingGenerator` protocol and its implementations (mobileclip, perception_encoder, `RandomEmbeddingGenerator`, custom-model example).
- `embedding_manager.embed_images` filters `sample_ids` by `kept_indices` before storing, keeping sample IDs and embeddings in sync. `compute_image_embedding` relies on the all-failed raise for a single broken file.

The in-memory PIL path (`embed_pil_images_batched`) never skips and keeps its 1:1 array return.

## Scope

This covers the **full-image embedding path only** (the originally-reported unhandled file open). The crop path (`image_crop_embedding.py`) is not included here.

## Testing

- `make static-checks` — ruff + format clean; mypy clean on all touched files (pre-existing unrelated `s3fs` stub errors aside).
- Added tests: mixed good/broken/missing files with `kept_indices` assertions, all-broken raises, single-image broken; updated existing tests for the new return type.

## Summary by CodeRabbit

* **New Features**
  * Image embedding outputs now include both the generated embeddings and the positions of successfully processed files.
  * Unreadable or invalid images are skipped while valid images continue processing.
  * If all supplied images fail, a clear error is raised.
* **Updates**
  * MobileCLIP, perception-encoder, and custom embedding generators now return the structured embedding result.
  * Dataset embedding storage now preserves correct sample-to-embedding associations when files are skipped.
* **Tests**
  * Updated the embedding-related test suite to validate the new result format and broken-file behavior.

### DIFF
--- a/lightly_studio/src/lightly_studio/dataset/image_embedding.py
+++ b/lightly_studio/src/lightly_studio/dataset/image_embedding.py
@@ -9,7 +9,7 @@ crop-specific path lives in ``image_crop_embedding.py`` and reuses the same
 from __future__ import annotations
 
 import os
-from collections.abc import Callable, Iterable, Sequence
+from collections.abc import Callable, Iterable, Iterator, Sequence
 from dataclasses import dataclass
 from typing import TypeVar
 
@@ -17,9 +17,10 @@ import fsspec
 import numpy as np
 import torch
 from numpy.typing import NDArray
-from PIL import Image
+from PIL import Image, UnidentifiedImageError
 from tqdm import tqdm
 
+from lightly_studio.core.file_outcome_report import FileOutcome, FileOutcomeReport
 from lightly_studio.utils import batching, parallelize
 
 _ItemT = TypeVar("_ItemT")
@@ -52,12 +53,31 @@ class _EmbeddingProgress:
     unit: str
 
 
+@dataclass(frozen=True)
+class ImageEmbeddingResult:
+    """Embeddings for the image files that could be read, plus which inputs they cover.
+
+    Broken files (unreadable/undecodable) are skipped rather than aborting the whole
+    run, so ``embeddings`` may hold fewer rows than the input file list. ``kept_indices``
+    maps each row back to its position in the input list, in input order, letting callers
+    realign any parallel per-file data (e.g. sample IDs) with the embeddings.
+
+    Attributes:
+        embeddings: Float32 array of shape ``(len(kept_indices), embedding_dimension)``.
+        kept_indices: Indices into the input file list of the files that were embedded,
+            in input order.
+    """
+
+    embeddings: NDArray[np.float32]
+    kept_indices: list[int]
+
+
 def embed_image_files_batched(
     filepaths: list[str],
     context: EmbeddingContext,
     show_progress: bool,
-) -> NDArray[np.float32]:
-    """Embed image files in batches, preserving input order.
+) -> ImageEmbeddingResult:
+    """Embed image files in batches, preserving input order and skipping broken files.
 
     Args:
         filepaths: Paths of the images to embed.
@@ -65,21 +85,40 @@ def embed_image_files_batched(
         show_progress: Whether to show a tqdm progress bar.
 
     Returns:
-        Float32 array of shape ``(len(filepaths), embedding_dimension)``.
+        An ``ImageEmbeddingResult`` whose embeddings cover only the readable files, with
+        ``kept_indices`` mapping each row back to its input position.
+
+    Raises:
+        AllInputFilesFailedError: If at least one file was attempted and every attempted
+            file was broken (i.e. no file could be read and embedded).
+        ValueError: If ``context.max_batch_size`` is not positive.
     """
 
-    def load_and_preprocess(filepath: str) -> torch.Tensor:
-        with fsspec.open(filepath, "rb") as file:
-            image = Image.open(file).convert("RGB")
-            return context.preprocess(image)
+    def load_and_preprocess(filepath: str) -> torch.Tensor | None:
+        try:
+            with fsspec.open(filepath, "rb") as file:
+                image = Image.open(file).convert("RGB")
+        except (OSError, UnidentifiedImageError):
+            return None
+        return context.preprocess(image)
 
-    return _embed_items_batched(
+    embeddings, kept_indices = _embed_items_batched(
         items=filepaths,
         preprocess_item=load_and_preprocess,
         context=context,
         show_progress=show_progress,
         progress=_EmbeddingProgress(desc="Generating embeddings", unit=" images"),
     )
+
+    kept = set(kept_indices)
+    report = FileOutcomeReport()
+    for index, filepath in enumerate(filepaths):
+        outcome = FileOutcome.ADDED if index in kept else FileOutcome.BROKEN
+        report.record(path=filepath, outcome=outcome)
+    report.raise_if_all_failed()
+    report.log_summary()
+
+    return ImageEmbeddingResult(embeddings=embeddings, kept_indices=kept_indices)
 
 
 def embed_pil_images_batched(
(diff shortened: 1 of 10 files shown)


lightly_studio/src/lightly_studio/dataset/image_embedding.py
```diff
@@ -52,35 +53,73 @@ class _EmbeddingProgress:
...
-    def load_and_preprocess(filepath: str) -> torch.Tensor:
-        with fsspec.open(filepath, "rb") as file:
-            image = Image.open(file).convert("RGB")
-            return context.preprocess(image)
+    def load_and_preprocess(filepath: str) -> torch.Tensor | None:
+        try:
+            with fsspec.open(filepath, "rb") as file:
+                image = Image.open(file).convert("RGB")
+        except (OSError, UnidentifiedImageError):
+            return None
```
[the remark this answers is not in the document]
JonasWurst: The existence of the file was checked in add_image. here it is sufficient to just catch all of them as broken. The main purpose is to not stop the embedding.

lightly_studio/src/lightly_studio/dataset/image_embedding.py
```diff
@@ -52,35 +53,73 @@ class _EmbeddingProgress:
...
         progress=_EmbeddingProgress(desc="Generating embeddings", unit=" images"),
     )
 
+    kept = set(kept_indices)
+    report = FileOutcomeReport()
+    for index, filepath in enumerate(filepaths):
+        outcome = FileOutcome.ADDED if index in kept else FileOutcome.BROKEN
+        report.record(path=filepath, outcome=outcome)
+    report.raise_if_all_failed()
+    report.log_summary()
```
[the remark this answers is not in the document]
JonasWurst: This is only used in this place, so we'll keep it as is for now.

lightly_studio/src/lightly_studio/dataset/image_embedding.py
```diff
@@ -52,35 +53,73 @@ class _EmbeddingProgress:
...
+    """Embeddings for the image files that could be read, plus which inputs they cover.
+
+    Broken files (unreadable/undecodable) are skipped rather than aborting the whole
+    run, so ``embeddings`` may hold fewer rows than the input file list. ``kept_indices``
+    maps each row back to its position in the input list, in input order, letting callers
+    realign any parallel per-file data (e.g. sample IDs) with the embeddings.
+
+    Attributes:
+        embeddings: Float32 array of shape ``(len(kept_indices), embedding_dimension)``.
+        kept_indices: Indices into the input file list of the files that were embedded,
```
michal-lightly, line 67: Why not sample_ids?

lightly_studio/src/lightly_studio/dataset/image_embedding.py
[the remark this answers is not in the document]
michal-lightly: Can we keep record of the error and report it to the user (sth like "x skiped because xyz")? It would be useful to know why a file can't be read (broken connection? insufficient rights? deleted after indexing? corrupt file? out of disk space?). Maybe for a follow-up.

lightly_studio/src/lightly_studio/dataset/image_embedding.py
```diff
@@ -97,33 +136,40 @@ def embed_pil_images_batched(
...
 
 def _embed_items_batched(
     items: Sequence[_ItemT],
-    preprocess_item: Callable[[_ItemT], torch.Tensor],
+    preprocess_item: Callable[[_ItemT], torch.Tensor | None],
     context: EmbeddingContext,
     show_progress: bool,
     progress: _EmbeddingProgress,
-) -> NDArray[np.float32]:
+) -> tuple[NDArray[np.float32], list[int]]:
```
michal-lightly: Use a dataclass instead of an ephemeral type. Index list should be always tied to the data it indexes.

lightly_studio/src/lightly_studio/dataset/image_embedding.py
```diff
@@ -52,35 +53,73 @@ class _EmbeddingProgress:
...
-    return _embed_items_batched(
+    embeddings, kept_indices = _embed_items_batched(
         items=filepaths,
         preprocess_item=load_and_preprocess,
         context=context,
         show_progress=show_progress,
         progress=_EmbeddingProgress(desc="Generating embeddings", unit=" images"),
     )
 
+    kept = set(kept_indices)
```
michal-lightly: Optional
```suggestion
    kept_indices_set = set(kept_indices)
```

lightly_studio/src/lightly_studio/dataset/image_embedding.py
```diff
@@ -136,13 +182,29 @@ def _embed_items_batched(
         # while memory stays bounded to a small multiple of the batch size.
         buffer_size=2 * context.max_batch_size,
     )
-    return _encode_preprocessed_batches(
-        preprocessed_tensors=preprocessed_tensors,
+    # Drop skipped items before batching
+    kept_indices: list[int] = []
+    kept_tensors = _keep_non_none(preprocessed_tensors, kept_indices=kept_indices)
```
michal-lightly:
```suggestion
    kept_tensors_iter = _keep_non_none(tensors=preprocessed_tensors, kept_indices=kept_indices)
```

lightly_studio/src/lightly_studio/dataset/image_embedding.py
```diff
@@ -136,13 +182,29 @@ def _embed_items_batched(
...
         context=context,
         show_progress=show_progress,
         progress=progress,
     )
+    return embeddings[: len(kept_indices)], kept_indices
+
+
+def _keep_non_none(
+    tensors: Iterable[torch.Tensor | None],
+    kept_indices: list[int],
```
michal-lightly: It would be better to name this differently to explicitly mark it as an output value. Please also make the docstring more verbose.

```suggestion
    out_kept_indices: list[int],
```

lightly_studio/src/lightly_studio/dataset/image_embedding.py
michal-lightly: Rename to indicate this is an upper bound.

Optional: I would suggest to do the array truncation in this function.

lightly_studio/tests/dataset/test_image_embedding.py
```diff
@@ -6,37 +6,35 @@
...
-    embeddings = image_embedding.embed_image_files_batched(
+    result = image_embedding.embed_image_files_batched(
         filepaths=[],
         context=EmbeddingContext(
             embedding_dimension=4,
             max_batch_size=2,
             device=torch.device("cpu"),
             preprocess=lambda image: torch.tensor([float(image.size[0])]),
-            encode_batch=lambda images_tensor: images_tensor.cpu().numpy(),
+            encode_batch=lambda images_tensor: images_tensor.numpy().astype(np.float32),
```
michal-lightly: Why the change here?

lightly_studio/tests/dataset/test_image_embedding.py
```diff
@@ -6,37 +6,35 @@
...
 def test_embed_image_files_batched__preserves_input_order(tmp_path: Path) -> None:
     # Each image has a distinct width, and preprocess encodes an image as its
     # width, so the embeddings must come back in input order across batches.
     widths = [5, 6, 7]
-    filepaths = []
-    for index, width in enumerate(widths):
-        path = tmp_path / f"image_{index}.png"
-        Image.new("RGB", (width, 10), color=(255, 0, 0)).save(path)
-        filepaths.append(str(path))
+    filepaths = _write_images(tmp_path, widths)
```
michal-lightly: Let's use keyword arguments

lightly_studio/tests/dataset/test_image_embedding.py
```diff
@@ -84,7 +79,55 @@ def preprocess(image: Image.Image) -> torch.Tensor:
...
 
-    assert embeddings[:, 0].tolist() == [float(width) for width in widths]
+    assert result.embeddings[:, 0].tolist() == [float(width) for width in widths]
+
+
+def test_embed_image_files_batched__skips_broken_files(tmp_path: Path) -> None:
+    # Broken files are skipped per-item: the embeddings cover only the readable files and
+    # kept_indices maps each row back to its input position, so callers stay in sync.
+    widths = [5, 6, 7, 8]
+    filepaths = _write_images(tmp_path, widths)
```
michal-lightly: Optional: Call out that these filepaths are the good ones:

```suggestion
    valid_filepaths = _write_images(tmp_path, widths)
```

michal-lightly (changes requested):

lightly_studio/src/lightly_studio/dataset/image_embedding.py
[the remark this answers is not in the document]
JonasWurst: The follow up: the embedding will happen during the image indexing. Hence also the loading will happen in the indexing loop.

lightly_studio/src/lightly_studio/dataset/image_embedding.py
JonasWurst: This would require more changes in the interface. Decision: keep the minimal changes and stick with the index

lightly_studio/src/lightly_studio/dataset/image_embedding.py
JonasWurst: Done in be51281 — `_embed_items_batched` now returns `ImageEmbeddingResult` instead of a bare tuple, so the kept indices always travel with the embeddings they index.

lightly_studio/src/lightly_studio/dataset/image_embedding.py
JonasWurst: Done — renamed to `kept_indices_set`.

lightly_studio/src/lightly_studio/dataset/image_embedding.py
JonasWurst: Done — renamed to `kept_tensors_iter` and switched to keyword arguments.

lightly_studio/src/lightly_studio/dataset/image_embedding.py
JonasWurst: Done — renamed to `out_kept_indices` and expanded the docstring to describe the in-place output contract (empty until iteration starts, complete once the iterator is exhausted).

lightly_studio/src/lightly_studio/dataset/image_embedding.py
JonasWurst: Done — renamed to `max_items` to mark it as an upper bound, and moved the array truncation into `_encode_preprocessed_batches`.

lightly_studio/tests/dataset/test_image_embedding.py
JonasWurst: Unnecessary churn — reverted to `.cpu().numpy()` to match the PIL empty-input test (encode_batch is never called for empty input anyway).

lightly_studio/tests/dataset/test_image_embedding.py
JonasWurst: Done — using keyword arguments for `_write_images` calls.

lightly_studio/tests/dataset/test_image_embedding.py
JonasWurst: Done — renamed to `valid_filepaths`.

michal-lightly (approved):

PR #1704, opened 2026-07-21
Display classification annotations in the annotations grid

LeonardoRosaa:
## What has changed and why?

Classification annotations were previously filtered out of the annotations grid. This PR removes that filter and renders classification annotations using the existing `AnnotationClassificationGridItem` component.

Screenshot 2026-07-21 at 08 03 44

## How has it been tested?

Unit tests

Manual test:

1. Add any classification annotation.
2. Go to the annotation grid - the classification annotation should be displayed.

- [x] Yes

## Summary by CodeRabbit

- **New Features**
  - Classification annotations are now shown in the annotations grid with dedicated classification tiles.
  - Classification tiles support selection, drag-and-drop into image-based search, and bulk relabeling with undo.

- **Bug Fixes**
  - Classification label pills now render consistently in tiles, while still honoring the active collection filter.

- **Tests**
  - Added/updated Vitest coverage for grid rendering, tile interactions, classification drag payloads, and bulk relabeling/undo flows.


lightly_studio_view/src/lib/components/SampleClassificationPills/SampleClassificationPills.svelte
```diff
@@ -10,15 +10,24 @@
...
         sample: Pick<ImageView, 'annotations'>;
         hasBottomOverlay?: boolean;
         hasRightOverlay?: boolean;
-    } = $props();
+        /**
+         * When provided, overrides the collection filter from `useAnnotationCollectionsFilter`.
+         * Pass `[]` to show all annotations regardless of the active source filter (e.g. in the
+         * annotations grid, where the tile already represents a single annotation).
+         */
+        selectedCollectionIds?: string[];
```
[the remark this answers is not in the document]
LeonardoRosaa: I wouldn’t split it. Most of the added lines are comments

lightly_studio_view/src/lib/components/AnnotationsGrid/AnnotationsGrid.svelte
```diff
@@ -356,17 +348,31 @@
...
-                                            cachedCollectionVersion={collectionVersion}
-                                            showLabel={showLabels}
-                                            selected={$pickedAnnotationIds[collection_id]?.has(
-                                                annotations[index].annotation.sample_id
-                                            )}
-                                            onCropWindowChange={handleCropWindowChange}
-                                        />
+                                        {#if isClassification}
+                                            <!-- One classification annotation = one tile (1:1 mapping, same as OD/seg). -->
+                                            <AnnotationClassificationGridItem
```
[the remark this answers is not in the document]
LeonardoRosaa: Discussed internally. We're going to address this in a follow-up issue.

ikondrat (commented):
The rendering branch looks reasonable, but several tests currently mock away the behavior they claim to cover. The inline comments focus on making those contracts observable without broadening the production change.

lightly_studio_view/src/lib/components/AnnotationsGrid/AnnotationsGrid.helpers.ts
```diff
@@ -24,3 +30,23 @@ export function buildAnnotationDragData(
...
+export function buildClassificationDragData({
+    annotation,
+    cropWindow,
+    cropUrl
+}: AnnotationDragDataParams): GridItemDragData | undefined {
+    if (!cropWindow) return undefined;
+    return {
+        url: cropUrl ?? cropWindow.sourceUrl,
+        fileName: `${annotation.annotation_label.annotation_label_name}-crop.png`
+    };
```
ikondrat, line 51: **P2: Add a focused test for this payload.** The classification path depends on omitting `annotationSampleId`; otherwise the drop handler attempts to read a nonexistent annotation embedding. `AnnotationsGrid.test.ts` replaces `GridItem` with a mock that ignores `dragData`, so the current suite does not exercise this contract. Please assert the URL/file name, absence of annotation IDs, and the `undefined` result without a crop window.

lightly_studio_view/src/lib/components/AnnotationsGrid/AnnotationClassificationGridItem/AnnotationClassificationGridItem.svelte
```diff
@@ -73,8 +70,13 @@
...
-        <SampleClassificationPills sample={{ annotations: [annotation.annotation] }} />
-    {/if}
+    <!-- One tile shows exactly one label — [annotation.annotation] wraps a single classification. -->
+    <!-- selectedCollectionIds=[] bypasses the images-grid source filter; the tile already
+         represents one annotation so no further filtering is needed. -->
+    <!-- The badge is always visible: for classification there is no bounding-box equivalent,
+         so the badge is the primary visual indicator regardless of the showLabel setting. -->
+    <SampleClassificationPills
+        sample={{ annotations: [annotation.annotation] }}
+        selectedCollectionIds={[]}
```
ikondrat, line 75: **P2: Cover this filter override instead of mocking it away.** `SampleClassificationPills.mock.svelte` ignores `selectedCollectionIds` and always renders, so the tests would still pass if this prop were removed and the real pills were filtered out. Please expose this prop from the mock and assert that `[]` is forwarded, or exercise the real child with a non-matching active source filter.

lightly_studio_view/src/lib/components/AnnotationsGrid/AnnotationsGrid.test.ts
```diff
@@ -0,0 +1,324 @@
...
+        mocks.annotationsData = [buildClassificationAnnotation('cls-1')];
+        mocks.pickedAnnotationIds.set({ 'col-1': new Set(['cls-1']) });
+        mocks.isEditingModeStore.set(true);
+
+        renderGrid();
+
+        const button = screen.getByTestId('mock-select-label');
+        await fireEvent.click(button);
+
+        expect(mocks.addAnnotationLabelChangeToUndoStackFn).toHaveBeenCalledOnce();
```
ikondrat: **P2: Test the undo behavior rather than the mocked delegation.** This test is named as if it verifies the undo stack, but `addAnnotationLabelChangeToUndoStack` is replaced with a spy and only that spy call is asserted. Keep the real helper and assert the resulting `addReversibleAction` payload, or rename the test to state that it only verifies delegation.

lightly_studio_view/src/lib/components/AnnotationsGrid/AnnotationsGrid.test.ts
```diff
@@ -0,0 +1,324 @@
...
+        expect(tiles[1]).toHaveAttribute('data-annotation-id', 'cls-2');
+    });
+
+    it('calls toggleSampleAnnotationCropSelection when a classification tile is clicked', () => {
+        mocks.annotationsData = [buildClassificationAnnotation('cls-1')];
+
+        renderGrid();
+
+        const gridItem = screen.getByTestId('annotation-grid-item');
+        fireEvent.click(gridItem);
```
ikondrat: **P2: Await this Svelte event.** `@testing-library/svelte` wraps `fireEvent.click` in an async `act`, so leaving the promise unawaited can make this test race component updates or cleanup. Mark the test `async` and use `await fireEvent.click(gridItem)`.

lightly_studio_view/src/lib/components/AnnotationsGrid/AnnotationClassificationGridItem/AnnotationClassificationGridItem.svelte
```diff
@@ -73,8 +70,13 @@
...
     style="width: {containerWidth}px; height: {containerHeight}px; background-image: url('{thumbnailUrl}'); background-size: cover; background-position: center;"
 >
-    {#if showLabel}
-        <!-- One tile shows exactly one label — [annotation.annotation] wraps a single classification. -->
-        <SampleClassificationPills sample={{ annotations: [annotation.annotation] }} />
-    {/if}
+    <!-- One tile shows exactly one label — [annotation.annotation] wraps a single classification. -->
+    <!-- selectedCollectionIds=[] bypasses the images-grid source filter; the tile already
+         represents one annotation so no further filtering is needed. -->
+    <!-- The badge is always visible: for classification there is no bounding-box equivalent,
```
ikondrat: I would avoid describing the SampleClassificationPills logic on usage level, either remove comments or add logic description to SampleClassificationPills (if it is required)

ikondrat (approved):

PR #1721, opened 2026-07-22
BE: add 'in' operator to MetadataFilter for categorical filtering

ikondrat:
## Summary

Stacked on p1.

- Adds an `in` operator to `MetadataFilter` that accepts a list of `str | bool | None` values
- Builds an `OR` predicate for non-null values using a standard SQL `IN` clause
- When `None` is in the list, uses a `LEFT OUTER JOIN` so that samples with no metadata entry for the field are also included
- Validates that the list is non-empty and that all values share the same type (prevents mixing strings with booleans or numbers)

## Test plan

- [ ] `make static-checks` passes
- [ ] New unit tests in `tests/resolvers/metadata_resolvers/test_metadata_filter.py` and `tests/api/routes/api/test_metadata.py`
- [ ] `tests/sample/test_sample_metadata_filtering.py` covers round-trip filtering with null and non-null values

Part of LIG-9585.

### DIFF
--- a/lightly_studio/src/lightly_studio/resolvers/metadata_resolver/metadata_filter.py
+++ b/lightly_studio/src/lightly_studio/resolvers/metadata_resolver/metadata_filter.py
@@ -3,7 +3,9 @@
 import operator
 from typing import Any, Callable, Literal, Protocol, TypeVar
 
-from pydantic import BaseModel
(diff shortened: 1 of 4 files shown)


ikondrat:
Superseded by a finer-grained stack (p0–p15).

PR #1727, opened 2026-07-22
FE: filter-aware bar colours and dual-series background bars

ikondrat:
## Summary

Stacked on p7. Final visual layer — makes the distribution charts aware of active sidebar filters.

**Dim unselected bars** (`buildEchartsOption.ts`): when the user has selected one or more categorical values, all other bars render in grey (`#4b5563`), matching the numeric Histogram behaviour. The white border on selected bars is removed because the green/grey contrast already communicates selection clearly.

**Keep full distribution visible** (`+layout.svelte`): introduces `distributionBaseFilter` — a derived filter that strips the analysis filters (metadata, annotation, dimension) while keeping the collection-scoping context (tagIds, sampleIds, etc.). Both distribution queries now use this base filter so bar heights always reflect the full dataset, not the currently filtered subset.

**Dual-series background bars** (`buildEchartsOption.ts`, `BarChart/types.ts`):
- `CategoryCount` gains an optional `filteredCount` field
- When the layout fetches a second distribution using the full `imageAnnotationCountsFilter` (including sidebar filters), it passes the filtered counts to each categorical group
- If any bar's `filteredCount` differs from its full count, the chart renders two ECharts series: a grey background series at full height and a green/grey foreground series at filtered height
- Tooltip updates to "Total / In filter" when the filter actually reduces any bar; falls back to the single "Count" line otherwise
- When no sidebar filter is active the second query returns identical counts, `hasActiveFilter` is false, and a single series is rendered — no visual change from the unfiltered state

## Test plan

- [ ] `make static-checks` (frontend) passes (includes Prettier fix for the ternary in `buildEchartsOption`)
- [ ] `npm run test:unit` passes (new tests in `buildEchartsOption.test.ts` for the dual-series logic)
- [ ] Manual: apply an annotation-class filter from the left sidebar, verify categorical bars show grey background at full height and green foreground at filtered height; clear the filter, verify bars return to solid colour

Part of LIG-9585.

### DIFF
--- a/lightly_studio_view/src/lib/components/BarChart/buildEchartsOption.test.ts
+++ b/lightly_studio_view/src/lib/components/BarChart/buildEchartsOption.test.ts
@@ -41,22 +41,55 @@ describe('buildEchartsOption', () => {
         expect(horizontal.xAxis.minInterval).toBe(1);
     });
 
-    it('adds selected and disabled treatments without changing typed category identity', () => {
+    it('dims unselected bars green when a selection is active; disabled bars keep reduced opacity', () => {
         const option = buildEchartsOption([
-            { id: 'selected', label: 'Missing', count: 3, selected: true, selectable: true },
+            { id: 'sel', label: 'Missing', count: 3, selected: true, selectable: true },
             { id: 'other', label: 'Other', count: 2, selected: false, selectable: false }
         ]) as {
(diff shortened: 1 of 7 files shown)


ikondrat:
Superseded by a finer-grained stack (p0–p15).

PR #1729, opened 2026-07-22
Apply the label color opacity to segmentation masks in the detail view

sam-watts:
## Problem

In `SampleAnnotation.svelte`, `segmentationMaskOpacity` returns a fixed `0.65` for any annotation carrying a segmentation mask, ignoring the label's custom alpha — while the bounding-box branch immediately beside it applies that alpha:

```ts
return segmentationMask ? 0.65 : ($customLabelColorsStore[colorLabel]?.alpha ?? 1.0) * 0.6;
```

The grid view's canvas renderer *does* apply the alpha, so the Color picker's **Opacity** slider works in the grid but has no effect in the sample detail view. The two views disagree about the same setting.

For a dataset whose annotations are all segmentation masks, the slider is inert in the one view where the mask sits directly on top of the pixels being inspected, and there's no other way to fade it. The per-annotation eye toggle resets on every sample, and `key_hide_annotations` is hold-to-hide.

We hit this labelling aerial imagery, where the mask covers the roof being judged.

## Fix

Read the alpha once and multiply both branches by it, keeping the existing `0.65` base constant:

```ts
const alpha = $customLabelColorsStore[colorLabel]?.alpha ?? 1.0;

return segmentationMask ? alpha * 0.65 : alpha * 0.6;
```

**No behaviour change by default:** with no custom color the alpha falls back to `1.0`, so the rendered opacity stays exactly `0.65`. Setting the alpha now scales the mask, and an alpha of `0` hides it.

## Tests

Three cases added to the existing `opacity when custom label color is absent` block in `SampleAnnotation.test.ts`, using a new factory that carries an actual `segmentation_mask` (the existing one has none, so no mask image is rendered for it):

- default alpha → opacity stays `0.65` (guards the no-change-by-default claim)
- custom alpha `0.8` → `0.52`
- custom alpha `0` → mask hidden

Verified the two alpha-dependent tests fail against the unpatched component (`expected 0.65 to be close to 0.52`, `expected 0.65 to be +0`) and pass with the fix. Full file: 15 passed. `prettier --check` and `eslint` clean on the changed directory.

Also verified in a browser against a real dataset: setting a layer to 0% now yields `opacity="0"` on the mask `<image>`, where it was previously pinned at `0.65` regardless of the slider.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected segmentation mask opacity so custom label transparency settings are consistently applied.
  * Segmentation masks now hide correctly when their configured transparency is set to zero.

* **Tests**
  * Added coverage for default, customized, and fully transparent segmentation mask opacity behavior.


sam-watts:
TLDR: Not sure if existing behaviour on main is intended? this PR just means that segmentation masks in the detail page now respect alpha settings, which makes additional labelling far easier

michal-lightly:
Thanks @sam-watts for the PR, we'll move it through our review process.

michal-lightly:
Hi @sam-watts , we have merged your change, you will see it in the next release. Thanks!

PR #1749, opened 2026-07-22
BE: add categorical metadata value-counts endpoint

ikondrat:
## Summary

Stacked on #1750.

Adds the `POST /metadata/value-counts` endpoint. Dedicated resolver, database, and API coverage follows in #1743.

## Validation

- `uv run ruff check …`

PR #1742, opened 2026-07-22
BE: add categorical metadata value primitives

ikondrat:
## Summary

First PR in the categorical metadata value-count stack.

Adds the response models and database JSON scalar extraction needed to query categorical metadata values.

## Validation

- `uv run ruff check …`
- Focused metadata and API tests

PR #1750, opened 2026-07-22
BE: add categorical metadata value-count resolver

ikondrat:
## Summary

Stacked on #1742.

Adds the categorical metadata value-count resolver and extracts shared metadata query helpers used by histogram and categorical resolvers.

Test will come with https://github.com/lightly-ai/lightly-studio/pull/1749

## Design Overview

The endpoint returns value counts for categorical metadata fields used to render bar charts in the distribution panel.

**Flow:**

1. UI calls `GET /metadata/info` to discover all metadata fields and their types
2. UI determines which categorical fields (`string`, `boolean`) to render
3. UI calls `POST /metadata/value-counts` with `fields` set to exactly those fields
4. Backend counts only the requested fields — cost is O(fields_shown × N_samples), not O(all_fields × N_samples)
5. UI renders one bar chart per field from the response

**Why `fields` is in the request body:**

Without it the endpoint would scan the full dataset for every categorical field on every filter change, regardless of what is displayed. Passing `fields` explicitly makes the caller responsible for what it needs, keeping the backend cost proportional to what is rendered. At 1M samples × 20 fields this is the difference between ~2 queries and ~40 queries per filter interaction.

**Faceted search behaviour:**

Each field's bar chart must show counts as if that field's own filter were not active — otherwise the user can only see values they have already selected, not the full set to choose from. The resolver handles this by stripping each field's own filter condition when computing that field's counts, while all other active filters still apply.

**Example request:**

```json
POST /api/collections/{id}/metadata/value-counts
{
  "fields": ["city", "active"],
  "filters": { "sample_filter": { "metadata_filters": [{ "key": "city", "op": "in", "value": ["Zurich"] }] } }
}
```

## Validation

- `uv run ruff check …`
- Focused metadata and API tests

## Summary by CodeRabbit

- **New Features**
  - Added categorical metadata value counts for collections, returning only the top values (removed missing/other totals).
  - Added the ability to restrict results to selected metadata fields.
- **Bug Fixes**
  - Improved metadata filtering so a field’s own filter no longer affects its counts, while other filters remain applied.
- **Tests**
  - Added comprehensive coverage for value counting, top-N limits, filter behavior, field selection, and empty/unknown-collection cases.

### DIFF
--- a/lightly_studio/src/lightly_studio/resolvers/metadata_resolver/__init__.py
+++ b/lightly_studio/src/lightly_studio/resolvers/metadata_resolver/__init__.py
@@ -1,15 +1,9 @@
 """Metadata resolver module."""
 
-from lightly_studio.resolvers.metadata_resolver.sample.bulk_update_metadata import (
+from lightly_studio.resolvers.metadata_resolver.sample import (
     bulk_update_metadata,
-)
-from lightly_studio.resolvers.metadata_resolver.sample.get_by_sample_id import (
     get_by_sample_id,
-)
-from lightly_studio.resolvers.metadata_resolver.sample.get_value_for_sample import (
     get_value_for_sample,
-)
-from lightly_studio.resolvers.metadata_resolver.sample.set_value_for_sample import (
     set_value_for_sample,
 )
 
--- a/lightly_studio/src/lightly_studio/resolvers/metadata_resolver/__init__.py
+++ b/lightly_studio/src/lightly_studio/resolvers/metadata_resolver/__init__.py
@@ -1,9 +1,15 @@
 """Metadata resolver module."""
 
-from lightly_studio.resolvers.metadata_resolver.sample import (
+from lightly_studio.resolvers.metadata_resolver.sample.bulk_update_metadata import (
     bulk_update_metadata,
+)
+from lightly_studio.resolvers.metadata_resolver.sample.get_by_sample_id import (
     get_by_sample_id,
+)
+from lightly_studio.resolvers.metadata_resolver.sample.get_value_for_sample import (
     get_value_for_sample,
+)
+from lightly_studio.resolvers.metadata_resolver.sample.set_value_for_sample import (
     set_value_for_sample,
 )
 
--- a/lightly_studio/src/lightly_studio/resolvers/metadata_resolver/sample/__init__.py
+++ b/lightly_studio/src/lightly_studio/resolvers/metadata_resolver/sample/__init__.py
@@ -6,9 +6,6 @@ from .bulk_update_metadata import (
 from .get_by_sample_id import (
     get_by_sample_id,
 )
-from .get_metadata_info import (
-    get_all_metadata_keys_and_schema,
-)
 from .get_metadata_values_for_key import (
     get_metadata_values_for_key,
 )
@@ -21,7 +18,6 @@ from .set_value_for_sample import (
 
 __all__ = [
     "bulk_update_metadata",
-    "get_all_metadata_keys_and_schema",
     "get_by_sample_id",
     "get_metadata_values_for_key",
     "get_value_for_sample",
--- a/lightly_studio/src/lightly_studio/resolvers/metadata_resolver/sample/categorical_value_counts.py
+++ b/lightly_studio/src/lightly_studio/resolvers/metadata_resolver/sample/categorical_value_counts.py
@@ -1,0 +1,152 @@
+"""Resolver for categorical sample metadata value counts."""
+
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+from uuid import UUID
+
+from sqlalchemy import func
+from sqlalchemy.sql.elements import ColumnElement
+from sqlmodel import Session, col, select
+
+from lightly_studio.database import db_json
+from lightly_studio.models.metadata import (
+    MetadataValueCountsView,
+    MetadataValueCountView,
+    SampleMetadataTable,
+)
+from lightly_studio.models.sample import SampleTable
+from lightly_studio.resolvers.metadata_resolver.sample import metadata_helpers
+
+if TYPE_CHECKING:
+    from lightly_studio.resolvers.image_filter import ImageFilter
+
+_CATEGORICAL_TYPES = ("string", "boolean")
+_TOP_VALUE_COUNT = 20
+
+
+def get_metadata_value_counts(
+    session: Session,
+    collection_id: UUID,
+    filters: ImageFilter | None = None,
+) -> dict[str, MetadataValueCountsView]:
+    """Count categorical metadata values for a collection.
+
+    Each field's own metadata filter is excluded while all other filters apply.
+    Results contain the 20 most frequent concrete values, with the remaining
+    concrete and missing values counted separately.
+
+    Args:
+        session: The database session.
+        collection_id: The collection whose sample metadata is aggregated.
+        filters: Optional image filters restricting the counted samples.
+
+    Returns:
+        A mapping from categorical metadata keys to their value counts.
+    """
+    schema = metadata_helpers.get_merged_schema(session=session, collection_id=collection_id)
+    result: dict[str, MetadataValueCountsView] = {}
+    for key, metadata_type in schema.items():
+        if metadata_type not in _CATEGORICAL_TYPES:
+            continue
+        field_filters = metadata_helpers.without_metadata_key_filter(
+            filters=filters, metadata_key=key
+        )
+        result[key] = _get_field_value_counts(
+            session=session,
+            collection_id=collection_id,
+            metadata_key=key,
+            metadata_type=metadata_type,
+            filters=field_filters,
+        )
+    return result
+
+
+def _get_field_value_counts(
+    session: Session,
+    collection_id: UUID,
+    metadata_key: str,
+    metadata_type: str,
+    filters: ImageFilter | None,
+) -> MetadataValueCountsView:
+    value_expr = db_json.json_extract_string(column=SampleMetadataTable.data, field=metadata_key)
+    total_count, concrete_count = _get_total_and_concrete_counts(
+        session=session,
+        collection_id=collection_id,
+        value_expr=value_expr,
+        filters=filters,
+    )
+    rows = _get_top_value_counts(
+        session=session,
+        collection_id=collection_id,
+        value_expr=value_expr,
+        filters=filters,
+    )
+    value_counts = [
+        MetadataValueCountView(
+            value=_parse_value(value=value, metadata_type=metadata_type), count=int(count)
+        )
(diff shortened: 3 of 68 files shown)


lightly_studio/src/lightly_studio/resolvers/metadata_resolver/sample/categorical_value_counts.py
```diff
@@ -0,0 +1,152 @@
...
+from lightly_studio.resolvers.metadata_resolver.sample import metadata_helpers
+
+if TYPE_CHECKING:
+    from lightly_studio.resolvers.image_filter import ImageFilter
+
+_CATEGORICAL_TYPES = ("string", "boolean")
+_TOP_VALUE_COUNT = 20
+
+
+def get_metadata_value_counts(
```
JonasWurst, line 26: I've already commented on the other PR. I really think counting all metadata even when do not dispaly them will be a huge problem for large datasets. Can you give numbers for a 1M dataset with say 20 metadata fields?

JonasWurst (changes requested):
Before reviewing this in detail, I'd like us to get to a common sense about the general design.

lightly_studio/src/lightly_studio/resolvers/metadata_resolver/sample/categorical_value_counts.py
```diff
@@ -0,0 +1,152 @@
...
+from lightly_studio.database import db_json
+from lightly_studio.models.metadata import (
+    MetadataValueCountsView,
+    MetadataValueCountView,
+    SampleMetadataTable,
+)
+from lightly_studio.models.sample import SampleTable
+from lightly_studio.resolvers.metadata_resolver.sample import metadata_helpers
+
+if TYPE_CHECKING:
```
michal-lightly: Do not use `if TYPE_CHECKING` if possible, it is an antipattern. I spend significant effort removing it from the model, we should avoid introducing it.

The problem is that ImageFilter is not loaded during runtime which causes issues with pydantic and openapi generation.

lightly_studio/src/lightly_studio/resolvers/metadata_resolver/sample/categorical_value_counts.py
ikondrat: We need entire dataset information still to see the diff on the plot 

I will check how it will be have on large dataset and how e can address it

JonasWurst: I'm fine on getting it on the full sample length. My main point is that we do not need to collect metadata counts for meta data that is not viewed atm.

lightly_studio/src/lightly_studio/resolvers/metadata_resolver/sample/categorical_value_counts.py
```diff
@@ -0,0 +1,155 @@
...
+    filters: ImageFilter | None,
+) -> MetadataValueCountsView:
+    value_expr = db_json.json_extract_string(column=SampleMetadataTable.data, field=metadata_key)
+    total_count, concrete_count = _get_total_and_concrete_counts(
+        session=session,
+        collection_id=collection_id,
+        value_expr=value_expr,
+        filters=filters,
+    )
+    rows = _get_top_value_counts(
```
JonasWurst, line 78: I'm wondering if that is a good design decision. I would rather always fetch all values, and let the frontend handle the top value task. Typically, it should not be so many distinct.

lightly_studio/src/lightly_studio/resolvers/metadata_resolver/sample/metadata_helpers.py
JonasWurst, line 1: I assume this is just moved and did not get changed. Did not double check this.

lightly_studio/src/lightly_studio/resolvers/metadata_resolver/__init__.py
JonasWurst, line 1: how is this related?

lightly_studio/src/lightly_studio/resolvers/metadata_resolver/sample/__init__.py
JonasWurst, line 1: why is this changed?

lightly_studio/tests/metadata/test_get_metadata_value_counts.py
```diff
@@ -0,0 +1,258 @@
...
+
+    assert set(counts) == {"city", "active"}
+    assert [(entry.value, entry.count) for entry in counts["city"].value_counts] == [
+        ("Zurich", 2),
+        ("", 1),
+        ("Missing", 1),
+        ("Other", 1),
+    ]
+    assert counts["city"].other_count == 0
+    assert counts["city"].missing_count == 2
```
JonasWurst: why is it two? we only added one image w/o meta no?

lightly_studio/tests/metadata/test_get_metadata_value_counts.py
```diff
@@ -0,0 +1,258 @@
...
+    assert [(entry.value, entry.count) for entry in counts["active"].value_counts] == [
+        (True, 3),
+        (False, 2),
+    ]
+    assert counts["active"].other_count == 0
+    assert counts["active"].missing_count == 2
```
JonasWurst: similar question

lightly_studio/tests/metadata/test_get_metadata_value_counts.py
```diff
@@ -0,0 +1,258 @@
...
+    counts = categorical_value_counts.get_metadata_value_counts(
+        session=db_session, collection_id=collection.collection_id
+    )["category"]
+
+    assert len(counts.value_counts) == 20
+    assert (counts.value_counts[0].value, counts.value_counts[0].count) == ("value-20", 2)
+    assert [entry.value for entry in counts.value_counts[1:]] == [
+        f"value-{index:02d}" for index in range(19)
+    ]
+    assert counts.other_count == 1
```
JonasWurst: what is the other count? how will it be used ion the frontend?

lightly_studio/src/lightly_studio/resolvers/metadata_resolver/sample/categorical_value_counts.py
```diff
@@ -0,0 +1,155 @@
...
+
+def get_metadata_value_counts(
+    session: Session,
+    collection_id: UUID,
+    filters: ImageFilter | None = None,
+    fields: list[str] | None = None,
+) -> dict[str, MetadataValueCountsView]:
+    """Count categorical metadata values for a collection.
+
+    Each field's own metadata filter is excluded while all other filters apply.
```
JonasWurst, line 34: Can you motivate/explain this?

lightly_studio/src/lightly_studio/resolvers/metadata_resolver/sample/categorical_value_counts.py
```diff
@@ -0,0 +1,155 @@
...
+_CATEGORICAL_TYPES = ("string", "boolean")
+_TOP_VALUE_COUNT = 20
+
+
+def get_metadata_value_counts(
+    session: Session,
+    collection_id: UUID,
+    filters: ImageFilter | None = None,
+    fields: list[str] | None = None,
+) -> dict[str, MetadataValueCountsView]:
```
JonasWurst, line 31: I'm not fully clear on this return type and how we make use of all the values

lightly_studio/src/lightly_studio/resolvers/metadata_resolver/sample/__init__.py
ikondrat: lt caused circuclar dependency issue and never used from module export

lightly_studio/src/lightly_studio/resolvers/metadata_resolver/sample/categorical_value_counts.py
ikondrat: I will be used directly in endpoint and the frontend https://github.com/lightly-ai/lightly-studio/pull/1749/changes

lightly_studio/src/lightly_studio/resolvers/metadata_resolver/sample/categorical_value_counts.py
ikondrat: If you just select everything once how it would work when you apply filters? we don't filter data on client because it would logic duplicaiton

ikondrat: I think keeping logic on backend is always better

lightly_studio/src/lightly_studio/resolvers/metadata_resolver/sample/metadata_helpers.py
ikondrat: yes it moved to avoid too big file

lightly_studio/tests/metadata/test_get_metadata_value_counts.py
ikondrat: There are actually two samples without a city value:

Line 48 — _create_explicit_null_sample: has a metadata row with {"city": None} — an explicit SQL NULL for city.

Line 49–53 — create_image with no metadata at all: no SampleMetadataTable row exists for this sample, so the outer join in the query produces NULL for the city column.

Both produce a NULL when the query extracts the city field, so both fall into missing_count. The "one image without meta" intuition is right for line 49–53, but the explicit-null sample on line 48 is a separate, subtler case — it has a metadata row,

lightly_studio/tests/metadata/test_get_metadata_value_counts.py
ikondrat: the same answer :)

lightly_studio/src/lightly_studio/resolvers/metadata_resolver/sample/categorical_value_counts.py
ikondrat: added a shot description

lightly_studio/src/lightly_studio/resolvers/metadata_resolver/sample/categorical_value_counts.py
JonasWurst: your call :)

lightly_studio/src/lightly_studio/resolvers/metadata_resolver/__init__.py
JonasWurst: Can you answer

ikondrat: It is not related and was cleaned up

JonasWurst (approved):

PR #1753, opened 2026-07-23
Refactor video embedding to a manual batch loop (drop DataLoader)

JonasWurst:
## Summary

Pure-refactor prefactor for the stacked per-item broken-file PR, mirroring how #1710 (the `EmbeddingResult` extraction) was stacked under #1706.

Replaces the `_VideoFileDataset` + `DataLoader` machinery in the perception-encoder video path with a manual per-video batch loop, structurally matching `embed_image_crops_batched`. At `num_workers=0` the `DataLoader` gave **no** async prefetch — only batching/collation — so nothing is lost, and the manual loop gives the stacked follow-up a natural place to skip individual broken videos.

- Extract frame sampling into a module-level `_load_video_frames` helper (the former `_VideoFileDataset._load_frames` + stack); it raises on an unusable video, preserving today's abort-on-broken behavior.
- `embed_videos` buffers preprocessed videos and flushes batches via a new `_flush_video_batch` helper.

**No behavior change.** Return type stays `NDArray[np.float32]` and a broken video still aborts the run.

## Testing

- `make static-checks` — ruff + format + mypy clean.
- `pytest tests/dataset/test_perception_encoder_embedding_generator.py` — 10 passed, unchanged (incl. the real `dog.mp4` embedding + video-classification tests), which is the proof this is a clean refactor.

## Summary by CodeRabbit

* **Performance**
  * Improved video embedding generation with synchronous batching for faster, more predictable processing.
  * Lower overhead and memory usage when embedding large video collections.
* **Quality / Reliability**
  * Improved robustness for videos with unusable timing or frames that can’t be decoded.
  * Preserves the embedding order to match the input list.
  * Returns an empty embeddings array when no videos are provided.

### DIFF
--- a/lightly_studio/src/lightly_studio/dataset/perception_encoder_embedding_generator.py
+++ b/lightly_studio/src/lightly_studio/dataset/perception_encoder_embedding_generator.py
@@ -173,7 +173,7 @@ class PerceptionEncoderEmbeddingGenerator(ImageEmbeddingGenerator, VideoEmbeddin
         embeddings = np.empty((total_videos, self._model.output_dim), dtype=np.float32)
         # Every video yields a fixed VIDEO_FRAMES_PER_SAMPLE-frame tensor of the same shape,
         # so the per-video tensors can be stacked directly into a batch.
-        preprocessed_videos = (
+        preprocessed_videos_iter = (
             _load_video_frames(filepath, self._preprocess) for filepath in filepaths
         )
         position = 0
@@ -181,7 +181,7 @@ class PerceptionEncoderEmbeddingGenerator(ImageEmbeddingGenerator, VideoEmbeddin
             tqdm(total=total_videos, desc="Generating embeddings", unit=" videos") as progress_bar,
             torch.no_grad(),
         ):
-            for batch in batching.batched(preprocessed_videos, batch_size=MAX_BATCH_SIZE):
+            for batch in batching.batched(preprocessed_videos_iter, batch_size=MAX_BATCH_SIZE):
                 videos_tensor = torch.stack(batch).to(self._device)
                 batch_embeddings = (
                     self._model.encode_video(videos_tensor, normalize=True).cpu().numpy()
--- a/lightly_studio/src/lightly_studio/dataset/perception_encoder_embedding_generator.py
+++ b/lightly_studio/src/lightly_studio/dataset/perception_encoder_embedding_generator.py
@@ -3,7 +3,7 @@
 from __future__ import annotations
 
 from pathlib import Path
-from typing import Callable
+from typing import Any, Callable
 from uuid import UUID
 
 import fsspec
@@ -12,7 +12,6 @@ import torch
 from av import container
 from numpy.typing import NDArray
 from PIL import Image
-from torch.utils.data import DataLoader, Dataset
 from tqdm import tqdm
 
 from lightly_studio.dataset.env import LIGHTLY_STUDIO_MODEL_CACHE_DIR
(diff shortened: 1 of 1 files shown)


lightly_studio/src/lightly_studio/dataset/perception_encoder_embedding_generator.py
```diff
@@ -240,32 +165,119 @@ def embed_videos(self, filepaths: list[str]) -> NDArray[np.float32]:
...
-                embeddings[position : position + batch_size] = batch_embeddings
-                position += batch_size
-                progress_bar.update(batch_size)
+            for index, filepath in enumerate(filepaths):
+                frames = _load_video_frames(filepath, self._preprocess)
+                if batch_buffer is None:
+                    batch_buffer = torch.empty((MAX_BATCH_SIZE, *frames.shape), dtype=frames.dtype)
+                batch_buffer[len(batch_indices)] = frames
+                batch_indices.append(index)
+                if len(batch_indices) >= MAX_BATCH_SIZE:
```
michal-lightly: The logic here should be much simpler with `lightly_studio/src/lightly_studio/utils/batching.py`, can you have a look? Please ask AI to have a look how we use it elsewhere.

lightly_studio/src/lightly_studio/dataset/perception_encoder_embedding_generator.py
```diff
@@ -240,32 +165,119 @@ def embed_videos(self, filepaths: list[str]) -> NDArray[np.float32]:
...
+            )
 
         return embeddings
+
+    def _flush_video_batch(
+        self,
+        batch_buffer: torch.Tensor | None,
+        batch_indices: list[int],
+        embeddings: NDArray[np.float32],
+        progress_bar: Any,
```
michal-lightly: Don't use Any, please ask AI to use the proper type.

lightly_studio/src/lightly_studio/dataset/perception_encoder_embedding_generator.py
```diff
@@ -240,32 +165,119 @@ def embed_videos(self, filepaths: list[str]) -> NDArray[np.float32]:
...
-                videos = videos_tensor.to(self._device, non_blocking=True)
-                batch_embeddings = self._model.encode_video(videos, normalize=True).cpu().numpy()
-                batch_size = videos.size(0)
-                embeddings[position : position + batch_size] = batch_embeddings
-                position += batch_size
-                progress_bar.update(batch_size)
+            for index, filepath in enumerate(filepaths):
+                frames = _load_video_frames(filepath, self._preprocess)
+                if batch_buffer is None:
+                    batch_buffer = torch.empty((MAX_BATCH_SIZE, *frames.shape), dtype=frames.dtype)
```
michal-lightly: How does this work if different videos have different shapes?

lightly_studio/src/lightly_studio/dataset/perception_encoder_embedding_generator.py
```diff
@@ -240,32 +165,119 @@ def embed_videos(self, filepaths: list[str]) -> NDArray[np.float32]:
...
+            embeddings[video_index] = batch_embeddings[batch_position]
+        progress_bar.update(filled)
+        batch_indices.clear()
+
+
+def _load_video_frames(
+    filepath: str,
+    preprocess: Callable[[Image.Image], torch.Tensor],
+) -> torch.Tensor:
+    """Sample uniformly spaced frames from a video and stack them into a model input.
```
michal-lightly, line 201: Is this functionality just moved or are there any logical changes? A note for the reviewer would be helpful here.

michal-lightly (changes requested):

lightly_studio/src/lightly_studio/dataset/perception_encoder_embedding_generator.py
JonasWurst: They are not of different shape as we are always sampling 8 frames fixed. If we can't we error out. This was the case before as well

lightly_studio/src/lightly_studio/dataset/perception_encoder_embedding_generator.py
JonasWurst: It just got moved

lightly_studio/src/lightly_studio/dataset/perception_encoder_embedding_generator.py
michal-lightly: They would have the same number of frames, but different width and height - no? Saying "same shape" seems incorrect to me.

lightly_studio/src/lightly_studio/dataset/perception_encoder_embedding_generator.py
```diff
@@ -240,32 +166,92 @@ def embed_videos(self, filepaths: list[str]) -> NDArray[np.float32]:
...
         total_videos = len(filepaths)
         if not total_videos:
             return np.empty((0, self._model.output_dim), dtype=np.float32)
 
         embeddings = np.empty((total_videos, self._model.output_dim), dtype=np.float32)
+        # Every video yields a fixed VIDEO_FRAMES_PER_SAMPLE-frame tensor of the same shape,
+        # so the per-video tensors can be stacked directly into a batch.
+        preprocessed_videos = (
+            _load_video_frames(filepath, self._preprocess) for filepath in filepaths
+        )
```
michal-lightly: This code decodes all video frames at once. Instead, I would suggest to iterate over filepaths below and decode only the current batch.

michal-lightly (changes requested):

lightly_studio/src/lightly_studio/dataset/perception_encoder_embedding_generator.py
JonasWurst: Width and height are always similar due to the preprocessing: `self._preprocess = transforms.get_image_transform(self._model.image_size)`

lightly_studio/src/lightly_studio/dataset/perception_encoder_embedding_generator.py
JonasWurst: It is a generator expression, it will only be consumed in the batching.

lightly_studio/src/lightly_studio/dataset/perception_encoder_embedding_generator.py
michal-lightly: Similar = equal?

lightly_studio/src/lightly_studio/dataset/perception_encoder_embedding_generator.py
michal-lightly: Indeed!

Optional: We could call it `preprocessed_videos_iter`

michal-lightly (approved):

PR #1754, opened 2026-07-23
Tolerate broken files per-item in video embedding

JonasWurst:
## Summary

Apply the per-item broken-file handling from the full-image (#1691) and crop (#1706) embedding paths to the **video** path.

- `_load_video_frames` returns `None` for an unreadable/undecodable video — catching `(OSError, FFmpegError, IndexError, StopIteration)` plus the no-usable-duration case — instead of raising, so a broken video skips itself rather than aborting the run.
- `embed_videos` records each video once as `BROKEN`/`ADDED` in a `FileOutcomeReport`, returns an `EmbeddingResult` carrying `kept_indices` (input positions of the embedded videos), and raises `AllInputFilesFailedError` when every video is broken.
- `EmbeddingResult` is propagated through the `VideoEmbeddingGenerator.embed_videos` protocol method and `RandomEmbeddingGenerator`.
- `embedding_manager.embed_videos` filters sample IDs by `kept_indices` before storing, keeping sample IDs and embeddings in sync.

`kept_indices` comes out already sorted (videos are processed in input order, unlike the crop path which groups by filepath), so no post-sort is needed.

## Stacking

This PR is **stacked on #1753** (the manual-batch-loop refactor), not `main`. Its base is `jonas-lig-10035-refactor-video-embedding-manual-loop`. Rebase onto `main` once #1753 merges. Review #1753 first — this PR's diff is only the tolerance change.

## Testing

- `make static-checks` — ruff + format + mypy clean.
- `pytest tests/dataset/{test_perception_encoder_embedding_generator,test_embedding_manager,test_embedding_generator}.py` — 41 passed.
- Added video tests: mixed good/broken/missing with `kept_indices` assertions, all-broken raises `AllInputFilesFailedError`, and a manager stub-drop case asserting only kept sample IDs are stored; unwrapped `.embeddings` in the existing video tests.

## Summary by CodeRabbit

- **New Features**
  - Video embedding now returns a structured result that includes both the generated embeddings and a `kept_indices` mapping for successfully processed inputs.
  - Improved fault-tolerance: missing or broken videos are skipped while still returning results for valid inputs.
- **Bug Fixes**
  - Embeddings are now stored only for kept samples, ensuring embeddings remain aligned with the correct videos when some inputs fail.
- **Tests**
  - Added/updated unit tests to verify skipping behavior, correct `kept_indices` alignment, and raising an error when all inputs fail.

### DIFF
--- a/lightly_studio/src/lightly_studio/dataset/embedding_manager.py
+++ b/lightly_studio/src/lightly_studio/dataset/embedding_manager.py
@@ -382,14 +382,16 @@ class EmbeddingManager:
         if len(filepaths) != len(sample_ids):
             raise ValueError("Could not fetch all video paths for the provided IDs.")
 
-        # Generate embeddings for the samples.
-        embeddings = model.embed_videos(filepaths=filepaths)
+        # Generate embeddings for the samples. Broken videos are skipped, so filter the
+        # sample IDs by kept_indices to keep them aligned with the returned embeddings.
+        result = model.embed_videos(filepaths=filepaths)
+        kept_sample_ids = [sample_ids[index] for index in result.kept_indices]
 
         _store_embeddings(
             session=session,
             model_id=model_id,
-            sample_ids=sample_ids,
-            embeddings=embeddings,
+            sample_ids=kept_sample_ids,
+            embeddings=result.embeddings,
         )
 
     def embed_and_store_pil_images(
--- a/lightly_studio/src/lightly_studio/dataset/embedding_manager.py
+++ b/lightly_studio/src/lightly_studio/dataset/embedding_manager.py
@@ -382,8 +382,7 @@ class EmbeddingManager:
         if len(filepaths) != len(sample_ids):
             raise ValueError("Could not fetch all video paths for the provided IDs.")
 
-        # Generate embeddings for the samples. Broken videos are skipped, so filter the
-        # sample IDs by kept_indices to keep them aligned with the returned embeddings.
+        # Generate embeddings for the samples.
         result = model.embed_videos(filepaths=filepaths)
         kept_sample_ids = [sample_ids[index] for index in result.kept_indices]
 
--- a/lightly_studio/src/lightly_studio/dataset/embedding_manager.py
+++ b/lightly_studio/src/lightly_studio/dataset/embedding_manager.py
@@ -380,14 +380,16 @@ class EmbeddingManager:
         if len(filepaths) != len(sample_ids):
             raise ValueError("Could not fetch all video paths for the provided IDs.")
 
-        # Generate embeddings for the samples.
-        embeddings = model.embed_videos(filepaths=filepaths)
+        # Generate embeddings for the samples. Broken videos are skipped, so filter the
+        # sample IDs by kept_indices to keep them aligned with the returned embeddings.
+        result = model.embed_videos(filepaths=filepaths)
+        kept_sample_ids = [sample_ids[index] for index in result.kept_indices]
 
         _store_embeddings(
             session=session,
             model_id=model_id,
-            sample_ids=sample_ids,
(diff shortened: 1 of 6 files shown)


lightly_studio/tests/dataset/test_embedding_manager.py
```diff
@@ -742,6 +742,46 @@ def test_embed_videos(
...
+            return EmbeddingResult(embeddings=embeddings, kept_indices=kept_indices)
+
+    video_collection = create_collection(session=db_session, sample_type=SampleType.VIDEO)
+    collection_id = video_collection.collection_id
+    video_ids = create_videos(
+        session=db_session,
+        collection_id=collection_id,
+        videos=[
+            VideoStub(path=f"/videos/video_{idx}.mp4", duration_s=1.0 + idx, fps=24.0)
+            for idx in range(3)
```
michal-lightly: Optional: I'd write out the three video stubs instead of the for loop.

lightly_studio/tests/dataset/test_embedding_manager.py
```diff
@@ -742,6 +742,46 @@ def test_embed_videos(
...
+    db_session: Session,
+) -> None:
+    """A generator that drops a broken video stores embeddings only for the kept ones."""
+
+    class DropsMiddleVideoGenerator(RandomEmbeddingGenerator):
+        def embed_videos(self, filepaths: list[str]) -> EmbeddingResult:
+            # Simulate the middle video being broken and skipped.
+            kept_indices = [index for index in range(len(filepaths)) if index != 1]
+            embeddings = np.zeros((len(kept_indices), self._dimension), dtype=np.float32)
+            return EmbeddingResult(embeddings=embeddings, kept_indices=kept_indices)
```
michal-lightly: Optional: Drop the for loop:
```py
return EmbeddingResult(
    embeddings=np.zeros((2, self._dimension), dtype=np.float32),
    kept_indices=[0, 2],
)
```

michal-lightly (approved):

lightly_studio/src/lightly_studio/dataset/embedding_manager.py
```diff
@@ -380,14 +380,16 @@ def embed_videos(
         if len(filepaths) != len(sample_ids):
             raise ValueError("Could not fetch all video paths for the provided IDs.")
 
-        # Generate embeddings for the samples.
-        embeddings = model.embed_videos(filepaths=filepaths)
+        # Generate embeddings for the samples. Broken videos are skipped, so filter the
+        # sample IDs by kept_indices to keep them aligned with the returned embeddings.
+        result = model.embed_videos(filepaths=filepaths)
+        kept_sample_ids = [sample_ids[index] for index in result.kept_indices]
```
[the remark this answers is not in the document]
JonasWurst: This is an intentional, uniform contract change rather than a regression specific to videos. The `embed_images` (#1691) and `embed_image_crops` (#1706) paths already return `EmbeddingResult` and are consumed via `result.kept_indices`/`result.embeddings` with no legacy `NDArray` fallback. `VideoEmbeddingGenerator` is an internal `Protocol`, and this PR just brings the video path in line with images/crops. Adding a legacy array-result shim only for videos would leave the three paths inconsistent, so we're deliberately not doing it here.

lightly_studio/tests/dataset/test_embedding_manager.py
JonasWurst: Done — hardcoded to `kept_indices=[0, 2]` and `np.zeros((2, ...))`.

lightly_studio/tests/dataset/test_embedding_manager.py
JonasWurst: Done — written out as three explicit `VideoStub`s.

PR #1778, opened 2026-07-27
Prefactor: one shared ordering helper for annotations

JonasWurst:
## What has changed and why?

The annotations grid resolver and the adjacency window query each built their own ORDER BY, so grid order and detail-view next/prev could drift apart. Both now derive it from `annotation_ordering.build_order_by` (file path, created at, annotation ID), which takes an optional leading order key so a metric expression can be prepended later without touching the call sites. Similarity distance now goes through that leading key.

Pure refactor, no user-observable change.

## How has it been tested?

- `test_adjacents.py` gains `test_get_adjacent_samples__annotations_match_grid_ordering`: fetches a page from `/annotations/payload`, walks `/samples/{id}/adjacents` from the first annotation, asserts the two sequences match. It passes on the pre-refactor code too, which is the point. Also the first annotation coverage in that file, which previously tested images and videos only.
- New unit tests for the helper in `tests/resolvers/annotations/test_annotation_ordering.py`: tiebreaker chain, leading-key placement, per-sample-type file path expressions.
- `make static-checks` and `make test` (1761 passed, 32 skipped).

- [ ] Yes

## Summary by CodeRabbit

- **Improvements**
  - Standardized annotation sorting across annotation grids, adjacent-navigation results, and annotation list/detail lookups.
  - Ensured adjacent-navigation follows the same ordering as the annotations grid, with consistent tie-breaking by file location, creation time, and sample identifier.
- **Tests**
  - Added coverage for the shared ordering logic, including helper behavior for supported/unsupported sample types and generated SQL expressions.
  - Added an integration test confirming adjacent-navigation annotation sequences match the grid/payload ordering.


lightly_studio/tests/api/routes/api/annotations/test_annotation_ordering.py
```diff
@@ -0,0 +1,104 @@
...
+    )
+
+    assert walked_sample_ids == grid_sample_ids
+
+
+def _walk_adjacents(
+    test_client: TestClient,
+    collection_id: UUID,
+    start_sample_id: UUID,
+) -> list[UUID]:
```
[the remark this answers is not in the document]
JonasWurst: only split if we benefit

lightly_studio/src/lightly_studio/resolvers/annotations/annotation_ordering.py
```diff
@@ -0,0 +1,85 @@
...
+    if sample_type == SampleType.IMAGE:
+        return col(ImageTable.file_path_abs)
+
+    if sample_type in (SampleType.VIDEO_FRAME, SampleType.VIDEO):
+        return col(VideoTable.file_path_abs)
+
+    raise NotImplementedError(f"Unsupported sample type: {sample_type}")
+
+
+def coalesced_file_path_abs_expression() -> OrderExpression:
```
michal-lightly, line 74: This function is strange and I think might break when we introduce groups that can mix different modalities. It would be best if it did not exist - if we could use `file_path_abs_expression` everywhere. Or even better, `get_default_order_by(sample_type)`.

It is an improvement over the previous state though so good to merge.

lightly_studio/tests/api/routes/api/test_adjacents.py
```diff
@@ -204,3 +205,79 @@ def test_get_adjacent_samples__returns_adjacents_for_videos(
...
+        root_collection_id=collection_id,
+        label_name="dog",
+    )
+
+    # Paths are created out of order so that ordering by file path is observable.
+    images = helpers_resolvers.create_images(
+        db_session=db_session,
+        collection_id=collection_id,
+        images=[
+            helpers_resolvers.ImageStub(path=path)
```
michal-lightly: We import classes usually directly.

Optional: Remove the for loop.

michal-lightly (approved):

lightly_studio/src/lightly_studio/resolvers/annotations/annotation_ordering.py
JonasWurst: Created a follow up

PR #1791, opened 2026-07-28
Add useExportDownload hook and ExportDownloadButton component

LeonardoRosaa:
## What has changed and why?

- `useExportDownload` - wraps any prepare-endpoint call; manages `isLoading` and `errorMessage` state; opens the download URL in a new tab on success.
- `ExportDownloadButton`- renders a spinner while loading, an error alert on failure, and the primary download button; used by every tab.

## How has it been tested?

Unit tests

- [ ] Yes

## Summary by CodeRabbit

- **New Features**
  - Added an “Export Download” button with built-in loading, disabled behavior, and destructive error display.
  - Introduced a download hook that manages async preparation, shows loading during export, and surfaces failures as user-friendly messages.
  - Clears any previous error automatically on a new download attempt.
- **Tests**
  - Added Vitest coverage for button rendering/click handling, loading and disabled states, error rendering, and hook behavior during success/failure.

### DIFF
--- a/lightly_studio_view/src/lib/components/ExportSamples/useExportDownload/useExportDownload.ts
+++ b/lightly_studio_view/src/lib/components/ExportSamples/useExportDownload/useExportDownload.ts
@@ -1,0 +1,26 @@
+import { writable, type Readable } from 'svelte/store';
+
+interface UseExportDownloadReturn {
+    isLoading: Readable<boolean>;
+    errorMessage: Readable<string>;
+    handleDownload: () => Promise<void>;
+}
+
+export function useExportDownload(prepare: () => Promise<void>): UseExportDownloadReturn {
+    const isLoading = writable(false);
+    const errorMessage = writable('');
+
+    const handleDownload = async () => {
+        isLoading.set(true);
+        errorMessage.set('');
+        try {
+            await prepare();
+        } catch (e) {
+            errorMessage.set(`Export failed: ${String(e)}`);
+        } finally {
+            isLoading.set(false);
+        }
+    };
+
+    return { isLoading, errorMessage, handleDownload };
+}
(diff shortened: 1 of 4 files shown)


lightly_studio_view/src/lib/components/ExportSamples/useExportDownload/useExportDownload.ts
```diff
@@ -0,0 +1,26 @@
...
+    const isLoading = writable(false);
+    const errorMessage = writable('');
+
+    const handleDownload = async () => {
+        isLoading.set(true);
+        errorMessage.set('');
+        try {
+            await prepare();
+        } catch (e) {
+            errorMessage.set(`Export failed: ${String(e)}`);
```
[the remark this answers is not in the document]
LeonardoRosaa: In this case, the behavior is that prepare will request the `/prepare` endpoint and start downloading the file via `window.open`. So, the current implementation is fine.

lightly_studio_view/src/lib/components/ExportSamples/useExportDownload/useExportDownload.ts
```diff
@@ -0,0 +1,35 @@
...
+    handleDownload: () => Promise<void>;
+}
+
+/**
+ * Hook that manages the loading and error state for triggering an export download.
+ *
+ * @param prepare - Async function that performs the actual export preparation (e.g. API call).
+ *                  Should throw on failure so the error can be caught and surfaced to the user.
+ */
+export function useExportDownload(prepare: () => Promise<void>): UseExportDownloadReturn {
```
ikondrat, line 18: hmm, don't we have generated query function for it? 

I mean query already contains info about error and isLoading state

ikondrat (approved):

lightly_studio_view/src/lib/components/ExportSamples/useExportDownload/useExportDownload.ts
LeonardoRosaa: Discussed internally - the hook is necessary to wrap two HTTP requests.

PR #1798, opened 2026-07-29
Backend, Part 1: Add TableParameter to the plugin parameter types

mrpositron:
## What has changed and why?

Adds a new `TableParameter` type to the plugin parameter types (backend part 1).

Operators sometimes need a variable number of homogeneous multi-field inputs — e.g. a list of segmentation prompts, each paired with the annotation label to assign to the resulting masks and a confidence threshold. The existing scalar parameters (`Int`/`Float`/`Bool`/`StringParameter`) can't express that.

`TableParameter` holds a list of rows, where each row maps every column name to a cell of that column's type:

```python
TableParameter(
    name="prompts",
    description="Prompt to segment with and the label to assign to the masks.",
    columns=[
        StringParameter(name="prompt", description="What to segment."),
        StringParameter(name="label", default="pedestrian", required=False),
        FloatParameter(name="threshold", default=0.5, required=False),
    ],
)
```

Details:

- A column is any built-in parameter (`Int`/`Float`/`Bool`/`StringParameter`), so each column carries its own type, description, default and required flag. Cell validation is delegated to the column's own parameter, so types stay in one place.
- Columns are validated at construction: at least one column, no duplicate names, and every column must be a `BuiltinParameter` (nested tables fail loudly rather than rendering as an empty column).
- Row validation reports the offending row index and column name, and normalizes cell ordering to match the declared column order.

No API/frontend changes yet — this is backend-only groundwork. Note for the follow-up GUI PR: since columns are typed, the table editor needs a per-type cell editor, not just text inputs.

## How has it been tested?

New unit tests in `lightly_studio/tests/plugins/test_parameter.py` covering:

- `asdict()` serialization of a table parameter incl. its nested columns
- column defaults / required flags, no-default and empty-list-default cases
- cell ordering normalized to the column order
- a mixed-type table (string/float/int/bool columns) validating through its defaults
- error cases: missing/empty columns, duplicate column names, non-`BuiltinParameter` column, nested `TableParameter` column, non-list default, non-dict row, missing/extra column in a row, cell type not matching its column, and correct row index reporting for a later row

## Did you update [CHANGELOG.md](../CHANGELOG.md)?

- [ ] Yes
- [x] Not needed (internal change)

## Summary by CodeRabbit

* **New Features**
  * Added support for table-based parameters with configurable built-in columns.
  * Table values now validate column definitions, row structure, required fields, and cell types.
  * Preserves the configured column order and provides clear error messages identifying invalid rows and values.

* **Tests**
[...]

PR #1799, opened 2026-07-29
Backend, Part 2: expose operator parameter columns over the API

mrpositron:
## What has changed and why?

Backend part 2 of the table parameter stack. Part 1 (#1798, `TableParameter` in `plugins/parameter.py`) has landed, so this now targets `main` directly.

Part 1 let a plugin *declare* a table parameter; the GUI parts render one. Neither is reachable from the other, because the columns never survived the trip to the browser:

`GET /operators/{operator_id}/parameters` was annotated with the base class `list[BaseParameter]`. Pydantic serialises against the *declared* type, so it validated each item against the base class and dropped `columns` — a field that only exists on `TableParameter`. An operator could declare a fully specified table and the GUI had no way to learn its columns existed.

- Return explicit `ParameterView` / `ParameterColumnView` models instead of the plugin dataclasses, which also keeps the API shape independent of the plugin dataclasses per `ai_guidelines/backend.md`.
- Migrate `useOperators.ts` to `ParameterView` and map `columns` onto a new `OperatorParameterColumn`. This has to happen in the same PR: the generated client no longer exports `BaseParameter`, so `check-typescript` / `test-frontend` would break on an intermediate commit.

`OperatorParameterColumn` passes every column field through unchanged, including `param_type` as `paramType`. Part 1 permits a column of any built-in type, so a `FloatParameter(name="threshold", default=0.5)` column has to keep both its type and its numeric default — narrowing either one here would discard it silently at the one boundary
where the loss is invisible to every later caller, which is the same class of bug this PR exists to fix. `paramType` holds the Python type name (`'str'`, `'float'`, …) and is deliberately typed `string` rather than `OperatorParameterType`, whose members spell `'string'`.

## How has it been tested?

`make static-checks` and `make test` in `lightly_studio`, plus `npm run check` in `lightly_studio_view` (0 errors).

New tests in `tests/api/routes/api/test_operator.py`:

- `test_get_operator_parameters__table_parameter` — asserts the full response JSON for a table parameter. This is the regression guard for the dropped columns: it fails against the old `list[BaseParameter]` annotation.
- `test_execute_operator__table_parameter_rows_reach_the_operator` — rows arrive in `execute()` as a list of dicts.

Existing multi-parameter test updated for the additive `"columns": null`.

- [ ] Yes

## Summary by CodeRabbit

* **New Features**
  * Operator parameter metadata now includes table columns, descriptions, defaults, required status, and types.
  * Added support for displaying nested table parameter definitions in the operator interface.

* **Bug Fixes**
  * Table parameter rows are now passed to operators without modification.
  * Parameter metadata is presented consistently, including correct handling of optional or unspecified types.

### DIFF
--- a/lightly_studio/src/lightly_studio/api/routes/api/operator.py
+++ b/lightly_studio/src/lightly_studio/api/routes/api/operator.py
@@ -14,10 +14,27 @@ from lightly_studio.plugins import operator_context
 from lightly_studio.plugins.base_operator import OperatorResult, OperatorStatus
 from lightly_studio.plugins.operator_context import AnyFilter, ExecutionContext
 from lightly_studio.plugins.operator_registry import RegisteredOperatorMetadata, operator_registry
-from lightly_studio.plugins.parameter import BaseParameter
+from lightly_studio.plugins.parameter import BaseParameter, TableParameter
 from lightly_studio.resolvers import collection_resolver
 
 operator_router = APIRouter(prefix="/operators", tags=["operators"])
+
+
+class ParameterColumnView(BaseModel):
+    """A single column of a table parameter as exposed to the GUI."""
+
+    name: str
+    description: str
+    default: Any = None
+    required: bool
+    param_type: str | None
+
+
+class ParameterView(ParameterColumnView):
+    """An operator parameter as exposed to the GUI."""
+
+    columns: list[ParameterColumnView] | None = None
+    """The columns of a table parameter, `None` for every other parameter type."""
 
 
 class OperatorContextRequest(BaseModel):
@@ -44,7 +61,7 @@ def get_operators() -> list[RegisteredOperatorMetadata]:
 
 
 @operator_router.get("/{operator_id}/parameters")
-def get_operator_parameters(operator_id: str) -> list[BaseParameter]:
+def get_operator_parameters(operator_id: str) -> list[ParameterView]:
     """Get the parameters for a registered operator."""
     operator = operator_registry.get_by_id(operator_id=operator_id)
     if operator is None:
@@ -52,7 +69,7 @@ def get_operator_parameters(operator_id: str) -> list[BaseParameter]:
             status_code=HTTP_STATUS_NOT_FOUND,
             detail=f"Operator '{operator_id}' not found",
         )
-    return operator.parameters
+    return [_parameter_view(parameter=parameter) for parameter in operator.parameters]
 
 
 @operator_router.post("/{operator_id}/execute", response_model=OperatorResult)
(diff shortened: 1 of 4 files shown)


lightly_studio/src/lightly_studio/api/routes/api/operator.py
```diff
@@ -44,15 +61,15 @@ def get_operators() -> list[RegisteredOperatorMetadata]:
 
 
 @operator_router.get("/{operator_id}/parameters")
-def get_operator_parameters(operator_id: str) -> list[BaseParameter]:
+def get_operator_parameters(operator_id: str) -> list[ParameterView]:
```
JonasWurst, line 64:
```suggestion

def get_operator_parameters(operator_id: str) -> list[TableParameter | BaseParameter]:

```

ParameterView, is that really required?

JonasWurst (commented):
Before moving on with the review. Please check my question

lightly_studio/src/lightly_studio/api/routes/api/operator.py
mrpositron: Union also works. But ParameterView will make things cleaner.

From Claude:
1. TypeScript can't tell the two types apart, so the frontend guesses by hand with an unchecked `'columns' in p ` check
2. Every field becomes optional, so you add `?? true` fallbacks that duplicate the Python defaults and go stale
3. The frontend imports a type named `BuiltinParameter_Any_`, auto-mangled from Python generics

Also, (ai_guidelines/backend.md#L28) already names *View as the convention for API response models.

lightly_studio/tests/api/routes/api/test_operator.py
```diff
@@ -391,6 +470,35 @@ def parameters(self) -> list[BaseParameter]:
...
+
+    @property
+    def parameters(self) -> list[BaseParameter]:
+        return [
+            TableParameter(
+                name="prompts",
+                description="Prompts and labels.",
+                columns=[
+                    StringParameter(name="prompt", description="What to segment."),
+                    StringParameter(name="label", default="pedestrian", required=False),
```
JonasWurst: use one other

lightly_studio/tests/api/routes/api/test_operator.py
```diff
@@ -111,17 +116,64 @@ def test_get_operator_parameters__multiple_parameters(
...
+    test_client: TestClient,
+    isolated_operator_registry: OperatorRegistry,
+) -> None:
+    """The full column definitions must reach the client.
+
+    Regression guard: `columns` lives on `TableParameter` only, so the route has to map the
+    parameters onto `ParameterView` to expose them. Returning the plugin dataclasses directly would
+    make pydantic serialize each item against the base class and silently drop the columns, leaving
+    the GUI without column information and without any error.
+    """
```
JonasWurst:
```suggestion
```

lightly_studio/tests/api/routes/api/test_operator.py
```diff
@@ -306,6 +358,33 @@ def test_execute_operator__filter_is_passed_through(
...
+def test_execute_operator__table_parameter_rows_reach_the_operator(
+    test_client: TestClient,
+    collection_id: UUID,
+    isolated_operator_registry: OperatorRegistry,
+) -> None:
+    """Table rows must arrive in `execute()` as a list of dicts.
+
+    `ExecuteOperatorRequest.parameters` is a `dict[str, Any]`, so the rows pass through untouched.
+    This pins that contract, since it is what plugins index into.
+    """
```
JonasWurst:
```suggestion
```

lightly_studio_view/src/lib/hooks/useOperators/useOperators.ts
```diff
@@ -1,13 +1,25 @@
-import type { BaseParameter, RegisteredOperatorMetadata } from '$lib/api/lightly_studio_local';
+import type {
+    ParameterColumnView,
+    ParameterView,
+    RegisteredOperatorMetadata
+} from '$lib/api/lightly_studio_local';
 
 export type OperatorParameterType = 'string' | 'int' | 'float' | 'bool';
 
+export type OperatorParameterColumn = {
```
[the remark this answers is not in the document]
JonasWurst: please check this.

lightly_studio_view/src/lib/hooks/useOperators/useOperators.ts
```diff
@@ -16,17 +28,27 @@ export type Operator = {
     parameters: OperatorParameter[];
 };
 
-const mapParameter = (parameter: BaseParameter): OperatorParameter => ({
+const mapColumn = (column: ParameterColumnView): OperatorParameterColumn => ({
```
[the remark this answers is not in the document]
JonasWurst: yes, please add

JonasWurst (commented):
Minor issues

JonasWurst (approved):

PR #1800, opened 2026-07-29
Frontend, Part 1: Add ParameterTable component for table operator parameters

mrpositron:
## What has changed and why?

LIG-10323 frontend part 1, stacked on backend parts 1 and 2 (`TableParameter` in `plugins/parameter.py`, and `columns` exposed over the API).

Adds `ParameterTable/`, which renders a table parameter as editable rows: one type-aware cell per column, "Add row", and a per-row delete button. It holds no internal row state and reports every change through `onUpdate`, matching the existing `ParameterInput` / `ParameterCheckbox` contract.

- `ParameterTableCell` renders each cell after its column's Python type: `str` as text, `int` / `float` as number inputs, `bool` as a checkbox. An unknown type degrades to text rather than breaking the table.
- New rows are pre-filled from the column defaults. Only empty cells of *required* columns are flagged, except that a boolean is never flagged (a checkbox always has a value, so treating `false` as blank would turn "required" into "must be checked") and an empty numeric cell always is (the backend rejects `''` against the column type regardless of `required`).
- Data columns have a `9rem` minimum, so a wide table scrolls horizontally instead of squeezing every input down to a few characters. Both axes scroll in one box, which is what puts the header inside it: the header pins vertically and scrolls with the rows horizontally for free, and the remove-button column pins right.
- Both sticky edges cover the scroll container's padding, which an inset of `0` cannot reach — insets resolve against the padding box, but a sticky box only ever shifts *inward*, and these already sit at the content-box edge. Each pairs a negative inset with a margin that grows the box rather than moving it.
- `min-w-0` in two places keeps that scrolling inside the table: a column minimum grows the grid's min-content width, and `Dialog.Content` is itself a grid, whose items don't shrink below min-content by default.
- `useOperators` is now exported from `$lib/hooks`, so these components import from the barrel instead of a deep path.

Not in this PR: registering `'table'` in `TYPE_CONFIG`. Until part 2 lands the dialog still renders a plain text input for a table parameter; the component is covered by its own tests meanwhile.

~1076 lines: 438 source, 638 tests. The `frontend/coverage` guardrail requires a test file per source file plus 80% on added lines, so the test volume is a floor. Coverage is 100%.

## How has it been tested?

55 unit tests across four files:

- **`parameterTypeConfig.test.ts`** (21) — type mapping, cell config, filled/valid checks, initial parameters.
- **`ParameterTable.test.ts`** (12) — table wiring: empty state, add/edit/remove, validation hint, required-only flagging.
- **`ParameterTableCell.test.ts`** (11) — per-type rendering, parsing, unknown-type fallback, `aria-invalid`.
- **`ParameterTable.helpers.test.ts`** (11) — cell defaults, grid template, cell validity.

Scroll and sticky behaviour is layout, which jsdom cannot exercise, so it was measured in the browser: 0px leak at both pinned edges in Chromium, WebKit and Firefox, at device pixel ratios 1–2, with no artifact at the container's rounded corners. Also checked by hand in the operator dialog via the part 3 example, and in Storybook.

- [ ] Yes


lightly_studio_view/src/lib/components/Operator/ParameterTable/ParameterTableCell/ParameterTableCell.svelte
```diff
@@ -0,0 +1,41 @@
+<script lang="ts">
+    import { Checkbox } from '$lib/components/ui/checkbox';
+    import { Input } from '$lib/components/ui/input';
+    import type { OperatorParameterColumn } from '$lib/hooks';
+    import { getCellConfig, type ParameterTableRow } from '../../parameterTypeConfig';
+
+    interface Props {
+        column: OperatorParameterColumn;
```
ikondrat, line 9: add explaining jsdoc comments please for params

ikondrat (commented):
I am wonder if we can rather use existing component to render the data in a table way + have some wrappers to add/remove data?

E.g. https://shadcn-svelte.com/docs/components/data-table

ikondrat (approved):
I would approve but not sure about target UX of this

I would rely on @JonasWurst

PR #1825, opened 2026-07-31
Raise a clear error when a DuckDB file is already locked by another process

gabrielfruet:
## What has changed and why?

DuckDB only allows one process to hold a database file open at a time. Running two lightly_studio processes against the same file (e.g. `lightly_studio gui` plus a dataset-loading script) makes the second one fail with a raw `duckdb.IOException`/`sqlalchemy.exc.OperationalError` traceback that doesn't say what's actually wrong. This catches that specific error and re-raises it as a `RuntimeError` naming the database path and telling the user to close the other process or point at a different DB via `LIGHTLY_STUDIO_DATABASE_URL`.

## How has it been tested?

Added `test_database_engine__duckdb_lock_conflict_raises_clear_error`, which mocks `SQLModel.metadata.create_all` to raise the exact exception shape DuckDB produces on a lock conflict, and asserts it's translated into the new `RuntimeError`.

I also reproduced the real conflict manually with two independent processes pointed at the same DuckDB file and confirmed the second one now prints the clear message instead of the raw traceback. A fully automated cross-process test turned out too OS-dependent to run reliably (subprocess-spawned children didn't consistently see the lock in this environment), so the automated test targets the translation logic directly instead.

- [x] Yes

## Summary by CodeRabbit

* **Bug Fixes**
  * Opening a DuckDB database already in use by another process now shows a clear, actionable error message instead of a raw database traceback.
  * Other database errors continue to be reported normally.

* **Documentation**
  * Added a changelog entry describing the improved database-lock error handling.

### DIFF
--- a/lightly_studio/src/lightly_studio/database/db_manager.py
+++ b/lightly_studio/src/lightly_studio/database/db_manager.py
@@ -320,12 +320,9 @@ def _create_duckdb_schema(engine: Engine, engine_url: str) -> None:
     except OperationalError as e:
         if isinstance(e.orig, duckdb.IOException) and "Could not set lock" in str(e.orig):
             raise RuntimeError(
-                f"Could not open the DuckDB database at {engine_url!r}: another process "
-                "already has it open. DuckDB only allows one process to access a "
-                "database file at a time, so close the other lightly_studio process "
-                "(e.g. the GUI server or another script) before running this one, or "
-                "point this process at a different database with "
-                "LIGHTLY_STUDIO_DATABASE_URL."
+                f"DuckDB database at {engine_url!r} is locked by another process. "
+                "Close the other lightly_studio process, or set "
+                "LIGHTLY_STUDIO_DATABASE_URL to point at a different database."
             ) from e
         raise
 
--- a/lightly_studio/src/lightly_studio/database/db_manager.py
+++ b/lightly_studio/src/lightly_studio/database/db_manager.py
@@ -66,6 +66,7 @@ class DatabaseEngine:
 
         Raises:
             FileNotFoundError: If must_exist is True and the database does not exist.
+            RuntimeError: If the DuckDB file is already open in another process.
         """
         if engine_url is not None:
             self._engine_url = engine_url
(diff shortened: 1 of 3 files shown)


lightly_studio/tests/database/test_db_manager.py
```diff
@@ -213,6 +215,34 @@ def test_close__removes_wal_and_allows_reconnect(
     db_manager.close()
 
 
+def test_database_engine__duckdb_lock_conflict_raises_clear_error(
+    tmp_path: Path,
```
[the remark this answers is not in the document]
gabrielfruet: Kept it as a flat function — every other test in this file is already flat (`test_connect`, `test_set_engine__file_db`, etc.), so wrapping just this one test in `TestDatabaseEngine` would be inconsistent with the rest of the file rather than following it.

lightly_studio/src/lightly_studio/database/db_manager.py
```diff
@@ -310,6 +313,23 @@ def _initialize_postgres_schema(
...
+        SQLModel.metadata.create_all(engine)
+    except OperationalError as e:
+        if isinstance(e.orig, duckdb.IOException) and "Could not set lock" in str(e.orig):
+            raise RuntimeError(
+                f"Could not open the DuckDB database at {engine_url!r}: another process "
+                "already has it open. DuckDB only allows one process to access a "
+                "database file at a time, so close the other lightly_studio process "
+                "(e.g. the GUI server or another script) before running this one, or "
+                "point this process at a different database with "
+                "LIGHTLY_STUDIO_DATABASE_URL."
```
JonasWurst: Can we shorten this message?

lightly_studio/tests/database/test_db_manager.py
```diff
@@ -213,6 +215,41 @@ def test_close__removes_wal_and_allows_reconnect(
...
+    mocker: MockerFixture,
+    patch_engine_singleton: None,  # noqa: ARG001
+) -> None:
+    """Test that a DuckDB lock conflict (another process has the file open) fails clearly."""
+    lock_error = duckdb.IOException(
```
JonasWurst, line 224: This can break silently. lets at least add some kind of regression test: assert issubclass(duckdb.IOException, duckdb.OperationalError)

JonasWurst (approved):
Please check the comments before merging

PR #1831, opened 2026-08-03
fast_track: add a full-suite coverage base

JonasWurst:
## What has changed and why?

Adds `shared/full-suite-coverage.ts`: a coverage guardrail factory that **reads**
the coverage report the CI workflow step from #1838 produces, instead of running
tests itself. It owns env resolution, the skip/fail rules, added-line extraction,
the per-file 90% comparison and summary formatting. A config contributes only
which files are its own and how to parse its report format, mapping report keys
to repo-relative paths so the base only ever sees those.

Purely additive. Nothing imports it yet, and the existing `coverage-base.ts`
factory is untouched apart from a three-line comment pointing new guardrails at
this one — so no guardrail changes behaviour in this PR. `backend/coverage` moves
across in #1832; `frontend/coverage` follows, which is when `coverage-base.ts`
can be deleted.

## How has it been tested?

`make static-checks` and `make test` in `fast_track` (219 tests). The new base is
covered by `shared/full-suite-coverage.test.ts` across every verdict branch: no
files in scope, env var unset, red suite, missing report, a file absent from the
report, and per-file judging at the threshold boundary.

- [ ] Yes

---

**Stack** — review and merge bottom-up:

1. ~~#1838 — CI produces the coverage report~~ **merged**
2. #1831 — shared coverage base (base: main)  **← this PR**
3. #1832 — backend guardrail moves onto it, and reads the report (base: #1831)

Reordered from the original stack so the report producer lands first and each
layer reads top-to-bottom.

Stacked manually: each PR targets the branch below it. GitHub's native
stacked-PR view is not enabled on this repository yet.


fast_track/src/guardrails/shared/full-suite-coverage.ts
```diff
@@ -0,0 +1,128 @@
...
+ * file is judged on its own added lines, so one well-covered file cannot carry
+ * an uncovered one.
+ */
+export function createCoverageGuardrail(config: CoverageConfig): Guardrail {
+    return {
+        name: config.name,
+        required: true,
+        needsPrContext: false,
+
+        async run(ctx: GuardrailContext): Promise<GuardrailOutcome> {
```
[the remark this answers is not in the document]
JonasWurst: Done in 42079b9c — extracted `loadReport(config)`, returning a `{ report } | { outcome }` lookup. `run` is now 6 lines.

fast_track/src/guardrails/shared/full-suite-coverage.ts
```diff
@@ -0,0 +1,128 @@
...
+    return files.flatMap((file) => {
+        // Deleted files have no lines to cover; patch-less files (binary, or a
+        // diff too large for the API) cannot be line-filtered.
+        if (file.status === 'deleted' || file.patch === undefined) return [];
+        const addedLines = extractAddedLines(file.patch);
+        return addedLines.size > 0 ? [{ path: file.path, addedLines }] : [];
+    });
+}
+
+function judge(scoped: ScopedFile[], report: LineCoverage): GuardrailOutcome {
```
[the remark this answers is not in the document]
JonasWurst: Done in 42079b9c — extracted `judgeFile`, so `judge` only maps and assembles the summary.

fast_track/src/guardrails/shared/full-suite-coverage.ts
```diff
@@ -0,0 +1,128 @@
+import { existsSync, readFileSync } from 'node:fs';
```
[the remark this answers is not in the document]
JonasWurst: Not doing this — kebab-case is the convention throughout `fast_track/src`: `coverage-base.ts`, `run-guardrails.ts`, `build-verdict.ts`, `author-opt-out.ts`, `api-context.ts`, `verdict-policy.ts`. Renaming one file would make it the odd one out.

The camelCase rule is in `ai_guidelines/frontend.md`, which scopes itself to the SvelteKit app. The cited `AGENTS.md:L43-45` is the review-guidelines section and says nothing about filenames.

fast_track/src/guardrails/shared/full-suite-coverage.ts
```diff
@@ -0,0 +1,128 @@
...
+    /** Lines the coverage tool considers executable. */
+    executable: Set<number>;
+    /** Executable lines that were actually executed. */
+    covered: Set<number>;
+}
+
+/** Per-file line coverage, keyed by repo-relative path. */
+export type LineCoverage = Map<string, FileLineCoverage>;
+
+export interface CoverageConfig {
```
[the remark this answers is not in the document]
JonasWurst: Not doing this, for the same scoping reason — the "avoid exporting/importing types" rule is `ai_guidelines/frontend.md` guidance for the SvelteKit app, and the cited `AGENTS.md:L43-45` does not cover it.

`fast_track` exports its config and param interfaces everywhere: `context/types.ts` (`Guardrail`, `ChangedFile`, `GuardrailContext`), `registry.ts` (`SelectOptions`), `api-context.ts` (`ApiGuardrailContextParams`), and `coverage-base.ts`'s own `CoverageConfig`. `CoverageConfig` here is an object a caller constructs, so it needs a name; `Parameters<typeof createCoverageGuardrail>[0]` at every call site would be worse.

fast_track/src/guardrails/shared/full-suite-coverage.ts
```diff
@@ -0,0 +1,128 @@
...
+            const scoped = filesInScope(config.filterFiles(files));
+            if (scoped.length === 0) {
+                return { status: 'pass', summary: '0 file(s) checked.' };
+            }
+
+            const reportPath = process.env[config.coverageJsonEnvVar];
+            if (reportPath === undefined || reportPath === '') {
+                return {
+                    status: 'pass',
+                    summary: `coverage skipped: ${config.coverageJsonEnvVar} not set (no full-suite report available outside CI).`
```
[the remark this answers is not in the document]
JonasWurst: Checked this against the workflow and the premise does not hold, in both halves.

**"The report variable is missing or misspelled in the CI workflow."** `BACKEND_COVERAGE_JSON` is set in the `env:` block of the guardrail step itself, as a literal `${{ github.workspace }}/...` expression with no condition. There is no run in which it is unset.

**"The same pass occurs if a red test step fails before exporting the path."** That `env:` block belongs to the *guardrail* step, not the test step, so it is evaluated regardless of how the test step ended. A red suite gives `outcome == 'failure'` -> `BACKEND_TESTS_PASSED=false` -> the guardrail already fails with `test suite failed`.

Every way to reach the guardrail without a report already fails closed:

| how | behaviour |
| --- | --- |
| test step skipped (no backend source `.py` changed) | `pass - 0 file(s) checked`, before env is read. The guardrail's filter is a strict subset of the workflow's `backend_src` paths filter, so a skipped step implies nothing to judge. |
| suite ran but died before writing the report (collection error, OOM, missing extra) | non-zero exit -> `BACKEND_TESTS_PASSED=false` -> fail, `test suite failed` |
| report written somewhere other than the configured path | fail, `coverage report missing: <path>` |

The only blind pass needs someone to edit the workflow. I did prototype a `GITHUB_ACTIONS` gate for exactly that regression, then dropped it: it is ~40 lines on a PR already over `diff-size`, guarding a state the workflow cannot currently produce, and the existing behaviour is not silent either — `coverage skipped: BACKEND_COVERAGE_JSON not set` shows up in the bot's comment table.

fast_track/src/guardrails/shared/coverage-base.ts
JonasWurst, line 1: those comments are not required. remove them. no change on this file

fast_track/src/guardrails/shared/full-suite-coverage.test.ts
JonasWurst, line 1: make all the tests DAMP. at the moment you need to always rememebr how the diff looked like and what the report format holds. Using more descriptive methods and keywords should help

fast_track/src/guardrails/shared/full-suite-coverage.ts
```diff
@@ -0,0 +1,134 @@
...
+export type LineCoverage = Map<string, FileLineCoverage>;
+
+export interface CoverageConfig {
+    name: string;
+    /** Env var holding the path to the report the workflow's test step produced. */
+    coverageJsonEnvVar: string;
+    /** Env var set to `false` when that test step ended red. */
+    testsPassedEnvVar: string;
+    filterFiles(files: ChangedFile[]): ChangedFile[];
+    /** Normalises the raw report, mapping its keys to repo-relative paths. */
```
JonasWurst: highlight that this is the main difference and to be implemented by the frontend/backend .ts

fast_track/src/guardrails/shared/coverage-base.ts
JonasWurst: Done in 8f87b9bf — `coverage-base.ts` is back to its original state, no diff on this file.

fast_track/src/guardrails/shared/full-suite-coverage.ts
JonasWurst: Done in 8f87b9bf — the `parseReport` doc now says it is the only language-specific piece, implemented by `backend/coverage.ts` (coverage.py JSON) and `frontend/coverage.ts` (vitest), and the factory doc lists what the base owns versus what a caller contributes.

fast_track/src/guardrails/shared/full-suite-coverage.test.ts
JonasWurst: Done in 8f87b9bf. The shared `PATCH`/`FILE` constants are gone; each test now spells out its own diff and report:

```ts
it("fails below the 90% threshold: 2 of 3 added lines covered", async () => {
    givenTheReportExists();
    const guardrail = coverageGuardrail({
        parseReport: () => reportFor({ coveredLines: [1, 2], uncoveredLines: [3] })
    });

    const result = await runOn(guardrail, fileAddingLines(3));

    expect(result.status).toBe("fail");
    expect(result.summary).toContain("66.7%");
});
```

- `fileAddingLines(n, path?)` / `fileRemovingLines(n)` — the diff, by what it does rather than raw hunk text.
- `reportFor({ coveredLines, uncoveredLines })` — the report in terms of what ran, so the reader never has to recall the `executable`/`covered` set shape; a line in neither list is not executable, which is what the "no executable added lines" case now shows.
- `givenTheReportExists()` / `givenTheReportPathIsSetButNothingIsThere()` / `givenTheTestSuiteWasRed()` — env and fs setup named after the state, not the mock calls.
- `coverageGuardrail(overrides)` and `runOn(guardrail, ...files)` replace `makeConfig`/`makeCtx`, and threshold test names carry their ratio ("9 of 10 added lines covered").

fast_track/src/guardrails/shared/full-suite-coverage.test.ts
```diff
@@ -0,0 +1,317 @@
...
+    });
+
+    it('skips files whose patch adds no lines', async () => {
+        const result = await runOn(coverageGuardrail(), fileRemovingLines(3));
+
+        expect(result.status).toBe('pass');
+        expect(result.summary).toContain('0 file(s) checked');
+    });
+
+    it('passes with a loud summary when the env var is unset (local run)', async () => {
```
michal-lightly: Side note: `loud summary`? :D

fast_track/src/guardrails/shared/full-suite-coverage.test.ts
```diff
@@ -0,0 +1,317 @@
...
+
+const COVERAGE_ENV = 'TEST_COVERAGE_JSON';
+const PASSED_ENV = 'TEST_TESTS_PASSED';
+const REPORT_PATH = '/tmp/coverage.json';
+
+const MODEL = 'pkg/src/model.py';
+const SERVICE = 'pkg/src/service.py';
+
+/** A changed file whose diff adds `count` new lines, numbered 1..count. */
+function fileAddingLines(count: number, path = MODEL): ChangedFile {
```
michal-lightly: Minor: Can you change the input to be the list of line numbers instead of the count? That makes the tests explicit (don't have to assume it is the first lines). Also for fileRemovingLines.

michal-lightly (approved):

fast_track/src/guardrails/shared/full-suite-coverage.test.ts
JonasWurst: Done in c35cf91b — `fileAddingLines`/`fileRemovingLines` now take the new-file line numbers, and emit one hunk per consecutive block, so a gapped list is a valid diff too:

```ts
const guardrail = coverageGuardrail({
    parseReport: () => reportFor({ coveredLines: [12, 13], uncoveredLines: [14] })
});

const result = await runOn(guardrail, fileAddingLines([12, 13, 14]));
```

Every test now names lines away from 1 so nothing reads as "the first N", and there is a new case for a diff with a gap: added 12, 13, 15 pass at 100% while the untouched, uncovered line 14 in between is excluded.

fast_track/src/guardrails/shared/full-suite-coverage.test.ts
JonasWurst: renamed to "passes with an explanatory summary when the env var is unset (local run)". Same in the backend test on #1832.

PR #1832, opened 2026-08-03
fast_track: move backend/coverage onto the full-suite base

JonasWurst:
## What has changed and why?

`backend/coverage` no longer runs pytest. CI runs the full suite and writes a
coverage report (#1838). The guardrail reads that report from
`BACKEND_COVERAGE_JSON`. This PR adds the workflow wiring for that variable. It
is what turns `backend/coverage` from `coverage skipped` into a real verdict.

The guardrail judges the added lines of each changed file at 90%, per file, not
pooled. `BACKEND_TESTS_PASSED=false` is a fail, because coverage from a red test
suite is partial data. The workflow writes `!= 'failure'`, not `== 'success'`,
so a skipped test step does not read as a red suite.

Two effects:

- The test-file matching code is gone: `findTestFile`, `matchesTestFile`,
  `subpathOverlap` and the recursive `readdir`. With it goes the hole where a
  changed file that the report did not list passed automatically. A full-suite
  report lists files that no test imports, so an untested new module now fails
  at its real 0%.
- The threshold goes from 80% to 90%. A full-suite run inflates the numbers,
  because `conftest.py` imports the whole application. Imports, decorators and
  class bodies read as covered whether or not a test exercises them.

`examples/` and `vendor/` leave the scope of the guardrail. `examples/` has no
`__init__.py`, so coverage.py's source scan never reaches it. `vendor/` is
third-party code we do not hold to a coverage bar. mypy skips both trees too.

`frontend/coverage` stays on the old `coverage-base.ts` factory. It moves in
LIG-10351, which is when that file can go.

## How has it been tested?

- `make static-checks` and `make test` in `fast_track` (210 tests).
- Against a real coverage report, not a fixture: a covered file passes at 100%,
  a vendored file that no test imports fails at 0%.
- `BASE_REF=origin/main~8 GUARDRAILS=backend/coverage make run-guardrails`:
  11 files judged, 10 at 100%, one at 79.1%.
- Counted the files in scope that the report does not list: 28 → 1 → 0 as
  `examples/` and then `vendor/` leave the scope. The 27 example scripts are
  absent because coverage.py cannot scan them; the last file is one vendored
  module in a subtree without `__init__.py`.

## Did you update [CHANGELOG.md](../CHANGELOG.md)?

- [ ] Yes
- [x] Not needed (internal change)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved backend coverage guardrails by using the shared full-suite coverage report.
[...]

PR #1833, opened 2026-08-03
CI: run the backend suite with coverage for the guardrail

JonasWurst:
## What has changed and why?

Adds `make test-coverage`, which writes `coverage.json` for the whole backend
suite, and the workflow step that runs it and hands the report to the guardrail
through `BACKEND_COVERAGE_JSON` / `BACKEND_TESTS_PASSED`. This is what turns
`backend/coverage` from `coverage skipped` into a real verdict.

- The step is `continue-on-error`: a red suite must not kill the job, because the
  bot still needs the verdict artifact to explain a withheld approval. Its
  outcome is compared with `!= 'failure'`, so a *skipped* step does not read as a
  red suite.
- `make test-coverage` deliberately has no `build` prerequisite — that target's
  `clean` step would delete the webapp dist an earlier workflow step just built.
- The step is gated on a new paths-filter output scoped to backend source rather
  than all of `lightly_studio/**`, so a docs- or test-only PR does not pay for a
  full suite the guardrail would score as `0 file(s) checked`.

Trade-off worth knowing: this adds one full backend suite to the guardrail job's
wall time, duplicating work the Unit Test workflow does on the same commit.
Reading it from that workflow instead needs a `workflow_run` trigger, and that
workflow skips the backend on frontend-only diffs, so "not produced yet" would be
indistinguishable from "never will be".

## How has it been tested?

`make test-coverage` locally: 1792 passed in ~40s, producing one combined
416-file `coverage.json`. That confirms pytest-cov merges xdist worker data
correctly under `-n auto`, which was the one part of the design that could not be
verified without a real run. The guardrail then judged that report end to end
(see the PR below this one). Workflow YAML parsed, and the step order and env
wiring asserted against it.

- [ ] Yes

---

**Stack** — review and merge bottom-up:

1. #1831 — shared coverage base (base: main)
2. #1832 — backend guardrail moves onto it (base: #1831)
3. #1833 — CI produces the report (base: #1832)  **← this PR**

Stacked manually: each PR targets the branch below it. GitHub's native
stacked-PR view is not enabled on this repository yet.

### DIFF
--- a/.github/workflows/fast_track_guardrails.yml
+++ b/.github/workflows/fast_track_guardrails.yml
@@ -44,6 +44,10 @@ jobs:
           filters: |
             backend:
               - 'lightly_studio/**'
+            # Superset of the upcoming backend/coverage guardrail's own scope, so the
+            # full suite runs only for PRs that guardrail will actually judge.
+            backend_src:
+              - 'lightly_studio/src/lightly_studio/**/*.py'
             frontend:
               - 'lightly_studio_view/**'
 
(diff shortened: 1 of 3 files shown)


JonasWurst:
Superseded by #1838. The stack was reordered so the CI step that produces the coverage report lands first; GitHub auto-closed this PR as merged when its head branch became an ancestor of its base during the force-push. Nothing from it landed on `main`.

PR #1834, opened 2026-08-03
Support creating captions with a temporal segment

horatiualmasan:
## What has changed and why?

Add optional temporal bounds to captions so a caption can be created with a time segment.

## How has it been tested?

New tests

- [ ] Yes

## Summary by CodeRabbit

* **New Features**
  * Captions can include optional start and end times.
  * Caption details display associated temporal-span information when available.
  * Captions without timing data remain supported.
  * Valid timing data is automatically associated with captions.

* **Bug Fixes**
  * Invalid or incomplete time ranges are rejected.
  * Associated timing data is removed when captions are deleted.

* **Tests**
  * Added coverage for valid, invalid, and missing temporal-span scenarios.

### DIFF
--- a/lightly_studio/src/lightly_studio/models/caption.py
+++ b/lightly_studio/src/lightly_studio/models/caption.py
@@ -1,12 +1,14 @@
 """This module defines the caption model."""
 
 from datetime import datetime, timezone
+from typing import Optional
 from uuid import UUID
 
 from sqlalchemy.orm import Mapped
 from sqlmodel import Field, Relationship, SQLModel
 
 from lightly_studio.models.sample import SampleTable
+from lightly_studio.models.temporal_span import TemporalSpanTable, TemporalSpanView
 
 
 class CaptionTable(SQLModel, table=True):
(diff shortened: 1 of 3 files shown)


lightly_studio/src/lightly_studio/models/caption.py
```diff
@@ -33,17 +35,32 @@ class CaptionTable(SQLModel, table=True):
...
+
 
 class CaptionCreate(SQLModel):
     """Input model for creating captions."""
 
     parent_sample_id: UUID
     text: str
 
+    # Optional temporal bounds for the caption's sample.
+    start_time_s: Optional[float] = None
```
ikondrat, line 56: I am wonder if using timestamp would be suitable and why it is float?



how it is going to be used?

horatiualmasan: You would define the interval in seconds (for example [0.0, 3.5] ). It would either be loaded alongside the captions if available or created from the GUI at some point (Here I would imagine dragging some sliders in the video timeline and getting the correct timestamp from there).

JonasWurst (approved):

PR #1856, opened 2026-08-05
Introduce dataset splitting prototype

michal-lightly:
## What has changed and why?

Adds the ability to split a dataset into named subsets (e.g. train/val/test).

- **UI:** a "Split dataset" dialog to name splits and set their **relative parts** (e.g. 8 : 1 : 1); the resulting share % and sample count are shown per split. Operates on the current filtered view. Warns which tags will be created and which will be cleared, and confirms before clearing tags that already hold selected samples.
- **Python SDK:** `DatasetQuery.random_split({"train": 8, "val": 1, "test": 1}, seed=...)`, composing with `match`/`slice`.

Splits are stored as ordinary sample tags — no new table or migration. One core implementation backs both the API route and the SDK method. The split is random (always a random seed from the UI; optional seed in the SDK) and re-running clears then reassigns the target tags.

## How has it been tested?

New unit tests for the core split algorithm, the selection-overlap query, the API routes, the SDK method, and the frontend dialog/hooks. Backend (pytest) and frontend (Vitest) suites and static checks pass.

Also manually in the UI.

Screenshot 2026-08-05 at 12 59 41
Screenshot 2026-08-05 at 13 00 37

- [x] Yes


IgorSusmelj (approved):
Tested and it works.

One small nit:
- make it 80, 10, 10 and not 8,1,1 as default for the splits (I'd always think in percentages and having numbers summing up to 100 is easier in my head)

PR #1872, opened 2026-08-06
Infer numeric metadata type when sorting instead of requiring cast_to_float

iunir:
Closes [LIG-9567](https://linear.app/lightly/issue/LIG-9567/orderbymetadatafield-infer-numeric-type-automatically-instead-of).

## What has changed and why?

Sorting by a numeric metadata field ordered lexicographically — `9` after `10` — unless the caller passed `cast_to_float=True`. That flag travelled all the way from the browser, as `is_numeric` on every sort request.

The type is already recorded per key in `metadata_schema`, and the join we need for sorting already brings that column along. So `OrderByMetadataField` now decides in SQL: order by the value as a float when the schema says the field is numerical, then by the raw value. Nothing to pass, nothing to resolve.

Three things for a reviewer:

- `cast_to_float` is gone from the constructor. Drop it from any call — it no longer controls anything.
- `is_numeric` on the request is now optional and ignored. It goes away in a follow-up together with the frontend code that sends it. Not a breaking change: these models use pydantic's default `extra='ignore'`, so a request still carrying it validates.
- The `CAST` wraps the `CASE` instead of sitting inside a branch, so a non-numerical value never reaches the cast whatever order the engine evaluates in.

## How much is there to review?

|  | added | removed |
|---|---:|---:|
| Executable source | 44 | 23 |
| Docstrings & comments | 38 | 17 |
| Tests | 411 | 53 |
| CHANGELOG | 14 | 0 |

The logic is the 44 added source lines, and 26 of those are `order_by.py`. Everything else is
one-line call-site updates: `to_column_element` became `to_column_elements`, so the two consumers
in `get_all_by_collection_id.py` and `get_adjacent_images_window.py` flatten a list now.

Read `OrderByMetadataField` first — the rest follows from it.

## How has it been tested?

New tests cover all three sort paths — image list, adjacent images, and `DatasetQuery` — with values `9 / 10 / 100`, where lexicographic and numeric ordering disagree. String metadata is pinned too, so the fallback can't silently break. Deliberately breaking the numeric detection fails 8 of them.

`make static-checks` and `make test` pass locally on DuckDB; CI covers Postgres, where the JSON extraction and the cast compile differently.

- [x] Yes

## Summary by CodeRabbit

* **New Features**
  * Metadata sorting now automatically follows the configured schema type.
  * Numeric metadata fields sort numerically, while text fields retain lexicographic ordering.
  * Missing metadata values consistently appear last.
  * Numeric sorting provides reliable tie-breaking and improved precision for large values.

* **Deprecations**
  * The `is_numeric` sorting option is deprecated and ignored; schema metadata determines numeric behavior automatically.
  * Explicit numeric-casting options have been removed from sorting configuration.


lightly_studio/src/lightly_studio/core/dataset_query/dataset_query.py
```diff
@@ -334,6 +335,11 @@ def _compose_query(self, query: SelectOfScalar[Table]) -> SelectOfScalar[Table]:
 
         # Apply ordering
         if self.order_by_expressions:
+            metadata_sort_types.resolve_cast_to_float(
+                session=self.session,
+                collection_id=self.dataset.collection_id,
+                order_by=self.order_by_expressions,
+            )
```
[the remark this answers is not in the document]
iunir: done

lightly_studio/src/lightly_studio/core/dataset_query/order_by.py
iunir, line 1: Image

lightly_studio/src/lightly_studio/models/sort.py
```diff
@@ -30,15 +30,17 @@ class SortFieldExpr(BaseModel):
...
-            Only relevant when ``source`` is ``"metadata"``.
+        is_numeric: Deprecated and ignored.  Numeric metadata fields are detected
+            from the collection's stored metadata schema.
     """
 
     source: Literal[SortFieldSource.image, SortFieldSource.metadata]
     field_name: str
     direction: SortDirection
-    is_numeric: bool
+    is_numeric: bool | None = Field(
```
LeonardoRosaa, line 40: I think we can remove it, no?

iunir: Yes, I've already planned removal https://linear.app/lightly/issue/LIG-10452/remove-the-deprecated-is-numeric-sort-field-from-the-api-and-frontend
Will do it after merge since it's require frontend changes

LeonardoRosaa (approved):

PR #1875, opened 2026-08-07
Add: custom embedding model serving for hosted Studio (LIG-10428)

JonasWurst:
## What has changed and why?

Prototype for [LIG-10428](https://linear.app/lightly/issue/LIG-10428/figure-out-custom-embedding-model-hosted), implementing the [decision record](https://linear.app/lightly/document/decision-record-custom-embedding-models-on-hosted-studio-d3daa5d97237).

Customers who index with their own embedding model write vectors straight to the DB. Everything that reads those vectors works, but **search does not**: embedding a query needs a live model, and a hosted instance has none. The chosen approach is that the customer serves the model over HTTP on their own network and the browser calls it directly — our backend is never in the path, so weights and queries never reach our infrastructure.

- **`embedding_model_service` table** keyed on `embedding_model_hash` rather than collection, so a hostname change is one update instead of one per collection (D-8). Plus its Alembic revision.
- **`serving_url` on `set_default_embedding_model`**, persisted where the model row is written so "what produced these vectors" and "where do I get more" cannot drift apart (D-9). An endpoint changes it without re-indexing.
- **Frontend probes `/info` on collection open**, routes text and image queries to the service, and disables search on model mismatch or timeout, re-probing when a search fails (D-12, D-13). It never falls back to the built-in model — those query vectors against custom-model embeddings return a confidently ordered list of unrelated results with no error, and dimensions often match so a dimension check would not catch it.
- **Write-on-read fix (D-18)**, the blocking prerequisite: `GET /has_embeddings` registered the env-default model and overwrote the record of what produced the stored vectors, silently hiding search on custom-model collections. `collection_has_embeddings` is now model-free, and `load_or_get_default_model` refuses to register a generator that does not match the collection's stored model.
- **Reference service** implementing the contract, a **standalone page** for answering the open Private Network Access question, and **customer-facing docs**.

### Notes for reviewers

**Size.** Above the 100-line target because a vertical slice needs the table, the write path, the read path and the frontend together — a partial slice is not demoable and cannot answer the open question the issue exists to settle. Roughly half the diff is tests and docs. The D-18 fix is separable and is the one part with blast radius beyond this feature; happy to split it out if preferred.

**Admin gating is deliberately absent.** D-16 calls for it, but auth is terminated by a proxy outside this repo and D-17 (whether that proxy distinguishes admins) is unresolved, so a real check cannot be written here. Left as a `TODO` on the endpoint rather than a placeholder dependency that would look like a gate without being one. Whoever can write a serving URL can redirect every user's queries to a host they control, with no trace on our servers.

**Open questions this does not close.** D-5 (Private Network Access) needs a browser against a real hosted origin — see below. D-20 stands: video collections are hard-wired to PerceptionEncoder and are documented as unsupported.

**Trust-based identity (D-11).** `embedding_model_hash` is a customer-declared string, not a digest. A forgotten version bump yields silently wrong results; accepted and documented.

## How has it been tested?

**Automated.** Backend `make static-checks` and `make test` (1879 passed). Frontend `make static-checks` and `npm run test:unit` (2505 passed). New coverage: URL validation, the service resolver, both endpoints, the D-18 behaviour in `embedding_manager` and `embedding_utils`, the contract client, and search routing/refusal in `useSearchEmbedding`.

**Contract round trip verified numerically.** The reference service returns vectors bit-identical (max abs diff `0.0`, text and image) to the in-process `MobileCLIPEmbeddingGenerator` used for indexing, so going over HTTP preserves the embedding space.

**Alembic revision is hand-written** — Docker was unavailable locally. Its DDL was verified to match `SQLModel.metadata` exactly, but `make migration-check-postgresql` has **not** been run. Please run it before merge.

**Not yet run: the browser.** `useEmbeddingService` has no direct unit test (it is a runes hook needing a component harness plus a `QueryClient`); it is covered indirectly through `useSearchEmbedding`. Manual verification planned:

1. Local: default model regression, then a custom model with `serving_url="http://localhost:8123"`, then the negative case — a custom-model collection reopened with no service must *disable* search, not return results.
2. Hosted: Studio on EC2 over **HTTPS** (mandatory — private-network requests from an `http://` public origin are blocked outright with no header opt-in, which looks identical to a PNA failure and would produce a false negative on D-5), with the model served locally.

`example_embedding_service_pna_check.html` isolates D-5 from CORS and TLS so a failure has one interpretation.

- [x] Yes

### DIFF
--- a/lightly_studio/src/lightly_studio/api/routes/api/embedding_model.py
+++ b/lightly_studio/src/lightly_studio/api/routes/api/embedding_model.py
@@ -1,0 +1,127 @@
+"""This module contains the API routes for embedding models and their serving endpoints.
+
+A collection's embedding model may live outside this process: customers who index with
+their own model run it behind an HTTP service on their own network, and the browser calls
+that service directly for search queries. These routes expose which model a collection was
+embedded with and where it is served from.
+"""
+
+from __future__ import annotations
+
+from typing import Annotated
+from uuid import UUID
+
+from fastapi import APIRouter, Depends, HTTPException, Path
+from pydantic import BaseModel
+
+from lightly_studio.api.routes.api import collection
+from lightly_studio.api.routes.api.status import (
+    HTTP_STATUS_BAD_REQUEST,
+    HTTP_STATUS_NOT_FOUND,
+)
+from lightly_studio.database.db_manager import SessionDep
+from lightly_studio.models.collection import CollectionTable
+from lightly_studio.resolvers import embedding_model_resolver, embedding_model_service_resolver
+
+embedding_model_router = APIRouter()
+
+
+class EmbeddingModelView(BaseModel):
+    """The model a collection's stored embeddings were produced with.
+
+    Attributes:
+        embedding_model_hash: A customer-declared identifier, not a digest of the
+            weights. The rule is to bump it whenever the output vectors change.
+        serving_url: Where a live copy of the model is served for search queries, or
+            None when the model runs inside this process. The browser calls it directly,
+            so this host must be reachable from the user's machine.
+    """
+
+    embedding_model_id: UUID
+    name: str
+    embedding_model_hash: str
+    embedding_dimension: int
+    serving_url: str | None
+
+
+class UpdateServingUrlRequest(BaseModel):
+    """Request body for pointing a model at an embedding service.
+
+    Attributes:
+        serving_url: Base URL of the service, or None to stop serving the model.
+    """
+
+    serving_url: str | None
+
+
+@embedding_model_router.get("/collections/{collection_id}/embedding_model")
+def read_embedding_model(
+    session: SessionDep,
+    collection_row: Annotated[
+        CollectionTable,
+        Path(title="Collection Id"),
+        Depends(collection.get_and_validate_collection_id),
+    ],
+) -> EmbeddingModelView | None:
+    """Return the collection's embedding model, or None if it has none."""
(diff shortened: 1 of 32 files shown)


lightly_studio/src/lightly_studio/api/routes/api/embedding_model.py
```diff
@@ -0,0 +1,127 @@
...
+from lightly_studio.resolvers import embedding_model_resolver, embedding_model_service_resolver
+
+embedding_model_router = APIRouter()
+
+
+class EmbeddingModelView(BaseModel):
+    """The model a collection's stored embeddings were produced with.
+
+    Attributes:
+        embedding_model_hash: A customer-declared identifier, not a digest of the
```
gabrielfruet, line 33: For future (already documented [here](https://linear.app/lightly/issue/LIG-10475/figure-out-improve-embeddings-in-lightlystudio)) I think we should instead of expecting the user to manage this `embedding_model_hash` we could just do a simple request for the Embedding model with 8 images or so to calculate their embeddings and then calculate the hash of the output, that would be called a **fingerprint** 

It's statistically impossible for two different models to have the same embedding for the same 8 images. The only failure mode is when the user explicitly set an specific embedding for these 8 images, which in anyway the user is already trying to make things go wrong haha. 

And we would also add a very nice UX feature, preventing users from attaching embedding models with same name, same hash but different embeddings.

PR #1882, opened 2026-08-07
Allow choosing the MobileCLIP model size for image embeddings

gabrielfruet:
## What has changed and why?

Image embeddings used a single hardcoded model, `mobileclip_s0`, the smallest and weakest MobileCLIP variant. This change lets a caller pick a bigger model when retrieval needs higher-quality embeddings.

`MobileCLIPEmbeddingGenerator` now accepts a `model_name` parameter (`mobileclip_s0`, `mobileclip_s1`, `mobileclip_s2`, or `mobileclip_b`) and a `max_batch_size` parameter. `lightly_studio` exports the class, so a caller can register a bigger model with `set_default_embedding_model` before dataset creation. The export is lazy: `import lightly_studio` still does not load torch until the class is used.

The embedding dimension now comes from the vendored model config file, not a hardcoded constant. An unsupported model name now raises a clear error before any download attempt, instead of failing with a confusing 404.

Docs for image datasets show how to pick a bigger model.

Video embeddings (PerceptionEncoder) are a separate follow-up, not part of this change.

## How has it been tested?

- Added tests: construct the generator with a non-default model, reject an invalid model name, and check that every vendored config reports the same embedding dimension (no network access needed for that last one).
- `make static-checks` and the full backend test suite pass.

- [x] Yes

## Summary by CodeRabbit

* **New Features**
  * Added support for selecting larger MobileCLIP embedding models.
  * Added configurable batch size to help manage memory usage.
  * Exposed `MobileCLIPEmbeddingGenerator` through the public SDK.
  * Added validation with clear errors for unsupported model names.

* **Documentation**
  * Documented available models, download and performance considerations, batch-size configuration, and default-model behavior.
  * Added an unreleased changelog entry.

* **Tests**
  * Added coverage for custom models, invalid model names, and embedding dimensions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Python SDK: `MobileCLIPEmbeddingGenerator` now accepts a `model_name` parameter to pick a bigger MobileCLIP variant (`mobileclip_s1`, `mobileclip_s2`, `mobileclip_b`) for higher-quality image embeddings.
+
 ### Changed
 
 ### Deprecated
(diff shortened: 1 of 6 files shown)


gabrielfruet:
we will need a design doc

PR #1888, opened 2026-08-10
Fix: relax image performance test threshold

IgorSusmelj:
- Increase the image render-time limit from 6 seconds to 6.5 seconds.
- Reduce flaky failures when CI performance is slightly slower than usual.

## Summary by CodeRabbit

* **Tests**
  * Updated the image rendering performance threshold to allow completion within 6.5 seconds.


gabrielfruet (approved):
LGTM

PR #1892, opened 2026-08-10
Add endpoint listing per-annotation evaluation metrics

JonasWurst:
Part 2 of 8 for Sort by annotation metric.

## What has changed and why?

Adds `GET /collections/{collection_id}/evaluation/metrics/annotation/info`, listing the
per-annotation metrics of every evaluation run involving an annotation source. It populates the
annotations grid's sort options later in this stack.

The query is scoped to the annotation source rather than the dataset, so it can only offer
options the backend can resolve to a side of a run's pairing. Null metric names are excluded and
runs left without any metric name are omitted.

The view is deliberately not shared with the sample-side `EvaluationRunMetricsInfoView`: this one
needs `run_id` (annotation sorting keys on the run ID, not the name) and `task_type`, and it
carries no value bounds, because how bounds interact with unmatched annotations is decided by the
threshold filter (LIG-10116). It uses a list-of-objects shape so LIG-10116 can add bounds to an
entry without changing the list shape or the frontend's mapping code.

**This PR adds code nothing calls yet.** The endpoint's only consumer is the sort menu in part 8.
That is the expected cost of slicing a vertical feature — it is not dead code.

## How has it been tested?

5 resolver tests in `tests/resolvers/evaluation_annotation_metric_resolver/test_get_metrics_info_by_collection_id.py`:
each side of the pairing is listed, null metric names are excluded, runs with no per-annotation
rows are omitted, and runs of other annotation sources are excluded. One shallow endpoint test in
`tests/api/routes/api/annotations/test_get_annotation_metrics_info.py` covers the response shape.

`cd lightly_studio && make static-checks && make test`: ruff clean, 1772 passing.

- [ ] Yes

## Summary by CodeRabbit

* **New Features**
  * Added an API endpoint to retrieve annotation evaluation metrics for a collection.
  * Results include evaluation run details and available metric names, including IoU and disagreement metrics.
  * Metrics are grouped by evaluation run, deduplicated, sorted, and exclude unavailable entries.

* **Tests**
  * Added coverage for ground-truth and prediction collections, unsupported sources, and runs without metrics.


lightly_studio/src/lightly_studio/models/evaluation_annotation_metric.py
```diff
@@ -35,6 +38,25 @@ class EvaluationAnnotationMetricTable(SQLModel, table=True):
...
+    """One per-annotation metric name recorded by an evaluation run."""
+
+    metric_name: str
+
+
+class EvaluationRunAnnotationMetricsInfoView(BaseModel):
+    """Per-annotation metrics recorded by one evaluation run.
+
+    Carries no value bounds: how bounds interact with unmatched annotations, whose
+    ordering value is coalesced to zero, is decided by the threshold filter (LIG-10116).
```
JonasWurst:
```suggestion
    ordering value is coalesced to zero, is decided by the threshold filter.
```

lightly_studio/tests/api/routes/api/annotations/test_get_annotation_metrics_info.py
```diff
@@ -0,0 +1,127 @@
...
+            {
+                "run_id": str(run.id),
+                "run_name": "detection eval",
+                "task_type": "object_detection",
+                "metrics": [
+                    {"metric_name": "disagreement"},
+                    {"metric_name": "iou"},
+                ],
+            }
+        ]
```
JonasWurst: write this out, do not use a for loop. the result dict can be defined first and then asserted

lightly_studio/tests/api/routes/api/annotations/test_annotation_metrics_info.py
JonasWurst, line 1: lets move the testing of the resolver logic to a different test for the resolver. the api test can be very shallow. if somwone is adding more logic, this will grow

lightly_studio/src/lightly_studio/api/routes/api/annotations/get_annotation_metrics_info.py
```diff
@@ -0,0 +1,43 @@
...
+
+from fastapi import APIRouter, Path
+
+from lightly_studio.database.db_manager import SessionDep
+from lightly_studio.models.evaluation_annotation_metric import (
+    EvaluationRunAnnotationMetricsInfoView,
+)
+from lightly_studio.resolvers import evaluation_annotation_metric_resolver
+
+get_annotation_metrics_info_router = APIRouter()
```
iunir:
```suggestion
annotation_metrics_info_router = APIRouter(prefix="/evaluation/metrics/annotation/info")
```

1. I would simplify name of router and **file name** as well, since `get` is method. Maybe we will later support `post` or `delete`.
2. I recommend to use router prefix, this way, routes themselves will be simpler

lightly_studio/src/lightly_studio/api/routes/api/annotations/get_annotation_metrics_info.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+@get_annotation_metrics_info_router.get(
+    "/evaluation/metrics/annotation/info",
```
iunir:
```suggestion
    "/",
```

lightly_studio/src/lightly_studio/resolvers/evaluation_annotation_metric_resolver/get_metrics_info_by_collection_id.py
```diff
@@ -0,0 +1,77 @@
...
+            or_(
+                col(EvaluationRunTable.gt_annotation_collection_id) == collection_id,
+                col(EvaluationRunTable.pred_annotation_collection_id) == collection_id,
+            ),
+            col(EvaluationAnnotationMetricTable.metric_name).is_not(None),
+        )
+        .distinct()
+    )
+
+    rows = session.execute(stmt).mappings().all()
```
iunir, line 57: Should we limit it? Maybe last 10000 records, to not overflow with too big DB output?

iunir (approved):
Not critical, address if you agree

lightly_studio/src/lightly_studio/api/routes/api/annotations/get_annotation_metrics_info.py
JonasWurst: Done: file is now `annotation_metrics_info.py`, router `annotation_metrics_info_router` with the prefix.

lightly_studio/src/lightly_studio/api/routes/api/annotations/get_annotation_metrics_info.py
JonasWurst: Used `""` instead of `"/"` so the path stays without a trailing slash (no redirect).

lightly_studio/src/lightly_studio/resolvers/evaluation_annotation_metric_resolver/get_metrics_info_by_collection_id.py
JonasWurst: I would leave it unlimited. The query is `DISTINCT` over (run, metric_name) for one annotation source, so the row count is runs x metric names, not annotations. A limit would silently drop sort options.

PR #1901, opened 2026-08-10
Remove the deprecated is_numeric sort field from the API and frontend

iunir:
## What has changed and why?

`is_numeric` on sort expressions did nothing. The frontend looked up each metadata field's type, computed the flag, and sent it on every sort request — but since #1872 the backend ignores it and reads the type from `metadata_schema` instead. Removing it: one less field for new clients to copy, three fewer hooks maintaining a dead value.

Backend: dropped from `SortFieldExpr` and `EvaluationMetricSortExpr`. Frontend: dropped from `ImageSortField` and the three places that set it. `useSortFields` still fetches metadata info for the sortable-field list — only the type computation went.

Both sides ship together: the API client is generated from the schema, so the field leaving the model breaks any frontend code still setting it.

Old clients are fine — pydantic's `extra='ignore'` drops the key.

## How has it been tested?

Frontend static-checks and unit tests clean, run against a freshly generated client. Backend suite passes. A couple of local failures (missing `gcsfs`/`s3fs` stubs, and the static-webapp test because I stubbed the frontend dist folder instead of building it) reproduce on `main` too.

No visible change — sorting behaves exactly as before.

- [x] Yes

## Summary by CodeRabbit

- **Breaking Changes**
  - Removed the deprecated numeric-sort option from image and metadata sorting controls and request formats.
  - Existing sorting fields, directions, and metadata selections continue to work as before.

- **Bug Fixes**
  - Legacy requests containing the removed option remain compatible; the option is ignored and excluded during validation.
  - Updated default sorting and sort controls now generate cleaner request data.

- **Documentation**
  - Updated the unreleased changelog to document the sorting request format change.


lightly_studio/src/lightly_studio/models/sort.py
```diff
@@ -30,17 +30,11 @@ class SortFieldExpr(BaseModel):
...
-        is_numeric: Deprecated and ignored.  Numeric metadata fields are detected
-            from the collection's stored metadata schema.
     """
 
     source: Literal[SortFieldSource.image, SortFieldSource.metadata]
     field_name: str
     direction: SortDirection
-    is_numeric: bool | None = Field(
-        default=None,
-        description="Deprecated and ignored, will be removed in the next major version.",
```
gabrielfruet, line 42: Why are we removing this if we are not planning a major release yet?

iunir: Thanks, good point.
I've deprecated it just yesterday. Actually this is continuation of #1872 (frontend alignment). 
Since we are not restricting `extra="forbid"` anywhere in lightly_studio, default `extra="ignore"` will be used. So old clients, sending is_numeric param will continue to work.

lightly_studio/tests/models/test_sort.py
```diff
@@ -202,16 +180,19 @@ def test_sort_expr_discriminated_union__routes_to_sort_field_expr() -> None:
     assert expr.field_name == "file_name"
 
 
-@pytest.mark.parametrize("is_numeric", [True, False, None])
-def test_sort_field_expr__accepts_deprecated_is_numeric(is_numeric: bool | None) -> None:
-    """Test that request bodies from older clients still validate."""
+def test_sort_field_expr__ignores_removed_is_numeric() -> None:
```
LeonardoRosaa: I think we can also remove this test: `test_sort_field_expr__ignores_removed_is_numeric`. If a user updates the package, the frontend and backend will be updated too, so it would not be a breaking change.

iunir: Thanks! Removed

LeonardoRosaa (approved):

PR #1902, opened 2026-08-10
Wire isPending on async action buttons

gabrielfruet:
## What has changed and why?

Migrates 8 raw/ad-hoc buttons to the shared `Button` component and wires its `isPending` prop for async actions: save/create caption, remove/rename/delete tag, run operator, trigger evaluation, and export samples. Each button now shows a busy indicator and disables itself while its action is in flight, instead of no feedback or a manually-implemented label swap/spinner.

`ExportSamples`'s samples-tab download previously only tracked the count-fetch loading state, not the export/download click itself — added a dedicated pending state for that.

This is the first of a few follow-up PRs consolidating the rest of the app onto the shared `Button` component; the remaining files are consistency-only (no behavior change) and will land separately.

## How has it been tested?

- `make static-checks` and the full frontend unit suite pass.
- Manually exercised each flow against a running dev server + example dataset: create/save/delete caption, assign/remove/rename/delete tag (both the per-sample panel and the bulk tags menu), and the samples export download — confirmed the pending indicator appears and the button re-enables after.
- `OperatorDialog` and `TriggerEvaluationDialog` weren't reachable live in this dataset (no registered plugins / no annotation sources), but use the identical `isPending`/`buttonProps` pattern verified elsewhere.

- [x] Yes

## Summary by CodeRabbit

- **New Features**
  - Added visible progress indicators while saving captions, removing or renaming tags, executing operators, triggering evaluations, and exporting samples.
  - Updated action controls with consistent loading, disabled, and accessibility behavior.
  - Tag actions now clearly indicate which item is being updated or removed.
  - Prevented duplicate caption submissions and maintained independent states for concurrent tag removals.

- **Documentation**
  - Added an Unreleased changelog entry describing the new busy states.

- **Tests**
  - Expanded coverage for progress indicators and concurrent action handling.


LeonardoRosaa (approved):
LGTM

PR #1912, opened 2026-08-11
Split the general E2E jobs into three shards

gabrielfruet:
## What has changed and why?

The `general` Playwright project has 67 tests and runs with `workers: 1`, so a shard's wall clock is just the sum of its tests. Across two shards the heavier one carried 35, against the step's 6-minute `timeout-minutes` budget. A third shard brings the worst case down to 25.

Both matrix legs get the extra shard, taking four jobs to six: `General 1-3 of 3` and `General Postgres 1-3 of 3`, each with its own success-marker cache key and report artifact.

Also added the `list` reporter alongside `html`. The HTML report is only uploaded on failure, so reading per-test durations means downloading the artifact; `list` streams each test to the job log as it finishes.

## How has it been tested?

`npx playwright test --project=general --shard=N/3 --list` locally: 25 / 20 / 22 tests, against 35 / 32 before. No test code changed, only the workflow matrix.

- [ ] Yes

### DIFF
--- a/.github/workflows/end2end_test.yml
+++ b/.github/workflows/end2end_test.yml
@@ -206,21 +206,29 @@ jobs:
             success_key: e2e-success-evaluations
             artifact_name: Playwright Report (Evaluations)
             postgres: false
-          - name: General Postgres 1 of 2
+          - name: General Postgres 1 of 3
             index_script: index_general.py
             playwright_project: general
-            shard: "1/2"
+            shard: "1/3"
             success_marker: .e2e-success-general-postgres-1
             success_key: e2e-success-general-postgres-1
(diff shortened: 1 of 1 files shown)


michal-lightly (changes requested):
Also Postgres tests should be split

.github/workflows/end2end_test.yml
```diff
@@ -375,7 +391,7 @@ jobs:
         working-directory: lightly_studio_view
         run: >-
           npx playwright test
-          --reporter=html
+          --reporter=html,list
```
gabrielfruet, line 394: MOT: it's easier for agents to debugging using this list rather than the html

michal-lightly (approved):

PR #1914, opened 2026-08-11
Track app launches from the backend in PostHog

gabrielfruet:
## What has changed and why?

We have no signal on how many people run `lightly-studio quickstart`.

`quickstart` and `gui` now send an `app_launched` event with a `launch_source` property from Python. The SDK sits behind a `Tracker` protocol in `lightly_studio/analytics/`, so swapping it later touches one module. `LIGHTLY_STUDIO_ANALYTICS_ENABLED=false` opts out of the Python package and the browser UI together.

- `posthog` is pinned below 7.0 on every Python: 7.x dropped 3.9, and 6.9 covers the whole `>=3.9,<3.15` range on its own.
- Events are keyed on a random UUID at `~/.cache/lightly-studio/install_id`, never written while tracking is off.
- The PostHog key is a plain default in `env.py`, not build-time injection. It is write-only and already in every PyPI wheel's webapp bundle.
- A source checkout reports to the dev project instead, set in `lightly_studio/.env.example`. If you already have a `lightly_studio/.env`, add the `LIGHTLY_STUDIO_POSTHOG_KEY` line by hand, or your local runs count as production launches.

## How has it been tested?

Full suites both sides (1883 backend, 2545 frontend), static checks clean. New tests cover the no-op paths, install ID reuse, one event per CLI invocation, and eight threads racing to report an event building one tracker between them. Against a closed port, delivery fails quietly and the CLI still exits in under four seconds. An autouse fixture in `tests/conftest.py` keeps the whole suite off the network whatever the local `.env` says.

- [x] Yes


lightly_studio_view/src/lib/hooks/usePostHog.ts
```diff
@@ -67,8 +67,27 @@ export const usePostHog = () => {
...
+     * @param properties - Properties to attach to all following events (e.g., launch_source)
+     *
+     * @example
+     * ```ts
+     * registerSuperProperties({ launch_source: 'quickstart' });
+     * ```
+     */
+    const registerSuperProperties = (properties: Record<string, unknown>) => {
+        if (!initialized) return;
+        posthog.register(properties);
```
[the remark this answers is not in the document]
gabrielfruet: Fixed, now uses `register_for_session`.

lightly_studio_view/src/lib/hooks/useLaunchSource/useLaunchSource.ts
```diff
@@ -0,0 +1,32 @@
...
+    const { trackEvent, registerSuperProperties } = usePostHog();
+
+    const trackLaunchSource = async () => {
+        try {
+            const { data } = await getLaunchSource();
+            if (!data) return;
+
+            registerSuperProperties({ launch_source: data.launch_source });
+            if (data.launch_source === LaunchSource.QUICKSTART) {
+                trackEvent('quickstart_launched');
```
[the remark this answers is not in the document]
gabrielfruet: Fixed. The backend returns a `launch_id` per process and the GUI keeps the last reported one in localStorage, so it's one event per invocation.

lightly_studio_view/src/lib/hooks/useLaunchSource/useLaunchSource.ts
```diff
@@ -0,0 +1,32 @@
...
+ * Registers `launch_source` as a super property so every later event in the session can be
+ * segmented by entry point, and captures a one-off event when the app came from the
+ * `lightly-studio quickstart` command.
+ */
+export const useLaunchSource = () => {
+    const { trackEvent, registerSuperProperties } = usePostHog();
+
+    const trackLaunchSource = async () => {
+        try {
+            const { data } = await getLaunchSource();
```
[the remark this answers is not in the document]
gabrielfruet: Leaving as is. `useFeatureFlags.ts` is the guide's own reference hook and calls the SDK directly. Nothing renders this result.

lightly_studio_view/src/lib/hooks/usePostHog.ts
```diff
@@ -72,3 +86,19 @@ export const usePostHog = () => {
...
+ */
+const analyticsEnabled = async (): Promise<boolean> => {
+    try {
+        const { data } = await getFeatures();
+        return data?.includes(ANALYTICS_FEATURE) ?? false;
+    } catch (e) {
+        console.warn('Could not read the active features, leaving analytics off', e);
+        return false;
+    }
+};
```
mihnea-necsulescu: there's already a `useFeatureFlags` hook that can be used instead

lightly_studio/src/lightly_studio/analytics/posthog_tracker.py
```diff
@@ -0,0 +1,90 @@
...
+import logging
+import platform
+from collections.abc import Mapping
+from importlib import metadata
+
+from posthog import Posthog
+
+from lightly_studio.analytics import install_id
+
+POSTHOG_HOST = "https://eu.i.posthog.com"
```
mihnea-necsulescu: this is already defined -> https://github.com/lightly-ai/lightly-studio/blob/main/lightly_studio_view/.env#L9

lightly_studio/src/lightly_studio/api/features.py
```diff
@@ -2,4 +2,19 @@
 
 from __future__ import annotations
 
-lightly_studio_active_features: list[str] = []
+from lightly_studio.dataset.env import LIGHTLY_STUDIO_ANALYTICS_ENABLED
+
+# The browser UI reads this back to decide whether to start PostHog, so that
```
mihnea-necsulescu: can we call it "GUI" instead of "browser UI"? I think that's how it's usually called in the docs, changelog, etc. (here and the other occurrences, too)

lightly_studio/tests/analytics/test_posthog_tracker.py
```diff
@@ -0,0 +1,109 @@
...
+    loggers = [logging.getLogger(name) for name in posthog_tracker._NOISY_LOGGERS]
+    saved = [(logger, logger.level, logger.propagate, list(logger.handlers)) for logger in loggers]
+    yield
+    for logger, level, propagate, handlers in saved:
+        logger.setLevel(level)
+        logger.propagate = propagate
+        logger.handlers = handlers
+
+
+class TestPostHogTracker:
```
mihnea-necsulescu: we don't usually use test classes

lightly_studio/src/lightly_studio/analytics/tracking.py
```diff
@@ -0,0 +1,113 @@
...
+    def track(self, event: str, properties: Mapping[str, object]) -> None:
+        """Report a single event."""
+        ...
+
+    def shutdown(self) -> None:
+        """Deliver anything still pending and release resources."""
+        ...
+
+
+class NoOpTracker:
```
mihnea-necsulescu:
```suggestion
class NoOpTracker(Tracker):
```

lightly_studio/src/lightly_studio/analytics/posthog_tracker.py
```diff
@@ -0,0 +1,90 @@
...
+REQUEST_TIMEOUT_SECONDS = 3
+
+# PostHog reports delivery failures at ERROR, and the retries around them at INFO, straight to the
+# user's terminal. An unreachable analytics endpoint is not the user's problem. Both names are
+# needed: PostHog's consumer wraps its send in `backoff.on_exception` without passing a logger, so
+# the retry lines go to `backoff`'s own logger rather than PostHog's.
+_NOISY_LOGGERS = ("posthog", "backoff")
+
+
+class PostHogTracker:
```
mihnea-necsulescu:
```suggestion
class PostHogTracker(Tracker):
```

mihnea-necsulescu (changes requested):
has this been tested? I looked up `app_launched` in PostHog on the dev project and couldn't see anything

lightly_studio_view/src/lib/hooks/usePostHog.ts
gabrielfruet: Applied. `useFeatureFlags` now also returns the in-flight promise so `init` can await the answer.

lightly_studio/src/lightly_studio/analytics/posthog_tracker.py
gabrielfruet: Moved to `LIGHTLY_STUDIO_POSTHOG_HOST` in `dataset/env.py`, next to the key. Vite and environs can't share a literal, so the string still lives in both.

lightly_studio/src/lightly_studio/api/features.py
gabrielfruet: Renamed, all five occurrences.

lightly_studio/tests/analytics/test_posthog_tracker.py
gabrielfruet: Flattened. Note `.agents/skills/python-guide` prescribes `class TestMyClass:` for a class under test, so the guideline and the convention disagree.

lightly_studio/src/lightly_studio/analytics/tracking.py
gabrielfruet: Applied. `Tracker` moved to `analytics/tracker.py` so `posthog_tracker` can import it without a cycle.

lightly_studio/src/lightly_studio/analytics/posthog_tracker.py
gabrielfruet: Applied, via the same protocol move as the `NoOpTracker` thread.

gabrielfruet:
It was, but against the production project. The checkout had no `lightly_studio/.env`, so the backend kept its built-in production key while the locally built frontend fell back to `PUBLIC_POSTHOG_DEV_KEY`. Hence nothing in dev.

`.env.example` now ships the dev key. With it copied, `quickstart` lands in dev with `distinct_id` matching the install ID. `gui` is the same call site with `launch_source=gui`.

lightly_studio/tests/analytics/test_posthog_tracker.py
mihnea-necsulescu: whoops, you're right, I forgot about those 🤦‍♂️ feel free to keep it as-is or revert to your previous changes, won't block the PR on this

mihnea-necsulescu (approved):

lightly_studio/tests/analytics/test_posthog_tracker.py
gabrielfruet: It's fine! I'll ask on dev-studio-discuss also to make sure that we should follow the class convention or, if not, modify what we currently have in the ai guidelines.

PR #1925, opened 2026-08-12
Tagging by embedding region now correctly limits samples to the selected area

LeonardoRosaa:
## What has changed and why?

In the embedding plot, users can draw a polygon to select a region of samples and then tag all samples inside that region. Before this fix, the polygon selection was ignored during the "add samples to tag" step - all samples matched by the other filter conditions were tagged instead of just the ones inside the drawn area.

The root cause: the polygon needed to be translated into a concrete list of sample IDs *before* the database query was built, but that translation step was missing. This fix adds the translation in the right place for both image filters and annotation filters.

## How has it been tested?

Unit tests

- [ ] Yes

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved collection tagging with embedding-region filters.
  * Image and annotation filters now tag only samples whose cached coordinates fall within the selected region.
  * Region-based filtering now works consistently across supported sample filter types.

### DIFF
--- a/lightly_studio/src/lightly_studio/api/routes/api/collection_tag.py
+++ b/lightly_studio/src/lightly_studio/api/routes/api/collection_tag.py
@@ -26,8 +26,10 @@ from lightly_studio.models.tag import (
     TagTable,
     TagView,
 )
-from lightly_studio.resolvers import tag_resolver
+from lightly_studio.resolvers import embedding_region_resolver, tag_resolver
+from lightly_studio.resolvers.annotations.annotations_filter import AnnotationsFilter
 from lightly_studio.resolvers.grid_filter import GridFilter
+from lightly_studio.resolvers.image_filter import ImageFilter
 
 tag_router = APIRouter()
 
@@ -234,7 +236,25 @@ def add_samples_to_tag_by_filter(
             detail=f"Tag {tag_id} not found in collection {collection_id}.",
         )
 
-    sample_ids_query = body.filter.build_sample_ids_query(collection_id=collection_id)
+    # Resolve any embedding-plot region selection to concrete sample ids before building
+    # the query (the point-in-polygon test needs the session, which `apply` lacks).
+    grid_filter = body.filter
+    if isinstance(grid_filter, ImageFilter) and grid_filter.sample_filter is not None:
+        if grid_filter.sample_filter.embedding_region is not None:
+            grid_filter.sample_filter.region_sample_ids = (
+                embedding_region_resolver.get_sample_ids_in_region(
+                    session=session,
+                    collection_id=collection_id,
+                    region=grid_filter.sample_filter.embedding_region,
+                )
+            )
+    elif isinstance(grid_filter, AnnotationsFilter) and grid_filter.embedding_region is not None:
+        grid_filter.region_sample_ids = embedding_region_resolver.get_sample_ids_in_region(
+            session=session,
+            collection_id=collection_id,
+            region=grid_filter.embedding_region,
+        )
+    sample_ids_query = grid_filter.build_sample_ids_query(collection_id=collection_id)
     tag_resolver.add_samples_to_tag_from_query(
         session=session, tag_id=tag_id, sample_ids_query=sample_ids_query
     )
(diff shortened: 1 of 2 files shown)


lightly_studio/src/lightly_studio/api/routes/api/collection_tag.py
```diff
@@ -234,7 +236,25 @@ def add_samples_to_tag_by_filter(
             detail=f"Tag {tag_id} not found in collection {collection_id}.",
         )
 
-    sample_ids_query = body.filter.build_sample_ids_query(collection_id=collection_id)
+    # Resolve any embedding-plot region selection to concrete sample ids before building
+    # the query (the point-in-polygon test needs the session, which `apply` lacks).
+    grid_filter = body.filter
+    if isinstance(grid_filter, ImageFilter) and grid_filter.sample_filter is not None:
+        if grid_filter.sample_filter.embedding_region is not None:
```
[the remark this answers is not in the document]
LeonardoRosaa: Introduced the resolve_region_sample_ids method to GridFilterBase. Video and video frame are bypassed since the lasso selections are resolved on the client side.

lightly_studio/tests/api/routes/api/test_collection_tag.py
```diff
@@ -141,6 +145,176 @@ def test_add_samples_by_filter__wrong_collection_returns_404(
...
+        sample_id=outside.sample_id,
+        embedding_model_id=embedding_model.embedding_model_id,
+        embedding=[2.1, 0.2, 0.3],
+    )
+    # Seed the 2D embedding cache with known coordinates.
+    cache_key, cached_sample_ids = sample_embedding_resolver.get_hash_by_collection_id(
+        session=db_session,
+        collection_id=annotation_collection_id,
+        embedding_model_id=embedding_model.embedding_model_id,
+    )
```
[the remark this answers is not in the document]
LeonardoRosaa: We would prefer DAMP tests

lightly_studio/src/lightly_studio/api/routes/api/collection_tag.py
```diff
@@ -234,7 +234,11 @@ def add_samples_to_tag_by_filter(
             detail=f"Tag {tag_id} not found in collection {collection_id}.",
         )
 
-    sample_ids_query = body.filter.build_sample_ids_query(collection_id=collection_id)
+    # Resolve any embedding-plot region selection to concrete sample ids before building
+    # the query (the point-in-polygon test needs the session, which `apply` lacks).
+    grid_filter = body.filter
+    grid_filter.resolve_region_sample_ids(session=session, collection_id=collection_id)
+    sample_ids_query = grid_filter.build_sample_ids_query(collection_id=collection_id)
```
michal-lightly, line 260: `filter.region_sample_ids` is easy to become stale, the lasso should be always applied when calling `filter.build_sample_ids_query()` - without assuming another function was called.

michal-lightly (changes requested):

lightly_studio/src/lightly_studio/api/routes/api/collection_tag.py
```diff
@@ -234,7 +236,28 @@ def add_samples_to_tag_by_filter(
             detail=f"Tag {tag_id} not found in collection {collection_id}.",
         )
 
-    sample_ids_query = body.filter.build_sample_ids_query(collection_id=collection_id)
+    # Resolve any embedding-plot region selection to concrete sample ids before building
+    # the query (the point-in-polygon test needs the session, which `apply` lacks).
+    grid_filter = body.filter
+    if (
+        isinstance(grid_filter, ImageFilter)
```
michal-lightly, line 243: Why do we support only `ImageFilter` and `AnnotationsFilter`? Shouldn't only annotations filter be handled differently?

(Also it is strange that it is different - annotation is a sample)

michal-lightly (changes requested):

lightly_studio/src/lightly_studio/api/routes/api/collection_tag.py
michal-lightly: Resolved offline. Video and frame filters do not inherit from the RegionsFilter. Will be resolved in a follow-up.

michal-lightly (approved):

PR #1926, opened 2026-08-12
Extract the keyset seek engine into a shared module

iunir:
## What has changed and why?

Second of six in the stack. Pure code movement.

The keyset (seek) machinery lived inside the image adjacency resolver, but nothing about it is
image-specific — a caller only supplies the sort keys and the base query. It moves to
`resolvers/adjacents/keyset_seek.py` so the annotation resolver can use it too.

Three cleanups on the way:

- The image resolver had two query builders that duplicated the collection join and the filters.
  Now one builder, parameterised by the select list.
- The anchor read moves in as `keyset_seek.fetch_anchor`, since both resolvers need it.
- `SortKey`'s column type now matches what `col()` returns, which drops a
  `cast(ColumnElement[Any], ...)` at every call site.

`get_adjacent_result`, `fetch_anchor` and `SortKey` are public; the rest is module-private.

**What to review.** Not everything here is untouched, so the two halves are worth separating:

- *Moved as-is* — the seek, the comparison and the ASC/DESC helper are the same code, only
  re-wrapped to 100 columns and renamed with a leading underscore. `get_adjacent_result` is the same
  four queries that were inline in `get_adjacent_images_keyset`, and `_count` is the count that was
  written out twice.
- *Actually new* — the merged query builder, `fetch_anchor`, the widened `SortKey`, and the
  docstrings. This is where a bug could hide.

## How has it been tested?

`make static-checks` and the full suite (1893 passed), plus the adjacency tests against Postgres.

**The image adjacency tests are untouched and still pass** — that is the argument that this is a
pure move.

- [ ] Yes

## Summary by CodeRabbit

* **Improvements**
  * Improved previous and next image navigation across sorted results.
  * Added support for consistent navigation when sorting by multiple fields, including ascending and descending orders.
  * Improved accuracy of each image’s position and total result count.
  * Ensured deterministic navigation when images share the same sort values.

### DIFF
--- a/lightly_studio/src/lightly_studio/resolvers/adjacents/keyset_seek.py
+++ b/lightly_studio/src/lightly_studio/resolvers/adjacents/keyset_seek.py
@@ -1,0 +1,146 @@
+"""Keyset (seek) engine shared by the adjacency resolvers.
+
+Reads the anchor row's sort-key values once, then jumps straight to its neighbours with
+``LIMIT 1`` queries instead of sorting the whole collection and scanning it with window
+functions. Position and total come from two ``COUNT``s. Callers supply the sort keys and
+the base query, so resolvers for different sample types can share the engine.
+"""
+
+from __future__ import annotations
+
+from typing import Any, Union
+from uuid import UUID
+
+import sqlalchemy
+from sqlalchemy.orm import Mapped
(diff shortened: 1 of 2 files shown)


ikondrat (approved):
lgtm

PR #1523, opened 2026-06-30
fix full scan for images prev/next

ikondrat:
## What has changed and why?

Fixes **LIG-9925** — the sample-adjacents endpoint (prev/next navigation in image details) took >2s at 250k samples and up to 37s at 20M annotations.

The fix for annotations will come in a separate PR

### Root cause

Every prev/next click hits `POST /samples/{sample_id}/adjacents`, which for images built a query that computed `lag` / `lead` / `row_number` **window functions over the entire collection**, wrapped it in a subquery, then `COUNT(*)`'d it and filtered down to the single `sample_id`:

```python
func.lag(sample_id).over(order_by=order_col)        # previous_sample_id
func.lead(sample_id).over(order_by=order_col)        # next_sample_id
func.row_number().over(order_by=order_col)           # row_number
```

Window functions are evaluated over the whole partition *before* any row can be filtered, so the `WHERE sample_id = :id` predicate can't be pushed down. The database read and sorted every row in the collection (`O(N log N)`) on **every single click**, then kept one row. No index could help, because the filter sits outside the window. This is the full scan.

### Fix

Replace the window query with **keyset (seek) pagination** for the common case (default `file_path_abs` sort, or any sort over plain image columns):

1. Fetch the current sample's sort-key value(s) once.
2. `next_sample_id` → `WHERE (file_path_abs, sample_id) > (:v, :id) ORDER BY file_path_abs, sample_id ASC LIMIT 1`
3. `previous_sample_id` → same with `<` and `DESC LIMIT 1`
4. `total_count` → index-only `COUNT(*)` over the filtered set
5. `current_sample_position` → `COUNT(*)` of rows before the anchor, `+ 1`

A unique `sample_id` tiebreaker is

### How to test

1. Create subset
```
 uv run scripts/download_coco_subset.py --num-images 10000 --output-dir ./datasets/coco_subset_10k_images
 ```
 
2. Adjust the index_general.py to use new subset with more images
I would disable embedding because it doesn't effect 
 ```
    annotations_json="datasets/**coco_subset_100k_images**/instances_train2017.json",
    images_path="datasets/**coco_subset_100k_images**/images",
    annotation_type=ls.AnnotationType.SEGMENTATION_MASK,
    **embed=False,**
 ``` 
 3. Start postgress instance
 ```
  make start-postgres
  ```
 4. Run app with postgre
 ```
 make start-e2e-postgres
 ``` 
 
 If you make the same and use dataset for main you will how laggy prev/next operations
 
 
[...]

PR #1927, opened 2026-08-12
Add composite indexes for annotation adjacency

iunir:
## What has changed and why?

First of six in the stack. Adds the two composite indexes that annotation prev/next needs to become a keyset (seek) lookup instead of sorting the whole collection on every click. Split out so the schema change can be reviewed and merged on its own.

- `annotation_base(parent_sample_id, created_at, sample_id)` — the adjacency sort keys that live on the annotation itself, so each parent's annotations can be read already ordered. The table had only single-column indexes.
- `video(file_path_abs, sample_id)` — the leading sort key for annotations on video frames, mirroring the `image` index added for the same reason in #1523.

Both indexes are inert until the resolver change in the follow-up PR. Measured there: the neighbour lookup on PostgreSQL with 4M annotations goes from ~6s to ~3ms.

Note that migrations only run on PostgreSQL. DuckDB builds its schema from the models via `create_all`, so new DuckDB databases pick these up automatically but existing DuckDB files do not gain them.

**What to review.** Two indexes and one migration; no code paths change. The thing to check is that each index's column order matches the sort keys the follow-up seeks on, since a wrong order leaves the index unusable for the seek.

## How has it been tested?

`make static-checks` and the test suite pass (1862 passed). `make migration-check-postgresql` applies the chain to an empty database and reports "No new upgrade operations detected", so the migration matches the models. `downgrade` drops both indexes.

- [ ] Yes

## Summary by CodeRabbit

* **Performance Improvements**
  * Improved responsiveness for annotation and video adjacency lookups.
  * Optimized searches involving sample relationships, creation times, and video file paths.
* **Maintenance**
  * Added database migration support to apply and safely revert these performance improvements.

### DIFF
--- a/lightly_studio/src/lightly_studio/migrations/versions/1786435200_b7c8d9e0f1a2_add_annotation_adjacency_indexes.py
+++ b/lightly_studio/src/lightly_studio/migrations/versions/1786435200_b7c8d9e0f1a2_add_annotation_adjacency_indexes.py
@@ -1,0 +1,59 @@
+"""add_annotation_adjacency_indexes.
+
+Adds the two composite indexes the keyset (seek) path for annotation prev/next needs:
+
+- ``annotation_base(parent_sample_id, created_at, sample_id)`` — the sort keys that live on
+  the annotation itself, so each parent's annotations can be read already ordered.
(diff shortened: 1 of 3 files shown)


IgorSusmelj (approved):
LGTM

PR #1931, opened 2026-08-13
Fix: Arrow keys navigate image detail view instead of moving text cursor in menu

IgorSusmelj:
## What has changed and why?

  - Prevent left and right arrow keys from changing samples while typing or using an open dialog or menu
  - Keep normal arrow-key navigation working in the sample detail view
  
## How has it been tested?

- New unit test and manual testing

- [x] Yes

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented left/right arrow keys from switching samples while editing text.
  * Arrow-key navigation is now ignored while drawing or interacting with dialogs, menus, listboxes, and popover content.
  * Improved keyboard behavior for text inputs, textareas, editable content, and buttons within overlays.
  * Prevented already-handled keyboard events from triggering sample navigation.
  * Keyboard navigation now responds more reliably when interacting with controls and modal interfaces.


lightly_studio_view/src/lib/components/SteppingNavigation/SteppingNavigation.svelte
```diff
@@ -9,8 +11,12 @@
     const { hasPrevious, hasNext, onNext, onPrevious, isDrawing }: SteppingNavigationProps =
         $props();
 
+    const isOverlayTarget = (target: EventTarget | null): boolean =>
+        target instanceof Element &&
+        target.closest('[role="dialog"], [role="menu"], [role="listbox"]') !== null;
+
     const handleKeyDownEvent = (event: KeyboardEvent) => {
-        if (isDrawing) return;
+        if (isDrawing || isTextInputTarget(event.target) || isOverlayTarget(event.target)) return;
```
iunir, line 19:
```suggestion
        if (event.defaultPrevented || isDrawing || isTextInputTarget(event.target) || isOverlayTarget(event.target)) return;
```

Because it can be prevented downstream in `lightly_studio_view/src/lib/components/VideoControls/VideoControls.svelte` `handleKeyDown`

lightly_studio_view/src/lib/components/SteppingNavigation/SteppingNavigation.svelte
```diff
@@ -9,8 +11,12 @@
     const { hasPrevious, hasNext, onNext, onPrevious, isDrawing }: SteppingNavigationProps =
         $props();
 
+    const isOverlayTarget = (target: EventTarget | null): boolean =>
+        target instanceof Element &&
+        target.closest('[role="dialog"], [role="menu"], [role="listbox"]') !== null;
```
iunir, line 16:
```suggestion
        target.closest('[role="dialog"], [role="menu"], [role="listbox"], ["data-popover-content"]') !== null;
```

Suggested by Claude, popovers render with no `role`

PR #1940, opened 2026-08-13
Simplify the README by pointing Python details at the docs

gabrielfruet:
## What has changed and why?

The README reproduced docs.lightly.ai/studio: seven Quickstart examples and a Python Interface tour of the dataset, sample, query and sampling APIs. Quickstart keeps two, a COCO import and an image folder. Everything else becomes a link index. Only mixing cloud and local sources lacked a docs page, so it moved to `dataset_setup/cloud_storage.md`.

After review, COCO opens the section instead of the image folder: PostHog showed users confused when the first example produced no annotations. The Python Interface section was linking most of the docs rather than the API, so the link lists moved out to `A selection of features`.

## How has it been tested?

Both examples were run end to end from the new text. The COCO one indexed 128 images and 900 annotations; `lightly-studio gui` then reopened the same database without reindexing.

Every docs URL in the README returns 200 and all five anchors exist on the deployed pages, so this no longer needs to wait for a docs release.

`mkdocs build --strict` would not run locally, no Artifact Registry token, but CI's Build Docs job runs it and passed.

- [ ] Yes

## Summary by CodeRabbit

* **Documentation**
  * Simplified the README quickstart with local-folder indexing and a COCO example.
  * Added categorized links to dataset setup, data operations, workflows, tutorials, notebooks, and API references.
  * Added guidance for combining images from S3, GCS, and local folders in a single dataset.
  * Streamlined the Python interface overview while retaining links to detailed reference material.

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,216 +90,10 @@ pip install lightly-studio
 
 ## 🚀 Quickstart
 
-LightlyStudio is a browser app that runs on your own computer. Use it in two simple steps:
-
-1. Load your data into the local database with a Python script.
-2. Start the server and explore the data in your browser.
-
-Prefer a guided, end-to-end walkthrough? Follow the tutorial
-[Curate a Traffic CCTV Dataset for YOLO Training](https://docs.lightly.ai/studio/tutorials/yolo-traffic-cctv-object-detection/) — from raw images to a trained model.
-Or try LightlyStudio without installing anything: [open the example notebook in Colab](https://colab.research.google.com/github/lightly-ai/lightly-studio/blob/main/lightly_studio/src/lightly_studio/examples/example_notebook.ipynb).
-
-Get started with one of these example workflows:
+LightlyStudio is a browser app that runs on your own computer. Load your data into the local
+database with a Python script, then start the server and explore it in your browser.
 
 <details open>
-<summary><strong>Evaluate object detection predictions on a COCO dataset</strong></summary>
-
-Create a file named `example_coco_od_evaluation.py`:
-
-```python
-import lightly_studio as ls
-from lightly_studio.core.dataset_query.image_sample_field import ImageSampleField
-from lightly_studio.evaluation.image_dataset_evaluate import ObjectDetectionEvaluationConfig
-
-
-# Download the example dataset (will be skipped if it already exists)
-dataset_path = ls.utils.download_example_dataset(download_dir="dataset_examples")
-
-images_path = f"{dataset_path}/coco_subset_128_images/images"
-evaluation_config = ObjectDetectionEvaluationConfig(
-    iou_threshold=0.5,
-    classwise=True,
-)
-
-dataset = ls.ImageDataset.load_or_create()
-dataset.add_images_from_path(path=images_path)
-# Add ground truth annotations
-dataset.add_annotations_from_coco(
-    annotations_json=f"{dataset_path}/coco_subset_128_images/instances_train2017.json",
-    images_root=images_path,
-    annotation_source="ground_truth",
-)
-# Add predictions annotations
-dataset.add_annotations_from_coco(
-    annotations_json=f"{dataset_path}/coco_subset_128_images/predictions_train2017.json",
-    images_root=images_path,
-    annotation_source="predictions",
-)
-# Optional: tag a subset of samples to run the evaluation on.
-dataset.query()[:10].add_tag("evaluated_samples")
-# Create query for tagged samples
-tagged_evaluation_query = dataset.query().match(ImageSampleField.tags.contains("evaluated_samples"))
-
-dataset.evaluate(query=tagged_evaluation_query).object_detection(
-    name="od_evaluation",
-    gt_annotation_source="ground_truth",
-    pred_annotation_source="predictions",
-    config=evaluation_config,
-)
-
-ls.start_gui()
-```
-</details>
-
-<details>
-<summary><strong>Index a COCO dataset</strong></summary>
-
-Create a file named `example_coco.py`:
-
-```python
-import lightly_studio as ls
-
-# Download the example dataset (will be skipped if it already exists)
-dataset_path = ls.utils.download_example_dataset(download_dir="dataset_examples")
-
-dataset = ls.ImageDataset.load_or_create()
-dataset.add_samples_from_coco(
-    annotations_json=f"{dataset_path}/coco_subset_128_images/instances_train2017.json",
-    images_path=f"{dataset_path}/coco_subset_128_images/images",
-)
-# Optional: tag a subset of samples to filter them in the GUI. 
-dataset.query()[:10].add_tag("sample_subset")
-
-ls.start_gui()
-```
-
-Run `python example_coco.py` and open the printed URL to inspect images with their annotations.
-
-To import COCO segmentation masks instead of object detections, set:
-
-```python
-annotation_type=ls.AnnotationType.SEGMENTATION_MASK
-```
-
-</details>
-
-<details>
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -351,185 +171,38 @@ Step-by-step guides covering complete workflows — from raw, unlabeled data to
...
+- [Image datasets](https://docs.lightly.ai/studio/dataset_setup/image_dataset/) — index folders, COCO, and YOLO data
+- [Video datasets](https://docs.lightly.ai/studio/dataset_setup/video_dataset/) — index videos and work with frames
+- [Cloud storage](https://docs.lightly.ai/studio/dataset_setup/cloud_storage/) — read from S3, GCS, or Azure
+- [Reuse datasets](https://docs.lightly.ai/studio/dataset_setup/reuse_datasets/) — append to an existing database, or point at a custom path
+- [Notebooks](https://docs.lightly.ai/studio/dataset_setup/notebooks/) — run the GUI inside Jupyter or Colab
+
+**Work with your data**
+
+- [Query, filter, and sort](https://docs.lightly.ai/studio/concepts_and_tools/search_and_filter/#query-in-python) — build reusable subsets in Python
+- [Lightly Query Language](https://docs.lightly.ai/studio/concepts_and_tools/lightly_query_language/) — the query syntax shared by the GUI and Python
```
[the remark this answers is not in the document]
gabrielfruet: Good catch, fixed.

lightly_studio/docs/docs/dataset_setup/cloud_storage.md
```diff
@@ -19,6 +19,19 @@ dataset.add_images_from_path(path="s3://my-bucket/images/")
...
+
+Sources can be mixed and appended at any time:
+
+```py
+import lightly_studio as ls
+
+dataset = ls.ImageDataset.load_or_create()
+dataset.add_images_from_path(path="s3://my-bucket/images/")
+dataset.add_images_from_path(path="gcs://other-bucket/more-images/")
+dataset.add_images_from_path(path="local-folder/not-in-the-cloud-yet")
```
michal-lightly: Optional: I suggest to make it clear this is a relative path (or use an absolute one)

```suggestion
dataset.add_images_from_path(path="local-folder-relative-path/not-in-the-cloud-yet")
```

README.md
```diff
@@ -96,58 +96,36 @@ lightly-studio quickstart
...
+
 ## Example workflows
 
-Get started with one of these example workflows:
+LightlyStudio is a browser app that runs on your own computer. Load your data into the local
+database with a Python script, then start the server and explore it in your browser.
 
 <details open>
-<summary><strong>Evaluate object detection predictions on a COCO dataset</strong></summary>
+<summary><strong>Index a folder of images for curation and labeling</strong></summary>
```
michal-lightly: Let's have the COCO example first. (We saw in posthog confused users when they did not see annotations).

Minor: Let's drop `for curation and labeling` - there's no information

README.md
```diff
@@ -96,58 +96,36 @@ lightly-studio quickstart
...
-    name="od_evaluation",
-    gt_annotation_source="ground_truth",
-    pred_annotation_source="predictions",
-    config=evaluation_config,
+dataset.add_images_from_path(
+    path=f"{dataset_path}/coco_subset_128_images/images",
 )
 
+# Start the UI server on localhost:8001.
+# Pass `host` and `port` parameters to customize it.
```
michal-lightly, line 123: Optional: Shall we also mention the `lightly-studio gui` command as an alternative?

README.md
```diff
@@ -166,7 +144,7 @@ dataset.add_samples_from_coco(
     annotations_json=f"{dataset_path}/coco_subset_128_images/instances_train2017.json",
     images_path=f"{dataset_path}/coco_subset_128_images/images",
 )
-# Optional: tag a subset of samples to filter them in the GUI. 
+# Optional: tag a subset of samples to filter them in the GUI.
```
michal-lightly: Optional: I would consider to remove the tagging completely from this example.

README.md
```diff
@@ -182,164 +160,14 @@ annotation_type=ls.AnnotationType.SEGMENTATION_MASK
...
-
-dataset_path = ls.utils.download_example_dataset(download_dir="dataset_examples")
-
-dataset = ls.VideoDataset.load_or_create()
-dataset.add_videos_from_path(path=f"{dataset_path}/youtube_vis_50_videos/train/videos")
-
-ls.start_gui()
-```
-
+Other starting points:
```
michal-lightly, line 161: Minor: Let's add something like "and more in our [docs]" so we don't need to edit this place too often.

README.md
```diff
@@ -182,164 +160,14 @@ annotation_type=ls.AnnotationType.SEGMENTATION_MASK
...
+[YOLO datasets](https://docs.lightly.ai/studio/dataset_setup/image_dataset/#from-an-annotation-format),
+[video folders](https://docs.lightly.ai/studio/dataset_setup/video_dataset/),
+[custom annotations](https://docs.lightly.ai/studio/concepts_and_tools/annotations/#annotations-in-python),
+[model evaluation](https://docs.lightly.ai/studio/concepts_and_tools/evaluation/#model-evaluation-in-python),
+and [notebooks](https://docs.lightly.ai/studio/dataset_setup/notebooks/).
 
-</details>
+Try LightlyStudio without installing anything: [open the example notebook in Colab](https://colab.research.google.com/github/lightly-ai/lightly-studio/blob/main/lightly_studio/src/lightly_studio/examples/example_notebook.ipynb).
```
michal-lightly: This link is already above ("No install"), I would consider removing this.

README.md
```diff
@@ -350,185 +178,32 @@ Step-by-step guides covering complete workflows — from raw, unlabeled data to
 
 ## 🐍 Python Interface
```
michal-lightly: This section should be reworked:
- It is not linking just Python API, but almost everything from the docs
- Should it be a docs index?
    - If yes, we can use cards similarly as we do in docs

What solution do you propose?

michal-lightly: One option:
* Keep just the one sentence about python API. Link the API reference.
* In a separate section, have the docs index (we can call it "A selection of features" or something similar). And remove the API reference table.

michal-lightly (changes requested):

README.md
gabrielfruet: I think it's worth it

michal-lightly: Minor: You swapped the examples - the comment for `ls.start_gui()` should be visible for the first example (it can be copied).

Optional: Regarding `lightly-studio gui` - you added a sentence. What I was thinking about was to extend the comment instead. I don't mind a new sentence though.

README.md
michal-lightly: This was not done but it seems you want to do it?

README.md
michal-lightly: Optional: I would rather link https://docs.lightly.ai/studio/dataset_setup/image_dataset/

michal-lightly (approved):

README.md
gabrielfruet: I missed this one, sorry

README.md
gabrielfruet: Fine for me, but then It would not open the `YOLO Object Detection` tab directly

README.md
gabrielfruet: I’ll add inline comments to both code blocks.

I added a sentence after the first code block with the `lightly-studio gui` command, but I didn't add to the second example.

README.md
gabrielfruet: Done

PR #1952, opened 2026-08-14
Rename json_extract_string to json_extract_key_as_text

iunir:
Split out of #1905. Bottom of the stack — merge this one first.

## What has changed and why?

`json_extract_string` sat next to `json_extract` and said nothing about which one takes a path and which one takes a literal key. Both return text, so the distinction was invisible at the call site, and the next commit adds more of these functions.

Rename the argument to `key` for the same reason, rename `_JsonTopLevelKeyType` to `_JsonKeyType` since it binds any object key rather than only a top-level one, and say in the module docstring what separates a field from a key.

No behaviour change.

## How has it been tested?

`make static-checks` and `make test`, plus the metadata suites under `--postgres`.

- [ ] Yes

## Summary by CodeRabbit

- **New Features**
  - Improved JSON metadata extraction to clearly support literal keys, including keys with special characters.
  - Updated metadata filtering and value retrieval to use refined key-based extraction behavior.
  - Renamed the public extraction interface to clarify that it retrieves values by literal key.

- **Bug Fixes**
  - Improved consistency when extracting metadata across supported database systems.

- **Tests**
  - Expanded coverage for special keys, database-specific behavior, JSON paths, and unsupported SQLite operations.

### DIFF
--- a/lightly_studio/src/lightly_studio/database/db_json.py
+++ b/lightly_studio/src/lightly_studio/database/db_json.py
@@ -70,20 +70,26 @@ class json_extract(GenericFunction[Any]):  # noqa: N801
         super().__init__(column)
 
 
-class json_extract_string(GenericFunction[str]):  # noqa: N801
-    """Extract a top-level JSON scalar as plain text.
+class json_extract_key_as_text(GenericFunction[str]):  # noqa: N801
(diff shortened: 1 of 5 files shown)


IgorSusmelj (approved):
LGTM

PR #1975, opened 2026-08-17
Scope the annotation source filter to the current collection

gabrielfruet:
## What has changed and why?

A dataset had three annotation sources, and unchecking them one by one hid their annotations in turn. Unchecking the last made all three reappear. An empty selection meant "draw everything".

That selection is shared by every grid, and only the images grid fills it, so reversing the rule would hide annotations elsewhere. The store behind `useAnnotationCollectionsFilter` now tracks the current collection's annotation sources and draws each only while checked. Untracked sources still draw, so video frame annotations stay on the videos grid. The new `useSeedAnnotationSourceFilter` hook fills the selection on every collection route.

Unchecking a source also clears annotation class filters. That predates this change; the selection scopes the annotation counts query, which prunes unlisted classes.

## How has it been tested?

New unit tests: unchecking every source draws nothing, an untracked source still draws, and seeding covers single- and empty-source collections. `make static-checks` and `npx vitest run` pass in lightly_studio_view (2596 tests).

Not checked manually against a running app. To reproduce, run the object-detection evaluation example, a COCO dataset with two annotation sources, then uncheck both on the images grid.

- [x] Yes

## Summary by CodeRabbit

* **Bug Fixes**
  * Annotation and category counts are now hidden when all annotation sources are unchecked.
  * Annotations correctly disappear when their sources are hidden, while annotations from unknown sources remain visible.
  * Empty-state messaging now distinguishes hidden sources from collections with no annotations.
  * Source selections stay synchronized as annotation sources are added or removed.

* **Tests**
  * Expanded coverage for source visibility, selection updates, hidden annotations, and count behavior.

### DIFF
--- a/lightly_studio_view/src/lib/components/SampleAnnotations/index.test.ts
+++ b/lightly_studio_view/src/lib/components/SampleAnnotations/index.test.ts
@@ -91,6 +91,8 @@ type Mock2dContext = {
     strokeRect: ReturnType<typeof vi.fn>;
 };
 
+const SAMPLE_SOURCE_ID = 'col-default';
+
 const createSample = (): ComponentProps<typeof SampleAnnotations>['sample'] =>
     ({
         width: 100,
@@ -98,6 +100,7 @@ const createSample = (): ComponentProps<typeof SampleAnnotations>['sample'] =>
         annotations: [
             {
                 sample_id: 'object-detection-1',
+                annotation_collection_id: SAMPLE_SOURCE_ID,
                 annotation_type: 'object_detection',
                 annotation_label: { annotation_label_name: 'car' },
                 object_detection_details: {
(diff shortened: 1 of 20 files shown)


gabrielfruet:
New behavior

https://github.com/user-attachments/assets/174167bf-c33e-4418-b52c-2bc10afe56dd

lightly_studio_view/src/lib/components/SampleAnnotations/index.test.ts
```diff
@@ -209,6 +217,42 @@ describe('SampleAnnotations', () => {
...
+        useAnnotationCollectionsFilter().setSelectedCollectionIds([]);
+
+        render(SampleAnnotations, {
+            props: {
+                sample: createSample()
+            }
+        });
+
+        // Give the idle-callback canvas mount a chance to run before asserting nothing drew.
+        await new Promise((resolve) => setTimeout(resolve, 50));
```
LeonardoRosaa: I think we can use `vi.useFakeTimers()` + `vi.runAllTimersAsync()` to avoid this promise.

LeonardoRosaa (approved):

PR #1762, opened 2026-07-23
docs: fix punctuation in CONTRIBUTING.md test section

amir-rezaei:
## Description
This PR fixes minor punctuation in `CONTRIBUTING.md` under the test commands section.

## Details
Fixed `To run static checks and unit tests use the following commands` -> `To run static checks and unit tests, use the following commands:`.

## Summary by CodeRabbit

* **Documentation**
  * Clarified the “Development Quickstart” instructions with minor punctuation updates.

PR #1977, opened 2026-08-18
Skip GCP auth in End2End Test when the build/success cache already hits

iunir:
## What has changed and why?

`Authenticate with GCP` ran unconditionally in both `prepare-caches` and `end-to-end`, before
either job even checked its own cache. Every step that actually needs those credentials was
already skipped on a cache hit, so this just moves auth after the cache lookup and gates it the
same way. On a cache hit (e.g. a docs-only PR like #1762), auth is skipped entirely — which also
means it no longer fails for fork PRs that don't have repo secrets.

## How has it been tested?

Checked the Makefile: only the keyring install, schema export, and optional-deps install steps
actually use the GCP credentials, and all three were already gated behind the same cache-hit
check auth is now gated on. `success-check` is untouched.

- [ ] Yes

### DIFF
--- a/.github/workflows/end2end_test.yml
+++ b/.github/workflows/end2end_test.yml
@@ -30,13 +30,6 @@ jobs:
         with:
           submodules: recursive
 
-      # Authenticate to download dev versions of lightly-mundig.
-      - name: Authenticate with GCP
-        uses: 'google-github-actions/auth@v3'
-        with:
-          project_id: boris-250909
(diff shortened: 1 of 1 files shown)


michal-lightly (changes requested):
The `Build Cache` step serves as a decision gate to skip the tests. An extra "Detect Changes" step just duplicates the functionality.

michal-lightly (approved):

PR #2016, opened 2026-08-19
Annotations sort: add adjancent ordering support endpoint

LeonardoRosaa:
## What has changed and why?

Introduces the `order_by` field to the annotation adjacent endpoint to respect the ordering from the annotations grid view while navigating between the previous and next annotation sample.

## How has it been tested?

Unit tests

- [ ] Yes

---

## Summary by CodeRabbit

* **New Features**
  * Added support for sorting adjacent annotation samples by evaluation metrics.
  * Added an optional annotation sorting option to adjacent-sample requests.
  * Annotations without metric coverage are placed last when sorting by evaluation metric.

* **Bug Fixes**
  * Improved ordering behavior so unmatched annotations appear before matched annotations when applicable.
  * Added validation for annotation sorting requests involving multiple collections.

* **Tests**
  * Added coverage for evaluation-metric ordering and annotation sorting options.

### DIFF
--- a/lightly_studio/src/lightly_studio/resolvers/annotation_resolver/get_adjacent_annotations_window.py
+++ b/lightly_studio/src/lightly_studio/resolvers/annotation_resolver/get_adjacent_annotations_window.py
@@ -30,7 +30,14 @@ def get_adjacent_annotations_window(
     filters: AnnotationsFilter,
     order_by: OrderByAnnotationEvaluationMetricField | None = None,
 ) -> AdjacentResultView | None:
-    """Get the adjacent annotations by sorting and windowing the whole filtered set."""
+    """Get the adjacent annotations by sorting and windowing the whole filtered set.
+
+    Args:
+        session: Database session.
+        sample_id: The anchor annotation whose neighbours we want.
+        filters: Annotation filters constraining the set; must scope the collection.
+        order_by: Optional leading sort key applied before the tiebreaker chain.
+    """
     return adjacents.get_sample_adjacent_info(
         session=session,
         sample_id=sample_id,
--- a/lightly_studio/src/lightly_studio/resolvers/annotation_resolver/get_adjacent_annotations_window.py
+++ b/lightly_studio/src/lightly_studio/resolvers/annotation_resolver/get_adjacent_annotations_window.py
@@ -10,10 +10,11 @@ from typing import Any
 from uuid import UUID
 
 import sqlmodel
-from sqlalchemy import func
+from sqlalchemy import ColumnElement, func
 from sqlmodel import Session, col, select
 from sqlmodel.sql.expression import Select, SelectOfScalar
 
+from lightly_studio.core.dataset_query.order_by import OrderByAnnotationEvaluationMetricField
 from lightly_studio.models.adjacents import AdjacentResultView
 from lightly_studio.models.annotation.annotation_base import AnnotationBaseTable
 from lightly_studio.models.image import ImageTable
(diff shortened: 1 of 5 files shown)


lightly_studio/src/lightly_studio/services/sample_services/sample_adjacents_service.py
```diff
@@ -39,6 +42,7 @@ class AdjacentRequest(BaseModel):
     ) = None
     text_embedding: list[float] | None = None
     sort_by: list[SortExpr] | None = None
+    annotation_sort_by: AnnotationEvaluationMetricSortExpr | None = None
```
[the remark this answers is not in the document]
LeonardoRosaa: That's fine. Addressed in the follow-up PR.

lightly_studio/src/lightly_studio/services/sample_services/sample_adjacents_service.py
```diff
@@ -101,10 +105,22 @@ def get_adjacent_samples(
...
+        annotation_order_by = (
+            annotation_metric_sort.sort_expr_to_order_by(
+                session=session,
+                annotation_collection_id=collection_ids[0],
+                sort_expr=request.annotation_sort_by,
+                annotation_id_column=col(AnnotationBaseTable.sample_id),
+            )
+            if request.annotation_sort_by and collection_ids
+            else None
+        )
```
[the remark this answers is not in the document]
LeonardoRosaa: Good catch. It is fine to assume that the collection_ids size is 1, since the front-end always passes only one collection_id. I added a guard to make sure it does not accept more than one collection_id.

lightly_studio/tests/services/sample_services/test_sample_adjacents_service.py
```diff
@@ -2,14 +2,17 @@
 
 from __future__ import annotations
 
+from unittest import mock
```
gabrielfruet: use pytest instead of unittest, to stay aligned with the python-guide skill

lightly_studio/src/lightly_studio/resolvers/annotation_resolver/get_adjacent_annotations_window.py
```diff
@@ -27,30 +28,48 @@ def get_adjacent_annotations_window(
...
     base_rows: SelectOfScalar[AnnotationBaseTable] = select(AnnotationBaseTable)
     filtered_rows = filters.apply(base_rows).subquery()
 
-    # TODO(Jonas, 07/2026): No leading order key here, so next/prev drifts from the grid
-    # whenever the grid sorts by similarity. A leading key added there must be added here.
+    # The metric join must match against filtered_rows.c.sample_id, not the ORM table column,
+    # because AnnotationBaseTable is not directly in the FROM — only the subquery is.
+    if order_by is not None:
+        order_by.annotation_id_column = filtered_rows.c.sample_id
+        leading_order_key: ColumnElement[Any] | None = order_by.to_column_elements()[0]
```
gabrielfruet: Add a note like

`# Safe only while order_by yields a single sort key.`

Because if we support OrderByMetadataField at some point, it could silently break

gabrielfruet (approved):
Minor things I think you should address. I'll approve and you can do these minor adjustments and then merge

lightly_studio/src/lightly_studio/resolvers/annotation_resolver/get_adjacent_annotations_window.py
LeonardoRosaa: Good catch :)

gabrielfruet: There is also another thing, this overrides the anotation_id_column. I couldn't figure out why, but it does not seem ideal 🙁 But I think it would need a big refactor so it's fine to leave it as it is

```python
order_by.annotation_id_column = filtered_rows.c.sample_id
```

PR #2031, opened 2026-08-21
Resolve the collection default from DefaultEmbeddingSpace

michal-lightly:
## What has changed and why?

Switches a collection's default embedding space from the on-the-fly "oldest model" derivation to the `default_embedding_space` table, and keeps the table populated. Behaviour-preserving.

- `get_default_by_collection_id` now reads the table; return type unchanged, so the embeddings2d and embedding-region callers are untouched.
- `embedding_manager.register_embedding_model` persists the default via `set_default` (in-memory generator cache stays the runtime lookup).
- Alembic revision `b1c2d3e4f5a6`: backfill each collection's current default (oldest model). Paired with the write-path change in the same release, so no collection is left without a default row.
- Downgrade empties the table; not lossy — `embedding_model.collection_id` still exists at this revision.
- Test helper `create_embedding_model` gains an opt-in `set_as_default`; default-resolving call sites opt in.

Stacked on #2030 — review/merge that first. Base retargets to `main` once #2030 lands.

## How has it been tested?

- `make static-checks` (ruff, mypy, format).
- `pytest` on the embedding resolver / manager / embeddings2d suite — 59 pass, 4 skipped.
- Single alembic head `b1c2d3e4f5a6`.

- [ ] Yes

## Summary by CodeRabbit

* **New Features**
  * Added persistent default embedding-space selection for each collection.
  * Existing collections are automatically assigned a default based on their earliest embedding model.
  * New embedding models can be explicitly designated as the collection default.

* **Bug Fixes**
  * 2D embedding and embedding-region views now consistently use the collection’s configured default embedding space.
  * Views gracefully return no results when a default embedding space is unavailable.

### DIFF
--- a/lightly_studio/src/lightly_studio/dataset/embedding_manager.py
+++ b/lightly_studio/src/lightly_studio/dataset/embedding_manager.py
@@ -24,6 +24,7 @@ from lightly_studio.models.sample_embedding import SampleEmbeddingCreate
 from lightly_studio.resolvers import (
     annotation_resolver,
     collection_resolver,
+    default_embedding_space_resolver,
     embedding_model_resolver,
     image_resolver,
     sample_embedding_resolver,
@@ -172,9 +173,22 @@ class EmbeddingManager:
         # Store the model in our dictionary
         self._models[model_id] = embedding_generator
 
-        # Set as default if requested or if it's the first model
+        # Set as default if requested or if it's the first model. The in-memory map is the
+        # runtime cache that pairs a collection with an already-loaded generator; the
+        # default_embedding_space table persists the same choice so query-layer callers
+        # (get_default_by_collection_id) survive across processes.
         if set_as_default or collection_id not in self._collection_id_to_default_model_id:
             self._collection_id_to_default_model_id[collection_id] = model_id
+        if (
+            set_as_default
+            or default_embedding_space_resolver.get_by_collection_id(
+                session=session, collection_id=collection_id
+            )
+            is None
+        ):
+            default_embedding_space_resolver.set_default(
+                session=session, collection_id=collection_id, embedding_model_id=model_id
+            )
 
         return db_model
 
--- a/lightly_studio/src/lightly_studio/dataset/embedding_manager.py
+++ b/lightly_studio/src/lightly_studio/dataset/embedding_manager.py
@@ -174,7 +174,7 @@ class EmbeddingManager:
 
         # Record the default in two places: the in-memory map caches the loaded generator
         # for this process, and the default_embedding_space table persists the choice for
-        # query-layer callers (get_default_by_collection_id).
+        # query-layer callers (default_embedding_space_resolver.get_by_collection_id).
         if set_as_default or collection_id not in self._collection_id_to_default_model_id:
             self._collection_id_to_default_model_id[collection_id] = model_id
         if (
--- a/lightly_studio/src/lightly_studio/dataset/embedding_manager.py
+++ b/lightly_studio/src/lightly_studio/dataset/embedding_manager.py
@@ -170,13 +170,11 @@ class EmbeddingManager:
         )
         model_id = db_model.embedding_model_id
 
-        # Store the model in our dictionary
         self._models[model_id] = embedding_generator
 
-        # Set as default if requested or if it's the first model. The in-memory map is the
-        # runtime cache that pairs a collection with an already-loaded generator; the
-        # default_embedding_space table persists the same choice so query-layer callers
-        # (get_default_by_collection_id) survive across processes.
+        # Record the default in two places: the in-memory map caches the loaded generator
+        # for this process, and the default_embedding_space table persists the choice for
+        # query-layer callers (get_default_by_collection_id).
         if set_as_default or collection_id not in self._collection_id_to_default_model_id:
             self._collection_id_to_default_model_id[collection_id] = model_id
         if (
--- a/lightly_studio/src/lightly_studio/migrations/versions/1787314050_b1c2d3e4f5a6_backfill_default_embedding_space.py
+++ b/lightly_studio/src/lightly_studio/migrations/versions/1787314050_b1c2d3e4f5a6_backfill_default_embedding_space.py
@@ -2,7 +2,7 @@
 
 Seeds the `default_embedding_space` table (created empty in `a05138ab5fc4`) with each
 collection's default: the oldest embedding model by `created_at ASC, embedding_model_id
-ASC`. This is the same rule `get_default_by_collection_id` applies, so reading the table
+ASC`. This is the same model the query resolved before, so reading the table
 resolves the same model the query used before. Collections created after this migration
 are populated by `embedding_manager.register_embedding_model`, so no collection is left
 without a default row.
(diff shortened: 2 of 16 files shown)


michal-lightly:
@coderabbit review

lightly_studio/src/lightly_studio/resolvers/dataset_resolver/table_coverage_utils.py
```diff
@@ -16,10 +16,11 @@
...
 # - setting (application-level, not collection-specific)
 # - two_dim_embeddings (cached projections, regenerated as needed)
-# - default_embedding_space (created empty; nothing writes it yet)
-# TODO(Michal, 08/2026): Move default_embedding_space into deep_copy and delete_dataset
-# (copy/delete its rows per collection) and shift it into _HANDLED_TABLES_COUNT once the
-# write path lands.
+# - default_embedding_space (populated, but not yet handled by deep_copy/delete_dataset — see TODO)
+# TODO(Michal, 08/2026): wire default_embedding_space into deep_copy and delete_dataset,
+# then move it into _HANDLED_TABLES_COUNT. Until then, on Postgres delete_dataset raises an
+# FK error for any dataset with embeddings, and deep_copy silently drops the copied
```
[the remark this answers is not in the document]
michal-lightly: Known and tracked — a follow-up deletes the default row in delete_dataset before the models it references.

lightly_studio/src/lightly_studio/resolvers/dataset_resolver/table_coverage_utils.py
```diff
@@ -16,10 +16,11 @@
...
+# collection's default.
```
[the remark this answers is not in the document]
michal-lightly: Known and tracked — a follow-up copies the default row into the new collection with both IDs remapped.

lightly_studio/src/lightly_studio/dataset/embedding_manager.py
```diff
@@ -169,12 +170,23 @@ def register_embedding_model(
...
+        if (
+            set_as_default
+            or default_embedding_space_resolver.get_by_collection_id(
+                session=session, collection_id=collection_id
+            )
+            is None
+        ):
+            default_embedding_space_resolver.set_default(
+                session=session, collection_id=collection_id, embedding_model_id=model_id
+            )
```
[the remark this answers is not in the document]
michal-lightly: The only path that registers a model here always sets it as the default, so the cache and table stay in agreement; unifying the two is a planned follow-up.

lightly_studio/src/lightly_studio/migrations/versions/1787314050_b1c2d3e4f5a6_backfill_default_embedding_space.py
```diff
@@ -0,0 +1,61 @@
...
+resolves the same model the query used before. Collections created after this migration
+are populated by `embedding_manager.register_embedding_model`, so no collection is left
+without a default row.
+
+Downgrade empties the table. It is not lossy at this revision: `embedding_model` still
+carries `collection_id`, so the default can be re-derived from the same oldest-model
+rule (the column drop lands in a later revision).
+
+DuckDB builds its schema with `create_all` and has no backfill step, so this migration
+only matters for tracked Postgres databases.
```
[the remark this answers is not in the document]
michal-lightly: Wont fix: upgrading an existing DuckDB database is not supported here — new DuckDB databases get a default when a model is registered, and existing ones must re-embed.

lightly_studio/src/lightly_studio/migrations/versions/1787314050_b1c2d3e4f5a6_backfill_default_embedding_space.py
```diff
@@ -0,0 +1,61 @@
...
+    op.get_bind().execute(
+        sa.text(
+            """
+            INSERT INTO default_embedding_space (collection_id, embedding_model_id)
+            SELECT DISTINCT ON (collection_id) collection_id, embedding_model_id
+            FROM embedding_model
+            ORDER BY collection_id, created_at ASC, embedding_model_id ASC
+            """
+        )
+    )
```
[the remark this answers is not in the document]
michal-lightly: Known and tracked — a follow-up makes delete_dataset remove the default row and deep_copy carry it over; both paths are Postgres-only.

lightly_studio/src/lightly_studio/dataset/embedding_manager.py
```diff
@@ -169,12 +170,23 @@ def register_embedding_model(
...
+        # for this process, and the default_embedding_space table persists the choice for
+        # query-layer callers (get_default_by_collection_id).
         if set_as_default or collection_id not in self._collection_id_to_default_model_id:
             self._collection_id_to_default_model_id[collection_id] = model_id
+        if (
+            set_as_default
+            or default_embedding_space_resolver.get_by_collection_id(
+                session=session, collection_id=collection_id
+            )
+            is None
```
iunir, line 185: Can this function be used to override existing default? 
```suggestion
```

This condition will prevent any change, I think we have to dump it

lightly_studio/src/lightly_studio/resolvers/embedding_model_resolver.py
```diff
@@ -58,20 +59,19 @@ def get_default_by_collection_id(
...
-            col(EmbeddingModelTable.created_at).asc(),
-            col(EmbeddingModelTable.embedding_model_id).asc(),
+        .join(
+            DefaultEmbeddingSpaceTable,
+            onclause=col(DefaultEmbeddingSpaceTable.embedding_model_id)
+            == col(EmbeddingModelTable.embedding_model_id),
         )
-        .limit(1)
+        .where(DefaultEmbeddingSpaceTable.collection_id == collection_id)
     ).first()
```
iunir: Since DuckDB is not gonna be backfilled, should we just fallback here to `embedding_model` table?

iunir (changes requested):

lightly_studio/src/lightly_studio/dataset/embedding_manager.py
michal-lightly: What is the issue? It seems OK to me.

Note that the class around will change. I'm not sure if this function will stay. If yes, I'd expect the "set_as_default" functionality to move - because it will be possible to register a model without setting a collection_id.

michal-lightly: One issue I see is that `self._collection_id_to_default_model_id` and the db contents of `default_embedding_space` can get out of sync. That will be resolved when the surrounding class changes.

iunir: Oh sorry, misread condition.

lightly_studio/src/lightly_studio/resolvers/embedding_model_resolver.py
michal-lightly: I'd suggest not. I don't think we can guarantee consistency. The collection_id field will move away from the table and then we could report a wrong model for a collection.

But your comment made me notice that this function should not exist, we should rather use `default_embedding_space_resolver.get_by_collection_id`. I'll try to do it in this PR.

iunir (approved):

PR #2033, opened 2026-08-21
Place nulls last for every sort expression

gabrielfruet:
## What has changed and why?

The base sort expression applied direction but never pinned NULL placement, so a `DESC` sort on a nullable column ordered NULLs first on PostgreSQL and last on DuckDB — the same request, two different first rows. `OrderByExpression` now pins NULLs last by default (with a `nulls_last` flag), and the two subclasses that used to paste their own `nullslast` wrapper drop it. This also fixes the evaluation-metric sort, whose docstring promised NULLs-last but never applied it.

First of a four-PR split of the video `sort_by` work. The nullable case surfaces there (video `duration_s`), but the fix is engine-level and stands alone.

## How has it been tested?

`make test` on the ordering suites. The `OrderByField` and evaluation-metric SQL-text assertions now expect `nulls last` — dialect-agnostic guards that fail if the wrapping is removed.

- [ ] Yes

## Summary by CodeRabbit

* **Bug Fixes**
  * Sorting now consistently places missing values last for ascending and descending orders.
  * Image filename and evaluation metric sorting now apply consistent missing-value behavior across ordering variations.
  * Ordering behavior is more consistent across supported database environments.

### DIFF
--- a/lightly_studio/src/lightly_studio/core/dataset_query/order_by.py
+++ b/lightly_studio/src/lightly_studio/core/dataset_query/order_by.py
@@ -7,7 +7,7 @@ from typing import Any, cast
 from uuid import UUID
 
 import sqlalchemy
-from sqlalchemy import ColumnElement, and_, case, func, nullslast, or_
+from sqlalchemy import ColumnElement, and_, case, func, or_
 from sqlalchemy import Select as SQLAlchemySelect
 from sqlalchemy.engine import Row
 from sqlalchemy.orm import Mapped, aliased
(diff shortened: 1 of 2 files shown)


lightly_studio/src/lightly_studio/core/dataset_query/order_by.py
```diff
@@ -166,9 +173,9 @@ class OrderByMetadataField(OrderByExpression):
         field_name: The key inside the JSON ``data`` column to sort by.
     """
 
-    def __init__(self, field_name: str) -> None:
+    def __init__(self, field_name: str, *, nulls_last: bool = True) -> None:
         """Initialize with the metadata field name."""
-        super().__init__()
+        super().__init__(nulls_last=nulls_last)
```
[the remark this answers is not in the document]
gabrielfruet: Done, qualified the three subclass docstrings with the nulls_last opt-out.

lightly_studio/src/lightly_studio/core/dataset_query/order_by.py
```diff
@@ -38,13 +38,16 @@
...
     """Base class for all order by expressions that can be applied to database queries."""
 
-    def __init__(self, *, ascending: bool = True) -> None:
+    def __init__(self, *, ascending: bool = True, nulls_last: bool = True) -> None:
         """Initialize the order by expression.
 
         Args:
             ascending: Whether to order in ascending (True) or descending (False) order.
+            nulls_last: Whether NULLs sort last. Pinned by default so DuckDB and
+                PostgreSQL, which disagree on the default placement, order the same way.
```
michal-lightly: Optional: It would be good to make this more explicit. I thought that `false` places nulls first.

michal-lightly: Actually is there a usecase for not placing nulls consistently? I'd consider dropping the parameter and always do it.

lightly_studio/src/lightly_studio/core/dataset_query/order_by.py
```diff
@@ -136,11 +142,12 @@ class OrderByField(OrderByExpression):
     Args:
         field: The field to order by.
         ascending: Whether to order in ascending (True) or descending (False) order.
+        nulls_last: Whether NULLs sort last.
     """
 
-    def __init__(self, field: Field) -> None:
+    def __init__(self, field: Field, *, nulls_last: bool = True) -> None:
```
michal-lightly: Style: We don't use `*`, we call all arguments with keywords by convention.

michal-lightly (approved):
💯

PR #2034, opened 2026-08-21
Accept video sort expressions in the shared engine

gabrielfruet:
## What has changed and why?

Groundwork so the shared sort engine accepts non-image sample types. Three image assumptions go: `apply_with_order_value` took a single-entity select (the video grid selects a video and its thumbnail frame together), and the metadata and evaluation-metric joins were pinned to `ImageTable.sample_id`. Both now key on `SampleTable`. Adds `VideoSortFieldExpr` and the video sort-field mapping — file name, path, width, height, duration, frame rate, created at, plus metadata.

Second of a four-PR split of the video `sort_by` work. No endpoint uses it yet.

## How has it been tested?

`make test`. New `test_sort.py` coverage for the video sort expressions; the order-by and dataset-query suites stay green.

- [ ] Yes

### DIFF
--- a/lightly_studio/src/lightly_studio/core/dataset_query/order_by.py
+++ b/lightly_studio/src/lightly_studio/core/dataset_query/order_by.py
@@ -95,7 +95,7 @@ class OrderByExpression(ABC):
         joined = self.apply_joins(query)
         return joined.order_by(*self.to_column_elements())
 
-    def apply_with_order_value(self, query: SelectOfScalar[T]) -> SQLAlchemySelect[tuple[T, Any]]:
+    def apply_with_order_value(self, query: SQLAlchemySelect[Any]) -> SQLAlchemySelect[Any]:
         """Apply this sort and append its value to the SELECT list.
 
         Behaves like ``apply`` but also appends the sort value to the SELECT list as
@@ -103,8 +103,11 @@ class OrderByExpression(ABC):
         multi-column ``Select`` (read rows with ``session.execute``) because the sort
         value is added to the row.
 
+        Accepts any ``Select`` rather than a single-entity one, because the video grid
+        selects the video and its thumbnail frame together.
+
         Args:
-            query: The SQLModel Select query to modify.
+            query: The SQLAlchemy Select query to modify.
 
         Returns:
             The query after joining, appending the labeled sort value, and ordering.
@@ -146,7 +149,7 @@ class OrderByField(OrderByExpression):
         self.field = field
 
     def _order_value_expression(self) -> ColumnElement[Any]:
-        """Return the image table column used for sorting."""
+        """Return the column wrapped by ``self.field`` — an image, video or sample column."""
         return cast(ColumnElement[Any], self.field.get_sqlmodel_field())
 
     def apply_joins(self, query: SelectT) -> SelectT:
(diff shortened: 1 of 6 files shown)


lightly_studio/src/lightly_studio/core/dataset_query/order_by.py
```diff
@@ -215,10 +218,15 @@ def _extracted_value(self) -> ColumnElement[Any]:
...
+        """Left-outer-join aliased ``SampleMetadataTable`` on ``sample_id``.
+
+        Joins through ``SampleTable`` rather than a concrete sample table so the same
+        expression works for image, video and video-frame queries. Every query that can
+        carry this expression joins its sample to filter by collection.
+        """
         return query.outerjoin(
             self._metadata_alias,
-            col(self._metadata_alias.sample_id) == col(ImageTable.sample_id),
+            col(self._metadata_alias.sample_id) == col(SampleTable.sample_id),
```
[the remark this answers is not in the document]
gabrielfruet: Decided not to act

lightly_studio/src/lightly_studio/core/dataset_query/order_by.py
```diff
@@ -95,16 +95,19 @@ def apply(self, query: SelectOfScalar[T]) -> SelectOfScalar[T]:
         joined = self.apply_joins(query)
         return joined.order_by(*self.to_column_elements())
 
-    def apply_with_order_value(self, query: SelectOfScalar[T]) -> SQLAlchemySelect[tuple[T, Any]]:
+    def apply_with_order_value(self, query: SQLAlchemySelect[Any]) -> SQLAlchemySelect[Any]:
```
michal-lightly, line 109: Is it possible not to lose the typing information? The rest of the file is preserving it.

Perhaps we can enumerate the cases?

lightly_studio/tests/models/test_sort.py
```diff
@@ -91,6 +101,69 @@ def test_sort_field_expr_to_order_by__all_fields_map() -> None:
...
+
+
+def test_sort_field_expr__rejects_video_source() -> None:
+    # Image queries do not have `VideoTable` in the FROM clause, so a video field
+    # would be cross-joined rather than sorted by.
+    with pytest.raises(ValidationError):
+        SortFieldExpr.model_validate({"source": "video", "field_name": "fps", "direction": "asc"})
+
+
+def test_video_sort_field_expr_to_order_by__all_fields_map() -> None:
```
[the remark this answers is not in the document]
michal-lightly: The test name is incorrect, the justification is partially correct.

Test for a function foo should be named `test_foo__something`. In this case it should be `test_sort_field_expr_to_order_by__video_all_fields_map`.

A second alternative (maybe a bit less philosophically clean) is to group the tests into classes, since we are essentially test `SortFieldExpr` and `VideoSortFieldExpr`. The classes would be `TestSortFieldExpr` and `TestVideoSortFieldExpr`.

Both options are fine by me.

lightly_studio/tests/models/test_sort.py
```diff
@@ -91,6 +101,69 @@ def test_sort_field_expr_to_order_by__all_fields_map() -> None:
...
 
+def test_sort_field_expr_to_order_by__image_field_name_not_shared_with_video() -> None:
+    # `duration_s` exists on videos only; the source must select the right registry.
+    expr = SortFieldExpr(
+        source=SortFieldSource.image,
+        field_name="duration_s",
+        direction=SortDirection.asc,
+    )
+    with pytest.raises(QueryExprError):
+        sort_field_expr_to_order_by(expr)
```
michal-lightly: Minor: Add the keyword `expr=expr`. Also elsewhere.

lightly_studio/src/lightly_studio/core/dataset_query/order_by.py
gabrielfruet: Using overloads now

lightly_studio/tests/models/test_sort.py
gabrielfruet: fixed

LeonardoRosaa (approved):

PR #2043, opened 2026-08-21
Add prepare-release tooling: changelog promotion

iunir:
Part of a stack ([gh-stack](https://github.github.com/gh-stack/)), 1 of 4, landing bottom-up onto `main`:

1. **#2043 - changelog promotion** (this PR)
2. #2044 - version, pyproject.toml, and uv.lock guards
3. #2045 - PR body rendering (draft notes + coverage checklist)
4. #2046 - CLI + GitHub Actions workflow

Please review/merge in order; each PR rebases onto the previous once it lands.

## What has changed and why?

First layer of the **Prepare Release** tooling: promote `CHANGELOG.md`'s
`[Unreleased]` section into a released version block, and assert the
result is structurally sound before anything downstream trusts it.

Lives in a standalone `.github/scripts/prepare_release` package
(stdlib only - it runs before `uv sync`), with its own
`pyproject.toml` for lint/type/test config.

- `changelog.py`: parses the six `[Unreleased]` subsections, suggests
  a semver bump, promotes them into a dated `## [X.Y.Z] - YYYY-MM-DD`
  heading, and inserts a fresh empty `[Unreleased]` skeleton above it.
- `assert_changelog_structure`: independently re-checks the result -
  one `[Unreleased]` heading, the new heading present, every earlier
  released block untouched - and fails loudly if not.

No CLI yet - that's wired up in the last PR of the stack.

## How has it been tested?

- 11 unit tests covering the happy path and each failure case.
- `ruff format --check`, `ruff check`, and `mypy --strict` pass.
- Hand-ran against this repo's real `CHANGELOG.md`.

---
Suggested reviewer: @michal-lightly. Alternate: @lukas-lightly.

## Summary by CodeRabbit

* **New Features**
  * Added automated changelog release preparation, promoting unreleased entries into versioned release sections.
  * Added extraction of released changelog content for release workflows.

* **Bug Fixes**
  * Added validation for malformed sections, duplicate releases, incorrect ordering, stray content, and unintended changes.
  * Release preparation now reports clear errors when changelog requirements are not met.

* **Chores**
  * Added consistent linting, testing, and type-checking configuration for release tooling.
  * Added cleanup rules for generated Python cache files.

### DIFF
--- a/.github/scripts/prepare_release/changelog.py
+++ b/.github/scripts/prepare_release/changelog.py
@@ -25,48 +25,6 @@ _ANY_H2_RE = re.compile(r"^## ", re.MULTILINE)
 _SUBSECTION_RE = re.compile(r"^### (?P<name>\w+)[ \t]*$", re.MULTILINE)
 
 
-def parse_unreleased_sections(changelog_text: str) -> dict[str, str]:
-    """Extracts the `[Unreleased]` section's six subsections.
-
-    Returns:
-        A mapping from subsection name (e.g. "Added") to its raw body text
-        (may be empty or whitespace-only).
-    """
-    body = _unreleased_body(changelog_text)
-    headings = list(_SUBSECTION_RE.finditer(body))
-    names = [h.group("name") for h in headings]
-    if names != list(SUBSECTIONS):
-        raise PrepareReleaseError(
-            f"[Unreleased] subsections are {names}, expected {list(SUBSECTIONS)} in that order"
-        )
-    if body[: headings[0].start()].strip():
-        raise PrepareReleaseError(
-            "unexpected content between the [Unreleased] heading and its first subsection"
-        )
-    sections = {}
-    for i, heading in enumerate(headings):
-        start = heading.end()
-        end = headings[i + 1].start() if i + 1 < len(headings) else len(body)
-        sections[heading.group("name")] = body[start:end]
-    return sections
-
-
-def suggest_bump(sections: dict[str, str]) -> tuple[str, str]:
-    """Suggests a semver bump from which [Unreleased] sections are non-empty.
-
-    Added / Changed / Removed / Deprecated entries suggest a minor bump;
-    only Fixed and/or Security entries suggest a patch. The operator still
-    confirms - this is a suggestion, not a decision.
-    """
-    non_empty = [name for name in SUBSECTIONS if sections.get(name, "").strip()]
-    if not non_empty:
-        raise PrepareReleaseError("the [Unreleased] section has no entries; nothing to release")
-    minor_triggers = [n for n in non_empty if n in ("Added", "Changed", "Removed", "Deprecated")]
-    bump = "minor" if minor_triggers else "patch"
-    reasoning = f"Non-empty [Unreleased] sections: {', '.join(non_empty)}."
-    return bump, reasoning
-
-
 def promote_changelog(changelog_text: str, version: str, date: str) -> str:
     """Promotes `[Unreleased]` to a released version block.
 
@@ -77,7 +35,7 @@ def promote_changelog(changelog_text: str, version: str, date: str) -> str:
     """
     unreleased_match = _single_unreleased_match(changelog_text)
     _, body_end = _unreleased_body_span(changelog_text, unreleased_match)
-    sections = parse_unreleased_sections(changelog_text)
+    sections = _parse_unreleased_sections(changelog_text)
 
     preamble = changelog_text[: unreleased_match.start()]
     rest = changelog_text[body_end:]
@@ -102,6 +60,7 @@ def assert_changelog_structure(original_text: str, new_text: str, version: str) 
     logic itself:
 
     * exactly one `## [Unreleased]` heading, with all six subheadings,
+    * the fresh `[Unreleased]` section has no entries,
     * exactly one new `## \\[version\\] - YYYY-MM-DD` heading, preceded by
       the `[Unreleased]` heading,
     * every previously-released version block is byte-identical to before.
@@ -113,7 +72,11 @@ def assert_changelog_structure(original_text: str, new_text: str, version: str) 
             f"{len(unreleased_matches)}"
         )
     # Raises PrepareReleaseError itself if a subheading is missing/misordered.
-    parse_unreleased_sections(new_text)
+    fresh_sections = _parse_unreleased_sections(new_text)
+    if any(content.strip() for content in fresh_sections.values()):
+        raise PrepareReleaseError(
+            "expected the fresh [Unreleased] section to have no entries after promotion"
+        )
 
     released_heading = re.compile(
         r"^## \\\[" + re.escape(version) + r"\\\] - \d{4}-\d{2}-\d{2}[ \t]*$",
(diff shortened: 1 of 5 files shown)


.github/scripts/prepare_release/changelog.py
```diff
@@ -0,0 +1,165 @@
...
+    if names != list(SUBSECTIONS):
+        raise PrepareReleaseError(
+            f"[Unreleased] subsections are {names}, expected {list(SUBSECTIONS)} in that order"
+        )
+    sections = {}
+    for i, heading in enumerate(headings):
+        start = heading.end()
+        end = headings[i + 1].start() if i + 1 < len(headings) else len(body)
+        sections[heading.group("name")] = body[start:end]
+    return sections
```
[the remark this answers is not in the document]
iunir: Good catch, confirmed by execution — content between the `## [Unreleased]` heading and the first `### Added` heading was never captured into `sections`, `preamble`, `skeleton`, or `promoted`, so it silently disappeared during promotion. Fixed in 4ab34330: `parse_unreleased_sections` now raises `PrepareReleaseError` if that gap is non-empty, with a regression test.

.github/scripts/prepare_release/changelog.py
```diff
@@ -0,0 +1,165 @@
...
+        )
+
+    original_match = _single_unreleased_match(original_text)
+    _, original_body_end = _unreleased_body_span(original_text, original_match)
+    previously_released = original_text[original_body_end:]
+    if not new_text.endswith(previously_released):
+        raise PrepareReleaseError(
+            "previously-released changelog blocks are not byte-identical after promotion"
+        )
+
```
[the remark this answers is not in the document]
iunir: Agreed — `assert_changelog_structure`'s whole point is to independently re-derive the invariants rather than trust `promote_changelog`'s construction, so it should catch a swapped order too. Fixed in 4ab34330, plus while in there I also tightened the released-heading check from "exists" to "exists exactly once" (it previously wouldn't catch promoting to an already-released version number — same class of gap, confirmed empirically). Added regression tests for both.

.github/scripts/tests/test_changelog.py
```diff
@@ -0,0 +1,133 @@
...
+
+
+def test_promote_changelog():
+    promoted = changelog.promote_changelog(SAMPLE_CHANGELOG, "1.1.0", "2026-08-21")
+
+    # Fresh, fully-empty [Unreleased] skeleton with all six subheadings.
+    assert promoted.count("## [Unreleased]") == 1
+    for name in changelog.SUBSECTIONS:
+        assert f"### {name}\n\n" in promoted
+
```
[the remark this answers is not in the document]
iunir: Fair point, the loop only checked the six subheadings exist somewhere in `promoted` (they'd match inside the promoted release block too, not just the fresh skeleton). Fixed in 4ab34330: now parses the fresh `[Unreleased]` section via `parse_unreleased_sections` and asserts every subsection body is blank.

.github/scripts/prepare_release/changelog.py
```diff
@@ -0,0 +1,176 @@
...
+        )
+    sections = {}
+    for i, heading in enumerate(headings):
+        start = heading.end()
+        end = headings[i + 1].start() if i + 1 < len(headings) else len(body)
+        sections[heading.group("name")] = body[start:end]
+    return sections
+
+
+def suggest_bump(sections: dict[str, str]) -> tuple[str, str]:
```
michal-lightly: Let's not guess from the changelog what bump should happen - always suggest a "patch" bump. We put frontend layout changes to "Changes" section which do not affect backwards compatibility. Let's remove this function.

.github/scripts/prepare_release/changelog.py
```diff
@@ -0,0 +1,176 @@
...
+# `## [Unreleased]` is the one heading kept unescaped; every released version
+# heading is `## \[X.Y.Z\] - YYYY-MM-DD`. Getting this backwards silently
+# breaks the convention instead of failing loudly, which is exactly what
+# `assert_changelog_structure` guards against.
+_UNRELEASED_RE = re.compile(r"^## \[Unreleased\][ \t]*$", re.MULTILINE)
+_ANY_H2_RE = re.compile(r"^## ", re.MULTILINE)
+_SUBSECTION_RE = re.compile(r"^### (?P<name>\w+)[ \t]*$", re.MULTILINE)
+
+
+def parse_unreleased_sections(changelog_text: str) -> dict[str, str]:
```
michal-lightly: I suppose this should be a "private" function - I suppose only `promote_changelog` and `assert_changelog_structure` will be called from outside?

.github/scripts/prepare_release/changelog.py
```diff
@@ -0,0 +1,176 @@
...
+    original_match = _single_unreleased_match(original_text)
+    _, original_body_end = _unreleased_body_span(original_text, original_match)
+    previously_released = original_text[original_body_end:]
+    if not new_text.endswith(previously_released):
+        raise PrepareReleaseError(
+            "previously-released changelog blocks are not byte-identical after promotion"
+        )
+
+
+def extract_released_section(changelog_text: str, version: str) -> str:
```
michal-lightly, line 117: What is this function for?

.github/scripts/prepare_release/changelog.py
```diff
@@ -0,0 +1,176 @@
...
+
+
+def parse_unreleased_sections(changelog_text: str) -> dict[str, str]:
+    """Extracts the `[Unreleased]` section's six subsections.
+
+    Returns:
+        A mapping from subsection name (e.g. "Added") to its raw body text
+        (may be empty or whitespace-only).
+    """
+    body = _unreleased_body(changelog_text)
```
michal-lightly: Minor: The PR would be easier to review if it used keyword arguments for functions (as we have in the guideline).

.github/scripts/pyproject.toml
```diff
@@ -0,0 +1,26 @@
+# Tooling config only - this directory is CI-internal tooling for the
+# "Prepare Release" workflow, not a distributed package, so there's no
+# [project] table here.
+
+[tool.pytest.ini_options]
+pythonpath = "."
+testpaths = ["tests"]
+
+[tool.ruff]
```
michal-lightly, line 9: How did you choose the linter and mypy settings?

michal-lightly (changes requested):

.github/scripts/pyproject.toml
iunir: Copied from lightly_studio/ruff.toml

.github/scripts/prepare_release/changelog.py
iunir: done

.github/scripts/prepare_release/changelog.py
iunir: It's consumed downstream by the release PR body renderer (render-pr-body in #2046)

.github/scripts/prepare_release/changelog.py
iunir: done

.github/scripts/prepare_release/changelog.py
iunir: dropped this function

michal-lightly (approved):

.github/scripts/prepare_release/changelog.py
```diff
@@ -0,0 +1,169 @@
...
+            f"expected the [Unreleased] heading to precede the new '[{version}]' release heading"
+        )
+
+    original_match = _single_unreleased_match(original_text)
+    _, original_body_end = _unreleased_body_span(original_text, original_match)
+    previously_released = original_text[original_body_end:]
+    if not new_text.endswith(previously_released):
+        raise PrepareReleaseError(
+            "previously-released changelog blocks are not byte-identical after promotion"
+        )
```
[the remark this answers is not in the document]
iunir: Confirmed and fixed in ce3a917b. Reproduced it empirically first: a construction bug that silently drops the original `Added`/`Changed` entries (while keeping the fresh skeleton empty, the release heading valid and correctly positioned, and the previously-released suffix untouched) passed `assert_changelog_structure` with no error before this fix.

Applied your proposed approach as-is: derives the expected promoted body from `original_text` via `_parse_unreleased_sections` and compares it against `extract_released_section(new_text, version)`. Added `test_assert_changelog_structure__dropped_entry_raises` as regression coverage.

PR #2052, opened 2026-08-24
Validate embedding vectors before writing to sample_embedding

iunir:
## What has changed and why?

Adds validation before embeddings are written to `sample_embedding`, in the shared `_store_embeddings` write path used by both image/annotation/video generators and the precomputed-embeddings import path. Rejects with a clear `ValueError` before any DB write:

- count mismatch between embeddings and sample IDs
- non-numeric dtype (e.g. strings), or a cast to float32 that overflows to Inf
- NaN/Inf values
- vector dimension not matching the embedding model's declared `embedding_dimension`

Non-float32 numeric dtypes (float64, ints) are safely cast to float32 rather than rejected outright — float64 is numpy's default, so real precomputed-embedding imports commonly arrive that way, and losing precision beyond float32 is expected for embedding storage.

Previously none of this was checked; bad rows landed in the DB and only failed later, at query time, with an opaque DuckDB error.

## How has it been tested?

- Added unit tests in `test_embedding_manager.py` covering each rejected case (count mismatch, non-numeric dtype, overflow-on-cast, NaN, Inf, dimension mismatch), the float64-safe-cast case, and a test confirming `_store_embeddings` writes nothing when validation fails.
- `make static-checks` and `make test` pass; existing built-in generator tests pass unchanged.

- [ ] Yes

Reviewer: @JonasWurst (most history on this file), alt: @horacio-almasan

## Bug Fixes

* Improved embedding validation and storage reliability.
* Numeric embeddings are consistently converted to the supported format, including higher-precision values.
* Invalid, non-numeric, non-finite, mismatched, incorrectly sized, or non-2-D embeddings are rejected before storage.
* Prevented invalid embedding data from being written.
* Clarified errors for embeddings with an invalid shape.

## Tests

* Expanded coverage for valid, empty, higher-precision, overflowing, non-finite, and invalid embedding batches.
* Added coverage for dimensionality checks, consistent embedding shapes, and reliable persistence.

### DIFF
--- a/lightly_studio/src/lightly_studio/dataset/embedding_manager.py
+++ b/lightly_studio/src/lightly_studio/dataset/embedding_manager.py
@@ -540,9 +540,9 @@ def _store_embeddings(
     so a failure leaves no partially embedded dataset behind.
 
     Raises:
-        ValueError: If the embeddings fail validation. See `_validate_embeddings_for_storage`.
+        ValueError: If the embeddings fail validation. See `_validate_and_coerce_embeddings`.
     """
-    _validate_embeddings_for_storage(
+    embeddings = _validate_and_coerce_embeddings(
         session=session, model_id=model_id, sample_ids=sample_ids, embeddings=embeddings
     )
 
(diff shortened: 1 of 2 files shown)


lightly_studio/src/lightly_studio/dataset/embedding_manager.py
```diff
@@ -565,6 +572,49 @@ def _store_embeddings(
...
+        )
+    if len(embeddings) == 0:
+        return
+
+    if embeddings.ndim != 2:  # noqa: PLR2004
+        raise ValueError(f"Embeddings must be a 2-D array, got {embeddings.ndim}-D.")
+    if embeddings.dtype != np.float32:
+        raise ValueError(f"Embeddings must be float32, got dtype {embeddings.dtype}.")
+    if not np.isfinite(embeddings).all():
+        raise ValueError("Embeddings must not contain NaN or infinite values.")
```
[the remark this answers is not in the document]
iunir: fixed

lightly_studio/tests/dataset/test_embedding_manager.py
```diff
@@ -846,3 +846,242 @@ def test_compute_image_embedding(
...
     ).all()
     assert len(stored_embeddings) == 0
+
+
+# Dimension of the model registered by _register_random_model, used to build
+# well-formed test embeddings below.
+_RANDOM_MODEL_DIMENSION = 3
+
+
+def _register_random_model(session: Session, collection: CollectionTable) -> UUID:
```
horatiualmasan: Minor: you can move the helper to the end of the file

horatiualmasan (approved):
LGTM

lightly_studio/tests/dataset/test_embedding_manager.py
iunir: Done — moved the helper to the end of the file.

PR #2055, opened 2026-08-24
Add recompute endpoint for stale evaluation runs

LeonardoRosaa:
## What has changed and why?

Adds a dedicated `POST /api/datasets/{dataset_id}/evaluation/runs/{run_id}/recompute` endpoint to refresh a stale evaluation run in place. The endpoint deletes the run’s existing `EvaluationAnnotationMetric` and `EvaluationSampleMetric` rows, re-runs the evaluation using the stored config, persists fresh metrics, and resets `stale_since` to `None`.

The delete methods do not commit on their own; the caller is responsible for committing or rolling back the transaction if something goes wrong.

## How has it been tested?

Unit tests

- [ ] Yes

---


LeonardoRosaa:
> If metric generation fails after any batch is written—for example, semantic segmentation encounters an invalid sample after its first 32 rows—the metric resolvers' create_many functions commit each batch, which also commits these pending deletes. The request then fails with the old metrics irreversibly removed and only a partial replacement persisted. Compute before deleting, or make deletion and all inserts one transaction whose callees do not commit independently.

I agree. However, it will require a large redesign. I'd avoid this ATM.

> Staleness is applied only in this API-oriented service, while the public Python API still mutates the same annotation collections directly: core/sample.py:302 calls annotation_resolver.create_many and core/sample.py:317 calls annotation_resolver.delete_annotation. Using Sample.add_annotation(s) or Sample.delete_annotation therefore leaves affected runs marked fresh even though their inputs changed. Centralize invalidation in the mutation resolver or cover every public mutation path.

Addressed in another PR.

> When an annotation mutation commits `mark_stale_by_collection_id` while recomputation is running, this unconditional clear can erase that newer timestamp after metrics were computed from an earlier snapshot, making outdated results appear fresh. Clear only if `stale_since` still equals the value observed when recomputation began, or serialize recomputation against annotation mutations.

Addressed the race condition

> Split this helper into focused operations for coverage selection, annotation loading, and EvaluationData assembly; at 34 lines it exceeds the repository's under-20-line guideline and combines several responsibilities.
> 
> AGENTS.md reference: [AGENTS.md:L10-L12](https://github.com/lightly-ai/lightly-studio/blob/5cad2baaf74d881e52650b045738bbcff9a79e0c/AGENTS.md#L10-L12)

IMO, the function is well-scoped. I wouldn't change it.

horatiualmasan (approved):

PR #2056, opened 2026-08-24
Let users recompute evaluation runs from the GUI

LeonardoRosaa:
## What has changed and why?

Users can recompute evaluation runs from the GUI when annotations in the source collections change and make the current results stale. This adds an inline warning icon and a recompute button in the evaluation metrics details.

Also invalidate the evaluation run query after annotation mutations.

recompute

## How has it been tested?

Unit tests + manual test

- [ ] Yes

---


LeonardoRosaa:
> When a run was created from a filtered request or derived collection, its original sample IDs are a subset of the two sources' full coverage. Replacing that scope with gt_covered & pred_covered silently recomputes the run over every jointly covered sample in the dataset, changing its counts and metrics while preserving the same run ID and name. Persist and reuse the original selection or its collection/filter definition.


> If metric generation fails or the request is interrupted after a batch is persisted, the metric resolvers' create_many() calls have already committed the preceding deletes and any completed batches. For example, an error in a later semantic-segmentation sample leaves the run with only a partial replacement while the endpoint reports failure. Compute/stage the complete replacement and commit the delete, inserts, and stale-state update in one transaction.

> If another request edits either annotation source after _fetch_annotations_for_run() but before this statement, that edit sets a newer stale_since, yet this unconditional update clears it. The recomputed metrics can therefore reflect the pre-edit annotations while the run is advertised as fresh. Clear staleness conditionally against the timestamp observed at recompute start, or lock/version the run.

All of those issues have been addressed in another PR.

horatiualmasan (approved):

PR #2065, opened 2026-08-25
Allow existig tag for preselected_tag_name sampling

horatiualmasan:
## What has changed and why?

When continuing sampling from an existing tagged selection with the `preselected_tag_name` parameter you can add the same tag name as both input and output. This allows the tag to grow obtaining the same functionality as the data pool from lightly one.

## How has it been tested?

Existing tests pass and new tests added.

- [x] Yes

## Summary by CodeRabbit

* **New Features**
  * Sampling results can now reuse the preselected tag, appending newly selected samples across repeated runs.
  * Existing result tags are extended when appropriate, while conflicts with unrelated tags are rejected.

* **Documentation**
  * Updated sampling guidance, API documentation, and changelog to explain tag reuse and conflict behavior.

* **Tests**
  * Added coverage for accumulating samples and validating conflicting tag names.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Sort the annotations grid by a per-annotation evaluation metric, such as IoU.
 - Python SDK: Order video queries by `VideoSampleField.created_at`.
 - Python SDK: Continue sampling from an existing tagged selection with the
-  `preselected_tag_name` parameter.
+  `preselected_tag_name` parameter. Passing the same name as `sampling_result_tag_name`
+  grows that tag with the newly selected samples instead of requiring a fresh tag.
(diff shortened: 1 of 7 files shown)


iunir (approved):
Looks good. My Claude found this. Check before merging, seems valid, but not critical


1. sampling_helpers.py:53 — create_result_tag does a redundant get_by_name lookup for a tag name that was already looked up moments earlier in sampling_via_database (either via _check_result_tag_name_free or _get_preselected_sample_ids). That's an extra DB round-trip per call that didn't exist before (old code called tag_resolver.create directly).
2. sampling_via_db.py:329 — the "don't silently merge into an unrelated existing tag" invariant is only enforced at this one call site, not inside create_result_tag/get_or_create_sample_tag_by_name itself. A future direct caller of the shared helper that skips this check could silently append samples into a same-named unrelated tag — the helper's docstring says callers "must have rejected any other name collision beforehand" but doesn't enforce it.

Issue #2019, opened 2026-08-20
[FEAT] Expose aggregate evaluation metrics and readable results in the Python API


lorinczszabolcs:
### 💡 Is your feature request related to a problem?

`dataset.evaluate().object_detection()`, `.classification()`, and `.semantic_segmentation()` run an evaluation and store per-sample metrics. The returned `EvaluationResult` gives only the run id and the input counts (`sample_count`, `gt_annotation_count`, `pred_annotation_count`). The Python SDK cannot read the results that matter:

- dataset-level mAP, accuracy, and mIoU
- per-class precision, recall, and F1
- the confusion matrix
- the list of past evaluation runs

Today these are reachable only through internal resolvers or REST routes. Otherwise the results stay in the GUI.

### 🧰 Describe the solution you'd like

Add readback methods to the existing `ImageDatasetEvaluate`:

- `evaluate().list_runs()` -> `list[EvaluationRunView]`
- `evaluate().confusion_matrix(run)` -> `ConfusionMatrix` (already computed internally, REST-only today)
- `evaluate().metrics(run)` -> ... for the aggregates: detection mAP and per-class AP; classification accuracy and per-class precision/recall/F1; segmentation per-class IoU, dataset mIoU, and pixel accuracy.

Then a user can assert model quality from a script and/or in CI, matching the Python-first workflow.

### 🛠 Alternatives you've considered

- The query DSL (`SampleEvaluationQuery`) filters samples by a metric. It does not return values or aggregates.
- The internal resolvers work, and the shipped `examples/example_evaluation.py` uses `evaluation_sample_metric_resolver`. The `get_confusion_matrix` resolver is reachable only through the REST route. But resolvers are not exported in the public `lightly_studio` API, and they are not in the API docs (`docs/api/`). So they do not seem to be the supported entry point for user code.

### 📝 Additional context

I checked this against `main` and no aggregate metric is computed today. Detection stores per-sample `tp/fp/fn` and per-match IoU at one threshold. Classification stores `disagreement`. Segmentation stores per-image mIoU. The confusion matrix is computed on read, but only through the REST route.

[...]

PR #2066, opened 2026-08-25
feat: read evaluation runs and the confusion matrix from the Python API

lorinczszabolcs:
## What has changed and why?

`dataset.evaluate()` can create evaluation runs, but the Python SDK cannot read the results. The runs, their configuration, and the confusion matrix are reachable only through the GUI or internal resolvers. This adds two read methods to `ImageDatasetEvaluate`:

- `evaluate().list_runs()` returns the stored runs, newest first, as `EvaluationRunView` objects (id, name, configuration, creation time, and resolved source names).
- `evaluate().confusion_matrix(run_id)` returns the confusion matrix of a run. Object detection and classification are supported; segmentation runs raise `NotImplementedError`.

A user can now assert model quality from a script or in CI, without the GUI.

This is the first of three staged PRs for #2019, in the order confirmed on the issue (readback → confusion-derived aggregates → mean average precision). It needs no schema change and no new dependency.

The first commit is a behavior-preserving refactor: it moves the run-to-view listing and the confusion-matrix task-type check into the resolvers, so the REST routes and the Python SDK share one implementation. The second commit adds the two read methods.

Part of #2019.

## How has it been tested?

- New unit tests for both methods and the shared resolver: `list_runs()` (populated and empty), `confusion_matrix()` (populated, run not found, unsupported task type), and the run-view resolver.
- `make static-checks` (ruff, ruff format, mypy) passes.
- The evaluation, service, and route test suites pass.
- The docs build passes with `--strict`; the new methods and the `EvaluationRunView` and `ConfusionMatrix` models render in the API reference.

- [x] Yes

## Summary by CodeRabbit

* **New Features**
  * Added Python SDK support for listing stored evaluation runs, ordered newest first.
  * Added confusion matrix retrieval for classification and object detection evaluations.
  * Confusion matrices include labels for missing ground-truth or prediction annotations.
  * Evaluation run results include source names and configuration details.
  * Added validation for missing runs, runs from another dataset, and unsupported segmentation confusion matrices.

* **Documentation**
  * Documented evaluation run listing, confusion matrix retrieval, supported task types, and related API models.

### DIFF
--- a/lightly_studio/docs/docs/concepts_and_tools/evaluation.md
+++ b/lightly_studio/docs/docs/concepts_and_tools/evaluation.md
@@ -155,4 +155,28 @@ result = dataset.evaluate(query=val_query).object_detection(
 Only samples that appear in both annotation sources **and** in the query are included. Samples
 missing from either source are silently skipped.
 
+### Reading evaluation results
+
+`evaluate()` also reads back the runs that were already stored. This lets a script assert model
+quality without the GUI.
+
+Use `list_runs()` to get the stored runs, newest first. Each run is an
+[`EvaluationRunView`](../api/evaluation.md#evaluationrunview) with its id, name, configuration,
+creation time, and source names. Use `confusion_matrix()` to get the confusion matrix of a run by
+its id.
+
+```python
+evaluator = dataset.evaluate()
+
+runs = evaluator.list_runs()
+for run in runs:
+    print(run.name, run.created_at, run.evaluation_run_configuration)
+
+matrix = evaluator.confusion_matrix(run_id=runs[0].id)
+print(matrix.row_labels, matrix.col_labels, matrix.counts)
+```
+
+`confusion_matrix()` supports object-detection and classification runs. It does not support
+segmentation runs.
+
 See [Evaluation API Reference](../api/evaluation.md) for the full API surface.
--- a/lightly_studio/src/lightly_studio/evaluation/image_dataset_evaluate.py
+++ b/lightly_studio/src/lightly_studio/evaluation/image_dataset_evaluate.py
@@ -17,15 +17,18 @@ from lightly_studio.evaluation import (
 )
 from lightly_studio.evaluation.evaluation_data import EvaluationData
 from lightly_studio.models.annotation.annotation_base import AnnotationBaseTable
+from lightly_studio.models.evaluation_confusion_matrix import ConfusionMatrix
 from lightly_studio.models.evaluation_run import (
     EvaluationRunCreate,
     EvaluationRunTable,
+    EvaluationRunView,
     EvaluationTaskType,
 )
 from lightly_studio.resolvers import (
     annotation_collection_coverage_resolver,
     annotation_resolver,
     collection_resolver,
+    evaluation_annotation_metric_resolver,
     evaluation_run_resolver,
 )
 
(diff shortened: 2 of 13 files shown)


lightly_studio/docs/docs/concepts_and_tools/evaluation.md
```diff
@@ -155,4 +155,28 @@ result = dataset.evaluate(query=val_query).object_detection(
...
+its id.
+
+```python
+evaluator = dataset.evaluate()
+
+runs = evaluator.list_runs()
+for run in runs:
+    print(run.name, run.created_at, run.evaluation_run_configuration)
+
+matrix = evaluator.confusion_matrix(run_id=runs[0].id)
```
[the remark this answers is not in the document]
lorinczszabolcs: I think it's not necessary to write defensive boilerplate in a teaching example, but let me know if you would like me to address it.

lightly_studio/src/lightly_studio/resolvers/evaluation_run_resolver/list_views_by_dataset_id.py
```diff
@@ -0,0 +1,48 @@
...
+        | {run.pred_annotation_collection_id for run in runs},
+    )
+    return [
+        EvaluationRunView(
+            id=run.id,
+            name=run.name,
+            evaluation_run_configuration=run.config_json,
+            created_at=run.created_at,
+            gt_annotation_source=collection_name_by_id[run.gt_annotation_collection_id],
+            pred_annotation_source=collection_name_by_id[run.pred_annotation_collection_id],
```
iunir: `get_names_by_ids` documents *"Missing IDs are omitted"* and `EvaluationRunView` types both fields `str | None`, but direct indexing raises `KeyError` instead — so that `| None` is unreachable, and one unresolvable id takes down the whole listing.

```python
gt_annotation_source=collection_name_by_id.get(run.gt_annotation_collection_id),
pred_annotation_source=collection_name_by_id.get(run.pred_annotation_collection_id),
```

Pre-existing behavior moved from the route, but now behind a public SDK method.

lightly_studio/src/lightly_studio/evaluation/image_dataset_evaluate.py
```diff
@@ -211,6 +214,57 @@ def semantic_segmentation(
         )
         return EvaluationResult.from_evaluation_data(data)
 
+    def list_runs(self) -> list[EvaluationRunView]:
```
iunir, line 217: `list_runs()` and `confusion_matrix()` never read `self.sample_ids`, but can't avoid paying for it: `ImageDataset.evaluate()` builds it eagerly at `core/image/image_dataset.py:730`, iterating every `ImageTable` row in the collection. So `dataset.evaluate().list_runs()` scans the whole dataset to return a handful of run rows - twice over, in the docs example.

Not introduced here: until now every method on the facade needed `sample_ids`. These are the first that don't.

`sample_ids` has one reader (`_prepare_evaluation_data`), so it could move behind a lazy property with `evaluate()` passing `query` through. Fine as a follow-up too, though PRs 2 and 3 will inherit the same cost.

iunir (changes requested):
Thanks a lot @lorinczszabolcs for contributing.
Please address key error one, and the second one (lazy sample_ids) is not critical

lightly_studio/src/lightly_studio/evaluation/image_dataset_evaluate.py
lorinczszabolcs: Thanks for the suggestion! I will create a follow-up PR before I continue with PRs 2 and 3.

iunir (approved):

PR #2069, opened 2026-08-25
Resolve classifier embedding model by dataset_id

michal-lightly:
## What has changed and why?

The long-term goal is to remove `embedding_model.collection_id` and enforce one embedding-model row per `(dataset_id, embedding_model_hash)`. This needs a staged rollout: dataset-scoped creators would break collection-scoped readers, while dataset-scoped readers must tolerate the existing per-collection duplicate rows until the data migration.

This PR moves the classifier reader to the new scope without changing the shared write contract:

- `create_classifier` resolves the collection default through `default_embedding_space`.
- Classifier hash lookups resolve within the dataset. While duplicate rows exist, the collection default is preferred so classifier-created sample embeddings keep using the matching physical row; dataset scope remains the fallback and final state.
- `get_or_create` keeps deduplicating per collection through `get_by_model_hash_deprecated`, so the remaining collection-scoped readers continue to work.

Follow-ups will migrate the remaining readers and creators, deduplicate embedding-model rows while repointing foreign keys, add the unique constraint, and remove `embedding_model.collection_id` plus the temporary compatibility path.

Extracted from #2067. Depends on the dataset-ID and default-embedding-space groundwork in `michal-lig-10548-add-dataset-id-to-model-table`.

## How has it been tested?

- `make static-checks`
- `pytest tests/few_shot_classifier tests/resolvers/test_embedding_model_resolver.py tests/dataset/test_embedding_manager.py`
- Focused regression rerun: 39 passed

- [ ] Yes

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved embedding model resolution for few-shot classifiers.
  * Classifiers now consistently use the collection’s default embedding model when available.
  * Prevented incorrect model selection when related collections share a dataset.
  * Improved handling of missing or unmatched embedding models across training, loading, fine-tuning, and inference.

* **Tests**
  * Added regression coverage for collection-scoped model lookup, collection isolation, and default model prioritization.

### DIFF
--- a/lightly_studio/src/lightly_studio/few_shot_classifier/classifier_manager.py
+++ b/lightly_studio/src/lightly_studio/few_shot_classifier/classifier_manager.py
@@ -133,7 +133,8 @@ class ClassifierManager:
         embedding_model = embedding_model_resolver.get_by_id(
             session=session, embedding_model_id=embedding_model_id
         )
-        assert embedding_model is not None, "Default embedding space references a missing model."
+        if embedding_model is None:
+            raise ValueError("Default embedding space references a missing model.")
         classifier = RandomForest(
             name=name,
             classes=class_list,
@@ -632,7 +633,8 @@ def _get_embedding_model_by_hash(
         default_embedding_model = embedding_model_resolver.get_by_id(
             session=session, embedding_model_id=default_embedding_model_id
         )
-        assert default_embedding_model is not None
+        if default_embedding_model is None:
+            raise ValueError("Default embedding space references a missing model.")
         if default_embedding_model.embedding_model_hash == embedding_model_hash:
             return default_embedding_model
 
--- a/lightly_studio/src/lightly_studio/few_shot_classifier/classifier_manager.py
+++ b/lightly_studio/src/lightly_studio/few_shot_classifier/classifier_manager.py
@@ -615,21 +615,12 @@ class ClassifierManager:
 def _get_embedding_model_by_hash(
     session: Session, embedding_model_hash: str, collection_id: UUID
 ) -> EmbeddingModelTable | None:
-    """Resolve the embedding model with the given hash for a classifier's collection.
+    """Resolve the embedding model with the given hash within the collection's dataset.
 
-    Sample embeddings are keyed to the collection's own model row, so match that row first
-    and fall back to the dataset only when the collection has none.
+    A classifier stores the hash of the model it was trained on. The model is looked up by
+    ``(dataset_id, embedding_model_hash)`` so it resolves regardless of whether it is the
+    collection's default embedding space.
     """
-    # TODO(Michal, 08/2026): Drop the collection-local lookup once the write path
-    # deduplicates per dataset. Until then a dataset can hold one row per collection for the
-    # same hash, and the dataset fallback returns the oldest, which may match no embeddings.
-    collection_model = embedding_model_resolver.get_by_model_hash_deprecated(
-        session=session,
-        collection_id=collection_id,
-        embedding_model_hash=embedding_model_hash,
-    )
-    if collection_model is not None:
-        return collection_model
     collection = collection_resolver.get_by_id(session=session, collection_id=collection_id)
     if collection is None:
         raise ValueError(f"Collection {collection_id} not found.")
--- a/lightly_studio/src/lightly_studio/few_shot_classifier/classifier_manager.py
+++ b/lightly_studio/src/lightly_studio/few_shot_classifier/classifier_manager.py
@@ -28,11 +28,13 @@ from lightly_studio.models.annotation_label import (
     AnnotationLabelCreate,
 )
 from lightly_studio.models.classifier import EmbeddingClassifier
+from lightly_studio.models.embedding_model import EmbeddingModelTable
 from lightly_studio.models.image import ImageTable
 from lightly_studio.resolvers import (
     annotation_label_resolver,
     annotation_resolver,
     collection_resolver,
+    default_embedding_space_resolver,
     embedding_model_resolver,
     image_resolver,
     sample_embedding_resolver,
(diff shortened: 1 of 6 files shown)


michal-lightly:
Reviewer note (from a Codex pass): `get_by_model_hash` now scopes by `dataset_id` and returns `.first()` (oldest). With no unique constraint yet, a dataset holding duplicate rows for the same hash could resolve to a model whose `embedding_model_id` differs from the one a collection's sample embeddings point at → empty embeddings at train/inference.

This is the intended design of the parent work (#2067) — the dedup + unique-constraint migration lands with the `collection_id` drop, as the resolver docstring notes. Flagging for visibility; not fixing here to keep this PR scoped.

lightly_studio/src/lightly_studio/resolvers/embedding_model_resolver.py
```diff
@@ -25,7 +25,7 @@ def get_or_create(session: Session, embedding_model: EmbeddingModelCreate) -> Em
     """Retrieve an existing EmbeddingModel by hash or create a new one if it does not exist."""
     db_model = get_by_model_hash(
         session=session,
-        collection_id=embedding_model.collection_id,
+        dataset_id=embedding_model.dataset_id,
```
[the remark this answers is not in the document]
michal-lightly: addressed

lightly_studio/src/lightly_studio/few_shot_classifier/classifier_manager.py
```diff
@@ -611,6 +612,25 @@ def load_classifier_from_buffer(
...
 
+def _get_embedding_model_by_hash(
+    session: Session, embedding_model_hash: str, collection_id: UUID
+) -> EmbeddingModelTable | None:
+    """Resolve the embedding model with the given hash within the collection's dataset.
+
+    A classifier stores the hash of the model it was trained on. The model is looked up by
+    ``(dataset_id, embedding_model_hash)`` so it resolves regardless of whether it is the
+    collection's default embedding space.
+    """
```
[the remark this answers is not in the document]
michal-lightly: Transition code, won't address.

michal-lightly:
@codex review

lightly_studio/src/lightly_studio/few_shot_classifier/classifier_manager.py
```diff
@@ -611,6 +612,37 @@ def load_classifier_from_buffer(
...
+        )
+        assert default_embedding_model is not None
+        if default_embedding_model.embedding_model_hash == embedding_model_hash:
+            return default_embedding_model
+
+    return embedding_model_resolver.get_by_model_hash(
+        session=session,
+        dataset_id=collection.dataset_id,
+        embedding_model_hash=embedding_model_hash,
+    )
```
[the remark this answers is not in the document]
michal-lightly: Exported/loaded classifiers using non-default embedding models are outside this PR’s supported scope. Classifiers created through `create_classifier` store the collection default model’s hash, and this helper preferentially resolves that same default row while per-collection duplicates exist. The scenario described here therefore depends on an out-of-scope loaded classifier (or changing the default after classifier creation), so I’m not planning to add another collection-scoped lookup in this PR.

lightly_studio/src/lightly_studio/few_shot_classifier/classifier_manager.py
```diff
@@ -122,17 +124,16 @@ def create_classifier(
...
             raise ValueError("No embedding model found for the given collection ID.")
-        # TODO(Horatiu, 05/2025): Handle multiple models correctly when
-        # available
-        if len(embedding_models) > 1:
-            raise ValueError("Multiple embedding models found for the given collection ID.")
-        embedding_model = embedding_models[0]
+        embedding_model = embedding_model_resolver.get_by_id(
+            session=session, embedding_model_id=embedding_model_id
+        )
+        assert embedding_model is not None, "Default embedding space references a missing model."
```
iunir: Please don't use assets outside of tests, just replace with `if` + `ValueError` or some other exception

lightly_studio/src/lightly_studio/few_shot_classifier/classifier_manager.py
```diff
@@ -611,6 +612,37 @@ def load_classifier_from_buffer(
...
+
+    # TODO(Michal, 08/2026): Remove once embedding models are unique per dataset and hash.
+    default_embedding_model_id = default_embedding_space_resolver.get_by_collection_id(
+        session=session, collection_id=collection_id
+    )
+    if default_embedding_model_id is not None:
+        default_embedding_model = embedding_model_resolver.get_by_id(
+            session=session, embedding_model_id=default_embedding_model_id
+        )
+        assert default_embedding_model is not None
```
iunir: same here

lightly_studio/src/lightly_studio/few_shot_classifier/classifier_manager.py
```diff
@@ -611,6 +612,37 @@ def load_classifier_from_buffer(
...
+        if default_embedding_model.embedding_model_hash == embedding_model_hash:
```
iunir, line 638: Is it even possible then 2 different models have same hash? (except unexpected rare case collision). I think we should just rely on hash without those tricky hash check

iunir (changes requested):

lightly_studio/src/lightly_studio/few_shot_classifier/classifier_manager.py
michal-lightly: Done

lightly_studio/src/lightly_studio/few_shot_classifier/classifier_manager.py
michal-lightly: Done

lightly_studio/src/lightly_studio/few_shot_classifier/classifier_manager.py
michal-lightly: Only during the migration. As discussed, we will remove the hash anyway.

iunir (approved):

PR #2077, opened 2026-08-25
Show the order value on the video tile

gabrielfruet:
Image tiles show the active sort value as a badge; video tiles don't. This adds the same badge to `VideoItem.svelte`, mirroring `SampleImageGridItem`.

When a numeric sort is active the badge shows its value. Otherwise the tile falls back to the existing similarity-score badge, and only one shows at a time. The right-overlay flag now accounts for the badge, not just similarity.

Render-only: `VideoView.order_value` already ships on the wire, so there's no plumbing to add.

Part of the video-grid sorting stack. It lands after the PR that adds the numeric video sort control (#2076), so the badge has a value to display.

Testing: frontend `make static-checks` and unit tests pass. No dedicated `VideoItem` test, same as the image tile badge.

## Summary by CodeRabbit

* **New Features**
  * Video and image items now display consistent badges for order values and similarity scores.
  * Order values use clear formatting: integers omit decimals, while decimal values show two places.
  * When both values are available, the order value takes priority.
  * Classification labels reserve space for badges, which remain consistently positioned above captions.
* **Tests**
  * Added coverage to verify badge formatting, fallback behavior, value precedence, and hidden badges when no values are available.


lightly_studio_view/src/lib/components/VideoItem/VideoItem.svelte
```diff
@@ -148,6 +148,19 @@
...
         (video.sample.annotations ?? []).filter(isWholeVideoClassificationAnnotation)
     );
+
+    const orderValue = $derived(video.order_value ?? undefined);
+    const shouldShowOrderValue = $derived(orderValue !== undefined);
+    const orderValueLabel = $derived(
+        orderValue !== undefined
+            ? Number.isInteger(orderValue)
+                ? String(orderValue)
+                : orderValue.toFixed(2)
```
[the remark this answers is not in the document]
gabrielfruet: I think we can raise an issue for this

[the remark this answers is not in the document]
gabrielfruet: Good catch. Extracted into a shared component in a20bb70d.

lightly_studio_view/src/lib/components/SampleValueBadge/SampleValueBadge.helpers.ts
```diff
@@ -0,0 +1,10 @@
+export function formatOrderValue(value: number): string {
+    return Number.isInteger(value) ? String(value) : value.toFixed(2);
+}
+
+export function hasValueBadge(
```
LeonardoRosaa, line 5: These parameters are optional. I think we can change the type to `number` rather than `number | null`.

lightly_studio_view/src/lib/components/SampleValueBadge/SampleValueBadge.svelte
```diff
@@ -0,0 +1,35 @@
+<script lang="ts">
+    import { cn, getSimilarityColor } from '$lib/utils';
+    import { formatOrderValue } from './SampleValueBadge.helpers';
+
+    interface Props {
```
LeonardoRosaa, line 5: Add fields description

lightly_studio_view/src/lib/components/VideoItem/VideoItem.svelte
```diff
@@ -182,20 +184,14 @@
...
-            class="absolute bottom-1 right-1 z-10 box-border flex h-5 items-center rounded bg-black/60 px-1.5 text-xs font-medium text-white backdrop-blur-sm"
-        >
-            <span
-                class="mr-1.5 block h-2 w-2 rounded-full"
-                style="background-color: {getSimilarityColor(video.similarity_score)}"
-            ></span>
-            {video.similarity_score.toFixed(2)}
-        </div>
-    {/if}
+    <SampleValueBadge
```
LeonardoRosaa, line 190: I would add a unit test.

lightly_studio_view/src/lib/components/SampleValueBadge/SampleValueBadge.helpers.ts
gabrielfruet: Image

The problem is that generated types are `number | null | undefined`.

So we would need to do use null coalescing on every call 

```python
hasValueBadge(sample.order_value ?? undefined, sample.similarity_score ?? undefined) 
```

I personally prefer the current way

LeonardoRosaa (approved):

PR #2095, opened 2026-08-26
Adjust collection embedding model table

michal-lightly:
## What has changed and why?

- Rename the persisted table to `collection_embedding_model` and track the default with `is_default`.
- Allow multiple embedding models per collection with a composite primary key.
- Add the PostgreSQL migration, backfill, and partial unique index for one default per collection.

Not done, scoped to a follow-up (next in stack):
- The resolver functions need to change to allow management of multiple models, not just the default. The current model registration is not completely consistent.

## How has it been tested?

- `make static-checks`
- Targeted resolver and dataset lifecycle tests: 6 passed, 30 skipped.

- [ ] Yes

## Summary by CodeRabbit

* **New Features**
  * Collections can now be linked to multiple embedding models.
  * One embedding model can be designated as the collection’s default.

* **Bug Fixes**
  * Changing the default embedding model now correctly preserves existing model links.
  * Database migrations retain all linked models and select the oldest as the initial default.

* **Tests**
  * Added coverage for migration backfills, default-model switching, and dataset copy/delete workflows.

### DIFF
--- a/lightly_studio/src/lightly_studio/migrations/versions/1787846400_d4e5f6a7b8c9_rename_default_embedding_space_to_collection_embedding_model.py
+++ b/lightly_studio/src/lightly_studio/migrations/versions/1787846400_d4e5f6a7b8c9_rename_default_embedding_space_to_collection_embedding_model.py
@@ -48,7 +48,9 @@ def upgrade() -> None:
         sa.Column("is_default", sa.Boolean(), nullable=False),
     )
     _backfill_defaults()
-    # At most one default embedding model per collection.
+    # At most one default embedding model per collection. Postgres-only: DuckDB cannot
+    # create partial indexes, so on DuckDB the invariant is not enforced at the database
+    # level and callers must maintain it.
     op.create_index(
         "uq_collection_embedding_model_default",
         "collection_embedding_model",
--- a/lightly_studio/src/lightly_studio/migrations/versions/1787846400_d4e5f6a7b8c9_rename_default_embedding_space_to_collection_embedding_model.py
+++ b/lightly_studio/src/lightly_studio/migrations/versions/1787846400_d4e5f6a7b8c9_rename_default_embedding_space_to_collection_embedding_model.py
@@ -1,14 +1,17 @@
 """rename default embedding space to collection embedding model.
 
 Renames the `default_embedding_space` table to `collection_embedding_model` and adds an
-`is_default` flag. The current contents are dropped and re-seeded from `embedding_model`
-(the oldest model per collection by `created_at ASC, embedding_model_id ASC`, the same
-rule the earlier backfill used), so every seeded row is a default and `is_default` is
-`true`.
+`is_default` flag. The current contents are dropped and re-seeded from `embedding_model`:
+one row per model, so a collection with several models gets a row for each. Only the
+oldest model per collection (by `created_at ASC, embedding_model_id ASC`) is the default,
+with `is_default = true`; the rest are `false`.
 
 Re-seeding from `embedding_model` keeps this migration independent of the table's prior
-contents. It reads `embedding_model.collection_id`, which still exists at this revision
-(its drop lands in a later revision).
+contents. The prior `default_embedding_space` rows held only the defaults, which this
+backfill reproduces from `embedding_model`, so dropping them loses nothing. It reads
+`embedding_model.collection_id`, which still exists at this revision (its drop lands in a
+later revision). This table is where `collection_id` moves to as it is removed from
+`embedding_model`.
 
 A partial unique index then enforces at most one default per collection. It is
 Postgres-only (DuckDB cannot create partial indexes), so it is not declared on
@@ -58,7 +61,7 @@ def upgrade() -> None:
         "collection_embedding_model",
         ["collection_id", "embedding_model_id"],
     )
-    _backfill_defaults()
+    _backfill_collection_embedding_models()
     # At most one default embedding model per collection. Postgres-only: DuckDB cannot
     # create partial indexes, so on DuckDB the invariant is not enforced at the database
     # level and callers must maintain it.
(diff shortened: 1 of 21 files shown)


lightly_studio/src/lightly_studio/resolvers/collection_embedding_model_resolver.py
```diff
@@ -1,42 +1,61 @@
-"""Handler for database operations related to a collection's default embedding space."""
+"""Handler for database operations related to a collection's default embedding space.
+
+TODO(Michal, 08/2026): Rename the functions and change the interface so that it reflects
+the new "all collection models" semantics, not the old "only default model" semantics.
```
michal-lightly, line 4: In scope for the follow-up - current interface does not allow adding a non-default model. The implementation below keeps the semantics of the previous resolver (only handling defaults).

michal-lightly:
@coderabbit review

lightly_studio/tests/database/test_db_migrations.py
```diff
@@ -75,6 +75,26 @@ def _reset_postgres_database(engine_url: str) -> None:
         raw_engine.dispose()
 
 
+def _restore_shared_database_to_head(engine: Engine, engine_url: str) -> None:
```
michal-lightly, line 78: This is a fix for failing Postgres tests. We use a shared db for all tests; the new test renamed a table, which made db cleanup on teardown fail. The fix is to restore the DB back to the head migration.

lightly_studio/src/lightly_studio/migrations/versions/1787846400_d4e5f6a7b8c9_rename_default_embedding_space_to_collection_embedding_model.py
```diff
@@ -0,0 +1,115 @@
...
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
+    """Upgrade schema."""
+    op.rename_table("default_embedding_space", "collection_embedding_model")
+    # Empty first so the new NOT NULL column needs no server default; the rows are
+    # re-seeded below.
+    op.get_bind().execute(sa.text("DELETE FROM collection_embedding_model"))
```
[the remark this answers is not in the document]
michal-lightly: Won't do. The `default_embedding_space` table was completely created from the `embedding_model` table and was not released. We decided to change the model; it is safe to repopulate the new table freshly from `embedding_model`.

lightly_studio/src/lightly_studio/migrations/versions/1787846400_d4e5f6a7b8c9_rename_default_embedding_space_to_collection_embedding_model.py
```diff
@@ -0,0 +1,115 @@
...
+only matters for tracked Postgres databases.
+
+Revision ID: d4e5f6a7b8c9
+Revises: c2d3e4f5a6b7
+Create Date: 2026-08-26 10:00:00.000000
+
+"""
+
+from collections.abc import Sequence
+from typing import Union
```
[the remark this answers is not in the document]
michal-lightly: SQLModel needs old syntax

lightly_studio/src/lightly_studio/resolvers/collection_embedding_model_resolver.py
```diff
@@ -1,42 +1,61 @@
...
+"""
 
 from __future__ import annotations
 
 from uuid import UUID
 
-from sqlmodel import Session, select
+from sqlmodel import Session, col, select
```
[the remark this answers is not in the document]
michal-lightly: That would break our largely applied convention.

lightly_studio/tests/resolvers/test_collection_embedding_model_resolver.py
```diff
@@ -1,7 +1,8 @@
 from __future__ import annotations
 
-from sqlmodel import Session
+from sqlmodel import Session, col, select
```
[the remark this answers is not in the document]
michal-lightly: That would break our largely applied convention.

lightly_studio/src/lightly_studio/resolvers/collection_embedding_model_resolver.py
```diff
@@ -63,9 +82,10 @@ def delete_by_collection_id(session: Session, collection_id: UUID) -> bool:
 def _get_by_collection_id(
```
iunir, line 82:
```suggestion
def get_default_by_collection_id(
```

lightly_studio/src/lightly_studio/resolvers/collection_embedding_model_resolver.py
iunir, line 64:
```suggestion
```

Why do we need this wrapper? Let's make another one public

iunir, line 67:
```suggestion
def delete_default_by_collection_id(session: Session, collection_id: UUID) -> bool:
```

Or even drop this function, it seems not used at all, only in tests

lightly_studio/tests/database/test_db_migrations.py
iunir, line 1: Not really sure if we need to write tests for migrations, it's nice to verify probably, but I don't think we need to run those tests all the time since migration files should never be changed.

lightly_studio/src/lightly_studio/resolvers/collection_embedding_model_resolver.py
michal-lightly: It deduplicates code for `set_default`. I would keep it. Note that resolver refactor is in the follow-up PR (already drafted).

michal-lightly: It's removed in the follow-up

lightly_studio/src/lightly_studio/resolvers/collection_embedding_model_resolver.py
michal-lightly: Addressed in the follow-up

lightly_studio/tests/database/test_db_migrations.py
michal-lightly: AI wrote it and I kept it. I think it does not hurt; at least it made the AI more certain the migration works. If they take time we can delete them.

michal-lightly: I think though that for more complex migrations we should write tests. We don't need to run them on every PR though.

iunir (approved):

PR #2096, opened 2026-08-27
LIG-10603: Compare metadata distributions by sample tags

ikondrat:
## What has changed and why?

Closes [LIG-10603](https://linear.app/lightly/issue/LIG-10603/compare-metadata-distributions-by-sample-tags).

The Distribution panel could already compare **annotation classes** across sample tags. This PR extends the same "Compare by" flow to the **Metadata** source, so you can see how a metadata field is distributed inside one or more tags versus the current view.

- **New hook** `useMetadataDistributionsBySampleTags` issues the existing metadata histogram and value-count requests once per selected tag, overriding only the request's tag scope and keeping the rest of the exploration filter intact. Comparison tags therefore never leak into the dataset filter. A failing sibling query does not discard the tags that did return.
- **Categorical metadata** reuses the grouped-bar pipeline. Buckets are matched by bucket id rather than label so the shared value axis stays correct, and `buildCategoricalComparisonSeries` disambiguates the synthetic `Missing`/`Other` buckets from literal `"Missing"`/`"Other"` values across all tag series.
- **Numerical metadata** renders grouped histograms. `Histogram` gains an optional `series` prop: named histograms that share the base bin edges and render side by side within each bin, so bars stay comparable across tags. Colors come from the same `assignSeriesColors` map the grouped bar chart uses, so a tag keeps its color across both chart types.
- Clearing the tag selection falls back to the existing single distribution; without the new prop the histogram rendering is unchanged.

Commits are split so each is reviewable on its own: query layer → histogram rendering → panel wiring.

## How has it been tested?

- New unit tests: `useMetadataDistributionsBySampleTags` (request scoping, both response shapes, partial results on a failed query), `buildCategoricalComparisonSeries` (Missing/Other disambiguation), `buildHistogramOption` (grouped series share the bin axis, stable distinct colors), and two `DatasetDistributionPanel` cases covering the categorical and numerical comparison rendering.
- `cd lightly_studio_view && make static-checks` — prettier, eslint and `svelte-check` all clean.
- `cd lightly_studio_view && npx vitest run` — full frontend unit suite passing (~353 test files).

_Screenshots to be added._

https://github.com/user-attachments/assets/3f2c9f7e-4890-49be-aa98-2261cd5e5da1

- [x] Yes

---

Suggested reviewer: Leonardo Rosa (recent history on `DatasetDistributionPanel` / `Histogram`); alternative: Igor Susmelj.

## New Features
* Compare metadata distributions across selected sample tags.
* View categorical metadata as grouped bar charts and numerical metadata as grouped histograms with shared bins.
* Expanded charts include comparison series, legends, and combined tooltips.
* Filter using values found only in comparison tags when supported.
* Select histogram ranges by dragging across bins.

## Bug Fixes
* Clarified labels for missing, “Missing,” and “Other” categorical values.
* Added clear loading and error states during comparisons.


mrpositron (approved):
LGTM 🚀

PR #2101, opened 2026-08-27
perf: materialize evaluation sample IDs lazily

lorinczszabolcs:
## What has changed and why?

`ImageDataset.evaluate()` built the selected sample IDs eagerly, scanning every image row up front. The readback methods (`list_runs()`, `confusion_matrix()`) never use those IDs, so `dataset.evaluate().list_runs()` scanned the whole dataset just to return a few run rows. Only the write methods (`object_detection`, `classification`, `semantic_segmentation`) need them.

This defers the work:

- `ImageDatasetEvaluate.sample_ids` is now a cached property, materialized once, only when a write method reads it. The constructor signature is unchanged.
- `evaluate()` passes a generator instead of a materialized list, so the dataset scan runs only on demand.
- The write path copies the set before its in-place coverage intersection, so the cached set is not mutated.

Part of the evaluation Python API work in #2019. This follows the readback PR #2066 (the first stage) and addresses the eager `sample_ids` scan noted in that review.

Reviewer note: `sample_ids` is fed a single-use generator by `evaluate()` (as before). It is materialized once and cached, so this is safe in normal use; it would only be re-read if a write's materialization failed mid-scan and was retried on the same evaluator, so callers should build a fresh evaluator after such an error rather than reuse one.

## How has it been tested?

- New tests: readback leaves the sample IDs unmaterialized while a write materializes them; and a write restricts to covered samples without shrinking the cached set.
- `make static-checks` (ruff, ruff format, mypy incl. tests) passes.
- The evaluation, service, and route test suites pass.
- The docs build passes with `--strict`.

- [ ] Yes
- [x] Not needed (internal change) — internal optimization; the readback methods it speeds up were only added in #2066, so no released behavior changes.

## Summary by CodeRabbit

* **Performance Improvements**
  * Evaluation now defers loading sample IDs until they are needed, improving efficiency for read-only operations.
  * Evaluation results remain reusable across multiple operations without prematurely consuming sample data.

* **Bug Fixes**
  * Fixed evaluation filtering so uncovered samples are excluded without altering the underlying sample set.

* **Tests**
  * Added coverage for lazy sample loading, read-only evaluation behavior, and preserving sample IDs after evaluation.


lorinczszabolcs:
Hi @iunir! The readback PR (#2066) merged, and this perf follow-up is up. I've got the next PRs ready locally: aggregate metrics, mAP, and segmentation metrics, but held off opening them so they stack on the reviewed/merged version and I avoid reworking them. Should I open the rest once this PR is reviewed and merged, or would you prefer them all up now for review together?

lightly_studio/src/lightly_studio/evaluation/image_dataset_evaluate.py
```diff
@@ -287,7 +294,8 @@ def _prepare_evaluation_data(
             config_json=config_json,
         )
 
-        selected_sample_ids = self.sample_ids
+        # Copy: the intersection below is in place, and must not mutate the cached set.
+        selected_sample_ids = set(self.sample_ids)
```
iunir:
```suggestion
        selected_sample_ids = set(self.sample_ids)
        gt_collection_id, pred_collection_id, evaluation_run = self._create_evaluation_run(
name=name,
gt_annotation_source=gt_annotation_source,
pred_annotation_source=pred_annotation_source,
task_type=task_type,
config_json=config_json,
)
```

Copy before the run row is committed: the intersection is in place, and a failed scan must not leave an orphan run.

iunir (changes requested):

iunir:
Thanks for contribution @lorinczszabolcs please address minor comment before we can merge

lorinczszabolcs:
Thanks for the review! Addressed.

iunir (approved):

iunir:
Thank you!

PR #2102, opened 2026-08-27
Align EmbeddingManager default model state

michal-lightly:
## What has changed and why?

- Keep the in-memory default model aligned with the database default when registering models.
- Use the persisted collection default to decide when the first registered model becomes the default.

## How has it been tested?

- `uv run pytest tests/dataset/test_embedding_manager.py` (41 passed)

- [ ] Yes

Alternative reviewer: @horatiualmasan

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved embedding model registration and default selection.
  * Preserved existing collection defaults when adding or registering additional embedding models.
  * Ensured all registered embedding models are correctly linked to their collections.
  * Restored default model settings when starting with an existing database configuration.
  * Persisted newly selected defaults so they remain available across application restarts.
  * Improved consistency between saved embedding model settings and the application’s active configuration.

### DIFF
--- a/lightly_studio/src/lightly_studio/dataset/embedding_manager.py
+++ b/lightly_studio/src/lightly_studio/dataset/embedding_manager.py
@@ -190,7 +190,7 @@ class EmbeddingManager:
         # Determine if the model should be set as default.
         # TODO(Michal, 08/2026): Currently there are two places storing the defaults:
         # the `_collection_id_to_default_model_id` dict and the database. The source
-        # of truth should completely move to the database,
+        # of truth should completely move to the database.
         has_no_default_in_db = (
             collection_embedding_model_resolver.get_default_by_collection_id(
                 session=session, collection_id=collection_id
--- a/lightly_studio/src/lightly_studio/dataset/embedding_manager.py
+++ b/lightly_studio/src/lightly_studio/dataset/embedding_manager.py
@@ -165,43 +165,43 @@ class EmbeddingManager:
         if collection is None:
             raise ValueError("Provided collection_id could not be found.")
 
-        embedding_space = embedding_generator.get_embedding_model_input()
-        embedding_model = EmbeddingModelCreate(
-            name=embedding_space.name,
-            parameter_count_in_mb=embedding_space.parameter_count_in_mb,
-            embedding_model_hash=embedding_space.embedding_model_hash,
-            embedding_dimension=embedding_space.embedding_dimension,
+        model_description = embedding_generator.get_embedding_model_input()
+        model_create = EmbeddingModelCreate(
+            name=model_description.name,
+            parameter_count_in_mb=model_description.parameter_count_in_mb,
+            embedding_model_hash=model_description.embedding_model_hash,
+            embedding_dimension=model_description.embedding_dimension,
             collection_id=collection_id,
             dataset_id=collection.dataset_id,
         )
         db_model = embedding_model_resolver.get_or_create(
             session=session,
-            embedding_model=embedding_model,
+            embedding_model=model_create,
         )
         model_id = db_model.embedding_model_id
 
         self._models[model_id] = embedding_generator
 
-        # Record the model in two places: the in-memory map caches the loaded generator for
-        # this process, and the collection_embedding_model table persists the link for
-        # query-layer callers (collection_embedding_model_resolver.get_default_by_collection_id).
-        if set_as_default or collection_id not in self._collection_id_to_default_model_id:
-            self._collection_id_to_default_model_id[collection_id] = model_id
+        # Register the model with the collection
         collection_embedding_model_resolver.get_or_add_collection_model(
             session=session, collection_id=collection_id, embedding_model_id=model_id
         )
-        # The first model linked to a collection becomes its default; an explicit request
-        # re-points it. "First model auto-defaults" is caller policy, kept out of the resolver.
-        if (
-            set_as_default
-            or collection_embedding_model_resolver.get_default_by_collection_id(
+
+        # Determine if the model should be set as default.
+        # TODO(Michal, 08/2026): Currently there are two places storing the defaults:
+        # the `_collection_id_to_default_model_id` dict and the database. The source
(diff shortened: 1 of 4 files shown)


michal-lightly:
@coderabbit review

lightly_studio/src/lightly_studio/dataset/embedding_manager.py
```diff
@@ -165,43 +165,46 @@ def register_embedding_model(
...
+        # Register the model with the collection.
+        collection_model_link = collection_embedding_model_resolver.get_or_create(
             session=session, collection_id=collection_id, embedding_model_id=model_id
         )
-        # The first model linked to a collection becomes its default; an explicit request
-        # re-points it. "First model auto-defaults" is caller policy, kept out of the resolver.
-        if (
-            set_as_default
-            or collection_embedding_model_resolver.get_default_by_collection_id(
+        if collection_model_link.is_default:
```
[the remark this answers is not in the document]
michal-lightly: It would be nice but at this point I think the change is not worth it - we'd like to rewrite the class.

lightly_studio/src/lightly_studio/dataset/embedding_manager.py
```diff
@@ -165,43 +165,46 @@ def register_embedding_model(
...
             raise ValueError("Provided collection_id could not be found.")
 
-        embedding_space = embedding_generator.get_embedding_model_input()
-        embedding_model = EmbeddingModelCreate(
-            name=embedding_space.name,
-            parameter_count_in_mb=embedding_space.parameter_count_in_mb,
-            embedding_model_hash=embedding_space.embedding_model_hash,
-            embedding_dimension=embedding_space.embedding_dimension,
+        model_description = embedding_generator.get_embedding_model_input()
+        model_create = EmbeddingModelCreate(
```
ikondrat, line 169: I am a bit confused by naming, do we pass constructor now?

michal-lightly: Not anymore. Just "EmbeddingSpaceDescription" from which the "constructor" is created.

The reason is that `dataset_id` has to be added on top of what the user specifies as the "model description".

ikondrat: I am confused by this `embedding_model=model_create`,

not sure why it is renamed

michal-lightly: Just a taste. Because it was shadowed (that's the only reason).

lightly_studio/src/lightly_studio/dataset/embedding_manager.py
```diff
@@ -165,43 +165,46 @@ def register_embedding_model(
...
+        collection_model_link = collection_embedding_model_resolver.get_or_create(
             session=session, collection_id=collection_id, embedding_model_id=model_id
         )
-        # The first model linked to a collection becomes its default; an explicit request
-        # re-points it. "First model auto-defaults" is caller policy, kept out of the resolver.
-        if (
-            set_as_default
-            or collection_embedding_model_resolver.get_default_by_collection_id(
+        if collection_model_link.is_default:
+            self._collection_id_to_default_model_id[collection_id] = model_id
```
iunir:
```suggestion
            self._collection_id_to_default_model_id[collection_id] = model_id
            return db_model
```

In that case there is no reason for the next checks, we can return right away

lightly_studio/src/lightly_studio/dataset/embedding_manager.py
```diff
@@ -165,43 +165,46 @@ def register_embedding_model(
...
+
+        # Determine if the model should be set as default.
+        # TODO(Michal, 08/2026): The default is stored both in the database and in
+        # `_collection_id_to_default_model_id`. The database is the source of truth,
+        # while the dictionary only tracks models loaded at runtime. Split these
+        # responsibilities during the EmbeddingManager refactor.
+        has_no_default_in_db = (
+            collection_embedding_model_resolver.get_default_by_collection_id(
```
iunir: For that reason I would write another SQLAlchemy function to check existence (it's more DB optimal)
```
def has_default_by_collection_id(session: Session, collection_id: UUID) -> bool:
    """Return whether the collection has a default embedding model."""
    stmt = select(
        exists().where(
            CollectionEmbeddingModelTable.collection_id == collection_id,
            col(CollectionEmbeddingModelTable.is_default).is_(True),
        )
    )
    return session.exec(stmt).one()
```

iunir (changes requested):

lightly_studio/src/lightly_studio/dataset/embedding_manager.py
michal-lightly: Done, returns early now.

lightly_studio/src/lightly_studio/dataset/embedding_manager.py
michal-lightly: Added has_default_by_collection_id and switched to it.

iunir (approved):

lightly_studio/src/lightly_studio/dataset/embedding_manager.py
```diff
@@ -165,43 +165,44 @@ def register_embedding_model(
...
+            return db_model
+
+        # Determine if the model should be set as default.
+        has_no_default_in_db = not collection_embedding_model_resolver.has_default_by_collection_id(
+            session=session, collection_id=collection_id
+        )
+        if set_as_default or has_no_default_in_db:
             collection_embedding_model_resolver.set_default(
                 session=session, collection_id=collection_id, embedding_model_id=model_id
             )
```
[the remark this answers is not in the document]
michal-lightly: Won't address in this PR. Concurrency guarantees are a concern for the whole codebase and should not be addressed in isolation.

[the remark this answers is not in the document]
iunir: good point

PR #2104, opened 2026-08-27
Label usage events with the cohort of the installation

iunir:
## What has changed and why?

A handful of devs and CI runs dominate the product signal, every event we report being anonymous and indistinguishable. Each event now carries a `user_cohort`:

- `staff` — `LIGHTLY_STUDIO_INTERNAL=1`, or an `internal` marker file in the model cache directory, which survives recreating the virtualenv.
- `ci` — `CI` or `GITHUB_ACTIONS` holding anything but an explicit false. A build matrix otherwise looks like one very active user.
- `source_build` — a PEP 610 `dir_info` record, covering editable and plain installs from a checkout. External contributors land here too.
- `user` — whenever nothing indicates otherwise.

The cohort also goes on the person via `$set`, which is what PostHog's internal user filter selects on. Our own suite reports nothing, the autouse fixture in `tests/conftest.py` keeping it off the network.

Still to do by hand in PostHog: set that filter to `user_cohort is not user`, and add a static cohort of the existing internal `install_id`s to exclude the data already collected.

## How has it been tested?

Unit tests for every cohort branch and every `direct_url.json` case. `cd lightly_studio && make static-checks && make test`: 2217 passed, 44 skipped, `ruff` and `mypy` clean. Four test files fail to collect on `main` alike, for missing optional `adlfs`/`gcsfs`/`s3fs`, and are excluded.

- [ ] Yes

---

Suggested reviewer: @gabrielfruet, who added the backend tracking in #1914. Alternative: @IgorSusmelj.

## Summary by CodeRabbit

* **New Features**
  * Usage tracking now classifies installations as staff, CI, source builds, or standard users.
  * Added options for marking machines or individual runs as internal.
  * Tracking supports unpublished and locally installed builds.

* **Bug Fixes**
  * Invalid or empty CI and internal-use settings no longer interfere with startup.
  * Version reporting now works reliably for local builds.

* **Documentation**
  * Updated guidance for internal markers, environment settings, usage tracking, and the required Uv version.


gabrielfruet:
I think this needs to get approval of either Michal/Jonas before we spend time on this

lightly_studio/src/lightly_studio/dataset/env.py
```diff
@@ -38,6 +38,15 @@
 # their configuration through different systems, so the value is written once on each side. Not an
 # environment variable: the key above is the only part worth pointing elsewhere.
 LIGHTLY_STUDIO_POSTHOG_HOST: str = "https://eu.i.posthog.com"
+# Marks this machine as a Lightly dev or staff machine, so internal usage can be filtered out of
+# the product metrics. See lightly_studio/analytics/cohort.py for the alternative marker file,
+# which survives recreating the virtualenv.
+try:
```
gabrielfruet: I think we should drop this try except, this failure mode exists for all other bools and I think it's fine for it to exist

gabrielfruet (approved):
I'll not block, but I'd suggest to remove the try except

lightly_studio/src/lightly_studio/dataset/env.py
iunir: thanks, fixed

PR #2123, opened 2026-08-28
Gate releases on green CI for the exact release commit

iunir:
Closes [LIG-10551](https://linear.app/lightly/issue/LIG-10551/public-repo-gate-releases-on-green-ci-for-the-exact-release-commit).

## What has changed and why?

Nothing in the release process checks that CI passed on the commit being
released. This adds a gate that does, given a tag, branch or sha.

**Why the release PR's own CI is not enough.** The version-bump PR runs Unit
Test and End2End Test, but a release is published from a ref, and that ref is
not necessarily the commit those runs tested:

- Squash-merging the release PR produces a **new** commit on `main`; the PR's
  runs belong to the PR head, not to what was merged.
- Publishing is a separate `workflow_dispatch` with no ordering against `main`'s
  runs. It can be dispatched while they are still going, or after they failed.
- The emergency path in `RELEASE.md` releases from a release branch, which may
  have no runs at all.

The gate asks about the sha rather than about the branch.

### Contents

- `.github/actions/check-release-ci/` - composite action, callable cross-repo so
  the private docs workflow can use it without duplicating the logic. Inputs:
  `ref`, `repository`, `skip-ci-check`, `github-token`; output: `sha`. The
  `README.md` there explains usage and records the acceptance commits.
- `.github/workflows/check_release_ci.yml` - job wrapper around the action, plus
  a `workflow_dispatch` for checking a ref by hand.
- `prepare_release check-ci` (`ci_gate.py`) - the rules, as a pure function with
  offline unit tests. `REQUIRED_CHECKS` is the single place the required checks
  are named.
- `unit_test.yml` - a `Static Checks + Tests Release Tooling` job so
  `.github/scripts` is linted, type checked and tested on PRs rather than only
  during a release, and a self-test of the gate (see below).
- `.github/scripts/Makefile` - `static-checks` / `test`, so `unit_test.yml` and
  `prepare_release.yml` do not each spell out the commands.

### Decisions, and the trap they avoid

The dangerous outcome here is a gate that silently passes everything, because
once it exists people stop checking by hand.

- **The required names are the aggregate jobs**, `CI Success Check` and `End2End
  Success Check` - not the workflow names `Unit Test` and `End2End Test`, which
  name no check run at all and would match nothing and pass vacuously.
- **Only a completed `success` counts.** Missing, still running, `cancelled`,
  `skipped` and `neutral` all block. Requiring the aggregates also settles the
  `skipped` question at the right level: both run with `if: always()`, so the
  per-job skips that `unit_test.yml`'s path filters produce are resolved inside
  `CI Success Check` itself. `da604e08` on `main` is a real example - every
  gated unit-test job `skipped`, aggregate green.
- **One completed green attempt on the sha is enough.** Every commit on `main`
  carries each check twice, from the merge-queue run and the push-to-main run,
  and `cancel-in-progress: true` routinely cancels the second - a cancelled
  aggregate job reports `failure`. "Newest attempt wins" would therefore block
  nearly every commit, and a gate that blocks everything gets overridden by
  habit, which loses the same vigilance by another route. Other attempts are
  reported next to the green one, so a genuine flake stays visible.

`skip-ci-check: true` skips the gate for the documented emergency path and
records the override, with the actor's name, in the log and the job summary.

## How has it been tested?

Unit tests cover the rules offline. Beyond that, the gate was run against real
commits, known-red first - a gate that returns success unconditionally also
passes a green commit, so checking a green one proves nothing on its own:

| Case | Commit | Result |
| --- | --- | --- |
| Both required checks failed | `1f9b2156` | blocks, both named and linked |
| One required check failed | `a311e602` | blocks, `CI Success Check` named and linked |
| Every gated unit-test job `skipped`, aggregate green | `da604e08` | passes |
| Green, with a cancelled push attempt | `2ae55628` | passes, other attempt flagged |
| Green | `b2dc2470` | passes |

Still-running and misspelled-requirement cases are unit tested; a commit whose
checks are still running is refused, not passed.

The first and last rows are pinned as a **self-test in CI**: whenever the gate
changes, `Static Checks + Tests Release Tooling` runs the action against the
known-red commit and fails the build unless the gate blocked it, then runs it
against the known-green commit. The shas are recorded in the action's
`README.md`.

Not yet exercised: `workflow_dispatch` on `check_release_ci.yml`, which GitHub
only offers once the workflow is on the default branch. I will dispatch it right
after merge as the final confirmation; the action itself is exercised on this PR
by the self-test.

- [ ] Yes

## Summary by CodeRabbit

* **New Features**
  * Added release validation that confirms required CI checks passed on the exact commit being released.
  * Added reusable and manually triggered workflows for checking release readiness.
  * Added clear reports and error details when checks are missing, incomplete, or unsuccessful.
  * Added an emergency override for bypassing CI validation when necessary.

* **Bug Fixes**
  * Prevents releases from proceeding when required checks have not completed successfully.

* **Tests**
  * Added automated coverage for successful, failed, missing, and in-progress CI results.

### DIFF
--- a/.github/actions/check-release-ci/README.md
+++ b/.github/actions/check-release-ci/README.md
@@ -1,0 +1,69 @@
+# Check release CI
+
+Fails unless every required CI check succeeded on the **exact commit** being
+released. Release workflows call it before tagging, publishing a wheel or
+deploying docs, so a release is never cut from an untested or red commit.
+
+## Why the release PR's own CI is not enough
+
+The version-bump PR runs Unit Test and End2End Test, but a release is published
+from a ref, and that ref is not necessarily the commit those runs tested:
+
+- Squash-merging the release PR produces a **new** commit on `main`. The PR's
+  runs belong to the PR head, not to what was merged.
+- `workflow_dispatch` puts no ordering between "merge" and "publish". A publish
+  can be dispatched while `main`'s runs are still going, or after they failed.
+- `RELEASE.md`'s emergency path releases from a release branch, which may have
+  no runs at all.
+
+The gate closes the gap by asking about the sha rather than about the branch.
+
+## Usage
+
+```yaml
+- uses: lightly-ai/lightly-studio/.github/actions/check-release-ci@main
+  with:
+    ref: v1.2.3            # tag, branch or sha being released
+    github-token: ${{ github.token }}
+```
+
+Cross-repo callers - including the private repo's docs workflow - use the
+action as above. Workflows in this repository can instead call the job wrapper
+`.github/workflows/check_release_ci.yml`, which also exposes the gate as a
+`workflow_dispatch` for manual checks. The runner needs `gh` and `python3`.
+
+Set `skip-ci-check: true` for the documented emergency path. It skips the gate
+entirely and records the override, with the actor's name, in the log and the
+job summary.
+
+## What counts as green
+
+The required checks are named once, in `REQUIRED_CHECKS` in
+[`.github/scripts/prepare_release/ci_gate.py`](../../scripts/prepare_release/ci_gate.py),
+which also carries the reasoning. In short:
+
+- The required names are the two **aggregate jobs**, `CI Success Check` and
+  `End2End Success Check` - not the workflow names `Unit Test` and `End2End
+  Test`, which name no check run and would pass vacuously.
+- Only a completed `success` counts. Missing, still running, `cancelled`,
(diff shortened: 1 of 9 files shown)


iunir:
Split into a reviewable stack: #2124 (test the release tooling in CI), #2125 (the gate rule), #2126 (the composite action). The gate now reads the required check names from main's branch ruleset instead of hardcoding them.

PR #2140, opened 2026-08-31
Simplified embedding generator interface with EmbeddingSpaceSpec

michal-lightly:
## What has changed and why?

Separates the embedding generator's public contract from the database model. Previously `EmbeddingSpaceDescription` was both a SQLModel base for the `embedding_model` table and the generator protocol's return type, forcing every custom generator to depend on a SQLModel type.

- `EmbeddingSpaceDescription` removed; its fields moved into `EmbeddingModelBase`.
- New `EmbeddingSpaceSpec` (`space_key`, `dimension`) as the generator return type, exported via `ls`.
- Protocol method `get_embedding_model_input` → `embedding_space_spec`.
- `space_key` is a caller-declared id (e.g. `mobileclip_s0`), stored in `name` and (truncated to the column limit) `embedding_model_hash`.
- Built-in generators drop the per-load checkpoint file hashing.

Identity is now name-based rather than content-addressed. A data migration rekeys existing `embedding_model` rows (`embedding_model_hash = name`) so registration matches them after upgrade. DuckDB and old few-shot-classifier exports are intentionally not covered. Deliberate for this beta API — no deprecation period.

## How has it been tested?

- `make static-checks` and `make migration-check-postgresql` pass.
- `test_embedding_manager.py` and the updated generator tests pass.

- [x] Yes

Suggested reviewer: @iunir

## Summary by CodeRabbit

* **New Features**
  * Added the beta `EmbeddingSpaceSpec` interface for defining embedding spaces by key and dimension.
  * Exposed `EmbeddingSpaceSpec` through the public Python SDK.
* **Documentation**
  * Updated embedding guides, API references, and examples to use the new interface.
* **Refactor**
  * Replaced the previous embedding model input method across supported embedding generators.
* **Tests**
  * Updated embedding generator tests to validate embedding space specifications.


lightly_studio/src/lightly_studio/dataset/embedding_manager.py
```diff
@@ -691,3 +695,18 @@ def _load_video_embedding_generator() -> VideoEmbeddingGenerator | None:
     except ImportError:
         logger.warning("Embedding functionality is disabled.")
         return None
+
+
+def _space_key_to_hash(space_key: str) -> str:
```
michal-lightly, line 700: The `hash` DB field will be deprecated. This function can get removed then.

lightly_studio/src/lightly_studio/migrations/versions/1788181633_f1a2b3c4d5e6_rekey_embedding_model_hash_to_space_key.py
```diff
@@ -0,0 +1,57 @@
...
+
+
+def upgrade() -> None:
+    """Upgrade schema."""
+    # Realign the hash to the space key, which equals the name for built-in generators.
+    # Only rows that still carry the old file hash are touched.
+    op.execute(
+        sa.text(
+            "UPDATE embedding_model SET embedding_model_hash = name "
+            "WHERE embedding_model_hash <> name"
```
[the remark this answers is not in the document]
michal-lightly: We don't have such long names in the production db. The migration script is run by us, let it rather fail.

lightly_studio/src/lightly_studio/dataset/embedding_manager.py
```diff
@@ -691,3 +695,18 @@ def _load_video_embedding_generator() -> VideoEmbeddingGenerator | None:
...
+    A `space_key` may be any string, but the column is a bounded `VARCHAR`, so keys
+    longer than the limit are truncated.
+
+    Args:
+        space_key: The embedding space key to store.
+
+    Returns:
+        The space key truncated to the column limit.
+    """
+    return space_key[:EMBEDDING_MODEL_HASH_MAX_LENGTH]
```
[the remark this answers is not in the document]
michal-lightly: We don't have such long names in the production db. Won't address. The hash field is planned for removal.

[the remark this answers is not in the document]
michal-lightly: We don't have such long names in the production db. Won't address. The hash field is planned for removal.

lightly_studio/src/lightly_studio/dataset/embedding_manager.py
```diff
@@ -165,11 +169,11 @@ def register_embedding_model(
         if collection is None:
             raise ValueError("Provided collection_id could not be found.")
 
-        model_description = embedding_generator.get_embedding_model_input()
+        space_spec = embedding_generator.embedding_space_spec()
         model_create = EmbeddingModelCreate(
-            name=model_description.name,
-            embedding_model_hash=model_description.embedding_model_hash,
-            embedding_dimension=model_description.embedding_dimension,
+            name=space_spec.space_key,
```
iunir, line 174: This I don't really like. 
I would like to be consistent with namings within repo. If it's industry standard we can rename DB column and fully migrate from name to space_key.

From my perspective `space_spec.name` is good enough and not ambiguous (uniqueness should be described in docs and code comments).

lightly_studio/src/lightly_studio/migrations/versions/1788181633_f1a2b3c4d5e6_rekey_embedding_model_hash_to_space_key.py
iunir, line 1: Why do we need this migration from name to hash if hash column removal PR already approved and going to be merged soon?

michal-lightly: Just because in the stack it is on the top. This PR was first.

iunir: But can we just remove this migration file?

michal-lightly: We can. It is a little work (need to update other migrations in the stack).

The only risk would be somehow not merging (or reverting) the PR on top - which is unlikely.

iunir (approved):
Please do remove migration. We don't have to fulfill column which is about to be removed

JonasWurst (approved):

lightly_studio/src/lightly_studio/dataset/embedding_manager.py
michal-lightly: Discussed offline. We'll accept slight inconsistency in exchange for a clearer user interface.

PR #2149, opened 2026-08-31
feat: add attach_embeddings to store precomputed vectors for existing samples

iunir:
## What has changed and why?

Adds `attach_embeddings(session, embedding_model_id, sample_ids, vectors)` to
`lightly_studio.dataset.embedding_manager`. It is a thin public wrapper around the
existing private `_store_embeddings`, for callers that already have precomputed
vectors and just need to attach them to existing samples. Previously the only way
to do this was to implement a 4-method `ImageEmbeddingGenerator` protocol with 3
no-op methods, just to supply vectors that were never computed by a generator.
Not exported from the top-level package; import it directly from
`lightly_studio.dataset.embedding_manager`.

## How has it been tested?

Added `test_attach_embeddings` to `tests/dataset/test_embedding_manager.py`: attaches
different vectors for the same batch of samples under two distinct embedding models,
then reads them back via `sample_embedding_resolver.get_by_sample_ids` and confirms
the correct vector values and correct `(sample_id, embedding_model_id)` pairing.

Ran `make static-checks` and `make test` in `lightly_studio/` (all passing except a
few pre-existing, unrelated failures caused by missing optional deps in this
environment: `tests/api/routes/api/test_enterprise.py`, `tests/dataset/test_fsspec_lister.py`,
`tests/test_enterprise.py`, `tests/test_cloud_credentials.py`, and `tests/api/test_server.py::test_server_static_webapp`,
none of which touch this change).

- [ ] Yes

---

Suggested reviewer: michal-lightly (most frequent recent author of `embedding_manager.py`). Alternative: JonasWurst.

## Summary by CodeRabbit

* **New Features**
  * Added support for attaching precomputed embedding vectors to existing samples.
  * Embeddings are validated for compatible dimensions, sample counts, numeric values, and finite values before storage.
  * Supports storing embeddings for multiple models while preserving the correct sample associations.

* **Tests**
  * Added coverage verifying embeddings are stored with the correct samples and models.

### DIFF
--- a/lightly_studio/src/lightly_studio/dataset/embedding_manager.py
+++ b/lightly_studio/src/lightly_studio/dataset/embedding_manager.py
@@ -541,6 +541,38 @@ class EmbeddingManager:
         return model
 
 
+def attach_embeddings(
+    session: Session,
+    embedding_model_id: UUID,
+    sample_ids: list[UUID],
+    vectors: NDArray[np.float32],
+) -> None:
+    """Store precomputed embedding vectors for existing samples.
+
+    Use this to attach vectors that were computed outside of LightlyStudio (for
+    example, by an external model or a previous run) directly to samples that
+    already exist in the database. Prefer the generator-based paths
+    (`embed_images`, `embed_videos`, `embed_and_store_pil_images`,
+    `embed_annotations`) when embeddings still need to be computed from an
+    `EmbeddingGenerator`.
+
+    Args:
+        session: Database session for resolver operations.
+        embedding_model_id: ID of a registered embedding model the vectors belong to.
+        sample_ids: Sample IDs the vectors are stored for.
+        vectors: Embedding vectors to store, in the same order as sample_ids.
+
+    Raises:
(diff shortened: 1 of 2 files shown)


lightly_studio/src/lightly_studio/dataset/embedding_manager.py
```diff
@@ -541,6 +541,38 @@ def _get_image_model(self, model_id: UUID) -> ImageEmbeddingGenerator:
...
+    Args:
+        session: Database session for resolver operations.
+        embedding_model_id: ID of a registered embedding model the vectors belong to.
+        sample_ids: Sample IDs the vectors are stored for.
+        vectors: Embedding vectors to store, in the same order as sample_ids.
+
+    Raises:
+        ValueError: If the vectors fail validation. See `_validate_and_coerce_embeddings`.
+    """
+    _store_embeddings(
```
michal-lightly, line 568: Why do we need a new function?

PR #2150, opened 2026-08-31
Expose set_default_embedding_space in the Python SDK

iunir:
## What has changed and why?

Adds a public, module-level `set_default_embedding_space(collection_id, embedding_model_id)` function to the Python SDK, next to `set_default_embedding_model`. It links the embedding model to the collection if needed (idempotent) and marks it as the collection's default, for callers who want to explicitly control a collection's default embedding space (e.g. as the last step of an import flow).

## How has it been tested?

Added two unit tests in `lightly_studio/tests/dataset/test_embedding_manager.py`:
- Setting the default when no default exists yet and the model isn't linked to the collection.
- Replacing an existing default with a different, already-linked model.

Ran `cd lightly_studio && make static-checks` (ruff, mypy) and `uv run pytest tests/dataset/test_embedding_manager.py` (43 passed). `make test` requires a full frontend build (`dist_lightly_studio_view_app`) unavailable in this environment, so the relevant test file was run directly instead, per repo guidance.

- [x] Yes

---

Suggested reviewer: @michal-lightly (most frequent recent author of `embedding_manager.py` and `collection_embedding_model_resolver.py`). Alternative: @JonasWurst.

## Summary by CodeRabbit

* **New Features**
  * Added support for explicitly setting a collection’s default embedding space through the Python SDK.
  * Automatically links an embedding model to the collection when needed.
  * Selecting a new default embedding space replaces the previous default.

* **Documentation**
  * Documented the new capability in the unreleased changelog.

### DIFF
--- a/lightly_studio/src/lightly_studio/dataset/embedding_manager.py
+++ b/lightly_studio/src/lightly_studio/dataset/embedding_manager.py
@@ -12,6 +12,7 @@ from PIL import Image
 from sqlmodel import Session
 from tqdm import tqdm
 
+from lightly_studio.database import db_manager
 from lightly_studio.dataset import env
 from lightly_studio.dataset.embedding_generator import (
     EmbeddingGenerator,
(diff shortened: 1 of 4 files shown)


lightly_studio/src/lightly_studio/dataset/embedding_manager.py
```diff
@@ -87,6 +88,26 @@ def set_default_embedding_model(embedding_generator: EmbeddingGenerator) -> None
...
+
+    Links the embedding model to the collection if it isn't already, then flags it as
+    the collection's default. Both steps commit internally, so call this after all
+    other pending writes on the session, typically as the last step of an import flow.
+
+    Args:
+        collection_id: The collection whose default embedding model is set.
+        embedding_model_id: The embedding model to make the default.
+    """
+    session = db_manager.persistent_session()
```
michal-lightly, line 105: We always pass the session argument to functions dealing with the db. There are some exceptions, let's discuss if this is one, depending on its indented use cases.

PR #2155, opened 2026-09-01
Warn in the annotation grid when the sort order is stale

JonasWurst:
## Stack context

The Label QA workflow has two gaps:

* Editing an annotation makes the evaluation run stale, but that was only visible in the evaluation panel. → The stack shows an orange warning in the annotation grid and the detail view.
* Sorting by a metric gave no numbers, so the ranking was unexplained. → The stack shows the sort value on the grid tiles.

Stack: lightly-ai/lightly-studio#2154 sort value on tiles → lightly-ai/lightly-studio#2155 stale warning in the grid → lightly-ai/lightly-studio#2156 stale warning in the detail view.

---

### This PR

Stacked on lightly-ai/lightly-studio#2154.

Puts the stale warning and the recompute button next to the grid's sort control, gated on the run the grid is sorted by. Recompute now also invalidates the annotation list query, so the grid reorders.

Also renames "source collections" to "annotation sources" in the evaluation panel, per the glossary.

* The warning is persistent, an icon with a tooltip rather than a banogner.
* The recompute button only appears together with the warning.
* Inline markup, no new component. This puts `AnnotationOrderBy.svelte` at 129 lines, over the 100-line rule — accepted knowingly.

Not addressed: an edited annotation keeps its old metric value and rank until a recompute, since staleness is tracked per run, not per annotation.

## How has it been tested?

Five new tests (warning and button when stale, nothing when up to date, nothing without a sort, recompute of the active run, annotation-list invalidation). `2838 passed`; prettier, eslint and `svelte-check` clean.

`make static-checks` fails only on the gitignored `.claude/settings.local.json`, which reproduces on a clean tree.

## Did you update [CHANGELOG.md](<../CHANGELOG.md>)?

- [X] Yes
- [ ] Not needed (internal change)

## Summary by CodeRabbit

* **New Features**
  * Added a warning and recompute option when the selected evaluation run is outdated while sorting annotations.
  * Annotation results now refresh automatically after an evaluation run is recomputed, ensuring the displayed order is up to date.

* **Bug Fixes**
  * Updated stale-run messaging to accurately refer to evaluated annotation sources.
  * Improved annotation sorting controls to use the correct dataset context.

PR #2156, opened 2026-09-01
Warn in the annotation detail view when the step order is stale

JonasWurst:
## Stack context

The Label QA workflow has two gaps:

- Editing an annotation makes the evaluation run stale, but that was only visible in the evaluation panel. → The stack shows an orange warning in the annotation grid and the detail view.
- Sorting by a metric gave no numbers, so the ranking was unexplained. → The stack shows the sort value on the grid tiles.

Stack: #2154 sort value on tiles → #2155 stale warning in the grid → #2156 stale warning in the detail view.

---

### This PR

Stacked on #2155.

Puts the same warning icon and tooltip beside the prev/next control in the detail view, so a reviewer stepping through the queue sees that the order went stale. The tooltip points back to the grid.

- No recompute button here: it would invalidate the sequence being walked, and the grid is one click away.
- The annotation's own metric value is not shown. The detail payload endpoint does not carry it, so that needs a backend change. Deferred.
- Inline markup, no new component. `AnnotationDetailsNavigation.svelte` is 94 lines.

Not addressed: the active sort lives in an unpersisted store, so a hard reload onto a detail URL shows no warning. Persisting it in the URL touches routing and grid state.

## How has it been tested?

Three new tests (warning when stale, absent when up to date, absent without a sort, no recompute button). `2841 passed`; prettier, eslint and `svelte-check` clean.

`make static-checks` fails only on the gitignored `.claude/settings.local.json`, which reproduces on a clean tree.

- [x] Yes

## Summary by CodeRabbit

* **New Features**
  * Added a warning in the annotation detail view when navigation is based on stale evaluation results.
  * Warning is hidden when navigation uses text similarity search or no evaluation-based sorting.
  * Stale warnings now use the collection’s dataset context for accurate evaluation-run status.

* **Tests**
  * Added coverage for stale and fresh evaluation runs, search navigation, sorting behavior, dataset selection, and disabled evaluation-run queries.


iunir (commented):
Can you add this e2e as well? Seems good finding.

`e2e/evaluations/annotation-sort-stale-warning.e2e-test.ts`

```
import { expect, test } from '../utils';
import { cocoDataset } from '../general/fixtures';

const RUN_NAME = 'evaluation-example-all-samples';
const IOU_SORT_LABEL = `${RUN_NAME}.iou`;

test('hides the detail-view stale warning while a similarity search drives prev/next', async ({
    page,
    annotationsPage,
    annotationDetailsPage
}) => {
    // Sort the grid by the run's per-annotation IoU. This is the sort prev/next follows.
    await page.getByTestId('sort-by-trigger').click();
    await page.getByRole('option', { name: IOU_SORT_LABEL, exact: true }).click();

    // Editing any annotation in the source marks every run over it stale
    // (update_annotation_label -> mark_stale_by_collection_id).
    await annotationsPage.startEditing();
    await annotationsPage.selectAnnotation(0);
    await annotationsPage.setLabel(cocoDataset.labels.bear.name);
    await expect(page.getByTestId('annotation-sort-stale-icon')).toBeVisible();

    // Positive control: with no search the sort is what prev/next follows, so the detail
    // view must warn. Without this, the final assertion could pass vacuously.
    await annotationsPage.clickAnnotation(0);
    await annotationDetailsPage.waitForNavigation();
    await expect(page.getByTestId('annotation-navigation-stale-icon')).toBeVisible();

    // Back to the grid client-side only: the active sort lives in an unpersisted
    // module-level store, so a reload would drop it and void the rest of the test.
    await page.goBack();
    await expect(page.getByTestId('annotations-grid')).toBeVisible();

    // A text search takes precedence: useAdjacentAnnotations drops annotation_sort_by
    // whenever textEmbedding is set, so prev/next is similarity-ordered from here on.
    const searchResponse = page.waitForResponse(
        (response) => response.url().includes('query_text=bear') && response.status() === 200
    );
    await page.getByTestId('text-embedding-search-input').fill('bear');
    await page.keyboard.press('Enter');
    await searchResponse;

    // The grid disables its own sort control for the same reason.
    await expect(page.getByTestId('sort-by-trigger')).toBeDisabled();

    await annotationsPage.clickAnnotation(0);
    await annotationDetailsPage.waitForNavigation();

    // The stale run has no bearing on the order being stepped through, so the warning
    // must not claim it does. This assertion fails on the current PR head.
    await expect(page.getByTestId('annotation-navigation-stale-icon')).toBeHidden();
});
```

lightly_studio_view/src/lib/components/AnnotationDetails/AnnotationDetailsNavigation/AnnotationDetailsNavigation.svelte
```diff
@@ -4,19 +4,41 @@
...
 
-    const { query: sampleAdjacentQuery } = $derived(
+    const { query: sampleAdjacentQuery, sortBy: adjacentSortBy } = $derived(
         useAdjacentAnnotations({
             sampleId: annotationId,
             collectionId
         })
     );
 
+    const runsQuery = useEvaluationRuns(() => ({ datasetId: collectionDatasetId }));
```
LeonardoRosaa: Could we avoid fetch if `adjacentSortBy?.evaluation_run_id` is null?

lightly_studio_view/src/lib/components/AnnotationDetails/AnnotationDetailsNavigation/AnnotationDetailsNavigation.svelte
```diff
@@ -4,19 +4,41 @@
...
+    import { Tooltip } from '$lib/components/ui/tooltip';
+    import { TriangleAlert } from '@lucide/svelte';
+
+    const STALE_WARNING =
+        'Annotations changed after this evaluation ran, so the order you are stepping through is ' +
+        'out of date. Go back to the grid view to recompute the evaluation.';
+
+    interface Props {
+        // The URL's dataset_id segment is a root collection ID, so it cannot query evaluation runs.
+        collectionDatasetId: string;
```
LeonardoRosaa, line 17: Should we rename it to `datasetId`?

LeonardoRosaa (approved):

lightly_studio_view/src/lib/components/AnnotationDetails/AnnotationDetailsNavigation/AnnotationDetailsNavigation.svelte
JonasWurst: Done in 9c53638a2 — `useEvaluationRuns` now takes `enabled`, and the component passes `enabled: activeRunId !== null`, so the runs query stays idle while the prev/next order follows no run.

lightly_studio_view/src/lib/components/AnnotationDetails/AnnotationDetailsNavigation/AnnotationDetailsNavigation.svelte
JonasWurst: Kept as is: `datasetId` in this component is already the URL `dataset_id` segment, which is the root collection ID, so the names would collide. `collectionDatasetId` is the name `DatasetGridHeader` already uses for the same value.

PR #2182, opened 2026-09-03
perf: remove the N+1 in the embed_videos video path fetch

lorinczszabolcs:
## What has changed and why?

`EmbeddingManager.embed_videos` fetched one video per query in a loop, an N+1: N video ids ran N separate database queries. It now uses the existing batched `video_resolver.get_many_by_id`, so the path fetch runs one query instead of N (a few for very large lists, since the helper batches at 8000). The returned paths and their order are unchanged.

This mainly reduces the query count and the database load of the path fetch, from N queries to about one. The wall-clock effect is small. In a measured run on a laptop GPU the full per-video embedding cost was about 65 ms and the path fetch was about 1.9 ms, so the fetch was about 3 percent of the end-to-end time. The embedding step, which decodes each video and runs a model forward pass, is the dominant cost. On a remote PostgreSQL the fetch share is somewhat higher, since each query pays a network round-trip.

The error path changes slightly. The old `get_by_id` used `.one()`, which raised `NoResultFound` on a missing id, so the `if sample is not None` branch was dead and a missing id never raised the `ValueError` the docstring promises. The batched version raises that documented `ValueError`. No test or caller depended on the old exception.

## How has it been tested?

- `ruff check`, `ruff format --check`, and `mypy` pass on the changed files.
- `pytest tests/dataset/test_embedding_manager.py` passes, 42 tests. This includes the existing `test_embed_videos` and a new `test_embed_videos__raises_when_a_video_is_missing`, which covers the missing-id error path that was previously unreachable.

- [x] Yes (a `### Changed` bullet under Unreleased)

## Summary by CodeRabbit

* **Performance**
  * Reduced database query overhead during video embedding by fetching video paths through a single batched lookup rather than individual queries.

* **Bug Fixes**
  * Embedding now clearly reports an error when a requested video cannot be found and avoids saving partial results.

* **Tests**
  * Added coverage for missing-video handling, error reporting, and preventing persisted embeddings after a failed operation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Python SDK (beta): Simplified embedding generator interface. Implement `embedding_space_spec`, returning the new `EmbeddingSpaceSpec` (`space_key`, `dimension`) instead of the former `get_embedding_model_input`.
 - Bump lightly-mundig to 0.1.15; its sampling algorithms are up to 7x faster.
 - Python SDK: Split a dataset into new sample tags with `DatasetQuery.split()`
+- Reduce the number of database queries when embedding videos by fetching the video paths in one query instead of one per video.
 
 ### Deprecated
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -40,6 +40,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Python SDK (beta): Simplified embedding generator interface. Implement `embedding_space_spec`, returning the new `EmbeddingSpaceSpec` (`space_key`, `dimension`) instead of the former `get_embedding_model_input`.
 - Bump lightly-mundig to 0.1.15; its sampling algorithms are up to 7x faster.
 - Python SDK: Split a dataset into new sample tags with `DatasetQuery.split()`
+- Reduce the number of database queries when embedding videos by fetching the video paths in a batched lookup instead of one query per video.
```
michal-lightly: The improvement is nice, but as the description says, it does not have a significant effect on the end user. Let's remove the mention from the changelog, it is an internal optimisation.

michal-lightly (approved):
Thank you for the contribution!

CHANGELOG.md
lorinczszabolcs: Sure! Removed the changelog entry.

PR #2184, opened 2026-09-03
Add backend routes for bulk classification labeling

JonasWurst:
## What has changed and why?

Backend groundwork for assigning one annotation class to many images at once. Nothing is user-visible yet; the GUI arrives later in this stack.

- Adds `POST /collections/{collection_id}/annotations/classifications/bulk` (a list of sample ids) and `.../bulk_by_filter` (a grid filter), mirroring how tagging already offers both. The filter variant means "apply to everything currently selected" never has to send tens of thousands of ids over the wire.
- Both return how many annotations were created and how many were skipped, so the GUI can tell the user "added to 28 of 40".
- Images that already carry the picked class in the target annotation source are skipped instead of getting a duplicate. There is no database constraint that could express this, so the check lives in the resolver.
- The skip check is deliberately limited to classifications: a sample that already has a *bounding box* of class "car" still gets a whole-image "car" classification, because those are different annotations.
- A missing annotation class is created on the fly, as the single-annotation path already does.
- Annotations are added alongside whatever is already there and nothing is ever overwritten or deleted.

Part of LIG-10438. Stack: **this PR** → panel component → wiring.

## How has it been tested?

- 32 new tests across resolver, service and route: fresh creation, the skip path and its counts, a second different class kept alongside the first, a bounding box of the same class *not* blocking the classification, on-the-fly class creation, default vs named annotation source, per-source scoping, duplicate ids collapsed, batching, and the filter path matching the id-list result.
- The bounding-box regression test was verified to have signal by removing the fix and confirming exactly those tests fail.
- `make static-checks` clean (ruff, mypy over 932 files, format check). `make test` passes except `test_server_static_webapp`, which needs a real frontend build and fails identically on a checkout without one.

- [ ] Yes

---

Suggested reviewer: @horatiualmasan (most history on these files). Alternative: @LeonardoRosaa.

---

**Stack (review bottom-up):**

## Summary by CodeRabbit

* **New Features**
  * Added bulk creation of classification annotations by sample IDs or grid filters.
  * Reports created and skipped annotations, while preserving existing annotation classes.
  * Supports default or named annotation collections and creates missing labels automatically.
  * Grid-filter actions now correctly resolve embedding-region selections to matching samples.

* **Bug Fixes**
  * Improved detection of samples that already have a specified annotation class, preventing duplicates.

### DIFF
--- a/lightly_studio/src/lightly_studio/api/routes/api/annotation.py
+++ b/lightly_studio/src/lightly_studio/api/routes/api/annotation.py
@@ -50,6 +50,7 @@ from lightly_studio.services.annotations_service.update_annotation import (
 annotations_router = APIRouter(prefix="/collections/{collection_id}", tags=["annotations"])
 annotations_router.include_router(annotations_module.create_annotation_router)
 annotations_router.include_router(annotations_module.annotation_metrics_info_router)
+annotations_router.include_router(annotations_module.create_classification_annotations_router)
 
 
 @annotations_router.get(
--- a/lightly_studio/src/lightly_studio/api/routes/api/annotations/__init__.py
+++ b/lightly_studio/src/lightly_studio/api/routes/api/annotations/__init__.py
@@ -4,8 +4,12 @@ from .annotation_metrics_info import (
 from .create_annotation import (
     create_annotation_router,
 )
+from .create_classification_annotations import (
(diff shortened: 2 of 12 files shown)


JonasWurst:

PR #2185, opened 2026-09-03
Add the bulk annotation class panel component

JonasWurst:
## What has changed and why?

The panel that lets a labeler add one annotation class to a whole selection of images, built on its own so it can be reviewed without the data plumbing. It makes no API calls and touches no existing component — everything is new files.

- A picker for the annotation source and one for the annotation class, both of which also accept a name that does not exist yet.
- Applying always goes through a confirmation, because this action has no undo.
- A short hint pointing at the annotation view for changing or removing annotations, since this panel only ever adds.
- Renders nothing when nothing is selected.

`AnnotationSourcePill` could not be reused: it reads four different contexts and hooks, so it cannot be driven from props. A small presentational `NamePicker` serves both pickers instead. Worth a look during review — consolidating the two pickers later is a reasonable follow-up.

Part of LIG-10438. Stack: backend routes → **this PR** → wiring.

## How has it been tested?

- 13 colocated unit tests: nothing rendered at an empty selection, the confirmation appearing before anything is applied, the chosen class and source being handed back, and the apply control disabled while in flight.
- A Storybook story covering the empty, populated, single-image and applying states.
- `make static-checks` and `make test` pass.

Reviewer note: this is a UI change, so **a screenshot would normally belong here** — the panel cannot be rendered meaningfully until the wiring PR on top of this one, so the screenshots live there instead.

- [ ] Yes

---

Suggested reviewer: @horatiualmasan (same reviewer across the stack). Alternative: @LeonardoRosaa.

---

**Stack (review bottom-up):**

### DIFF
--- a/lightly_studio_view/src/lib/components/BulkAnnotationClassPanel/BulkAnnotationClassPanel.helpers.test.ts
+++ b/lightly_studio_view/src/lib/components/BulkAnnotationClassPanel/BulkAnnotationClassPanel.helpers.test.ts
@@ -1,0 +1,38 @@
+import { describe, expect, it } from 'vitest';
+import { summarizeApply, withSelectedOption } from './BulkAnnotationClassPanel.helpers';
+
+describe('summarizeApply', () => {
+    const selectionClassCounts = [
+        { className: 'dog', sampleCount: 3 },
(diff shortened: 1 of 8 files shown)


JonasWurst:

PR #2186, opened 2026-09-03
Wire the bulk annotation class panel into the image grid

JonasWurst:
## What has changed and why?

Connects the panel to real data and puts it in the image grid, completing bulk labeling from the image view.

- Select images in the image grid, turn on annotation editing, pick a class and a source, and every selected image gets that class. The class is added alongside anything already there.
- When the selection is an untouched "select all", the request goes out as a filter rather than a list of ids, so selecting a whole large collection stays cheap. Any manual deselection falls back to sending ids. This mirrors how bulk tagging already decides.
- After applying, a message reports what happened, including how many images already had the class and were skipped.
- The panel sits beside the grid the same way the annotation grid's selection panel does.

**Also fixes a pre-existing bug (not introduced by this stack):** creating an annotation in a brand-new annotation source did not make that source's entry appear in the left navigation until the page was reloaded. The navigation is fed by a different query than the one the annotation hooks were refreshing. Fixed in a shared helper, so the single-annotation path (`useCreateAnnotation`) is fixed too. Called out because it is a second purpose in this PR — happy to split it out if reviewers prefer.

Note on the diff: `Images.svelte` shows a large change that is almost entirely re-indentation from two new wrapper elements. The real change there is three lines.

Part of LIG-10438. Stack: backend routes → panel component → **this PR**.

## How has it been tested?

- Unit tests for the filter-vs-ids decision in both directions, including the "select all then deselect one" case falling back to ids; the panel staying hidden when editing mode is off or nothing is selected; and the set of caches refreshed after applying, including the navigation one.
- Manually in the app: applying to a fresh selection, applying a class some images already had and seeing the skip count, adding a second different class to an already-classified image, and confirming the class pills on the tiles update without a reload.
- The focus-ring clipping fix was verified by measuring the ring rectangle against every clipping ancestor rather than by eye.
- `make static-checks` clean (svelte-check 0 errors). `make test`: 368 files, 2872 tests passing.

**Screenshots: to be added before merge.**

- [x] Yes

---

Suggested reviewer: @horatiualmasan (same reviewer across the stack). Alternative: @LeonardoRosaa.

---

**Stack (review bottom-up):**


JonasWurst:

PR #2192, opened 2026-09-03
perf: batch the metadata-weighting sampling read to remove an N+1

lorinczszabolcs:
## What has changed and why?

The `MetadataWeightingStrategy` branch of `_add_strategy_to_mundig` read one metadata value per sample in a loop, an N+1 over the whole candidate pool. It now reads all values for the key in one query via `metadata_resolver.get_metadata_values_for_key`, then looks each sample up in memory. The weights keep the same values and the same order, so mundig gets identical input.

Here the loop is the cost, not a small fraction. A measured run on the in-memory DuckDB backend:

| Samples | Before (one query per sample) | Weighting step as a share of the full sampling call |
|---|---|---|
| 2,000 | 2,000 queries, 3.1 s | 98% |
| 5,000 | 5,000 queries, 6.7 s | 87% |

After the change the weighting step is one query, so the metadata-weighting sampling call is about 8 to 15 times faster end to end in this measurement. On PostgreSQL each per-sample query is a network round-trip, so the gain is larger. The read reuses the existing collection-scoped `get_metadata_values_for_key`.

The error path is unchanged: a missing or non-numeric value still raises the same `ValueError`, now covered by a test.

## How has it been tested?

- `ruff check`, `ruff format --check`, and `mypy` pass on the changed files.
- `pytest tests/sampling/test_sampling.py` passes, including the existing weighting tests that assert the exact weights and order (`[16.0, 50.0, 35.0]`, and video frames `[0.0, 1.0, 2.0, 0.0, 1.0]`) and a new `test_metadata_weighting__non_numeric_raises`.

- [x] Yes (a `### Changed` bullet under Unreleased)

## Summary by CodeRabbit

* **Performance**
  * Metadata-weighted sampling is faster by retrieving metadata values in bulk and limiting retrieval to relevant samples.
  * Metadata lookups are processed in batches when needed, helping avoid oversized requests.

* **Bug Fixes**
  * Metadata-weighted sampling now reports an error when metadata values are non-numeric.

* **Documentation**
  * Updated the unreleased changelog with the sampling performance improvement.


JonasWurst:
Thank you for the contribution!

lightly_studio/src/lightly_studio/resolvers/metadata_resolver/sample/get_metadata_values_for_key.py
```diff
@@ -11,19 +12,24 @@
...
 ) -> tuple[dict[UUID, Any], str | None]:
     """Get metadata values and schema type for one key in a collection.
 
     Args:
         session: The database session.
         collection_id: The collection's UUID.
         key: The metadata key to retrieve.
+        sample_ids: If given, read only these samples instead of the whole
+            collection. The read is batched so a long list stays within the
+            database parameter limit.
```
JonasWurst: Minor: This is quite verbose and does not fit the arg description above. Let's aim for a descriptive one-liner, the batching details are not required here.

lightly_studio/src/lightly_studio/sampling/sampling_via_db.py
```diff
@@ -566,10 +566,18 @@ def _add_strategy_to_mundig(
             mundig=mundig,
         )
     elif isinstance(strat, MetadataWeightingStrategy):
-        weights: list[float] = []
```
JonasWurst, line 569: Minor: This change is not required, unless I'm missing something.

lightly_studio/src/lightly_studio/sampling/sampling_via_db.py
```diff
@@ -566,10 +566,18 @@ def _add_strategy_to_mundig(
...
         metadata_key = strat.metadata_key
+        # Read the selected samples' values in one batched query, then look each up in
+        # memory, instead of one query per sample.
```
JonasWurst: Nit: I think your proposed change is clear, we do not need a comment for it. :)
```suggestion
```

JonasWurst (changes requested):
This looks already pretty good, thank you!

A few minor points, I would ask you to update before we can merge this.

lorinczszabolcs:
Thanks for the review! All three addressed.

JonasWurst (approved):
LGTM, thanks for contributing!

PR #2206, opened 2026-09-04
perf: decode video frames in parallel in embed_videos

lorinczszabolcs:
## What has changed and why?

`PerceptionEncoderEmbeddingGenerator.embed_videos` decoded video frames one file at a time before running inference. The image path was already made parallel in #2086, but the video path was not, so video embedding was decode-bound.

This change decodes the frames on a bounded thread pool with `parallelize.thread_imap_lazy` and `executor.get_media_worker_count()`, the same utilities the image path uses. Video decoding is heavy work that releases the GIL, so it runs across cores, and the read-ahead also overlaps decoding with model inference. The results stay in input order, and missing or broken files are still skipped and reported. The embeddings are unchanged.

The embedder was copied to `embed/perception_encoder_embedder.py` in #2195 with the same sequential decode. I left that copy untouched to avoid colliding with the in-progress `embed/` refactor (#2015).

## How has it been tested?

- Added a test that embeds more than one batch of videos with broken and missing files interleaved, so the parallel path is checked for correct order and skips across batch boundaries.
- `make static-checks` passes (ruff, mypy, format).
- The video embedding tests pass.
- Local benchmark on an RTX 4070 laptop GPU with real 1080p and 720p clips: embedding a 24-video mix went from about 1006 to about 191 ms per video (about 5.3x). Video decode is the dominant cost on real videos, so the parallel decode gives most of the speedup.

- [x] Yes

## Summary by CodeRabbit

- **Performance**
  - Video embedding is now faster by decoding video frames in parallel.

- **Reliability**
  - Missing or broken video files continue to be skipped and reported correctly.
  - Embedding results preserve the original input order, including across batches with skipped files.

- **Documentation**
  - Added an Unreleased changelog entry describing the video embedding performance improvement.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 ### Changed
+- Speed up video embedding by decoding video frames in parallel.
 
 ### Deprecated
 
--- a/lightly_studio/src/lightly_studio/dataset/perception_encoder_embedding_generator.py
+++ b/lightly_studio/src/lightly_studio/dataset/perception_encoder_embedding_generator.py
@@ -22,7 +22,7 @@ from lightly_studio.dataset.env import LIGHTLY_STUDIO_MODEL_CACHE_DIR
 from lightly_studio.embed import image_crop_embedding, image_embedding
 from lightly_studio.embed.image_embedding import EmbeddingContext
 from lightly_studio.embed.types import EmbeddingResult, EmbeddingSpaceSpec, ImageCrop
-from lightly_studio.utils import batching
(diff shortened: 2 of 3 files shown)


michal-lightly:
Hi @lorinczszabolcs, we are currently refactoring backend logic around embeddings. The PerceptionEncoderEmbeddingGenerator will be soon deprecated, during the migration period we are implementing PerceptionEncoderEmbedder that will replace it.

This PR will therefore need to wait a little bit and then be applied to PerceptionEncoderEmbedder. Thanks for your patience!

lorinczszabolcs:
Sounds good, thanks for the quick feedback!

lorinczszabolcs:
Hi @michal-lightly,  closing this in favor of #2378, since I noticed the video embedding path now uses `PerceptionEncoderEmbedder`.

PR #2207, opened 2026-09-04
Finish the Capability enum rename before it becomes public API

iunir:
## Description

`Capability` and the `Embedder` ABCs are merged but not exported from `lightly_studio/__init__.py` and not referenced in the docs, so this rename is free today and a deprecation once they are public.

- `CROP_PATH` → `IMAGE_CROP_PATH`, `CropPathEmbedder` → `ImageCropPathEmbedder`, `embed_crops` → `embed_image_crops`, `get_crop_path_embedder` → `get_image_crop_path_embedder`. Subject-first matches the existing public `ImageCrop` type. Deliberately not `IMAGE_REGION_*`: a mask is also a region, so naming the rectangle "region" would leave a future non-rectangular capability with no name. `ImageCrop` keeps its name.
- Dropped "or URL" from the `IMAGE_PATH` and `VIDEO_PATH` docstrings. An fsspec path is resolved with our credentials; a presigned URL is fetched by whoever holds it. Conflating the two is what makes people assume `IMAGE_PATH` is legal over the wire — per LIG-10640 it is not, and `PATH` and `PIL` stay local-only.
- Added `VIDEO_BYTES`, declared ahead of a local implementation.
- `IMAGE_URL` and `VIDEO_URL` are **deferred, not merged into the path members**. The transport distinction above still holds; what is unsettled is how an implementer is meant to tell the two apart in practice, given fsspec also accepts `http(s)` paths. They land once signed URLs have a concrete call site to document them against. No `IMAGE_CROP_URL` in any case: crops have no endpoint of their own, we crop with padding on our side and post the result as image bytes.
- Dropped the `Ingest:` / `Interactive:` prefixes, and named the members per phase on the enum instead.

Existing member and class order is untouched.

## Linear

LIG-10735

## Test plan

`make static-checks` and `uv run pytest tests/embed` pass (41 tests). CI green on e919ec2.

### DIFF
--- a/lightly_studio/src/lightly_studio/embed/embedder.py
+++ b/lightly_studio/src/lightly_studio/embed/embedder.py
@@ -48,7 +48,6 @@ class Capability(str, Enum):
     """Video as raw file bytes. No local implementation yet."""
 
 
-
 class Embedder(ABC):
     """Base class for every embedder.
 
--- a/lightly_studio/src/lightly_studio/embed/embedder.py
+++ b/lightly_studio/src/lightly_studio/embed/embedder.py
@@ -24,6 +24,12 @@ class Capability(str, Enum):
     Members are named ``SUBJECT[_TRANSPORT]``, with the transport omitted when a
     subject has a single form. ``PATH`` and ``PIL`` are local-only; ``URL`` and
     ``BYTES`` are the transports an embedding backend can accept.
+
+    Capabilities are used both at ingest, when data is loaded into the database,
+    and interactively while the GUI runs. Which of the two a capability serves is
+    a property of the call site rather than of the capability itself:
+    ``IMAGE_BYTES`` backs interactive search today and bulk ingest over the wire
+    later.
     """
 
     IMAGE_PATH = "image_path"
--- a/lightly_studio/src/lightly_studio/embed/embedder.py
+++ b/lightly_studio/src/lightly_studio/embed/embedder.py
@@ -21,15 +21,9 @@ class Capability(str, Enum):
 
     <span class="doc-badge doc-badge--beta">Beta</span>
 
-    Members are named ``SUBJECT[_TRANSPORT]``, with the transport omitted when a
-    subject has a single form. ``PATH`` and ``PIL`` are local-only; ``URL`` and
-    ``BYTES`` are the transports an embedding backend can accept.
-
-    Capabilities are used both at ingest, when data is loaded into the database,
-    and interactively while the GUI runs. Which of the two a capability serves is
-    a property of the call site rather than of the capability itself:
-    ``IMAGE_BYTES`` backs interactive search today and bulk ingest over the wire
-    later.
+    Some capabilities are used at ingest time (IMAGE_PATH, IMAGE_CROP_PATH,
+    VIDEO_PATH, IMAGE_PIL, VIDEO_BYTES) and some interactively while the GUI
+    runs (TEXT, IMAGE_BYTES).
     """
 
     IMAGE_PATH = "image_path"
@@ -50,14 +44,9 @@ class Capability(str, Enum):
     IMAGE_BYTES = "image_bytes"
     """Image as raw file bytes."""
 
-    IMAGE_URL = "image_url"
-    """Image by URL. No local implementation yet."""
-
     VIDEO_BYTES = "video_bytes"
     """Video as raw file bytes. No local implementation yet."""
 
-    VIDEO_URL = "video_url"
-    """Video by URL. No local implementation yet."""
 
 
 class Embedder(ABC):
--- a/lightly_studio/src/lightly_studio/embed/embedder.py
+++ b/lightly_studio/src/lightly_studio/embed/embedder.py
@@ -1,9 +1,9 @@
 """Capability-split embedder interfaces.
 
 Defines the ``Embedder`` base class and one abstract subclass per input a model
-can embed: images and crops by path, videos, video frames, text, and images by
-bytes. A concrete model implements only the capabilities it supports, and
-callers pick an embedder by the capability they need.
+can embed: images and image crops by path, videos, video frames, text, and
+images by bytes. A concrete model implements only the capabilities it supports,
+and callers pick an embedder by the capability they need.
 """
 
 from __future__ import annotations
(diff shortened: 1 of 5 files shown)


lightly_studio/src/lightly_studio/embed/embedder.py
```diff
@@ -21,27 +21,37 @@ class Capability(str, Enum):
...
     TEXT = "text"
-    """Interactive: text query."""
+    """Text query."""
 
     IMAGE_BYTES = "image_bytes"
-    """Interactive: image as raw file bytes."""
+    """Image as raw file bytes."""
+
+    IMAGE_URL = "image_url"
+    """Image by URL. No local implementation yet."""
```
michal-lightly: We should not need this. It will be the same as IMAGE_PATH, because fsspec path for AWS/GCP/etc is a url.

Should we rather rename IMAGE_PATH to IMAGE_URI or maybe IMAGE_FSSPEC ?

lightly_studio/src/lightly_studio/embed/embedder.py
```diff
@@ -21,27 +21,37 @@ class Capability(str, Enum):
...
 
-    Some capabilities are used at ingest, when data is loaded into the database.
-    Others are needed interactively while the GUI runs.
+    Members are named ``SUBJECT[_TRANSPORT]``, with the transport omitted when a
+    subject has a single form. ``PATH`` and ``PIL`` are local-only; ``URL`` and
+    ``BYTES`` are the transports an embedding backend can accept.
     """
 
     IMAGE_PATH = "image_path"
-    """Ingest: image by fsspec path or URL."""
```
michal-lightly, line 29: Minor: I'd keep some documentation on when the capability is used. Ingest/interactive was minimal. maybe it should be mentioned elsewhere? Can be a separate PR later.

michal-lightly (changes requested):

lightly_studio/src/lightly_studio/embed/embedder.py
iunir: This can be needed in case of signed URLs. If we store in our private bucket and generating signed  read only URL so our client can access it

lightly_studio/src/lightly_studio/embed/embedder.py
iunir: Agreed, added it back at the class level in 3a78104 rather than deferring — it's a docstring, not worth its own PR.

The ticket's objection was to the per-member prefixes binding one phase to one capability, not to documenting the phases at all: `IMAGE_BYTES` was labelled `Interactive:` while bulk ingest over the wire is going to use it. So the phases are now described on `Capability` itself, with the note that which one a capability serves is a property of the call site:

> Capabilities are used both at ingest, when data is loaded into the database, and interactively while the GUI runs. Which of the two a capability serves is a property of the call site rather than of the capability itself: `IMAGE_BYTES` backs interactive search today and bulk ingest over the wire later.

lightly_studio/src/lightly_studio/embed/embedder.py
```diff
@@ -21,27 +21,43 @@ class Capability(str, Enum):
 
     <span class="doc-badge doc-badge--beta">Beta</span>
 
-    Some capabilities are used at ingest, when data is loaded into the database.
-    Others are needed interactively while the GUI runs.
+    Members are named ``SUBJECT[_TRANSPORT]``, with the transport omitted when a
+    subject has a single form. ``PATH`` and ``PIL`` are local-only; ``URL`` and
+    ``BYTES`` are the transports an embedding backend can accept.
```
michal-lightly: This can be removed

lightly_studio/src/lightly_studio/embed/embedder.py
```diff
@@ -21,27 +21,43 @@ class Capability(str, Enum):
...
+
+    Capabilities are used both at ingest, when data is loaded into the database,
+    and interactively while the GUI runs. Which of the two a capability serves is
+    a property of the call site rather than of the capability itself:
+    ``IMAGE_BYTES`` backs interactive search today and bulk ingest over the wire
+    later.
```
michal-lightly: Minor: This is not a very useful description (vague), and contains AI bloat. It should say:
```
Some capabilities are used at ingest time (IMAGE_PATH, CROP_PATH, VIDEO_PATH, IMAGE_PIL)
and some interactively while the GUI runs (TEXT, IMAGE_BYTES).
```

lightly_studio/src/lightly_studio/embed/embedder.py
michal-lightly: Then let's add it once we have signed URLs? Because otherwise; how does the user know if they should implement IMAGE_PATH or IMAGE_URL?

Or do you suggest we change the current implementation and split it by local path / cloud path (with the same implementation for the builtin models)?

iunir: This is an open question, not sure if we need to split local vs cloud path, but URL is something https://....

michal-lightly: Then my suggestion would be to add it later, since it is not clear yet when/how it would be used - and how we would document it.

Btw fsspec supports http paths too.

lightly_studio/src/lightly_studio/embed/embedder.py
```diff
@@ -21,27 +21,32 @@ class Capability(str, Enum):
 
     <span class="doc-badge doc-badge--beta">Beta</span>
 
-    Some capabilities are used at ingest, when data is loaded into the database.
-    Others are needed interactively while the GUI runs.
+    Some capabilities are used at ingest time (IMAGE_PATH, IMAGE_CROP_PATH,
+    VIDEO_PATH, IMAGE_PIL, VIDEO_BYTES) and some interactively while the GUI
+    runs (TEXT, IMAGE_BYTES).
```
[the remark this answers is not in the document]
iunir: Not applying this one — it would revert an explicit review instruction.

An earlier revision of this docstring said exactly what you're asking for:

> Which of the two a capability serves is a property of the call site rather than of the capability itself: `IMAGE_BYTES` backs interactive search today and bulk ingest over the wire later.

@michal-lightly asked for that to go in https://github.com/lightly-ai/lightly-studio/pull/2207#discussion_r3934165164, called it vague and bloated, and gave the exact replacement wording now in the file. Adding the caveat back would put us in a loop.

On the substance: the reading you're guarding against is the one the ticket set out to remove, and the current text is already weaker than what it replaced. The old per-member `Ingest:` / `Interactive:` prefixes did attach a phase to each member; a single sentence naming which capabilities are used where does not. `VIDEO_BYTES` is listed under ingest precisely because bulk ingest over the wire will use it, which is the case that would otherwise look interactive.

michal-lightly (approved):

PR #2211, opened 2026-09-04
Introduce the uv workspace and the lightly-embed member

iunir:
Resolves [LIG-10689](https://linear.app/lightly/issue/LIG-10689/introduce-the-uv-workspace-and-the-lightly-embed-member).

## What has changed and why?

`lightly-embed` will be the server a customer runs in front of their own embedding model
([LIG-10688](https://linear.app/lightly/issue/LIG-10688)). It ships as its own wheel from this
repo, so that it can be released together with `lightly-studio` and share one definition of the
wire contract. This PR is only the packaging: no protocol code yet, that is
[LIG-10691](https://linear.app/lightly/issue/LIG-10691).

- **Root `pyproject.toml`** with `[tool.uv.workspace]`, members `lightly_studio` and
  `lightly_embed`. The root is virtual: it ships no code. The `[tool.uv]` settings
  (`required-version`, `constraint-dependencies`) move there unchanged, because uv reads those
  from the workspace root only.
- **`uv.lock` moves to the repository root**, since one lockfile now covers both members. Every
  path that named `lightly_studio/uv.lock` moves with it: the two release workflows, the e2e
  cache key, the unit test path filter, `.gitattributes` and the Fast Track diff-size exclusions.
- **`lightly_embed/`**: `fastapi`, `uvicorn`, `pydantic` and nothing else, `requires-python`
  matching `lightly-studio`, plus ruff, mypy and pytest mirroring the other member.
- **Unit tests CI**: a new gated job runs the member's static checks, tests and the wheel
  assertion on 3.9 and 3.14, and is wired into the required `CI Success Check`.
- **Release CI**: a new job builds the wheel from the released commit and attaches it to the run.
  Uploading is behind `PUBLISH_TO_PYPI: "false"`, so
  [LIG-10690](https://linear.app/lightly/issue/LIG-10690) is a flag flip rather than new
  machinery written under time pressure.

The invariant the package exists for is that its wheel installs next to a customer's own CUDA and
torch pins. That is asserted twice, because a monorepo makes it easy to break by accident:

- `lightly_embed/tests/test_packaging.py` asserts the *declared* dependency set from the
  distribution metadata, so an added direct dependency has to be a deliberate decision.
- `make -C lightly_embed check-wheel-dependencies` resolves the *built wheel's* full transitive
  tree and fails on anything torch-, duckdb-, opencv- or lightly-studio-shaped. It runs in the
  unit test workflow for each supported Python, and again in the release job as the last gate.

Two things deliberately left out:

- **`lightly-studio` does not depend on `lightly-embed` yet.** The workspace makes it possible;
  adding it before the package is on PyPI would break `pip install lightly-studio`. That
  dependency lands with the wire models in
  [LIG-10641](https://linear.app/lightly/issue/LIG-10641).
- **No changelog entry**, since nothing user-facing ships: the package is empty and unpublished.

`lightly-embed` starts at `0.1.0`. Whether its version should track `lightly-studio`'s is still
the open question on the story, and nothing here depends on the answer.

## Notes for the reviewer

- **The `no-fast-track` label is set.** The PR adds ~370 lines, over the diff-size guardrail, and
  it touches CI. Roughly 180 of those lines are the new member's own files, 120 are the two
  workflows, and the rest is the lockfile path move. Splitting it would mean landing a workspace
  with one member, which changes the release flow twice for no benefit.
- Risk is concentrated in the paths that moved. `uv build` in a workspace defaults to the
  workspace root, so both `uv build` call sites in `lightly_studio/Makefile` now pass
  `--out-dir dist` to keep the studio's wheel where it was.
- `lightly_embed/ruff.toml` extends `lightly_studio/ruff.toml` rather than copying it. The
  `tests/**/*.py` per-file ignores are repeated because ruff resolves relative patterns against
  the directory of the file that declares them.

## How has it been tested?

Locally, on macOS with uv 0.12.6:

- `make -C lightly_embed static-checks`, `test` and `check-wheel-dependencies` pass. The resolved
  wheel tree is 14 packages, all of them `fastapi`/`uvicorn`/`pydantic` transitives.
- `make -C lightly_studio static-checks` passes (938 files), a test slice runs, and
  `uv build --out-dir dist` still writes `lightly_studio/dist`. The built `lightly-studio` wheel's
  metadata does not mention `lightly-embed`.
- `make -C .github/scripts static-checks test` (56 tests) and `fast_track` `static-checks` and
  `test` (205 tests) pass with the moved lockfile path.
- The `uv.lock` diff is the two new `[[package]]` entries plus the workspace manifest; the release
  job's `assert-lock-diff` guard already covers every other package block, including the new one.

CI covers the rest: the new member's job on both ends of the matrix, and the unchanged
`lightly-studio` jobs resolving from the moved lockfile.

- [ ] Yes

## Summary by CodeRabbit

* **New Features**
  * Added the `lightly-embed` package for serving custom embedding models to LightlyStudio.
  * Added installation and package metadata supporting Python 3.9–3.14 with minimal HTTP-serving dependencies.

* **Build & Release**
  * Added automated packaging, validation, and publication workflows for `lightly-embed`.
  * Added shared workspace tooling for building, testing, linting, and formatting both packages.

* **Documentation**
  * Added setup and workspace guidance for contributors and package users.
  * Documented that the initial package release does not yet provide a public API.


iunir:
Superseded by a two-PR stack, so that the workspace conversion and the new package can be reviewed separately:

- #2217 — convert the repository to a uv workspace
- #2218 — add the lightly-embed workspace member, based on #2217

Every review comment from this PR is addressed in the stack. This one closed itself when its branch was renamed; it is kept only for that history.

PR #2215, opened 2026-09-07
Define point-cloud labeling domain contracts and fixtures

ikondrat:
## What has changed and why?

Fixes [LIG-10658](https://linear.app/lightly/issue/LIG-10658/define-point-cloud-labeling-domain-contracts-and-fixtures).

- Defines framework-independent point-cloud, camera calibration, cuboid annotation, track, selection, hover, tool, and dirty-state contracts.
- Adds immutable constructors and explicit coordinate, timestamp, source, and camera validation.
- Adds deterministic empty, normal, large (350,000-point), partial, malformed, annotation, and workspace fixtures.
- Adds structured-cloneable worker transfer helpers so renderers can receive disposable typed-array copies.
- Documents canonical coordinates, units, timestamp semantics, calibration, ownership, and persistence usage.

The prototype reference was the browser-processing implementation in https://github.com/lightly-ai/kondrat-3d-point-cloud-with-browser-processing.

## How has it been tested?

- `npx prettier --check src/lib/components/PointCloudLabelingWorkspace/domain`
- `npx eslint src/lib/components/PointCloudLabelingWorkspace/domain`
- `npx vitest run src/lib/components/PointCloudLabelingWorkspace/domain` (14 tests)
- Strict isolated TypeScript compilation and declaration generation
- `git diff --check`

The full frontend suite still has pre-existing generated-client type errors and two unrelated navigation test failures.

Suggested reviewers: Leonardo Rosa. Alternative: Igor Susmelj.

## Proposed review process by commit diffrence
It is recommended to review in a stacked commit diff to optimise workflow and avoid to many overlaps with branch updates starting from the first commit https://github.com/lightly-ai/lightly-studio/pull/2215/commits

## Summary by CodeRabbit

* **New Features**
  * Added structured support for point-cloud, camera, and cuboid annotation data.
  * Added coordinate-frame validation using meter-based, right-handed 3D conventions.
  * Added validation for timestamps, camera calibration, point attributes, and annotation geometry.
  * Added immutable point-cloud data with calculated bounds and transferable buffers.
  * Added interaction state for tools, camera modes, selection, hover, and pending annotation changes.

* **Documentation**
  * Documented point-cloud labeling contracts and interaction-state conventions.

* **Tests**
  * Added coverage for validation, immutability, serialization, coordinate compatibility, fixtures, and buffer transfer.


lightly_studio_view/src/lib/components/PointCloudLabelingWorkspace/domain/pointCloudFrame.ts
```diff
@@ -0,0 +1,78 @@
...
+
+function freezeMetadata<T extends object>(value: T): T {
+    Object.values(value).forEach((child) => {
+        if (child !== null && typeof child === 'object') freezeMetadata(child);
+    });
+    return Object.freeze(value);
+}
+
+/** Copies, validates, bounds, and freezes provider output for shared domain consumption. */
+export function createPointCloudFrame(input: PointCloudFrameInput): PointCloudFrame {
```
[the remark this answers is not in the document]
ikondrat: Addressed in `1030a3527`, with the metadata typing follow-up in `696ec9a6c`. Frame validation and metadata cloning now live in focused helpers, leaving the constructor as orchestration only.

lightly_studio_view/src/lib/components/PointCloudLabelingWorkspace/domain/fixtures.ts
```diff
@@ -0,0 +1,133 @@
...
+        coordinateFrame: canonicalCoordinateFrame('lidar-0'),
+        positions,
+        intensity: new Float32Array(pointCount).fill(0.5),
+        color: new Float32Array(pointCount * 3).fill(0.25),
+        cameras: [createCameraFixture()]
+    };
+}
+
+/** Creates empty, normal, large, partial, and malformed frame fixtures. */
+export function createFrameFixtures() {
```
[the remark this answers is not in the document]
ikondrat: Addressed in `bf8b38cc9`. Added the named `PointCloudFrameFixtures` return interface and a named `WorkspaceFixture` contract for the related factory.

lightly_studio_view/src/lib/components/PointCloudLabelingWorkspace/domain/coordinates.ts
```diff
@@ -0,0 +1,59 @@
...
+        actual.unit !== expected.unit ||
+        actual.convention !== expected.convention
+    ) {
+        throw new Error('Coordinate frames differ; an explicit rigid transform is required.');
+    }
+}
+
+/** Validates and freezes a vector so shared geometry cannot be mutated. */
+export function freezeVector(value: Vector3): Vector3 {
+    if (value.length !== 3 || !value.every(Number.isFinite)) {
```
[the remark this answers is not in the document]
ikondrat: Addressed in `3206e30b6`. `freezeVector` and `freezeRotation` now materialize tuple inputs with `Array.from()` before validation, and regression coverage rejects sparse vectors and quaternions.

lightly_studio_view/src/lib/components/PointCloudLabelingWorkspace/domain/contracts.ts
```diff
@@ -0,0 +1,234 @@
+/** A canonical position or translation: x forward, y left, z up, in metres. */
+export type Vector3 = readonly [number, number, number];
```
michal-lightly, line 2: Isn't there a canonical vector type for typescript that also had implemented arithmetics?

lightly_studio_view/src/lib/components/PointCloudLabelingWorkspace/domain/contracts.ts
```diff
@@ -0,0 +1,234 @@
...
+    /** Packed xyz triples, no stride or padding. All coordinates must be finite. */
+    readonly positions: PackedPointAttribute;
+    /** One normalized [0, 1] value per point, when available. */
+    readonly intensity?: PackedPointAttribute;
+    /** Packed linear RGB triples in [0, 1], when available. */
+    readonly color?: PackedPointAttribute;
+    /** Null exactly when the frame has no points. */
+    readonly bounds: Bounds3 | null;
+    /** Camera context associated with this recording time. */
+    readonly cameras: readonly CameraFrame[];
```
michal-lightly, line 82: Should we have different cameras per frame? It is more robust (cameras can in theory move wrt the Lidar) but in the typical usecase I'd expect them fixed for one recording. Is it worth the complexity?

lightly_studio_view/src/lib/components/PointCloudLabelingWorkspace/domain/contracts.ts
```diff
@@ -0,0 +1,234 @@
+/** A canonical position or translation: x forward, y left, z up, in metres. */
+export type Vector3 = readonly [number, number, number];
+/** An active local-to-parent rotation in unit quaternion order x, y, z, w. */
+export type Quaternion = readonly [number, number, number, number];
```
michal-lightly, line 4: Side note: Quaternion is fine. I've used in the past also "rotation vector" - the vector is the axis of rotation, and the directed length is the rotation amount in radians. Pros is that it is easier to interpret and less data, cons is that it anyway needs to be converted to quaternions when doing math. You can mention it to AI in case it thought it would be easier that way (I doubt that though).

lightly_studio_view/src/lib/components/PointCloudLabelingWorkspace/domain/contracts.ts
```diff
@@ -0,0 +1,234 @@
...
+
+/** Identifies the origin, axes, handedness, and unit used by geometry. */
+export interface CoordinateFrame {
+    /** Stable identifier for the physical or recording frame. */
+    readonly id: string;
+    /** The only convention accepted by the canonical domain model. */
+    readonly convention: 'right-handed-x-forward-y-left-z-up';
+    /** The canonical distance unit. */
+    readonly unit: 'metre';
```
michal-lightly: Optional: It seems unusual to use british "metre" instead of american "meter".

lightly_studio_view/src/lib/components/PointCloudLabelingWorkspace/domain/contracts.test.ts
```diff
@@ -0,0 +1,91 @@
...
+} from './fixtures';
+
+describe('point-cloud domain contracts', () => {
+    it('covers empty, populated, large, partial and malformed frames', () => {
+        const fixtures = createFrameFixtures();
+        expect(fixtures.empty.bounds).toBeNull();
+        expect(fixtures.normal.bounds).toEqual({ min: [0, 0, 0], max: [3, 0, 1.5] });
+        expect(fixtures.large.positions.length).toBe(1_050_000);
+        expect(fixtures.partial.intensity).toBeUndefined();
+        expect(fixtures.partial.cameras[0].image).toBeNull();
```
michal-lightly: This test tests contents of a fixture, that is in a completely different file. In Python we have a convention (we try) that a test tests only data set up inside the test. Could we do that here?

michal-lightly (approved):

LeonardoRosaa (approved):

lightly_studio_view/src/lib/components/PointCloudLabelingWorkspace/domain/contracts.ts
ikondrat: no there is no canonical vector in js, there is array which behaves a bit like vector but not entirely

lightly_studio_view/src/lib/components/PointCloudLabelingWorkspace/domain/contracts.ts
ikondrat: we don't know for sure, because machines with Lidars can also move 

excavator can also operate on the move 

I would keep it for now, we have only one file for now, but I can imagine gravis carring machines can move as well and then this started to be important

lightly_studio_view/src/lib/components/PointCloudLabelingWorkspace/domain/contracts.test.ts
ikondrat: yep make sense

lightly_studio_view/src/lib/components/PointCloudLabelingWorkspace/domain/contracts.ts
ikondrat: It looks like https://threejs.org/docs/#Quaternion has support in Three js and that was probably the motivation to use it directly

PR #2218, opened 2026-09-07
Add the lightly-studio-embed workspace member

iunir:
**Top of a two-PR stack, based on #2217**, which converts the repo to a uv workspace. Read that
one first: this diff is against its branch.

## What has changed and why?

`lightly-studio-embed` will be the server a customer runs in front of their own embedding model.
It ships as its own wheel, from this repo so that the wire contract has a single definition that
both halves change together. This PR is only the packaging — no protocol code yet.

- **`lightly_studio_embed/`**: `fastapi`, `uvicorn`, `pydantic` and nothing else,
  `requires-python` matching `lightly-studio`, ruff extending the studio's config, mypy strict,
  pytest.
- **Unit tests CI**: a gated job runs the member's static checks and tests on 3.9 and 3.14, wired
  into the required `CI Success Check`. It runs a Makefile the PR supplies, so it gets the
  read-only token treatment `check-release-tooling` already uses.
- **No release CI.** The package must not be published on every `lightly-studio` release — it
  changes only when the protocol does, so a customer pinning it would see churn from releases
  that changed nothing in it. `publish_release.yml` is therefore byte-identical to `main`, and
  the release trigger is designed separately.
- **`make/python.mk`** holds the check targets every Python member shares, so they are defined
  once instead of copied per member. Root `make static-checks`, `type-check` and `lint-fix`
  iterate a `PYTHON_MEMBERS` list, so one command covers every Python package.

`lightly-studio` deliberately does **not** depend on `lightly-studio-embed` yet — declaring it
before the package is on PyPI would break `pip install lightly-studio`. It lands with the wire
models.

## Notes for the reviewer

- **`no-fast-track` is set**: 259 added lines excluding the lockfile, against a 215 limit. There
  is no logic in it at all — the largest files are the CI job (65), `pyproject.toml` (59),
  `make/python.mk` (37) and CONTRIBUTING (25). The package itself is a docstring and a
  `py.typed`.
- `lightly_studio/Makefile` loses 35 lines to `make/python.mk` and gains an `include`. The
  commands it runs are unchanged, which the backend CI jobs on this PR exercise.
- `lightly_studio_embed/ruff.toml` extends the studio's config instead of copying it. It repeats
  the `tests/**/*.py` ignores ahead of there being any tests, because ruff resolves relative
  patterns against the directory that declares them — so the extended entry covers only the
  studio's tests, and without this the first test written here fails on `D103`.
- Per review, both dependency guards were removed: the packaging test only restated the
  dependency lines above it in `pyproject.toml`, and the wheel check was a blocklist of five
  name fragments, so neither bounded the wheel's weight in general. The dependency list itself
  is three reviewable lines, and anything transitive shows in the `uv.lock` diff.
- The package has no code yet, so it has no tests. `make test` accepts pytest's "no tests
  collected" exit code until `serve()` lands; a real failure still fails the target.
- Per the thread on the Makefile: the backend coverage guardrail hardcodes `lightly_studio/`, so
  it does not see this package, and it fails quietly — coverage simply is not measured. Harmless
  while the package has no code, so it is a follow-up rather than a blocker here.

## How has it been tested?

Locally with uv 0.12.6: the member's `static-checks`, `test` and `build` pass, and root
`make static-checks` covers both members (940 files for the studio, 1 for the member). The
`uv.lock` diff is purely additive, and the `lightly-studio` wheel's metadata does not mention
`lightly-studio-embed`.

The backend CI jobs run on this PR because `lightly_studio/Makefile` changed, which is what
exercises the shared `make/python.mk` against the studio's full suite rather than only the
member's.

- [ ] Yes

## Summary by CodeRabbit

- **New Features**
  - Added the foundation for a lightweight embedding server package that can serve local embedding models over HTTP.
  - Defined a minimal runtime dependency set to better coexist with environment-specific machine-learning installations.
  - Added packaging and build support for creating a distributable wheel.

- **Documentation**
  - Added setup and development guidance for the embedding server package.
  - Documented its current status, including that no public API or package release is available yet.

- **Chores**
  - Expanded automated testing, linting, type checking, and packaging validation across supported Python versions.


lightly_studio_embed/Makefile
```diff
@@ -0,0 +1,77 @@
+.PHONY: clean
```
michal-lightly, line 1: I'm wondering if we could have one test command for all python code.

I.e.:
- `make test` runs pytest for all python packages
- `make type-check` type checks everything
- etc.

The concern with having them separate is that we have to duplicate the setup otherwise - and this will grow if we add more sibling packages. (E.g. fast-track would not work for lightly_studio_embed.)

We might need to nest the package one lever deeper for that to be possible (e.g. ./backend/lightly_studio and ./backend/lightly_studio_embed) - or the other way (which is perhaps cleaner) - bring the makefile and setup configs to the top level.

iunir: (Re-posting — I replied to this yesterday and the comment did not survive, apologies if you saw a notification for it.)

Agreed on the direction, and I have taken the cheap half here: the root Makefile now has `static-checks` and `type-check` alongside `test`, `lint`, `format` and `build`, each iterating a `PYTHON_MEMBERS` list. So there is one command per check for all Python code, and adding a sibling is one entry in that list.

What is still duplicated is each member's target *definitions* — the nine `uv run` one-liners here mirror the studio's. Hoisting those is the real fix and I would rather not do it in this PR: it rewrites `lightly_studio/Makefile`, which this stack deliberately left alone.

On fast-track you are right, and it is worth spelling out: `fast_track/src/guardrails/backend/shared.ts` hardcodes `BACKEND_PREFIX = 'lightly_studio/'` and `coverage.ts` builds `src/lightly_studio/` on top of it, so the backend coverage guardrail sees nothing under `lightly_studio_embed/`. Not fatal today — one source file, one test — but it silently under-reports as soon as `serve()` lands.

Between your two options I would take hoisting the config to the top level over nesting into `./backend/`: nesting moves `lightly_studio` itself and breaks every path in CI, the release workflows and the guardrails at once, whereas hoisting is additive and can be done per-tool. Happy to file a follow-up covering the shared Makefile plus making the guardrail prefixes a list rather than a constant — or say so if you would rather it block this PR.

.github/workflows/unit_test.yml
```diff
@@ -73,6 +74,13 @@ jobs:
             release_tooling:
               - '.github/scripts/**'
               - '.github/workflows/unit_test.yml'
+            # The second workspace member. Shares the lockfile, not the dependencies.
+            lightly_studio_embed:
+              - '.github/workflows/unit_test.yml'
+              - 'lightly_studio/ruff.toml'
+              - 'lightly_studio_embed/**'
+              - 'pyproject.toml'
+              - 'uv.lock'
```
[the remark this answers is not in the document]
iunir: Good catch, fixed. `'Makefile'` is now in both the `backend` and `lightly_studio_embed` filters, kept in sorted position.

Worth noting for anyone reading this later: no job in this workflow actually invokes the root Makefile — they all run `make` inside a member directory. So the hole was not that CI would run the wrong commands, but that the root Makefile now defines the aggregate Python checks (`static-checks`, `type-check`) and the pre-commit hook calls `make format` there, and a change to any of it would have landed with every gated job skipped and `unit-success` accepting that.

lightly_studio_embed/Makefile
```diff
@@ -0,0 +1,81 @@
...
+		--no-header --no-annotate --quiet --output-file build/wheel-requirements.txt
+	@awk -F'[=@ ]' -v self="$(PACKAGE_NAME)" \
+		'!/^#/ && NF && $$1 !~ /\// && tolower($$1) != self { print tolower($$1) }' \
+		build/wheel-requirements.txt > build/wheel-dependencies.txt
+	@for name in $(FORBIDDEN_WHEEL_DEPENDENCIES); do \
+		if grep -q "$$name" build/wheel-dependencies.txt; then \
+			echo "Error: the lightly-studio-embed wheel resolves '$$name'. Resolved tree:"; \
+			cat build/wheel-dependencies.txt; \
+			exit 1; \
+		fi; \
```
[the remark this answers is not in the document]
iunir: Fixed by moving the parsing to Python — `scripts/check_wheel_dependencies.py`. The target is now a single `uv run` line with no shell built-ins, so it behaves the same on every platform, and `clean` no longer needs a `build/` branch because `uv pip compile` reads the wheel path from stdin instead of an intermediate file.

It also closed a real gap rather than just a portability one. The shell version dropped any output line containing a slash, to skip the wheel's own path — which would equally have dropped a URL dependency such as `torch @ https://download.pytorch.org/whl/torch.whl`, precisely the kind of heavy pin this check exists to catch. The Python version looks for the slash only in the part before `@`, so that case now resolves to `torch` and fails the check.

Verified by hand: `torchvision` matches via `torch`, `nvidia-cublas-cu12` via `nvidia`, the package's own name is excluded (it contains `lightly-studio`), comments and the wheel path are ignored, and two wheels in `dist/` is a hard error rather than a resolve of both.

lightly_studio_embed/Makefile
iunir: Good idea, did it

make/python.mk
```diff
@@ -0,0 +1,37 @@
+# Targets shared by every Python workspace member. Each member includes this file rather than
+# copying the commands, so a new sibling package gets them for free and they cannot drift.
+#
+# Members add what is genuinely theirs on top: `build`, `test`, and anything package specific.
```
michal-lightly, line 4: Nice, I did not know that make targets can be shared like this.

.github/workflows/unit_test.yml
```diff
@@ -352,6 +364,63 @@ jobs:
...
+        env:
+          UV_PYTHON: ${{ matrix.python }}
+        run: make test
+
+      # The invariant the whole package exists for. Asserted on the built wheel, because that
+      # is what a customer installs onto an inference box next to their own torch pins.
+      - name: Assert the wheel pulls in nothing heavy
+        env:
+          UV_PYTHON: ${{ matrix.python }}
+        run: make check-wheel-dependencies WHEEL_PYTHON_VERSION=${{ matrix.python }}
```
michal-lightly: I think we don't need to have a check for this in CI

lightly_studio_embed/scripts/check_wheel_dependencies.py
```diff
@@ -0,0 +1,143 @@
...
+import argparse
+import re
+import subprocess
+import sys
+from pathlib import Path
+
+# Matched as substrings, so that variants such as duckdb-engine, torchvision or
+# opencv-python-headless are caught too. `cuda` and `nvidia` cover the CUDA wheels, which can
+# arrive without torch pulling them in.
+FORBIDDEN_DEPENDENCIES = ("lightly-studio", "torch", "duckdb", "opencv", "cuda", "nvidia")
```
michal-lightly: Let's remove this file. The test is ad-hoc; other packages can make the package heavy too. And it is unlikely we would not notice a problem.

michal-lightly: If you'd really want to keep some form of a test we can have a 5-line pytest test, checking that torch cannot be imported. But I still suggest to completely remove it.

lightly_studio_embed/tests/test_packaging.py
```diff
@@ -0,0 +1,27 @@
+"""Asserts the dependency floor of the published wheel.
```
michal-lightly: What is the value of this test? I would suggest to remove it too - it checks the declaration in pyproject.toml. It seems the AI overindexed on checking that the package is light.

michal-lightly (changes requested):

lightly_studio_embed/tests/test_packaging.py
iunir: deleted

lightly_studio_embed/Makefile
```diff
@@ -0,0 +1,22 @@
...
+	rm -rf dist
+endif
+
+# Cleans first, so that dist/ only ever holds the artifact this build produced.
+.PHONY: build
+build: clean
+	@echo "Building a distribution package..."
+	uv build --package lightly-studio-embed --out-dir dist
+
+# Exit code 5 is "no tests collected", which is the honest state until serve() adds code and
```
michal-lightly: Optional: Will this be removed when we add a real test? If yes, a TODO would be nice.

lightly_studio_embed/ruff.toml
```diff
@@ -0,0 +1,8 @@
+# One style for both members. When they ever need to diverge, copy what differs rather than
+# letting the two files drift apart silently.
```
michal-lightly: Please rewrite this such that no context about this PR is needed (especially the meaning of `both members`).

lightly_studio_embed/ruff.toml
```diff
@@ -0,0 +1,8 @@
...
+extend = "../lightly_studio/ruff.toml"
+
+# Relative patterns resolve against the directory of the file that declares them, so the
+# extended `tests/**/*.py` entry only covers lightly_studio's tests and has to be repeated.
```
michal-lightly: Same here, describe the end state, not the change.

michal-lightly (approved):

PR #2228, opened 2026-09-07
Persist the remote embedding backend API key

iunir:
## What has changed and why?

A remote embedding backend needs a bearer token, and the token had nowhere to live. This PR owns only the persistence layer.

- Nullable `api_key` column on `embedding_model`, next to `remote_embedder_url`, plus the Alembic migration.
- `embedding_model_resolver.set_api_key` to set, rotate or clear the key. Reading goes through the existing `get_by_id`.
- Both leak paths closed: `repr=False` keeps the key out of log lines that dump the row, and a test walks every OpenAPI schema so no route can return it.

Column instead of its own table because no route returns `EmbeddingModelTable` today. The cost is duplication per dataset, since `embedding_model` is unique on `(dataset_id, name)`; if backends get shared widely, a `remote_service` table is the better shape.

Not here: the `Authorization` header (LIG-10641), the GUI/Python surface (LIG-10740), encryption at rest.

## How has it been tested?

New tests for the resolver round trip, the key's absence from `repr`/`str`/a log line and from every OpenAPI schema, and the migration applying forward and back on Postgres.

```
cd lightly_studio
make static-checks               # pass
make test                        # 2430 passed, 44 skipped
make migration-check-postgresql  # No new upgrade operations detected
uv run pytest --postgres tests/database/test_db_migrations.py tests/models/test_embedding_model.py tests/resolvers/test_embedding_model_resolver.py  # 24 passed
```

- [ ] Yes

---

Suggested reviewer: @michal-lightly (wrote the surrounding embedding model code).

## Summary by CodeRabbit

* **New Features**
  * Added support for securely storing API keys for remote embedding models.
  * API keys can be set, replaced, or cleared for existing embedding models.
  * Existing local embedding models continue to work without an API key by default.

* **Security**
  * API keys are excluded from logs, string representations, and API response schemas.

* **Bug Fixes**
  * Added validation when attempting to update an unknown embedding model.

### DIFF
--- a/lightly_studio/src/lightly_studio/models/embedding_model.py
+++ b/lightly_studio/src/lightly_studio/models/embedding_model.py
@@ -10,12 +10,24 @@ from sqlmodel import Field, SQLModel
 
 
 class EmbeddingModelBase(SQLModel):
-    """Base class for the EmbeddingModel."""
+    """Base class for the EmbeddingModel.
+
+    Attributes:
+        name: The name identifying the embedding model within its dataset.
+        embedding_dimension: The dimension of the embeddings the model produces.
+        dataset_id: The dataset owning the embedding model.
+        remote_embedder_url: The base URL of the remote embedding backend, if the model is
+            served remotely.
+        api_key: The bearer token sent to `remote_embedder_url`. It is a secret: it must never
+            reach a response model, so a route returning an embedding model needs a view model
+            that omits it. `repr=False` keeps it out of log lines that dump the row.
+    """
 
     name: str
     embedding_dimension: int
     dataset_id: UUID = Field(foreign_key="dataset.dataset_id", index=True)
     remote_embedder_url: str | None = None
+    api_key: str | None = Field(default=None, repr=False)
 
 
 class EmbeddingModelCreate(EmbeddingModelBase):
--- a/lightly_studio/src/lightly_studio/models/embedding_model.py
+++ b/lightly_studio/src/lightly_studio/models/embedding_model.py
@@ -10,23 +10,13 @@ from sqlmodel import Field, SQLModel
 
 
 class EmbeddingModelBase(SQLModel):
-    """Base class for the EmbeddingModel.
-
-    Attributes:
-        name: The name identifying the embedding model within its dataset.
-        embedding_dimension: The dimension of the embeddings the model produces.
-        dataset_id: The dataset owning the embedding model.
-        remote_embedder_url: The base URL of the remote embedding backend, if the model is
-            served remotely.
-        api_key: The bearer token sent to `remote_embedder_url`. It is a secret: it must never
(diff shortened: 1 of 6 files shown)


lightly_studio/src/lightly_studio/models/embedding_model.py
```diff
@@ -10,12 +10,24 @@
...
+        api_key: The bearer token sent to `remote_embedder_url`. It is a secret: it must never
+            reach a response model, so a route returning an embedding model needs a view model
+            that omits it. `repr=False` keeps it out of log lines that dump the row.
+    """
 
     name: str
     embedding_dimension: int
     dataset_id: UUID = Field(foreign_key="dataset.dataset_id", index=True)
     remote_embedder_url: str | None = None
+    api_key: str | None = Field(default=None, repr=False)
```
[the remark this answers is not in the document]
iunir: Not acting on this. The mechanism is real, the vulnerability is not.

`_copy_embedding_models` does carry `api_key` verbatim, but there is no boundary to cross: `DatasetTable` has a single column, `dataset_id`, so there is no owner or tenant. Anyone who can call `POST /collections/{id}/deep-copy` can already read every dataset on the instance, and the key is not reachable through any route (the OpenAPI test in this PR enforces that). CWE-200 needs an unauthorized actor and there isn't one.

Verbatim copying is also the documented contract of the module: "New columns are copied verbatim automatically; a new *foreign-key* column must be added to its table's `overrides`". `api_key` is not an FK.

The suggested fix would make things worse: clearing `api_key` while `remote_embedder_url` is still copied leaves the duplicated dataset pointing at a backend it can no longer authenticate against, with no route to set the key again yet (LIG-10740 owns that). A deep copy is a clone of the same config for the same person on the same instance, so keeping the credential is the behavior that works.

The real residue is the duplication cost the ticket already accepted: the token lives in N rows and rotation touches all of them. That is an argument for the `remote_service` table once backend sharing becomes common, not for a one-line override here.

lightly_studio/tests/models/test_embedding_model.py
```diff
@@ -0,0 +1,120 @@
...
+        ),
+    )
+
+    assert created.api_key is None
+
+
+def test_embedding_model_table__api_key_is_nullable() -> None:
+    """The column must be nullable so existing rows need no backfill."""
+    columns = SQLModel.metadata.tables[EmbeddingModelTable.__tablename__].columns
+    assert columns["api_key"].nullable is True
```
michal-lightly: Optional: I'd consider removing this, resolver is tested separately, and it also tests nullability.

michal-lightly (approved):

lightly_studio/tests/models/test_embedding_model.py
iunir: Agreed, removed

PR #2251, opened 2026-09-09
Guard requests at the ASGI layer and add serve() (embed server stack 4/4)

iunir:
## What has changed and why?

- **`middleware.py`** checks the bearer token and the body size at the ASGI layer, ahead of
  routing. Neither can be a dependency: FastAPI reads the whole body before it solves
  dependencies, so both checks would run after an unknown peer had already decided how much the
  server buffers.
- The token is compared as **raw bytes**, so a byte that is not ASCII answers 401 instead of 500.
  A blank `api_key` is rejected.
- The body is **counted as it streams**, so a chunked request with no `Content-Length` is bounded
  too.
- **`serve()`** runs the application under uvicorn and can terminate TLS itself. On a non-loopback
  bind it warns when there is no key, or a key but no TLS to carry it, since the bearer token
  travels in the request. `ssl_keyfile` without `ssl_certfile` raises.
- The README gains the usage example, the package changelog the `serve` entry.

## How has it been tested?

New tests cover the token cases, the unauthenticated request whose body is never read, both 413s
including a chunked multipart body with no `Content-Length`, and the warnings, TLS forwarding and
`ValueError` of `serve()`. `make static-checks` and `make test` pass.

Smoke-tested against real `serve()` processes: every mounted route answers, 401 without the key,
404 for the unimplemented video route, 413 for a chunked 13 MB body, and HTTPS with a self-signed
certificate while plain HTTP is refused.

- [x] Yes — `lightly_studio_serve/CHANGELOG.md`, which the package keeps on its own cadence.

---

**Stack** ([#2260](https://github.com/lightly-ai/lightly-studio/stacks/2260)) — #2257, #2258,
#2259, then this PR.

## Summary by CodeRabbit

* **New Features**
  * Added a server entry point with configurable host, port, TLS certificates, and bearer-token authentication.
  * Added support for serving embedded services over HTTP or HTTPS.
  * Added configurable request-body size limits with appropriate errors for unauthorized or oversized requests.
* **Documentation**
  * Documented the public API, supported endpoints, HTTP usage, multipart uploads, TLS, authentication, capability handling, and error responses.


lightly_studio_serve/src/lightly_studio_serve/middleware.py
```diff
@@ -0,0 +1,124 @@
...
+    async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
+        """Check the declared length, then pass the request on and count the body."""
+        if scope["type"] != "http":
+            await self.app(scope, receive, send)
+            return
+        declared = _content_length(scope=scope)
+        if declared is not None and declared > self.max_request_bytes:
+            await _too_large(max_request_bytes=self.max_request_bytes)(scope, receive, send)
+            return
+        await self.app(scope, self._counted(receive=receive), send)
```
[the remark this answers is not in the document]
iunir: The behaviour is real, the exposure is not, and the suggested fix would create one.

`max_request_bytes` bounds what the server *holds*, and a route that never reads the body never
makes it hold anything. Under uvicorn an unread body accumulates only to `HIGH_WATER_LIMIT`
(64 KiB) before the transport is paused, and once the response is complete h11 discards further
`Data` events (`if self.conn.our_state is h11.DONE: continue`). Measured against a real `serve()`
process with `max_request_bytes=1024`: a client pushed 256 MB of chunked body at
`GET /v1/describe`, got its 200, and peak RSS moved 2.5 MB.

Buffering the whole body before delegation would make every request hold up to
`max_request_bytes` — 32 MiB by default — in memory before routing. That is the exposure this
middleware exists to prevent, and it defeats the streaming multipart path, where starlette spools
parts to disk for the same reason.

The declared-length check covers a body-less route that announces its size
(`test_create_app__declared_length_over_max_on_a_route_that_reads_no_body`). The receive-driven
count covers the route that reads a body
(`test_create_app__chunked_body_over_max_request_bytes`). What is left is a chunked body at a
route with no body semantics: 64 KiB of memory, after which the bytes are read and discarded.
That residue is bandwidth, which a body size limit does not address in any design.

Not changing this.

lightly_studio_serve/tests/test_server.py
```diff
@@ -207,6 +242,88 @@ def test_create_app__batch_over_max_batch_size() -> None:
...
+
+    assert client.get("/v1/describe").status_code == 401
+
+
+def test_create_app__wrong_bearer_token() -> None:
+    client = TestClient(create_app(embedder=FakeTextEmbedder(), api_key="secret"))
+
+    response = client.get("/v1/describe", headers={"Authorization": "Bearer wrong"})
+
+    assert response.status_code == 401
```
michal-lightly: Minor: It would be nice to use the constants like `status.HTTP_401_UNAUTHORIZED` from fastapi library for better test readability.

lightly_studio_serve/README.md
```diff
@@ -3,16 +3,70 @@
...
+        ...
+
+
+serve(
+    MyEmbedder(),
+    host="0.0.0.0",
+    port=8080,
+    api_key="the-key-you-paste-into-lightlystudio",
+    ssl_certfile="cert.pem",
+    ssl_keyfile="key.pem",
```
michal-lightly, line 56: Let's make it more clear that ssl is optional.

michal-lightly (approved):

lightly_studio_serve/README.md
iunir: Done

lightly_studio_serve/tests/test_server.py
iunir: Done

PR #2252, opened 2026-09-09
Add McapGroupSequence model

horatiualmasan:
## What has changed and why?

Adds the mcap_group_sequence table: an optional 1:1 specialization of sequence that marks a sequence as MCAP-backed and points at the recording holding its bag file.

## How has it been tested?

New tests

- [ ] Yes

## Summary by CodeRabbit

* **New Features**
  * Added support for creating and retrieving MCAP-backed sequence groups.
  * MCAP sequences can now link samples to recordings while preserving support for classic sequences.
  * Dataset duplication now copies MCAP recordings and sequence links with remapped identifiers.
  * Dataset deletion now removes associated recordings and MCAP sequence links.

* **Bug Fixes**
  * Added database support and migration for MCAP group sequence records.

* **Tests**
  * Added coverage for creation, retrieval, duplication, and deletion of MCAP-backed sequences.

### DIFF
--- a/lightly_studio/src/lightly_studio/models/mcap_group_sequence.py
+++ b/lightly_studio/src/lightly_studio/models/mcap_group_sequence.py
@@ -1,7 +1,8 @@
-"""MCAP specialisation of a sequence: the bag path lives here, not on ``sequence``.
+"""MCAP specialisation of a sequence: links to the recording holding the bag path.
 
-A sequence with this 1:1 row has an ``.mcap`` path. A sequence without it stays valid
-(classic / non-MCAP sequences). Identity is the parent sequence's ``sample_id``.
+A sequence with this 1:1 row references a ``recording`` (whose ``uri`` is the bag path). A
+sequence without it stays valid (classic / non-MCAP sequences). Identity is the parent
+sequence's ``sample_id``.
 """
 
 from uuid import UUID
@@ -9,15 +10,10 @@ from uuid import UUID
 from sqlmodel import Field, SQLModel
 
 
-class McapGroupSequenceBase(SQLModel):
-    """Shared fields for the MCAP group sequence specialisation."""
-
-    mcap_path: str
-    """Path or URI of the bag. Required, non-empty. Not unique."""
-
-
-class McapGroupSequenceTable(McapGroupSequenceBase, table=True):
-    """1:1 link from a sequence to its MCAP bag path."""
+class McapGroupSequenceTable(SQLModel, table=True):
+    """1:1 link from a sequence to its MCAP recording."""
 
     __tablename__ = "mcap_group_sequence"
     sample_id: UUID = Field(foreign_key="sequence.sample_id", primary_key=True)
+    recording_id: UUID = Field(foreign_key="recording.recording_id", index=True)
+    """FK to the recording holding the bag path. Required. Not unique."""
--- a/lightly_studio/src/lightly_studio/models/mcap_group_sequence.py
+++ b/lightly_studio/src/lightly_studio/models/mcap_group_sequence.py
@@ -1,0 +1,23 @@
+"""MCAP specialisation of a sequence: the bag path lives here, not on ``sequence``.
+
+A sequence with this 1:1 row has an ``.mcap`` path. A sequence without it stays valid
+(classic / non-MCAP sequences). Identity is the parent sequence's ``sample_id``.
+"""
+
+from uuid import UUID
+
+from sqlmodel import Field, SQLModel
+
+
+class McapGroupSequenceBase(SQLModel):
+    """Shared fields for the MCAP group sequence specialisation."""
+
+    mcap_path: str
+    """Path or URI of the bag. Required, non-empty. Not unique."""
+
+
+class McapGroupSequenceTable(McapGroupSequenceBase, table=True):
+    """1:1 link from a sequence to its MCAP bag path."""
+
+    __tablename__ = "mcap_group_sequence"
+    sample_id: UUID = Field(foreign_key="sequence.sample_id", primary_key=True)
--- a/lightly_studio/src/lightly_studio/resolvers/mcap_group_sequence_resolver/create.py
+++ b/lightly_studio/src/lightly_studio/resolvers/mcap_group_sequence_resolver/create.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+from urllib import parse
 from uuid import UUID
 
 from sqlmodel import Session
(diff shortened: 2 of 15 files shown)


lightly_studio/src/lightly_studio/models/mcap_group_sequence.py
```diff
@@ -0,0 +1,23 @@
...
+    """Shared fields for the MCAP group sequence specialisation."""
+
+    mcap_path: str
+    """Path or URI of the bag. Required, non-empty. Not unique."""
+
+
+class McapGroupSequenceTable(McapGroupSequenceBase, table=True):
+    """1:1 link from a sequence to its MCAP bag path."""
+
+    __tablename__ = "mcap_group_sequence"
```
ikondrat: I would avoid "sequence" name usage, it can cause confusion because we have built-in sequence data data in postgres [SEQUENCE](https://www.postgresql.org/docs/current/sql-createsequence.html)

lightly_studio/src/lightly_studio/models/mcap_group_sequence.py
```diff
@@ -0,0 +1,23 @@
...
+    sample_id: UUID = Field(foreign_key="sequence.sample_id", primary_key=True)
```
ikondrat: Can we stop using the same sample_id usage for different entities, if I understand correctly here it should 
be rather 

`sequence_id`

ikondrat (commented):
I think this PR currently places mcap_path on the wrong entity. 

McapGroupSequenceTable combines two independent concepts:
- a logical sequence, which defines playback order;
- a physical recording, which tells us where frame payloads are stored.

This makes resolving an MCAP frame unnecessarily indirect:
```
frame → group → sequence → McapGroupSequence.mcap_path
```
It also assumes that every sequence corresponds to exactly one MCAP file, which will make support for ROS bags, PCD sequences, split recordings, or multiple recording sources difficult.

I propose the following changes.

1. Introduce a format-independent recording entity
```
class RecordingTable(SQLModel, table=True):
    recording_id: UUID
    dataset_id: UUID
    format: str
    uri: str
```

Example:
```
recording_id = recording-1
dataset_id   = dataset-1
format       = "mcap"
uri          = "s3://bucket/drive.mcap"
```

Motivation: A recording is the physical source from which data is read. Making it independent of Sequence allows the same model to represent:
- an MCAP file;
- a ROS bag directory;
- a PCD file sequence;
- a LAS/LAZ file;
- another recording format added later.

Using uri instead of mcap_path also allows local paths, object-storage URIs, directories, and remote resources without encoding MCAP-specific semantics in the field name.

dataset_id gives the recording explicit ownership and supports routes such as:

`GET /datasets/{dataset_id}/recordings/{recording_id}`

2. Rename McapTable to McapFrameTable

The current model contains:
```
sample_id
channel_id
log_time_ns
capture_timestamp_ns
keyframe_log_time_ns
```

These fields identify a particular message or frame inside an MCAP recording. They do not describe an MCAP file.

Motivation: McapTable sounds like it represents the entire MCAP container. McapFrameTable makes it clear that each row is an annotatable frame/message locator.

3. Add a direct recording_id foreign key to every MCAP frame

```
class McapFrameTable(SQLModel, table=True):
    sample_id: UUID = Field(
        foreign_key="sample.sample_id",
        primary_key=True,
    )
    recording_id: UUID = Field(
        foreign_key="recording.recording_id",
        index=True,
    )

    channel_id: int
    log_time_ns: int
    occurrence: int = 0
    capture_timestamp_ns: int
    keyframe_log_time_ns: int | None = None
```

Motivation: The backend must be able to resolve the physical source directly from a frame:
`frame → recording → URI`

Group and sequence membership should not be required to read the frame. A frame may be read independently, moved between groups, or included in another sequence without changing its physical location.

This also reduces source resolution to one direct relationship instead of traversing three link tables.

4. Add occurrence to the MCAP locator

The complete MCAP locator becomes:
```
recording_id
channel_id
log_time_ns
occurrence
```

Motivation: Multiple messages on the same channel may have the same log_time_ns. The frontend provider already uses an occurrence value to distinguish these messages. Without storing it, a database row may not uniquely identify the intended message.

An alternative would be a direct message or chunk offset, but the persisted locator must uniquely identify one message.

5. Remove mcap_path from McapGroupSequenceTable
I would remove McapGroupSequenceTable from this PR unless it contains additional MCAP sequence-specific information beyond the path.

Motivation: Once every frame has a direct recording_id, the path can be resolved without sequence membership. Keeping it on McapGroupSequenceTable would create two possible sources of truth:

```
frame.recording_id → recording.uri
sequence → mcap_group_sequence.mcap_path
```

These references could eventually disagree.
It would also incorrectly restrict a logical sequence to a single MCAP source. A sequence may contain frames from multiple files or formats.

6. Keep Group and Sequence as logical entities

The existing logical relationships remain useful:

```
SampleGroupLink:
    component frame → synchronized group

SampleSequenceLink:
    group → sequence + seq_number

```

Motivation: These tables answer different questions:

```
Recording:
    Where are the bytes stored?

Group:
    Which camera, LiDAR, and radar frames belong to the same timestep?

Sequence:
    In what order should those timesteps be played?
```
A group can contain frames from different channels, recordings, and formats. The storage model therefore stays independent of synchronization.


7. Add an optional generic sequence-to-recording link only if required

If the product needs to explicitly record which sources participate in a sequence, it could later add:

```
class SequenceRecordingLinkTable(SQLModel, table=True):
    sequence_sample_id: UUID
    recording_id: UUID
```

Motivation: This relationship can help list or validate the recordings used by a sequence, but it should not be needed to locate an individual frame.

It should be many-to-many because:
- one recording can contain multiple logical sequences;
- one sequence can use multiple recordings.

8. Add a recording stream model in a follow-up PR
A future model could normalize channels across formats:

```
class RecordingStreamTable(SQLModel, table=True):
    stream_id: UUID
    recording_id: UUID

    source_key: str
    component_name: str
    modality: str
    schema: str | None
    encoding: str | None
```

Example:

```
source_key     = "5"                       # MCAP channel ID
component_name = "lidar"                   # stable product role
modality       = "point_cloud"
schema         = "sensor_msgs/PointCloud2"
```

For a ROS bag:
```
source_key     = "/lidar/points"
component_name = "lidar"
modality       = "point_cloud"
```

Motivation: Physical channel identifiers are format-specific and unstable between recordings. The product needs stable component roles such as lidar and front_camera.

I suggest keeping this out of PR #2252 unless it is immediately required, so this PR remains focused on recording ownership and source resolution.

Proposed final relationships
```
Dataset
└── Recording
    └── McapFrame
        ├── channel_id
        ├── log_time_ns
        └── occurrence

Sequence
└── Group
    ├── LiDAR frame  ──→ Recording A
    ├── Camera frame ──→ Recording A
    └── Radar frame  ──→ Recording B
```

This gives us two clear and independent paths:

```
Payload resolution:
frame → recording → URI

Playback:
sequence → ordered groups → component frames
```

API impact
This would support separate metadata and binary endpoints:

```
GET /datasets/{dataset_id}/recordings/{recording_id}
GET /datasets/{dataset_id}/recordings/{recording_id}/content
GET /datasets/{dataset_id}/sequences/{sequence_id}/groups?offset=0&limit=12
```


The sequence endpoint returns logical frame descriptors. The recording endpoint serves binary data, including MCAP byte-range requests.

Motivation: This separates the control plane from the data plane:
- sequence/group endpoints define what to display;
- recording endpoints provide the underlying bytes.

For MCAP, the browser can fetch indexed chunks with HTTP Range and decode multiple camera/LiDAR frames from the same chunk. It does not require one binary request per frame.

Suggested scope for this PR

I would keep PR #2252 focused on:
1. Adding RecordingTable.
2. Renaming McapTable to McapFrameTable.
3. Adding recording_id and occurrence to MCAP frames.
4. Removing mcap_path from the logical sequence model.

Recording streams, playback endpoints, and frontend integration can follow in separate PRs.

lightly_studio/src/lightly_studio/resolvers/mcap_group_sequence_resolver/create.py
```diff
@@ -0,0 +1,48 @@
...
+    Raises:
+        ValueError: If the collection does not exist or is not of type SEQUENCE, or if
+            the recording does not exist.
+    """
+    collection_resolver.check_collection_type(
+        session=session,
+        collection_id=collection_id,
+        expected_type=SampleType.SEQUENCE,
+    )
+    if session.get(RecordingTable, recording_id) is None:
```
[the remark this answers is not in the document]
horatiualmasan: Adding recording from different datasets does not usually happens, will not handle now.

ikondrat (approved):

PR #2257, opened 2026-09-09
Move the embedder interface into lightly-studio-embed (embed stack 1/4)

iunir:
## What has changed and why?

Part 1 of a 4-PR stack that splits #2240 (LIG-10691) into reviewable pieces. **A pure move: no
class, member, method or docstring is added, removed or reordered.**

`lightly-studio` and `lightly-studio-embed` each declared their own embedder hierarchy, kept
behaviourally identical by hand. This moves the shared half into the workspace member and imports
it back, so a model written against either package is the same type.

`lightly_studio/embed/embedder.py` and `lightly_studio/embed/types.py` are gone, with no re-export
shim left behind. GitHub renders both as renames:

- **`types.py` is byte-identical** — 0 lines changed.
- **`embedder.py` is +6/-2**, all of it the import block: `lightly_studio.embed.types` becomes
  `lightly_studio_embed.types`, and `PIL.Image` moves under `TYPE_CHECKING`.

Everything the server needs on top of this — `ready`, `VideoBytesEmbedder` — lands in #2259, where
the endpoints that use it are.

All 25 callers import straight from `lightly_studio_embed`. `lightly_studio` keeps exporting
`EmbeddingSpaceSpec` and `ImageCrop`, and now `EmbeddingResult` too, so the documented snippets
stay inside one namespace rather than naming the second package. That export was missing while the
other two were present, which is why the tutorials reached into `lightly_studio.embed.types`.

## Notes for the reviewer

- **The `TYPE_CHECKING` guard is the one deviation from a literal copy.** A plain
  `from PIL.Image import Image` would make Pillow an install dependency of every embedding server,
  including one that only embeds text. Under the guard it is a check-time dependency, pinned in
  the dev group.
- **`Capability` stays in `embedder.py`**, where it already was, rather than moving next to the
  other value types. #2258 imports it from there.
- **This costs the package a numpy dependency**, and the README and pyproject claims that it
  "depends on nothing but an HTTP server" are corrected. `EmbeddingResult.embeddings` has to stay
  an array: the ingest paths do `embeddings[kept_indices]`, `embeddings[:position]` and
  `.tolist()` at ~20 sites, and mypy rejects all three against a `Sequence[Sequence[float]]`.
- **The docs handler needed the member's `src` on its path.** griffe cannot load a class it has no
  source for, so without it the three supporting-type sections of `api/embeddings.md` render as an
  empty module heading — silently, since `mkdocs build --strict` still passes. They now document
  the types where they are defined. Verified by building the docs and diffing the rendered page
  against a `main` build: dataclass signatures, attribute docs and `Beta` badges all present.

## How has it been tested?

The move is covered by the existing suite, which passes unchanged: **2459 passed, 46 skipped**,
including `tests/embed/test_embedder_registry.py` against the moved base classes. `mypy` reports
no new errors — the ones that remain are the pre-existing missing-`cloud-storage`-extras errors
also present on `main`, verified by diffing both error sets against a clean `main` worktree.
`make static-checks` passes in both packages, and `mkdocs build --strict` is clean.

- [ ] Yes

---

1. #2257 — move the embedder interface into the workspace member
2. #2258 — the protocol wire models and the result checks
3. #2259 — `create_app` and the embed endpoints


lightly_studio/pyproject.toml
```diff
@@ -61,8 +61,15 @@ dependencies = [
     # Pinned below 7.0, which dropped Python 3.9. 6.9 covers every Python this package supports,
     # so one major runs everywhere rather than two that have to stay behaviourally identical.
     "posthog>=6.9,<7.0",
+    # The embedder base classes and the value types they pass live here, so a customer's own
+    # server and LightlyStudio's in-process embedders derive from one definition. Resolved
+    # from the workspace, so the two cannot drift apart between releases.
+    "lightly-studio-embed",
```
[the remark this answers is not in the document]
michal-lightly: I think the comment is valid, we should say `>=0.1.0` (if it works in this local setup, before we publish to PyPI).

Optional: Please change the comment, it is muddled. The first sentence is not that useful. The note that it uses local version for local development is useful.

lightly_studio_embed/src/lightly_studio_embed/types.py
```diff
@@ -1,19 +1,67 @@
...
+    """Interactive and ingest: image as raw file bytes."""
+
+    VIDEO_BYTES = "video_bytes"
+    """Ingest: video as raw file bytes. No local implementation yet."""
+
+    IMAGE_URL = "image_url"
+    """Ingest: image the server fetches from a presigned URL. No local implementation."""
+
+    VIDEO_URL = "video_url"
+    """Ingest: video the server fetches from a presigned URL. No local implementation."""
```
michal-lightly: We agreed not to include these two I believe. I would suggest to add them once we add signed url support, together with corresponding protocols.

I think they might be needed but I'm not 100% sure - because fsspec does support urls, it could be used as is today.

lightly_studio_embed/src/lightly_studio_embed/embedder.py
```diff
@@ -1,50 +1,27 @@
...
-bytes. A concrete model implements only the capabilities it supports, and
-callers pick an embedder by the capability they need.
+can embed: images and crops by path, videos, PIL images, text, and images and
+videos by bytes. A concrete model implements only the capabilities it supports,
+and callers pick an embedder by the capability they need.
+
+LightlyStudio's own embedders and a customer's embedding server derive from the
+same classes here, so a model written against either is the same type. The path
+and PIL capabilities are in-process only: an fsspec path is resolved with
+LightlyStudio's credentials, and a PIL image never crosses a network.
```
michal-lightly: The AI comments in this PR (not just this one) are cryptic, they are difficult to read. Can you ask it to rewrite it with "ASD-STE100 Simplified Technical English"?

lightly_studio_embed/pyproject.toml
```diff
@@ -30,6 +34,9 @@ dependencies = [
 # The same floors as the repository's other Python packages, so the tooling behaves alike.
 dev = [
     "mypy>=1.19.1",
+    # Types `ImagePILEmbedder`, which is in-process only, so Pillow is a check-time
+    # dependency rather than an install one. Same floor as `lightly-studio` resolves.
+    "pillow>=11.0.0",
```
michal-lightly: I think this is a premature optimisation. It also forces us to use `if TYPE_CHECKING` which has disadvantages.

Let's make it a hard dependency. If that becomes a problem I'd rather chose a different solution.

michal-lightly (changes requested):

lightly_studio_embed/src/lightly_studio_embed/embedder.py
iunir: Done

lightly_studio/docs/docs/api/embeddings.md
```diff
@@ -47,18 +47,18 @@ See the [Embeddings page](../concepts_and_tools/embeddings.md) for more details.
 
 ### EmbeddingSpaceSpec
 
-::: lightly_studio.dataset.embedding_generator
+::: lightly_studio_embed.types
```
michal-lightly, line 50: Make sure these docs changes are merged after we rerelease docs.

michal-lightly (approved):

PR #2258, opened 2026-09-09
Define the embedding protocol wire models and the result checks (embed stack 2/4)

iunir:
## What has changed and why?

Part 2 of a 4-PR stack that splits #2240 (LIG-10691) into reviewable pieces. Stacked on #2257,
which has landed.

The shared definition of the contract, version 1. Still no server: this is what both halves
validate against.

- **`protocol.py`** — the request and response bodies, every route path and the multipart field
  the bytes endpoints read. `RemoteEmbedder` (LIG-10641) validates against these same models on
  the client side, so the two halves cannot drift apart.
- **`errors.py`** — `EmbedderContractError`, for a result the protocol does not allow.
- **`validation.py`** — the two things a response body cannot show about itself: that the embedder
  returned a matrix of numbers at all, and that each kept index names an item of *this* request. A
  broken embedder then fails in the customer's own process rather than as a similarity search
  misbehaving inside LightlyStudio.

Most of the line count is declarations and docstrings.

### One rule, one home

`EmbeddingsResponse` owns every rule that a body does show: one row per kept index, indices that
ascend and never go negative, the declared vector length, and values a `float32` can hold. That is
where the rules have to live, because a client reads this model from a server we did not write.
`validation.py` therefore does not repeat any of them — it builds the model and turns a
`ValidationError` into an `EmbedderContractError`, so the server still answers 500 and names the
embedder in its own process.

Three of those rules are new, and each one was reachable:

- **A negative index.** `kept_indices=[-1]` validated cleanly and would have selected the *last*
  item of the request on any `items[index]` lookup.
- **A boolean mask for `kept_indices`.** A lax `int` field reads `true` as `1`. An embedder that
  derives `kept_indices` from a numpy comparison, such as `scores > 0.5`, holds `np.bool_` values,
  and the field read those as `0` and `1`. Every vector then lined up with the wrong item. A
  `mode="before"` validator rejects a bool ahead of the coercion, which is the only point where the
  bool is still visible. `dimension` and the `ServerLimits` fields keep the lax conversion on
  purpose: a bool there reads as 1, and the next check fails and names the problem, so a guard buys
  nothing. Strict mode would reject the bool, but it would also reject the `np.int64` that an
  embedder reports as its dimension.
- **A value outside `float32`.** Both ends store the vectors as `float32`, so `1e300` reached the
  database as an infinity. Reading the array as `float64` never caught that, despite the comment
  which said it did.

`dimension` must now be positive on both responses. `dimension=0` cleared every other check and
put zero-width vectors on the wire.

`_as_indices` and `_as_array` are gone, and with them the transformation they did. `validation.py`
checks the type instead of converting it: a ragged or non-numeric result is not a numeric array,
which is now the error it gets. The wire model handles `np.int64` and rejects a float index on the
way in, so nothing is lost.

### Capabilities

**The protocol no longer carries its own capability enum.** `WireCapability` duplicated three of
`Capability`'s strings, so `DescribeResponse` now uses `Capability` itself (moved in #2257) and
`WIRE_CAPABILITIES` names the kinds that can cross the wire: every one except `IMAGE_PIL`, which is
an object in this process and nothing else — LightlyStudio would have to encode it to send it, and
`IMAGE_BYTES` is that transport already. An fsspec path *does* cross the wire, so a GPU host that
holds its own credentials can serve the ingest kinds once an endpoint mounts them. A validator
rejects `IMAGE_PIL`, so a server advertising it is refused rather than trusted. That check matters
most on the client side, where the response comes from a server we did not write.

`WireCapability`'s two URL kinds are dropped rather than merged in, along with `EmbedUrlsRequest`
and the two URL paths: nothing implements them, no capability names them, and `RemoteEmbedder` will
not call them. They come back with signed URL support and its own protocol.

**`CapabilityNotImplementedError` is gone.** A capability an embedder does not implement is an
interface it does not subclass, so the error had no case left to cover. The 501 path in stack 3
goes with it.

`__init__.py` re-exports only the embedder classes and the value types. `errors` and `protocol` are
imported from their submodules, so the top level stays a small surface rather than one bag of
symbols.

## Notes for the reviewer

- Dropping the enum costs the wire field its type-level guarantee, since `Capability` also names
  `IMAGE_PIL`. The validator buys it back at runtime, and covers a server written in another
  language too, which the type never did.
- `DEFAULT_MAX_BATCH_SIZE` is 1024, with a TODO on the design problem it surfaces: one ceiling for
  every endpoint is too coarse, and a batch of text queries does not belong under the same number
  as a batch of videos.
- `protocol_version` accepts any string on purpose. A client that can parse `/v1/describe` can
  report the version it met; a rejected body would only say the response did not parse. The field
  docstring says who compares it.
- `validation.py` keeps one explicit `ndim` check. A flat array of the right size lines no row up
  with a kept index, and `.tolist()` would only reach the wire model as a confusing type error.
- The package now has tests, so the Makefile drops its tolerance of pytest's "no tests collected"
  exit code.

## How has it been tested?

37 tests. `make static-checks` and `make test` pass on 3.9 and 3.14.

Nineteen over `protocol.py`: five on `DescribeResponse` (a normal body, all wire capabilities
accepted, `IMAGE_PIL` rejected, a zero dimension rejected, an `np.int64` dimension kept, an unknown
`protocol_version` kept), one on `WIRE_CAPABILITIES` itself, and twelve on `EmbeddingsResponse` —
the row-count mismatch, non-ascending indices, a negative index, a JSON boolean mask, a numpy
boolean mask, a float index, the wrong width, a zero dimension, a non-finite value, a value too
large for `float32`, and a value exactly at the `float32` limit that must pass.

Eighteen over `build_embeddings_response`: the happy path, an embedder that kept nothing, an empty
array that still has a row, indices out of range, a negative index, non-ascending indices, the
row-count mismatch, the wrong width, a non-matrix array, a non-finite value, a value too large for
`float32`, a float index, a Python and a numpy boolean mask, `np.int64` indices that must pass,
ragged rows, values that are not numbers, and a `None` among the values.

- [ ] Yes

---

1. ~~#2257 — move the embedder interface into the workspace member~~ (merged)
2. #2258 — the protocol wire models and the result checks
3. #2259 — `create_app` and the embed endpoints

### DIFF
--- a/lightly_studio_embed/src/lightly_studio_embed/errors.py
+++ b/lightly_studio_embed/src/lightly_studio_embed/errors.py
@@ -1,0 +1,11 @@
+"""The errors that an embedder can raise or cause, in one module."""
+
+from __future__ import annotations
+
+
+class EmbedderContractError(RuntimeError):
+    """The embedder returned a result that the protocol does not allow.
+
+    The server answers 500 and names the problem. It does not send the result to
+    LightlyStudio.
+    """
--- a/lightly_studio_embed/src/lightly_studio_embed/protocol.py
+++ b/lightly_studio_embed/src/lightly_studio_embed/protocol.py
@@ -1,0 +1,199 @@
+"""Wire models of the LightlyStudio embedding protocol, version 1.
+
+The module holds the request bodies, the response bodies, the route paths and the
+multipart field name. The server in this package sends these models.
+``lightly-studio`` reads the same models. One definition keeps the two halves equal.
+
+A server serves the capability ``{subject}_{transport}`` at
+``POST /v1/embed/{subjects}/{transport}``. The capability ``text`` has no transport
+segment. Text is already a payload.
+
+``EmbeddingsResponse`` holds every rule that a response body shows.
+``lightly_studio_embed.validation`` adds only the rules that the body cannot show.
+"""
+
+from __future__ import annotations
+
+import math
+from typing import Annotated, Any
+
+import numpy as np
+from pydantic import BaseModel, Field, field_validator, model_validator
+
+from lightly_studio_embed.embedder import Capability
+
+PROTOCOL_VERSION = "1.0"
+
+BASE_PATH = "/v1"
+
+DESCRIBE_PATH = f"{BASE_PATH}/describe"
+
+EMBED_TEXTS_PATH = f"{BASE_PATH}/embed/texts"
+
+EMBED_IMAGES_BYTES_PATH = f"{BASE_PATH}/embed/images/bytes"
+
+EMBED_VIDEOS_BYTES_PATH = f"{BASE_PATH}/embed/videos/bytes"
+
+# The multipart field that the bytes endpoints read. One part per item, in input order.
+FILES_FIELD_NAME = "files"
+
+# A number, because starlette renamed its constant for this status.
+STATUS_PAYLOAD_TOO_LARGE = 413
+
+# TODO(Iunir, 09/2026): Report a limit per capability. Text queries and videos do not
+# belong under one ceiling.
+DEFAULT_MAX_BATCH_SIZE = 1024
+
+DEFAULT_MAX_REQUEST_BYTES = 32 * 1024 * 1024
+
+# Both ends store the vectors as `float32`. A larger value becomes an infinity there.
+MAX_ABS_EMBEDDING_VALUE = float(np.finfo(np.float32).max)
+
+# `IMAGE_PIL` is an object in memory, so it cannot cross the wire. An fsspec path can.
+WIRE_CAPABILITIES = frozenset(Capability) - {Capability.IMAGE_PIL}
+
+# A kept index names a position in the request, so it is never negative.
+_KeptIndex = Annotated[int, Field(ge=0)]
+
+
+class ServerLimits(BaseModel):
+    """The limits that the server applies and reports. A client does not have to guess them."""
+
+    max_batch_size: int = Field(default=DEFAULT_MAX_BATCH_SIZE, gt=0)
+    """The largest number of items in one request."""
+
+    max_request_bytes: int = Field(default=DEFAULT_MAX_REQUEST_BYTES, gt=0)
+    """The largest request body that the server accepts, in bytes."""
+
+
+class DescribeResponse(BaseModel):
+    """The body of ``GET /v1/describe``. Every server has this endpoint."""
+
+    protocol_version: str = PROTOCOL_VERSION
+    """The version of this contract that the server speaks.
+
+    A client compares this against its own ``PROTOCOL_VERSION`` first. The model accepts
+    any string, so a client can name the version that it met.
+    """
+
+    space_key: str
+    """The identifier of the embedding space that the server produces."""
+
+    dimension: int = Field(gt=0)
+    """The length of every embedding vector that the server returns."""
+
+    # TODO(Iunir, 09/2026): A ServerStatus enum may replace this flag. A model that
+    # failed to load is not the same as a model that is still loading.
+    ready: bool
+    """Whether the model is loaded. ``False`` means try again soon. It is not an error."""
+
+    capabilities: list[Capability]
+    """The input kinds that this server accepts. These are the endpoints that it mounts.
+
+    The model permits only the members of ``WIRE_CAPABILITIES``. LightlyStudio cannot
+    request the other kinds.
+    """
+
+    limits: ServerLimits
+    """The limits that a client must apply to its batches."""
+
+    @field_validator("capabilities")
+    @classmethod
+    def _reject_unservable_capabilities(cls, value: list[Capability]) -> list[Capability]:
+        unservable = [capability for capability in value if capability not in WIRE_CAPABILITIES]
+        if unservable:
+            raise ValueError(
+                f"{unservable} cannot be served over HTTP; only {sorted(WIRE_CAPABILITIES)} can."
+            )
+        return value
+
+
+class EmbedTextsRequest(BaseModel):
+    """The body of ``POST /v1/embed/texts``."""
+
+    texts: list[str]
+    """The strings to embed. The server returns the embeddings in this order."""
+
+
+class EmbeddingsResponse(BaseModel):
+    """The body of every embed endpoint.
+
+    Each response repeats ``space_key`` and ``dimension``. A client reads them to find a
+    server that changed its weights and kept its ``space_key``.
+    """
+
+    space_key: str
+    """The identifier of the embedding space of these vectors."""
+
+    dimension: int = Field(gt=0)
+    """The length of every vector in ``embeddings``."""
+
+    kept_indices: list[_KeptIndex]
+    """The indices of the request items that the server embedded, in ascending order.
+
+    The server omits an item that it cannot decode. It does not fail the batch.
+    """
+
+    embeddings: list[list[float]]
+    """One row for each entry in ``kept_indices``, in the same order."""
+
+    @field_validator("kept_indices", mode="before")
+    @classmethod
+    def _reject_boolean_mask(cls, value: Any) -> Any:
+        """Reject a mask before the field turns it into indices.
+
+        An embedder that derives ``kept_indices`` from a numpy comparison, such as
+        ``scores > 0.5``, holds ``np.bool_`` values. A lax ``int`` field reads those as
+        ``0`` and ``1``. Every vector then lines up with the wrong item, and the database
+        keeps the wrong vector for the sample. The check runs before the conversion,
+        which is the only point where the bool is still visible.
+        """
+        if not isinstance(value, (list, tuple, np.ndarray)):
+            return value
+        for item in value:
+            if isinstance(item, (bool, np.bool_)):
+                raise ValueError(f"kept_indices holds {item!r}, a bool, not an index.")
+        return value
+
+    @model_validator(mode="after")
+    def _check_rows_against_indices(self) -> EmbeddingsResponse:  # noqa: N804
+        """Check the rules that hold between the fields.
+
+        A field type does not show that a response is correct. A client reads this model
+        from a server that it does not control. Without these rules a client aligns the
+        wrong vector with an item.
+        """
+        if len(self.embeddings) != len(self.kept_indices):
+            raise ValueError(
+                f"Got {len(self.embeddings)} embeddings for {len(self.kept_indices)} kept "
+                f"indices. The two must have the same length."
+            )
+        if any(
+            current >= following
+            for current, following in zip(self.kept_indices, self.kept_indices[1:])
+        ):
+            raise ValueError(f"kept_indices must be ascending and unique, got {self.kept_indices}.")
+        for index, row in enumerate(self.embeddings):
+            if len(row) != self.dimension:
+                raise ValueError(
+                    f"Embedding {index} has {len(row)} values. The server declares "
+                    f"dimension {self.dimension}."
+                )
+            _check_values(row=row, index=index)
+        return self
+
+
+def _check_values(row: list[float], index: int) -> None:
+    """Check one vector. A value outside ``float32`` is as wrong as a NaN.
+
+    Both ends store the vectors as ``float32``. A larger value becomes an infinity there,
+    and a similarity search then ranks it.
+    """
+    for value in row:
+        if not math.isfinite(value):
+            raise ValueError(f"Embedding {index} holds a value that is not finite.")
+        if abs(value) > MAX_ABS_EMBEDDING_VALUE:
+            raise ValueError(
+                f"Embedding {index} holds {value}, over the {MAX_ABS_EMBEDDING_VALUE} that "
+                f"the float32 of an embedding can hold. It would become an infinity."
+            )
--- a/lightly_studio_embed/src/lightly_studio_embed/validation.py
+++ b/lightly_studio_embed/src/lightly_studio_embed/validation.py
@@ -1,0 +1,108 @@
+"""Checks the result of an embedder before the server sends it.
+
+A broken result is cheap to find here, in the process of the customer. It is expensive to
+find later, as a similarity search that gives bad results in LightlyStudio.
+
+``EmbeddingsResponse`` holds every rule that a response body shows. This module does not
+repeat them. It checks the two rules that the body cannot show: the embedder returned a
+matrix of numbers, and each kept index names an item of this request.
+"""
+
+from __future__ import annotations
+
+import numpy as np
+from numpy.typing import NDArray
+from pydantic import ValidationError
+
+from lightly_studio_embed.errors import EmbedderContractError
+from lightly_studio_embed.protocol import EmbeddingsResponse
+from lightly_studio_embed.types import EmbeddingResult
+
+# Float, signed integer and unsigned integer. A `bool` array is a mask, and an `object`
+# array holds anything.
+_NUMERIC_DTYPE_KINDS = frozenset("fiu")
+
+
+def build_embeddings_response(
+    result: EmbeddingResult, space_key: str, dimension: int, item_count: int
+) -> EmbeddingsResponse:
+    """Check the result of an embedder and convert it into a wire response.
+
+    Args:
+        result: What the embedder returned.
+        space_key: The embedding space of the embedder. The response repeats it.
+        dimension: The vector length that the embedder declares.
+        item_count: The number of items in the request.
+
+    Returns:
+        The response body for an embed endpoint.
+
+    Raises:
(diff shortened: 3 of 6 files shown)


lightly_studio_embed/src/lightly_studio_embed/__init__.py
```diff
@@ -10,18 +10,37 @@
     TextEmbedder,
     VideoPathEmbedder,
 )
+from lightly_studio_embed.errors import CapabilityNotImplementedError, EmbedderContractError
```
michal-lightly: I'd suggest not to reexport these (errors, also protocol classes). For the intended use-case, users would need just the Embedder classes and the `serve` method. If they need anything else they can import it from a submodule - it is then nicely groupped instead on one bag of symbols.

lightly_studio_embed/src/lightly_studio_embed/errors.py
```diff
@@ -0,0 +1,20 @@
...
+
+class EmbedderContractError(RuntimeError):
+    """Raised when an embedder returns a result the protocol does not allow.
+
+    The server answers 500 and names the problem, rather than shipping the result to
+    LightlyStudio.
+    """
+
+
+class CapabilityNotImplementedError(NotImplementedError):
```
michal-lightly: This class should not exist. If the capability is not implemented the users should not subclass the corresponding capability interface.

lightly_studio_embed/src/lightly_studio_embed/protocol.py
```diff
@@ -0,0 +1,151 @@
...
+EMBED_TEXTS_PATH = f"{BASE_PATH}/embed/texts"
+
+EMBED_IMAGES_BYTES_PATH = f"{BASE_PATH}/embed/images/bytes"
+
+EMBED_VIDEOS_BYTES_PATH = f"{BASE_PATH}/embed/videos/bytes"
+
+# Specified for a server that implements the URL transports; `serve` mounts neither.
+EMBED_IMAGES_URLS_PATH = f"{BASE_PATH}/embed/images/urls"
+
+EMBED_VIDEOS_URLS_PATH = f"{BASE_PATH}/embed/videos/urls"
```
michal-lightly: Let's not introduce this yet. Or let's discuss how this should interact with IMAGE_PATH functionality.

lightly_studio_embed/src/lightly_studio_embed/protocol.py
```diff
@@ -0,0 +1,151 @@
...
+
+# Multipart form field the bytes endpoints read, one part per item in input order.
+FILES_FIELD_NAME = "files"
+
+# Spelled out, because starlette renamed its constant for this status and the old name warns
+# on new versions while the new one is missing on the versions `requires-python` still allows.
+STATUS_PAYLOAD_TOO_LARGE = 413
+
+DEFAULT_MAX_BATCH_SIZE = 64
```
michal-lightly: Let's make this bigger, perhaps 1024?

This surfaces a design problem - different endpoints might have different batch size limits. Let's add a todo, we can address it later.

lightly_studio_embed/src/lightly_studio_embed/protocol.py
```diff
@@ -0,0 +1,151 @@
...
+
+DEFAULT_MAX_REQUEST_BYTES = 32 * 1024 * 1024
+
+
+# The capabilities an embedding server can accept over HTTP. The rest of ``Capability``
+# is in-process only: an fsspec path is resolved with LightlyStudio's credentials and a
+# PIL image is an object in memory, so neither can cross the wire.
```
michal-lightly: We should allow all (except maybe IMAGE_PIL). If the user wants to use a remote embedding server for data ingest - perhaps they have a cloud machine with a strong GPU and s3 access - then they can use the endpoint for IMAGE_PATHS.

We can optionally also add it later.

lightly_studio_embed/src/lightly_studio_embed/validation.py
```diff
@@ -0,0 +1,124 @@
...
+from lightly_studio_embed.errors import EmbedderContractError
+from lightly_studio_embed.protocol import EmbeddingsResponse
+from lightly_studio_embed.types import EmbeddingResult
+
+# Embeddings are one row per kept input, so anything but a matrix is a broken result.
+_EMBEDDINGS_NDIM = 2
+
+
+def build_embeddings_response(
+    *, result: EmbeddingResult, space_key: str, dimension: int, item_count: int
```
michal-lightly: Style: Do not include `*`, we call by keyword args by convention. Also elsewhere in this PR.

lightly_studio_embed/src/lightly_studio_embed/validation.py
```diff
@@ -0,0 +1,124 @@
+"""Checks an embedder's result against the contract before it leaves the server.
```
michal-lightly: Internal note: I have reviewed until here, will continue later.

michal-lightly (changes requested):

lightly_studio_embed/src/lightly_studio_embed/validation.py
```diff
@@ -0,0 +1,127 @@
+"""Checks the result of an embedder against the contract before the server sends it.
+
+A wrong dimension or a NaN is cheap to find here, in the process of the customer. It is
+expensive to find later, as a similarity search that gives bad results in LightlyStudio.
```
michal-lightly: This comment should explain more. We use mypy types, so the validation functions do not make sense - we do not need to check that an int variable is int.

However, this context can run on customer infrastructure where mypy types are not enforced. So it is nice that we double-check the shape.

lightly_studio_embed/src/lightly_studio_embed/validation.py
```diff
@@ -0,0 +1,127 @@
...
+
+    Returns:
+        The response body for an embed endpoint.
+
+    Raises:
+        EmbedderContractError: If the result does not agree with the declared dimension,
+            with the number of kept indices, or with the items of the request.
+    """
+    kept_indices = _as_indices(kept_indices=result.kept_indices)
+    embeddings = _as_array(embeddings=result.embeddings)
```
michal-lightly: Let's remove `_as_indices` and `_as_array`.

If we want to double-check the type we should do just checks, no data transformation.

lightly_studio_embed/src/lightly_studio_embed/validation.py
```diff
@@ -0,0 +1,127 @@
...
+
+def _validate_embeddings(
+    *, embeddings: NDArray[np.float64], kept_count: int, dimension: int
+) -> None:
+    # An embedder that kept no input can return an empty array of any width. The array must
+    # still have no rows. `np.empty((1, 0))` is also empty, but it would put one row on the
+    # wire against no index.
+    if kept_count == 0 and embeddings.ndim > 0 and embeddings.shape[0] == 0:
+        return
+    if embeddings.ndim != _EMBEDDINGS_NDIM:
```
michal-lightly: Optional: We can use just the constant 2 here

lightly_studio_embed/src/lightly_studio_embed/protocol.py
```diff
@@ -0,0 +1,160 @@
...
+    """The indices of the request items that the server embedded, in ascending order.
+
+    The server omits an item that it cannot decode. It does not fail the batch.
+    """
+
+    embeddings: list[list[float]]
+    """One row for each entry in ``kept_indices``, in the same order."""
+
+    @model_validator(mode="after")
+    def _check_rows_against_indices(self) -> EmbeddingsResponse:  # noqa: N804
```
michal-lightly, line 159: These checks are duplicate with validation.py. Let's have them only at one place.

lightly_studio_embed/tests/test_protocol.py
```diff
@@ -0,0 +1,93 @@
...
+    WIRE_CAPABILITIES,
+    DescribeResponse,
+    EmbeddingsResponse,
+    ServerLimits,
+)
+
+SPACE_KEY = "acme/model@v1"
+
+
+def _describe(*, capabilities: list[Capability]) -> DescribeResponse:
```
michal-lightly: Style: We list private functions at the bottom by convention.

michal-lightly (commented):
Adding comments for validation. If you want, validation can be added in a separate PR.

lightly_studio_embed/src/lightly_studio_embed/protocol.py
```diff
@@ -0,0 +1,160 @@
...
+    server that changed its weights but kept its ``space_key``.
+    """
+
+    space_key: str
+    """The identifier of the embedding space of these vectors."""
+
+    dimension: int
+    """The length of every vector in ``embeddings``."""
+
+    kept_indices: list[int]
```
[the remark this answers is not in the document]
iunir: Both parts confirmed and fixed in `bcb9d09..5cf83de`.

- **Negative indices**: `kept_indices` is now `list[_KeptIndex]` where `_KeptIndex = Annotated[int, Field(ge=0)]`. The bound sits on the item rather than in the validator so it also reaches the generated OpenAPI schema. Verified before the fix that `kept_indices=[-1]` validated cleanly, and that `[-2, -1]` did too, since the validator only tested ordering.
- **Boolean entries**: a `mode="before"` field validator now rejects them ahead of the coercion, `np.bool_` included, since an embedder in-process can produce one and `np.bool_` is not a `bool`.

Tests added on `EmbeddingsResponse` directly: `test_embeddings_response__negative_kept_index`, `test_embeddings_response__kept_indices_are_a_json_boolean_mask` (via `model_validate` with raw JSON `false`/`true`) and `test_embeddings_response__kept_indices_are_a_numpy_boolean_mask`.

Two related gaps came out of the same review and are fixed alongside: `dimension` now carries `gt=0` on both `EmbeddingsResponse` and `DescribeResponse`, and a value outside the `float32` range is rejected rather than reaching storage as an infinity. `validation.py` no longer repeats any of these rules — it builds the model and converts the `ValidationError`, so each rule has one home.

lightly_studio_embed/src/lightly_studio_embed/protocol.py
```diff
@@ -0,0 +1,211 @@
...
+_KeptIndex = Annotated[int, Field(ge=0)]
+
+
+class ServerLimits(BaseModel):
+    """The limits that the server applies and reports. A client does not have to guess them."""
+
+    max_batch_size: int = Field(default=DEFAULT_MAX_BATCH_SIZE, gt=0)
+    """The largest number of items in one request."""
+
+    max_request_bytes: int = Field(default=DEFAULT_MAX_REQUEST_BYTES, gt=0)
```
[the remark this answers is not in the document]
iunir: Confirmed on `5cf83de` and fixed in `f0bd3aa`. All four fields did accept a JSON boolean:

```
DescribeResponse.model_validate({... 'dimension': True, 'limits': {'max_batch_size': True, 'max_request_bytes': True}})
-> dimension=1 max_batch_size=1 max_request_bytes=1
EmbeddingsResponse.model_validate({... 'dimension': True ...})  -> dimension=1
```

I did not use `strict=True`, because it rejects more than booleans. Verified:

```
class Strict(BaseModel):
    dimension: int = Field(gt=0, strict=True)
Strict(dimension=np.int64(512))
-> ValidationError: Input should be a valid integer [type=int_type, input_value=512]
```

An embedder reports its dimension from the model it wraps (`dimension=self._model.output_dim`), and a numpy integer there is entirely reasonable. Strict mode would refuse that embedder at `/v1/describe` — a regression worse than the bug, on the public SDK surface.

Instead there is now one shared annotated type carrying a `BeforeValidator`, applied to all four fields:

```python
_WireInt = Annotated[int, BeforeValidator(_reject_bool)]
_KeptIndex = Annotated[_WireInt, Field(ge=0)]
```

The before-validator sees the raw value, which is the only point where the boolean is still distinguishable, and it also catches `np.bool_` — not a `bool`, and producible by an embedder in-process. `np.int64` still converts normally.

This also subsumes the bespoke `kept_indices` mask validator from the previous round: the per-item rule rejects `[false, true]` at `kept_indices.0`, and a raw `np.ndarray` of bools as well.

Four tests added: `test_describe_response__rejects_a_boolean_dimension`, `test_server_limits__rejects_boolean_limits` (both limits), `test_embeddings_response__rejects_a_boolean_dimension`, and `test_describe_response__keeps_a_numpy_integer_dimension` to pin the case strict mode would have broken.

iunir:
@coderabbitai Your review comments are hard to read. Please rewrite every finding on this PR in ASD-STE100 Simplified Technical English, and use that style for all future reviews and replies in this repository.

Apply these rules:

**Sentences**
- One idea per sentence. Maximum 20 words.
- Active voice only. Name the thing that does the action.
- Present tense. Say what is true now, not what "would have been".
- No gerund as a noun. Write "the field accepts a boolean", not "accepting a boolean".
- No noun cluster longer than three words. Write "the limit on the batch size", not "the advertised batch size ceiling limit".

**Words**
- One word, one meaning. Do not change to a synonym for variety.
- Use the simplest verb. "Use", not "utilise". "Start", not "initiate".
- Keep code identifiers, type names, library names and HTTP status codes exactly as they appear in the code. These are Technical Names, and STE always permits them.
- No idiom, no metaphor, no humour.

**Order of a finding**
1. One sentence that names what is wrong.
2. One sentence that names the effect on the data or on the user.
3. The condition that triggers it, with concrete values.
4. The change to make.

Put the finding first. Do not open with a severity banner or a category label.

One note on scope: a strict ASD-STE100 vocabulary of ~875 words cannot express "pydantic", "coercion" or "validator". Treat those as Technical Names and keep them. The sentence rules matter more than the word list here.

lightly_studio_embed/src/lightly_studio_embed/protocol.py
iunir: done

lightly_studio_embed/src/lightly_studio_embed/protocol.py
```diff
@@ -0,0 +1,197 @@
...
+
+EMBED_VIDEOS_BYTES_PATH = f"{BASE_PATH}/embed/videos/bytes"
+
+# The multipart field that the bytes endpoints read. One part per item, in input order.
+FILES_FIELD_NAME = "files"
+
+# A number, because starlette renamed its constant for this status.
+STATUS_PAYLOAD_TOO_LARGE = 413
+
+# TODO(Michal, 09/2026): Report a limit per capability. Text queries and videos do not
```
michal-lightly: Minor: The todo should be in your name :)

lightly_studio_embed/src/lightly_studio_embed/protocol.py
```diff
@@ -0,0 +1,197 @@
...
+DEFAULT_MAX_REQUEST_BYTES = 32 * 1024 * 1024
+
+# Both ends store the vectors as `float32`. A larger value becomes an infinity there.
+MAX_ABS_EMBEDDING_VALUE = float(np.finfo(np.float32).max)
+
+# `IMAGE_PIL` is an object in memory, so it cannot cross the wire. An fsspec path can.
+WIRE_CAPABILITIES = frozenset(Capability) - {Capability.IMAGE_PIL}
+
+
+def _reject_bool(value: Any) -> Any:
```
michal-lightly: Style: Move function to the bottom.

michal-lightly: I don't understand when this function would be needed. Under what scenario could the user pass a bool accidentally for dimension or indices? Seems very unlikely.

lightly_studio_embed/src/lightly_studio_embed/protocol.py
```diff
@@ -0,0 +1,197 @@
...
+    """
+
+    space_key: str
+    """The identifier of the embedding space that the server produces."""
+
+    dimension: _WireInt = Field(gt=0)
+    """The length of every embedding vector that the server returns."""
+
+    ready: bool
+    """Whether the model is loaded. ``False`` means try again soon. It is not an error."""
```
michal-lightly, line 88: Let's go with this in this PR. We might need to extend this to ServerStatus enum later.

lightly_studio_embed/src/lightly_studio_embed/validation.py
```diff
@@ -0,0 +1,108 @@
+"""Checks the result of an embedder before the server sends it.
+
+A wrong dimension or a NaN is cheap to find here, in the process of the customer. It is
+expensive to find later, as a similarity search that gives bad results in LightlyStudio.
```
michal-lightly: Outdated comment, the check is implemented in a different file.

michal-lightly (approved):

PR #2259, opened 2026-09-09
Mount the embed endpoints with create_app (embed stack 3/4)

iunir:
## What has changed and why?

Part 3 of a 4-PR stack that splits #2240 (LIG-10691) into reviewable pieces. #2257 and #2258 are
merged, and the package is `lightly-studio-serve` since #2289, so this branch is rebased onto it.

`create_app` inspects which capability classes the embedder implements and mounts exactly the
matching endpoints, so `/v1/describe` advertises what the server actually answers: a text-only
model exposes one embed route and nothing else. `/v1/embed/texts` takes JSON; `images/bytes` and
`videos/bytes` take a `multipart/form-data` body with one part per item in the `files` field. The
identity on every response comes from `embedder.embedding_space_spec()`, so the wire fields and
LightlyStudio's own embedders read the same value object.

Two additions to the moved interface land here rather than in #2257, which was a pure move,
because this is the PR that needs them: **`Embedder.ready`**, which the readiness guard and
`/v1/describe` read, and **`VideoBytesEmbedder`**, which `Capability.VIDEO_BYTES` named but no
class implemented.

## Notes for the reviewer

- **No 501, and 400 rather than 422.** A capability an embedder does not implement is an interface
  it does not subclass, so no endpoint for it is ever mounted. A malformed request gets 400; the
  protocol reserves 422 for input the server cannot use, and the two carry opposite advice for a
  client.
- **Readiness and the batch ceiling.** While `ready` is `False`, `describe` reports `ready: false`
  and the embed endpoints answer 503 with a `Retry-After`. Over the advertised `max_batch_size`,
  the request is answered 413.
- The bytes handlers are **synchronous**, so FastAPI runs the forward pass in a worker thread
  rather than blocking the event loop, and each part is closed right after it is read.
- **Annotation crops get no endpoint.** LightlyStudio crops with a margin and posts the result to
  `images/bytes`, so an image-capable embedder covers crops with no extra work.
- `python-multipart` joins the dependencies, since starlette parses the bytes bodies with it. The
  floor is the first release past the parser DoS advisories, which Python 3.9 cannot reach, so it
  is pinned separately there and the comment names the exposure.
- **No auth and no body size ceiling yet**, and no `serve()`: both arrive in #2251.

## How has it been tested?

13 new tests over a `TestClient`: the `describe` body for a text-only and a bytes embedder,
`ready: false`, 404 on the unmounted bytes paths, all three round trips, a per-item skip reflected
in `kept_indices`, the 413, the 503 with its `Retry-After`, the 500s for a failing model and a
wrong dimension, and the 400. `make static-checks` and `make test` pass on 3.9 and 3.14.

- [ ] Yes

---

1. ~~#2257 — move the embedder interface into the workspace member~~ (merged)
2. ~~#2258 — the protocol wire models and the result checks~~ (merged)
3. #2259 — `create_app` and the embed endpoints ← this PR


lightly_studio_serve/src/lightly_studio_serve/server.py
```diff
@@ -0,0 +1,228 @@
...
+    )
+
+
+def _check_batch_size(*, item_count: int, limits: ServerLimits) -> None:
+    if item_count > limits.max_batch_size:
+        raise HTTPException(
+            status_code=protocol.STATUS_PAYLOAD_TOO_LARGE,
+            detail=f"The request carries {item_count} items, over the advertised "
+            f"max_batch_size of {limits.max_batch_size}.",
+        )
```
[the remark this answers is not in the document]
iunir: Correct that nothing here enforces `max_request_bytes`, and that is deliberate for this PR: the ceiling lands in #2251, the next PR in the stack. It adds a `RequestSizeLimit` ASGI middleware that rejects an oversized declared `Content-Length` and caps the streamed body before parsing, wires it in `create_app`, and covers it with the regression tests this comment asks for — oversized JSON, oversized multipart, and a chunked body with no `Content-Length`. This PR is the routes only; the ASGI guards are split out so each stays reviewable on its own.

lightly_studio_serve/pyproject.toml
```diff
@@ -18,21 +18,31 @@ classifiers = [
...
     "numpy>=1.26.4",
     "pydantic>=2.0",
+    # Starlette parses the multipart bodies of the bytes endpoints through it. The floor is
+    # the first release fixing CVE-2026-53540, after a run of parser DoS advisories
+    # (CVE-2026-40347, CVE-2026-42561, CVE-2026-53537..53539). Python 3.9 is stuck on
+    # 0.0.20, the last release that supports it, so a 3.9 host keeps those weaknesses in its
+    # multipart parser.
+    # TODO(Iunir, 09/2026): Drop the 3.9 pin once `lightly-studio` drops Python 3.9.
+    "python-multipart>=0.0.31; python_version >= '3.10'",
+    "python-multipart>=0.0.20; python_version < '3.10'",
```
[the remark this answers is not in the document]
iunir: we will soon drop 3.9 support, let's ignore it

lightly_studio_serve/pyproject.toml
```diff
@@ -29,12 +29,24 @@ dependencies = [
     # Types `ImagePILEmbedder`. The floor is the version that `lightly-studio` resolves.
     "pillow>=11.0.0",
     "pydantic>=2.0",
+    # Starlette parses the multipart bodies of the bytes endpoints with this package. The
+    # floor is the first release that fixes CVE-2026-53540. Earlier releases have several
+    # parser denial-of-service problems (CVE-2026-40347, CVE-2026-42561 and
+    # CVE-2026-53537 to CVE-2026-53539). Python 3.9 cannot reach that floor, because 0.0.20
+    # is the last release that supports Python 3.9. A host on Python 3.9 therefore keeps
+    # those problems in its multipart parser.
```
michal-lightly: Optional: Can we make this more to the point?

lightly_studio_serve/pyproject.toml
```diff
@@ -29,12 +29,24 @@ dependencies = [
...
+    "python-multipart>=0.0.31; python_version >= '3.10'",
+    "python-multipart>=0.0.20; python_version < '3.10'",
     "uvicorn>=0.32.1",
 ]
 
 [dependency-groups]
 # The same floors as the repository's other Python packages, so the tooling behaves alike.
 dev = [
+    # The FastAPI test client is an httpx transport.
+    "httpx>=0.27.2",
```
michal-lightly, line 46: Shouldn't httpx be installed as a normal dependency?

lightly_studio_serve/src/lightly_studio_serve/server.py
```diff
@@ -0,0 +1,213 @@
...
+    app.add_exception_handler(EmbedderContractError, _handle_contract_error)
+    app.add_exception_handler(RequestValidationError, _handle_invalid_request)
+    router = APIRouter()
+    _mount_describe(router=router, embedder=embedder, limits=resolved_limits)
+    _mount_embed_routes(router=router, embedder=embedder, limits=resolved_limits)
+    app.include_router(router)
+    return app
+
+
+def _mount_describe(*, router: APIRouter, embedder: Embedder, limits: ServerLimits) -> None:
```
michal-lightly: Style: Remove `*`. Also elsewhere.

lightly_studio_serve/src/lightly_studio_serve/server.py
```diff
@@ -0,0 +1,213 @@
...
+        )
+
+
+def _readiness(*, embedder: Embedder) -> Callable[[], None]:
+    def verify() -> None:
+        if not embedder.ready:
+            raise HTTPException(
+                status_code=status.HTTP_503_SERVICE_UNAVAILABLE,
+                detail="The model is still loading.",
+                headers={"Retry-After": _RETRY_AFTER_SECONDS},
```
michal-lightly: Should we also have max retries?

Also I think it would be more practical to load the Embedder synchronously (i.e. don't support ready=False). The user needs to instantiate the Embedder, they can simply start the server once the Embedder is fully loaded. That would simplify the contract and prevent handling special cases (like the _readiness function does here). WDYT?

iunir: I think this is something optional to use, but nice to have. Imagine server got restarted and lightly-studio-serve loaded faster than the model itself, then readiness can be translated through API

lightly_studio_serve/pyproject.toml
iunir: it's needed only for testing

lightly_studio_serve/tests/test_server.py
```diff
@@ -0,0 +1,270 @@
...
+    def embed_video_bytes(self, videos: list[bytes]) -> EmbeddingResult:
+        self.received = videos
+        return _rows(count=len(videos))
+
+
+class BrokenTextEmbedder(FakeTextEmbedder):
+    """A model that fails, the way torch does for an op its backend lacks."""
+
+    def embed_text(self, texts: list[str]) -> EmbeddingResult:
+        raise NotImplementedError("could not run 'aten::foo' on the 'MPS' backend")
```
michal-lightly: Optional: The error message could say something more descriptive, e.g. just `mock_error_message`. I had to return here when seeing "MPS" in a test assert.

michal-lightly (approved):

PR #2280, opened 2026-09-10
perf: export video frames to image files in parallel

lorinczszabolcs:
## What has changed and why?

This speeds up `VideoFrameDataset.export().to_image_files()` and makes its output file names unique.

`to_image_files` grouped frames by their parent video and exported the videos one at a time. Each video is opened once and decoded in a single pass, and every requested frame is converted to a PIL image and written to a file. The videos are independent, so it now decodes and writes them on a bounded worker pool with `thread_imap_lazy(max_workers=get_media_worker_count())`, the same helper the image and crop embedding paths already use. The pool preserves input order, so the returned paths stay in the same order.

A local benchmark drove the real per-video export function, first sequentially and then through the worker pool, on 16 copies of a real 1080p clip (10 frames exported per video, 16 workers on a 20-core machine):

| Videos (1080p) | Before (sequential) | After (parallel) |
|---|---|---|
| 16 | 120.8 s | 11.9 s (10.1x) |

The exported frame images are identical whether the videos are decoded sequentially or in parallel. The speedup is across videos, so it scales with the number of videos up to the worker count. A single-video export runs the same single sequential decode pass, so it does not benefit from the parallelism.

The export also now adds a stable, path-derived token to each file name, so it is `{video_name}-{token}-{decode_index}-{video_format}{extension}`. Two videos with the same name in different directories previously wrote the same output path and overwrote each other, tracked by a TODO in the file. Exporting videos concurrently would turn that overwrite into a write race, so the token keeps every output path distinct.

## How has it been tested?

- `pytest tests/export` passes, including a new test that exports several videos in order, and a new test that exports two videos with the same name from different directories and checks that all frames are written without collision.
- The benchmark above compared wall-clock time and a hash of every output file between the sequential and the parallel run.

- [x] Yes (a `### Changed` bullet for the speedup and a `### Fixed` bullet for the unique file names)

## Summary by CodeRabbit

- **New Features**
  - Video frames can now be exported to image files with improved processing speed through parallel video decoding.
  - Exported frame filenames include a path-specific identifier, preventing overwrites when videos share the same name.

- **Bug Fixes**
  - Resolved filename collisions for same-named videos in different folders.
  - Export progress is now consolidated across the full frame-export operation.


lightly_studio/src/lightly_studio/export/video_frame_dataset_export.py
```diff
@@ -102,17 +103,27 @@ def to_image_files(self, output_dir: PathLike, extension: str = ".png") -> list[
...
+                fs=fs,
+                output_dir=output_dir_str,
+                extension=extension_lower,
             )
+
+        # Decode independent videos on a worker pool. thread_imap_lazy preserves order.
+        total_frames = sum(len(frame_samples) for frame_samples in frames_by_video.values())
+        exported_paths: list[str] = []
+        with tqdm(total=total_frames, desc="Exporting frames", unit=" frames") as pbar:
+            for video_paths in parallelize.thread_imap_lazy(
```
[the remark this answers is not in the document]
lorinczszabolcs: The collision is pre-existing (Horatiu's TODO on this line). This PR only makes concurrent writes possible. I can fix the naming here if you would like.

lorinczszabolcs:
I addressed the CodeRabbit comment by deriving a stable token from the video path and adding it to each exported frame name, so the output is now `{video_name}-{token}-{decode_index}-{video_format}{ext}`.

This adds the token to every export, not only when two videos actually share a basename. I can instead add it only for the videos involved in a real collision, so when no collision happens, it keeps the current `{video_name}-{decode_index}` names and the token appears only where files would otherwise overwrite. I'm open to other suggestions.

lightly_studio/src/lightly_studio/export/video_frame_dataset_export.py
```diff
@@ -102,17 +106,27 @@ def to_image_files(self, output_dir: PathLike, extension: str = ".png") -> list[
...
+                function=export_one,
+                iterable=frames_by_video.items(),
+                max_workers=executor.get_media_worker_count(),
+            ):
+                exported_paths.extend(video_paths)
+                pbar.update(len(video_paths))
```
mrpositron: The per-frame `pbar.update(1)` inside the decode loop was replaced by a single `pbar.update(len(video_paths))` after each video finishes, so the bar no longer moves while a video is decoding.

lightly_studio/src/lightly_studio/export/video_frame_dataset_export.py
```diff
@@ -102,17 +106,27 @@ def to_image_files(self, output_dir: PathLike, extension: str = ".png") -> list[
...
-        for video_path, frame_samples in frames_by_video.items():
-            exported_paths.extend(
-                _export_frames_from_video(
-                    video_path=video_path,
-                    frames=frame_samples,
-                    fs=fs,
-                    output_dir=output_dir_str,
-                    extension=extension_lower,
-                )
+        def export_one(item: tuple[str, list[VideoFrameSample]]) -> list[str]:
```
mrpositron: `export_one` is a closure in the middle of the public method; the python-guide puts private (https://github.com/lightly-ai/lightly-studio/blob/main/.agents/skills/python-guide/SKILL.md) helpers at module level below the public API.

lightly_studio/src/lightly_studio/export/video_frame_dataset_export.py
```diff
@@ -102,17 +106,27 @@ def to_image_files(self, output_dir: PathLike, extension: str = ".png") -> list[
...
+        # Decode independent videos on a worker pool. thread_imap_lazy preserves order.
+        total_frames = sum(len(frame_samples) for frame_samples in frames_by_video.values())
+        exported_paths: list[str] = []
+        with tqdm(total=total_frames, desc="Exporting frames", unit=" frames") as pbar:
+            for video_paths in parallelize.thread_imap_lazy(
+                function=export_one,
+                iterable=frames_by_video.items(),
+                max_workers=executor.get_media_worker_count(),
+            ):
+                exported_paths.extend(video_paths)
```
mrpositron: `video_paths` holds exported image paths. It binds `_export_frames_from_video`'s return value: image files, not videos. Easy to misread as source paths at `exported_paths.extend(video_paths)`, especially since the function genuinely groups by video path a few lines above. Rename to frame_paths.

mrpositron (commented):
Thanks for your contribution! Added some comments.

lightly_studio/src/lightly_studio/export/video_frame_dataset_export.py
```diff
@@ -161,9 +175,9 @@ def _export_frames_from_video(
     else:
         total_frame_count = max_frame_number
     zero_padding = len(str(total_frame_count))
-    # TODO(Horatiu 08/2026): Using the filename will overwrite files if two videos with the same
-    # name are in different directories. Add video sample id to the name.
     video_filename = Path(video_path).name
+    # A path-derived token keeps output names unique when two videos share a basename.
+    video_token = _video_path_token(video_path)
```
mrpositron: Please use `_video_path_token(video_path=video_path)` here, following the Python guide’s (https://github.com/lightly-ai/lightly-studio/blob/main/.agents/skills/python-guide/SKILL.md) keyword-argument convention. Apply the same convention to the new project-helper calls in the tests.

lorinczszabolcs:
@mrpositron Thank you for the review! Addressed both.

PR #2289, opened 2026-09-11
Rename lightly-studio-embed to lightly-studio-serve

iunir:
## What has changed and why?

Renames the workspace member and its distribution, per [LIG-10820](https://linear.app/lightly/issue/LIG-10820/rename-lightly-studio-embed-to-lightly-studio-serve):

- distribution `lightly-studio-embed` → `lightly-studio-serve`
- import package `lightly_studio_embed` → `lightly_studio_serve`
- directory `lightly_studio_embed/` → `lightly_studio_serve/`

`from lightly_studio_serve import serve` reads better than the old name, and the module then matches the one function the package exists to expose.

**Why now.** Nothing is published under the old name, so the rename is free today and expensive later:

- Neither `lightly-studio-embed` nor `lightly-studio-serve` exists on PyPI, so the new name is still free.
- A trusted publisher binds to a PyPI project name, so renaming after that is set up means redoing it on both PyPI and TestPyPI.
- No released artifact depends on the package. `lightly-studio` 1.1.0 on PyPI carries no dependency on it; the dependency only exists on `main`, resolved from the workspace.

**Not affected.** `lightly_studio` re-exports `EmbeddingResult`, `EmbeddingSpaceSpec` and `ImageCrop`, so the public `lightly_studio` API is unchanged. Only a direct import of the unreleased module breaks. No CHANGELOG entry, since neither the package nor the dependency on it has shipped.

**Left out on purpose.** The package stays at version `0.1.0`. `0.x` already says "no API stability promised", and lowering it would make the `lightly-studio-serve>=0.1.0` floor in `lightly_studio/pyproject.toml` unsatisfiable from PyPI.

**Mostly mechanical** — every changed line outside `uv.lock` carries the rename and nothing else. Four spots are not a find-and-replace and are worth a look:

- `.github/workflows/unit_test.yml` — the `paths-filter` key and path, the job, its `needs` entry and the gate variable.
- `lightly_studio/docs/mkdocs.yml` — the mkdocstrings search path, and the three `::: lightly_studio_serve.types` blocks in `api/embeddings.md`.
- Root `Makefile` and `pyproject.toml` — `PYTHON_MEMBERS` and `tool.uv.workspace.members`.
- `AGENTS.md` and `.claude/hooks/skill-reminder.py` — the path globs that pick the Python skill.

## How has it been tested?

- `make static-checks` and `make test` in `lightly_studio_serve` — pass (37 tests).
- `lightly_studio` tests that touch the renamed imports — 126 pass.
- `make docs` builds with `--strict`, so mkdocstrings resolves the renamed references.
- `uv build --package lightly-studio-serve` produces both a wheel and an sdist.
- `uv lock` renames the member entry with nothing re-resolved; `uv lock --check` is clean.
- `grep` finds no `lightly_studio_embed` / `lightly-studio-embed` left in the tree.

Note: `make static-checks` from the repository root fails locally with 16 pre-existing `adlfs`/`gcsfs`/`s3fs` stub errors in four test files this branch does not touch. Unrelated to the rename.

- [ ] Yes

---

@michal-lightly for review, since you reviewed #2257 and #2258, the earlier parts of this stack. @IgorSusmelj as an alternative.

## Summary by CodeRabbit

- **New Features**
  - Added shared embedding result, space specification, and image crop types for consistent integration.
  - Added clearer error handling for invalid embedder responses.

- **Updates**
  - Renamed the embedding package to LightlyStudio Serve across installation, build, validation, and integration workflows.
  - Updated examples and API references to use the new package name.

- **Documentation**
  - Updated setup instructions, package references, and API documentation for LightlyStudio Serve.

- **Tests**
  - Updated test coverage and CI configuration to reflect the package rename.


michal-lightly:
@codex review

lightly_studio/src/lightly_studio/dataset/embedding_manager.py
```diff
@@ -79,7 +79,7 @@ def set_default_embedding_model(embedding_generator: EmbeddingGenerator) -> None
 
     Call this before you add a dataset (for example, before ImageDataset.load_or_create)
     or before you launch the GUI to use your own generator instead of the model set by
-    LIGHTLY_STUDIO_EMBEDDINGS_MODEL_TYPE. The override applies to every collection.
+    LIGHTLY_STUDIO_SERVEDINGS_MODEL_TYPE. The override applies to every collection.
```
michal-lightly: This is wrong :D

iunir: Good catch, thank you — and it was worse than the docstring. My sed wasn't anchored to a word boundary, so `LIGHTLY_STUDIO_EMBED` also matched the prefix of `LIGHTLY_STUDIO_EMBEDDINGS_MODEL_TYPE`. That hit `dataset/env.py:11`, where the mangled name was the *string literal* passed to `env.str(...)`, so the rename silently changed the environment variable customers actually set. Tests stayed green because nothing asserts that literal.

Restored in all 7 files (11 lines): `env.py`, `embedding_manager.py`, two examples, two benchmarks — the benchmark also had a `lightly_studio_embedding_io_benchmark_` tmpdir prefix mangled the same way.

To make sure nothing else slipped through, I re-derived the whole branch from `main` with a word-anchored transform (`lightly_studio_embed(?![A-Za-z0-9])` and friends) and diffed it against what I had pushed — all 50 changed files now match exactly. The diff shrank from 51 files / 87 lines to 47 / 78, which is the four files that should never have been touched dropping out.

michal-lightly (approved):

PR #2290, opened 2026-09-11
Fallback to loading the default model from the DB if not present in-memory during text/image search

mihnea-necsulescu:
## What has changed and why?

Default models were previously looked up only in the `_collection_id_to_default_model_id` map which would be cleared on server restarts. This means that updating any hosted instance (without reindexing, for example() would make text/image search not functional.
In this PR, the DB is also checked when the default model is not found in the map before raising.

## How has it been tested?

New tests + Manually

- [ ] Yes

## Summary by CodeRabbit

- **Improvements**
  - Image and text embedding operations now consistently use the collection’s registered default embedding model, including when it is not already cached.
  - Embedding requests now maintain database-backed session handling throughout processing.

- **Bug Fixes**
  - Improved handling when a collection has no default embedding model by providing a clear error instead of proceeding without one.


ikondrat (approved):

michal-lightly (changes requested):
There is a simpler solution

mihnea-necsulescu:
addressed with #2293

PR #2294, opened 2026-09-11
Guard lightly-studio's workspace dependency on lightly-studio-serve

iunir:
## What has changed and why?

uv resolves a workspace dependency from the workspace and drops the declared specifier: `uv.lock` records `lightly-studio-serve` with no `specifier`, where every registry dependency next to it keeps one. So `uv lock --check` passes whatever the requirement says, and the range that ships in the wheel is never exercised here.

One check in Prepare Release, next to `check-labelformat-pin`, skipped for packages with no workspace dependency:

- `check-dependency-range` — the declared range admits the version the workspace builds and tests against.

This matters now that #2312 caps the requirement at `<0.2.0`. An open floor could never exclude anything; a cap can, so without this the next bump of the member ships a wheel declaring a range that excludes the version it was tested against.

Comparison is `packaging`, supplied by `uv run --with` in the workflow step. It is the only subcommand that needs a third-party import, so `cli.py` imports the module from inside its handler and every other subcommand stays runnable on a bare `python3`, as `version.py` documents.

## How has it been tested?

`make static-checks` and `make test` in `.github/scripts` — 100 pass, 12 new. Reproduced the drift the check exists for: requirement `>=0.1.0,<0.2.0` with the member at `0.2.0`, where `uv lock --check` passes and the new check fails naming both versions. Ran the check against the real tree for both packages: `lightly-studio` passes, `lightly-studio-serve` is a no-op.

- [ ] Yes

---

@michal-lightly for review. @IgorSusmelj as an alternative.

## Summary by CodeRabbit

- **New Features**
  - Added release validation to confirm workspace dependency versions satisfy declared requirements.
  - Release workflows now run dependency range checks automatically and include results in review checklists.
  - Added safeguards to verify dependency minimum versions are available on PyPI.

- **Tests**
  - Added coverage for dependency range validation, requirement parsing, publication checks, and workspace configuration output.


iunir:
Superseded by #2372, which does the same guard as a six-line test on the built metadata instead of a `check-dependency-range` subcommand — no new module, no `workspace_dependencies` field on the package registry, no workflow step. It also runs on every PR rather than at release, so drift fails in the release PR that introduces it.

The other half of this ticket, `check-dependency-published`, is moot: `lightly-studio-serve` 0.1.0 is on PyPI, so the floor names a release that exists.

Issue #2050, opened 2026-08-22
[FEAT] Add dataset.compute_image_quality_metadata() for per-image quality metrics


lorinczszabolcs:
### 💡 Is your feature request related to a problem?

LightlyStudio has no built-in way to compute per-image quality metrics. A user cannot sort or filter images by blur, brightness, contrast, or resolution from the SDK. The docs show how to compute a few basic statistics (luminance and per-channel color ratios) by hand with PIL, but there is no built-in method for image-quality metrics. So a common first curation step, the removal (or selection) of low-quality images, stays manual.

### 🧰 Describe the solution you'd like

Add a `dataset.compute_image_quality_metadata(...)` method that stores one numeric value per metric, per image. It mirrors the `dataset.compute_typicality_metadata()` method.

A first version can compute:
- sharpness (variance of the Laplacian)
- brightness (mean luminance)
- contrast (pixel standard deviation)
- entropy of the gray histogram
- per-channel color means
- aspect ratio, resolution, and file size

Each value is stored as float metadata. Then the distribution panel, the metadata filters, and the `metadata_weighting` sampling strategy use it with no extra work. For example, a low sharpness value marks a blurry image.

### 🛠 Alternatives you've considered

- The docs recipe computes luminance and per-channel color ratios by hand with PIL. It is manual, it covers only a few basic statistics, and each user must rewrite it.
- `dataset.compute_typicality_metadata()` finds atypical images from embeddings. It does not measure physical image quality such as blur or exposure.

### 📝 Additional context

I checked this against `main`. There is no quality-metric code today, only the manual recipe in the docs. The write path already exists: `metadata_resolver.bulk_update_metadata`, exposed as `dataset.update_metadata`. Numeric metadata already drives the distribution panel histograms, the filters, and the `metadata_weighting` sampling.

The method mirrors `compute_typicality_metadata`. It opens each image with `fsspec.open`. This works for local and cloud paths, as the ingest code does. Then it computes the metrics with numpy and OpenCV.

It needs no new dependency. `opencv-python-headless` and `fsspec` are already dependencies, and the project already uses numpy and PIL. It needs no schema change. The main cost is the image decode. The decode is I/O-bound, so a thread pool fits, as the ingest code shows. The values are stored as float, because the metadata schema pins the type on first write.

Prior art: the FiftyOne Data Quality workflow computes similar per-image metrics (brightness, blurriness, aspect ratio, entropy) to flag low-quality images.

[...]

PR #2309, opened 2026-09-12
feat(sdk): compute image quality metadata

breken-ai:
## What has changed and why?

Refs #2050

LightlyStudio has no built-in way to compute per-image quality metrics, so the "remove or select low-quality images" step stays a hand-written PIL loop (the recipe in the Metadata docs). This PR adds `ImageDataset.compute_image_quality_metadata()`, mirroring `compute_typicality_metadata()`: one call stores a set of pixel statistics as float metadata, and the distribution panel, the metadata filters and `metadata_weighting` sampling pick them up with no extra work.

- New module `lightly_studio/metadata/compute_image_quality.py`, called from a new `ImageDataset.compute_image_quality_metadata(overwrite=False, max_workers=None)` method. The method lives on `ImageDataset` rather than `Dataset` because it decodes `ImageTable.file_path_abs`.
- Each image is read once through the same fsspec path as ingest (`fsspec.core.url_to_fs` on the caller thread, `filesystem.exists` / `filesystem.open` in the worker, `remote_storage.configure_connections` up front) and decoded once; every metric comes from that one decode. Pillow + numpy only, no second image stack (`cv2` is not imported anywhere in `src` today, so I did not introduce it).
- Bounded concurrency and memory: `parallelize.thread_imap_lazy` with `buffer_size == max_workers`, so at most `max_workers` decoded images are in flight, and results are written in batches of 1000 through `metadata_resolver.bulk_update_metadata`. Default workers: 4 for local files, `LIGHTLY_STUDIO_REMOTE_IMAGE_PROBE_WORKERS` when any path is remote.
- Failures are separate from values: a missing or undecodable image (the shared `BROKEN_IMAGE_ERRORS`) gets no metric keys at all and is counted in a `FileOutcomeReport` log summary; if every attempted image fails, `AllInputFilesFailedError` is raised like ingest and embedding do. A corrupt file can never look like a dark or blurry image.
- Provenance and resume: every written sample also gets the integer `image_quality_version` (currently `1`). Without `overwrite`, samples already at the current version are skipped, so an interrupted run resumes; samples with another version are recomputed, so a changed definition never silently mixes with old values. `overwrite=True` recomputes everything.
- Docs: a "Compute image quality metadata" section in the Metadata page with the definitions table, the normalization rules, and an example that tags the 100 least sharp images of a dataset with `metadata_weighting(strength=-1)` rather than a global blur threshold.

### Metric definitions

All metrics are computed on the 8-bit RGB image after `ImageOps.exif_transpose` and `convert("RGB")` (alpha is discarded, transparent pixels contribute their stored colour). `L` is Pillow's `convert("L")` luminance: ITU-R BT.601 `0.299 R + 0.587 G + 0.114 B`, rounded to an integer in `[0, 255]`. All values are stored as `float`, so the metadata schema type is stable across images.

| Key | Definition | Range |
|---|---|---|
| `brightness` | mean of `L` | `[0, 255]` |
| `contrast` | population standard deviation of `L` | `[0, 127.5]` |
| `sharpness` | variance of the 4-neighbour Laplacian `4L(x,y) - L(x-1,y) - L(x+1,y) - L(x,y-1) - L(x,y+1)` over interior pixels; `0.0` for a constant image and for images with fewer than 3 px on a side | `[0, inf)` |
| `entropy` | Shannon entropy of the 256-bin histogram of `L`, in bits | `[0, 8]` |
| `red_mean`, `green_mean`, `blue_mean` | per-channel mean | `[0, 255]` |
| `aspect_ratio` | `width / height` after EXIF orientation | `(0, inf)` |
| `image_quality_version` | integer version of the definitions the values were computed with | |

The docstrings and the docs say explicitly that these describe pixels and are not a good/bad classifier, and that Laplacian sharpness depends on image scale and content.

### Deliberately out of scope (first slice)

- `file_size`: not cheaply available for remote files without an extra `info()` round trip per image; can be a follow-up with its own I/O budget.
- `width` / `height` as metadata: already first-class sample fields with their own GUI filters. `aspect_ratio` is the one derived value stored, since after EXIF orientation it can differ from the stored dimensions.
- An HTTP route / GUI entry point: the issue asks for the SDK method; the API can wrap the same module later exactly as `metadata/typicality` does.
- Video frames and other sample types.

## How has it been tested?

New tests (all deterministic fixtures, no example dataset needed):

- `tests/metadata/test_compute_image_quality.py` (14 tests): constant gray (expected brightness, zero contrast / sharpness / entropy), 2x2-cell checkerboard (brightness 127.5, contrast 127.5, entropy exactly 1 bit, sharpness exactly `510**2`, derived by hand in the test), blurred checkerboard is strictly less sharp than the original, pure-red channel means and BT.601 luminance 76, EXIF orientation 6 changes `aspect_ratio` from 2.0 to 0.5, RGBA alpha discarded, 2-row image gets sharpness 0.0; end-to-end on DuckDB: values and schema type `float`, missing and broken images get no keys and the log summary reads `computed=1, already_present=0, missing=1, broken=1`, resume / recompute-on-version-change / `overwrite=True` (spied `_compute_one` call counts 2 -> 2 -> 3 -> 5), batched writes (`WRITE_BATCH_SIZE=2` over 3 images yields writes of 2 then 1), all-failed raises `AllInputFilesFailedError`, an unexpected worker error propagates and writes nothing, `max_workers=0` raises `ValueError`.
- `tests/core/image/test_image_dataset__image_quality.py` (2 tests): the public `ImageDataset` method end to end, including a file corrupted after ingest.

Before the change (contract test): `ImportError: cannot import name 'compute_image_quality' from 'lightly_studio.metadata'`. After: `16 passed`.

Full backend suite passes locally except `test_server_static_webapp`, which needs the built frontend (see the evidence record); `make static-checks` (uv lock check, ruff, mypy on 984 files, ruff format) passes; `make -C docs docs` (strict) builds and validates the new anchor.

- [x] Yes

## Evidence record

- Repository / issue: lightly-ai/lightly-studio, issue #2050 ("[FEAT] Add dataset.compute_image_quality_metadata() for per-image quality metrics").
- Upstream SHA / patch SHA: base `62996452a31cf19560ed6339c5b0a4904b9a043b` (origin/main) / patch `bbbf582fcdb84858ebfbb50ae07faf10171f579b`.
- Issue ownership and competing PR result (2026-09-12 21:50 UTC): issue open, no assignees, zero comments, zero cross-references in the timeline; `gh search prs -- 2050` and `gh search prs -- "compute_image_quality_metadata brightness blur entropy"` returned nothing; no open PR touches image quality. Re-checked immediately before opening this PR.
- Reproducer red: `uv run pytest tests/core/image/test_image_dataset__image_quality.py` on the base tree -> `ImportError: cannot import name 'compute_image_quality' from 'lightly_studio.metadata'` (1 error during collection).
- Patched green: `uv run pytest tests/metadata/test_compute_image_quality.py tests/core/image/test_image_dataset__image_quality.py` -> `16 passed`.
- Full checks executed: `make static-checks` in `lightly_studio` (exit 0: `uv lock --check`, `ruff check`, `mypy .` "Success: no issues found in 984 source files", `ruff format --check`); full backend suite `uv run pytest -n auto` -> `2536 passed, 53 skipped, 1 failed` in 2:39; the one failure is `tests/api/test_server.py::test_server_static_webapp` at `GET /favicon.png` (404), which needs the built frontend in `dist_lightly_studio_view_app` and was run here against the `make mock-webapp-dist` stub (an empty `index.html`); it imports nothing this PR touches; `make -C docs docs` (mkdocs `--strict`) -> built in 17 s, exit 0. Python 3.9.6, uv 0.12.6 (the repo pin; installed into a local tool dir because the machine's uv is 0.10.6).
- Checks not executed and why: `make test-postgres` (needs a Docker Postgres container; the change uses only the existing `bulk_update_metadata` / `get_metadata_values_for_key` resolvers, no new SQL); frontend and e2e suites (no frontend change); remote-storage paths against a real bucket (the fsspec calls are the same three used by `_probe_image` in ingest, exercised here on the local filesystem).

https://claude.ai/code/session_01DkE5qM85ht9Uoq3aAqcKf7

## Summary by CodeRabbit

- **New Features**
  - Added image-quality metadata computation for brightness, contrast, sharpness, entropy, color-channel means, and aspect ratio.
  - Supports resuming or recomputing metadata with overwrite controls.
  - Handles unreadable images without blocking successful results.

- **Documentation**
  - Added guidance covering available metrics, value ranges, sampling, preprocessing, and recomputation behavior.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Add `ImageDataset.compute_image_quality_metadata()` to store brightness, contrast, sharpness, entropy, per-channel means and aspect ratio of every image as float metadata.
 - Add a Split dataset menu action to divide images or videos into tagged splits using the current filters.
 - Add a Glossary page to the docs.
 
--- a/lightly_studio/docs/docs/concepts_and_tools/metadata.md
+++ b/lightly_studio/docs/docs/concepts_and_tools/metadata.md
@@ -129,6 +129,59 @@ how to do this given a python dictionary (e.g. loaded from JSON) or a CSV file.
     dataset.update_metadata(sample_metadata)
     ```
 
+### Compute image quality metadata
+
+[ImageDataset.compute_image_quality_metadata](../api/dataset.md#lightly_studio.ImageDataset.compute_image_quality_metadata)
+decodes every image once and stores a set of pixel statistics as float metadata. Once stored,
+the values show up in the distribution panel, the metadata filters, and the
+`metadata_weighting` sampling strategy like any other numeric metadata.
+
+```python
+import lightly_studio as ls
+
+dataset = ls.ImageDataset.load()
+dataset.compute_image_quality_metadata()
+```
+
+All metrics are computed on the 8-bit RGB image after the EXIF orientation is applied and any
+alpha channel is discarded. `L` is Pillow's `convert("L")` luminance (ITU-R BT.601,
+`0.299 R + 0.587 G + 0.114 B`, rounded to an integer in `[0, 255]`).
+
+| Key | Definition | Range |
+|---|---|---|
+| `brightness` | Mean of `L` | `[0, 255]` |
+| `contrast` | Standard deviation of `L` | `[0, 127.5]` |
+| `sharpness` | Variance of the 4-neighbour Laplacian of `L` over the interior pixels | `[0, ∞)` |
+| `entropy` | Shannon entropy of the 256-bin histogram of `L`, in bits | `[0, 8]` |
+| `red_mean`, `green_mean`, `blue_mean` | Mean of each channel | `[0, 255]` |
+| `aspect_ratio` | `width / height` after EXIF orientation | `(0, ∞)` |
+| `image_quality_version` | Integer version of the metric definitions the values were computed with | |
+
+The metrics describe the pixels, they do not decide whether an image is good or bad. Sharpness
+in particular depends on the image size and content: a downscaled image or a flat scene scores
+low without being blurry. Compare values between the images of one dataset instead of applying
+one threshold everywhere. For example, to review the 100 least sharp images of a dataset, tag
+them by `sharpness`:
+
+```python
+import lightly_studio as ls
+
+dataset = ls.ImageDataset.load()
+dataset.compute_image_quality_metadata()
+
+dataset.query().sampling().metadata_weighting(
+    n_samples_to_select=100,
+    sampling_result_tag_name="least_sharp",
+    metadata_key="sharpness",
+    strength=-1,
+)
+```
+
+An image that is missing or cannot be decoded gets no values and is listed in the log summary,
+so it never looks like a dark or blurry image. Re-running the method skips images that already
+carry values of the current `image_quality_version` and recomputes the rest; pass
+`overwrite=True` to recompute every image.
+
 ### Add metadata calculated from images
 
 You can also iterate the dataset, open each image with PIL, compute derived statistics, and write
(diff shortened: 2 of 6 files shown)


lorinczszabolcs:
Hi @ikondrat! I opened the parent issue #2050 and was waiting for maintainer feedback on the approach before creating a PR, but it looks like a bot opened one in the meantime.
- I had a look and it is close to what I would have proposed, so I wouldn't duplicate effort by creating a competing PR.
- I can fold in the CodeRabbit findings and any changes required after review, in case the bot doesn't pick it up, just let me know.

One question for the future is if you would prefer contributors opening a PR alongside the issue, or waiting for feedback on the issue first? Thanks!

PR #2311, opened 2026-09-14
Fix embeddings read as Vector objects on PostgreSQL

JonasWurst:
## What has changed and why?

- pgvector 0.5.0 changed its psycopg binary loader to return a `Vector` object instead of a numpy array. Fresh installs on Python 3.10+ resolve to it, since `pgvector>=0.4.0` has no upper bound.
- The PostgreSQL read path in `sample_embedding_resolver` uses a raw binary cursor that bypasses SQLAlchemy, so nothing normalized the value and everything reading embeddings crashed: sampling, typicality, similarity, few-shot classification, the 2D embedding view. DuckDB goes through the ORM and was unaffected.
- `_read_embedding_rows_binary` now coerces to a 1-D float32 array, so it honors its return type on both 0.4.x and 0.5.x. No version cap: the regression test is what holds the line if a future release changes the loader again.

## How has it been tested?

New test asserts both read paths return `np.ndarray` / `float32` / `ndim == 1`.

Against a Postgres container with pgvector 0.5.0 on Python 3.10 (`pytest tests/resolvers/test_sample_embedding_resolver.py tests/sampling --postgres`): fails without the fix with the reported `TypeError: float() argument must be ... not 'Vector'`, passes with it. Also green on pgvector 0.4.2 and on DuckDB.

- [x] Yes

Alternative reviewer: @michal-lightly

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed PostgreSQL crashes affecting sampling, typicality, similarity, few-shot classification, and the 2D embedding view when using incompatible pgvector versions.
  - Embeddings are now consistently returned as one-dimensional, 32-bit floating-point arrays, improving compatibility across supported PostgreSQL configurations.
  - PostgreSQL support now uses pgvector versions below 0.5.0 to prevent unsupported embedding formats and related processing failures.
  - Improved embedding model resolution when checking whether collections contain embeddings.


CHANGELOG.md
```diff
@@ -32,6 +32,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Fix sampling, typicality, similarity, few-shot classification and the 2D embedding view crashing on PostgreSQL. New installs on Python 3.10 or later get pgvector 0.5.0, which returns embeddings in a wrong format.
```
michal-lightly: This line should not be here if we are merging before the release.

lightly_studio/src/lightly_studio/resolvers/sample_embedding_resolver.py
```diff
@@ -242,6 +243,17 @@ def _read_embedding_rows_binary(
     with connection.cursor(binary=True) as cursor:
         cursor.execute(sql, params)
         return [
-            SampleEmbeddingRow(sample_id=sample_id, embedding=embedding)
+            SampleEmbeddingRow(sample_id=sample_id, embedding=_to_embedding(embedding))
             for sample_id, embedding in cursor
         ]
+
+
+def _to_embedding(value: Any) -> Embedding:
```
michal-lightly, line 251: Let's merge as a fix, but let's add a TODO. Any reader of db embeddings will run into the issue, and solving it only for sample_embedding_resolver.py won't prevent it in the future.

michal-lightly (approved):

lightly_studio/src/lightly_studio/resolvers/sample_embedding_resolver.py
ikondrat: I would rather fix pgvector version, the fix looks very hacky

CHANGELOG.md
JonasWurst: Discussed to keep it

JonasWurst:
#2313

PR #2312, opened 2026-09-14
Cap lightly-studio's requirement on lightly-studio-serve

iunir:
## What has changed and why?

Caps the requirement at `<0.2.0` instead of an open floor, and drops the comment above it.

uv resolves this dependency from the workspace and drops the declared specifier, so nothing here exercises the range — it binds only an install from PyPI. A wheel then keeps that range for good: with an open floor, a later breaking `lightly-studio-serve` resolves under an already-installed `lightly-studio` and `import lightly_studio` fails. Published wheels can only be yanked, not fixed. 1.1.0 predates the dependency, so nothing is exposed yet and the next release is the first that would ship a range.

`<0.2.0` because the package is 0.x, where the minor is the breaking unit. The old comment claimed `tool.uv.sources` resolves from the workspace "for local development"; it applies to every uv command here, so it went rather than got corrected.

Complements #2294 — an open floor could never exclude anything, a cap can, so `check-dependency-range` is what keeps this from drifting from the version the workspace builds against.

## How has it been tested?

`uv lock --check` exits 0 with the lockfile untouched, which is also the demonstration that uv never validates the range. #2294's checks pass against this tree; the range admits `0.1.0` and `0.1.7`, rejects `0.2.0`.

- [ ] Yes

---

@michal-lightly for review, same as #2294. @IgorSusmelj as an alternative.

## Summary by CodeRabbit

* **Chores**
  * Constrained the supported `lightly-studio-serve` dependency to versions below 0.2.0.
  * Removed outdated dependency-resolution documentation.


ikondrat (approved):

lightly_studio/pyproject.toml
```diff
@@ -61,10 +61,7 @@ dependencies = [
     # Pinned below 7.0, which dropped Python 3.9. 6.9 covers every Python this package supports,
     # so one major runs everywhere rather than two that have to stay behaviourally identical.
     "posthog>=6.9,<7.0",
-    # Holds the embedder base classes and their value types. `tool.uv.sources` below
-    # resolves it from the workspace for local development. The bound applies to an
-    # install from PyPI, so publish `lightly-studio-serve` before you release this package.
-    "lightly-studio-serve>=0.1.0",
+    "lightly-studio-serve>=0.1.0,<0.2.0",
```
michal-lightly, line 64: It would be fine to also pin the version exactly for now.

michal-lightly (approved):

lightly_studio/pyproject.toml
iunir: In this early days of serve package, I believe we have to let it to be patch updated without need to re-release studio...

iunir: Both block the same thing — a breaking release resolving under an installed `lightly-studio`. The only difference is whether 0.1.x patches flow, so `==` adds protection only against a patch release of ours that breaks compatibility by accident.

It is not free, though: once #2294 lands, its range check compares the requirement against the version the workspace builds, so any bump of the member would block the next `lightly-studio` release until someone edits the pin — and anyone with both installed could not take a serve bugfix without waiting for a studio release. Happy to switch if you would rather not lean on our own patch discipline while the package is this young, but my preference is to keep the range.

PR #2317, opened 2026-09-14
Report backend exceptions to PostHog

LeonardoRosaa:
## What has changed and why?

- Enables PostHog's automatic capture of unhandled Python exceptions.
- Reports exceptions handled by the FastAPI exception handlers.

## How has it been tested?

Unit tests

- [ ] Yes

## Summary by CodeRabbit

* **New Features**
  * Added exception tracking with stack traces for application errors.
  * Enabled automatic capture of application exceptions in analytics.
  * Unhandled errors now return a consistent 500 response.

* **Bug Fixes**
  * Sensitive SQLAlchemy query parameters are redacted from reported exception details.
  * Invalid value errors now return a 400 response.
  * Exception-reporting failures no longer interrupt application behavior.

* **Tests**
  * Added coverage for exception capture, parameter redaction, tracking failures, and error responses.


lightly_studio/src/lightly_studio/analytics/posthog_tracker.py
```diff
@@ -46,6 +46,7 @@ def __init__(self, project_api_key: str, host: str) -> None:
             host=host,
             max_retries=MAX_RETRIES,
             timeout=REQUEST_TIMEOUT_SECONDS,
+            enable_exception_autocapture=True,
```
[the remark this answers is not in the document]
LeonardoRosaa: Added `unhandled_exception_handler` handler

lightly_studio/src/lightly_studio/api/routes/api/exceptions.py
```diff
@@ -29,9 +30,10 @@ def _log_error_details(
     """Log detailed error information with request context."""
     # Log the error with different levels based on status code
     logger.error(f"Server Error {status_code}: {exc}")
+    tracking.track_exception(exc)
```
mihnea-necsulescu, line 33: I think it would be cleaner to call this separately inside the handler after `_log_error_details` instead of here

lightly_studio/src/lightly_studio/api/routes/api/exceptions.py
```diff
@@ -134,3 +136,15 @@ async def _query_expr_error_handler(_request: Request, _exc: QueryExprError) ->
             status_code=HTTP_STATUS_BAD_REQUEST,
             content={"error": str(_exc) or "Invalid query expression."},
         )
+
+    @app.exception_handler(Exception)
+    async def _unhandled_exception_handler(_request: Request, _exc: Exception) -> JSONResponse:
```
mihnea-necsulescu, line 149: would it be possible for errors to be reported twice in PostHog? `enable_exception_autocapture` is set to true + this handler also tracks exceptions - how would it work?

LeonardoRosaa: Technically, auto-capture should only capture exceptions that escape the main process or a thread. These handlers catch FastAPI exceptions, so it is not possible for the errors to be reported twice.

lightly_studio/src/lightly_studio/api/routes/api/exceptions.py
LeonardoRosaa: I disagree. In most cases, this information isn't important to the reader; they only need to know that the handler calls `_log_error_details`. Also, this avoids some duplicated code, so I'd keep it as is.

lightly_studio/src/lightly_studio/analytics/posthog_tracker.py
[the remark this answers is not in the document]
mihnea-necsulescu: "Exposure of Sensitive Information to an Unauthorized Actor" - who would be the unauthorized actor in this scenario? us? 

- for hosted deployments, since we would be hosting customer data, PostHog would be within our trust boundary
- for on-prem deployments, analytics are disabled anyway
- OSS usage may be a problem, though, as database exceptions contain parameters, too, which would leak private user information (e.g., users' dataset names, internal paths, etc.); setting `hide_parameters` in SQLAlchemy also does not seem ideal, as that would affect all logs. parameters live in a `[parameters: ...]` block inside exceptions - stripping that block in `before_send` should solve this, right?

lightly_studio/src/lightly_studio/api/routes/api/exceptions.py
mihnea-necsulescu: oh I thought this was called just for the new handler, my bad - I'd update the function name then, as it does not only log anymore

and can you pass `exc` as a kwarg?

lightly_studio/src/lightly_studio/analytics/posthog_tracker.py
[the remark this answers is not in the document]
mihnea-necsulescu: your regex may not do it, though, as it stops at the first ']' - parameters containing lists may not get properly redacted

capturing till the end of the string may be better - `_SQLALCHEMY_PARAMETERS = re.compile(r"\[parameters:.*\Z", re.DOTALL)` even fi the "background on this error" may be swallowed; that can still be seen in the actual logs

lightly_studio/src/lightly_studio/api/routes/api/exceptions.py
LeonardoRosaa: Done

mihnea-necsulescu (approved):

PR #2343, opened 2026-09-15
Add McapFileReader to locate messages in indexed MCAP files

horatiualmasan:
## What has changed and why?

Adds `McapFileReader`: open an indexed MCAP through fsspec (local or object storage), `load_data_for_topics` in one pass, `get_frame_locators` with optional `sync_timestamps`. It returns where messages are, not payloads. `keyframe_log_time_ns` stays `None` until PR6.

## How has it been tested?

- `uv run --extra mcap --extra cloud-storage pytest tests/core/mcap/test_reader.py`

- [ ] Yes

## Summary by CodeRabbit

* **New Features**
  * Added support for reading frame locations from local and remote MCAP files.
  * Added filtering by topic and optional start and end times.
  * Added timestamp synchronization to retrieve matching frames, including clear handling for missing matches.
  * Added support for URI-based storage options.

* **Bug Fixes**
  * Added clear errors for unavailable topics, unloaded data, invalid files, missing file metadata, unsupported non-seekable streams, and MCAP files without required indexing.


lightly_studio/src/lightly_studio/core/mcap/reader.py
```diff
@@ -0,0 +1,249 @@
...
+from mcap.summary import Summary
+
+from lightly_studio.core.mcap import matching, topic_kind
+from lightly_studio.core.mcap.errors import DataNotLoadedError, McapAccessError, TopicNotFoundError
+from lightly_studio.core.mcap.matching import MatchFunction
+from lightly_studio.core.mcap.type_definitions import FrameLocator, TopicInfo
+from lightly_studio.type_definitions import PathLike
+
+
+class McapFileReader:
```
ikondrat, line 21: hmm unfortuanelly I can't use McapFileReader for streaming because it assumes to keep connection and for streaming it is critical to have connection per request.

so streaming probably will be handled differently, we can refactor then it later

ikondrat (approved):

PR #2351, opened 2026-09-15
Include the filter in annotation count query keys

gabrielfruet:
## What has changed and why?

The distribution panel and the sidebar annotation class counts ignored the filter. Their query key held only the annotation type and count mode, so selecting a tag kept the cached counts and sent no request.

`buildImageAnnotationCountsQueryKey` now appends the full request (collection ID, filter, annotation type, count mode) to the key. The key still starts with `useImageAnnotationCountsQueryKey` or the caller's override, so annotation mutations keep invalidating these queries.

## How has it been tested?

`npm run test:unit -- --run` in `lightly_studio_view` passes (382 files, 2982 tests), and so does `make static-checks`.

Manual check with the `example_split_work.py` dataset and the Distribution panel open:

- Clicking tag "Alice" sent 5 count requests (sidebar plus the four distribution queries) with the tag ID in `body.filter`. The chart and sidebar counts changed, green_light from 120 to 38. On main the same click sends none.
- Adding tag "Bob" sent 5 more with both tag IDs (110 annotations).
- Switching "Count by" to Samples sent the 4 distribution requests with `count_mode: samples`.
- Deleting an annotation, then Undo, each refetched the sidebar counts.

- [x] Yes

## Summary by CodeRabbit

* **Bug Fixes**
  * Distribution and annotation class counts now update correctly when filters change.
  * Counts remain accurate when filtering by collection, tags, annotation types, or count modes.
  * Labels with no matching annotations now display a count of zero.
  * Count and distribution requests are skipped when all annotation sources are hidden.
  * Invalid label selections are retained until updated counts finish loading.

* **Documentation**
  * Updated the unreleased changelog with filtering-related count improvements.


gabrielfruet:
https://github.com/user-attachments/assets/bc9984c5-c508-455a-8e1f-8464b25781a6

lightly_studio_view/src/routes/datasets/[dataset_id]/[collection_type]/[collection_id]/+layout.svelte
```diff
@@ -573,7 +580,7 @@
         filter: imageAnnotationCountsFilter,
         countMode: distributionCountMode,
         queryKey: distributionAllQueryKey,
-        enabled: distributionPanelVisible
+        enabled: distributionPanelVisible && !$allSourcesHidden
```
Comment (LeonardoRosaa) [inline lightly_studio_view/src/routes/datasets/[dataset_id]/[collection_type]/[collection_id]/+layout.svelte]: This condition is repeated in 4 places. We could extract it.

Comment (gabrielfruet) [reply lightly_studio_view/src/routes/datasets/[dataset_id]/[collection_type]/[collection_id]/+layout.svelte]: Extracted as `distributionCountsEnabled`, thanks.

LeonardoRosaa (approved):

PR #2352, opened 2026-09-16
Drop TagKind uniqueness in TagTable

lukas-lightly:
## What has changed and why?
TagKind should not be part of the uniqueness constraint.

## How has it been tested?
Existing tests.

- [ ] Yes

## Summary by CodeRabbit

- **New Features**
  - Tag names are now unique within each collection, regardless of tag kind.
  - Tag lookups are scoped to a specific collection.
  - Existing duplicate tags are consolidated during database migration, preserving their links.

- **Bug Fixes**
  - Renaming a tag to a name already used in the same collection is now correctly rejected.

- **Tests**
  - Added coverage for cross-kind uniqueness, collection-scoped lookups, renaming conflicts, and migration behavior.


lightly_studio/src/lightly_studio/migrations/versions/1789500000_c9d0e1f2a3b4_make_tag_name_unique_per_collection.py
```diff
@@ -0,0 +1,100 @@
+"""make tag names unique per collection.
```
lukas-lightly, line 1: This is not a very good migration but I'm convinced such a duplicate tag couldn't exist. Annotations live in different collection than samples, so sample and annotation tags can't mix - unless someone messed with the DB outside of our API.

horatiualmasan (approved):
LGTM

PR #2356, opened 2026-09-16
Introduce endpoint to fetch summary about the mcap sequence by MCAP sequence id

ikondrat:
## What has changed and why?

This PR introduces endpoint to provide information about indexed recording by recording_id

## How has it been tested?

By unit tests

The best way to review is to follow commits in https://github.com/lightly-ai/lightly-studio/pull/2356/commits starting from my commit

- [ ] Yes

## Summary by CodeRabbit

- **New Features**
  - Added an MCAP sequences API for retrieving sequence summaries by dataset and sequence.
  - Summaries include recording details, file format, earliest indexed timestamp, camera channels, lidar channels, channel IDs, slot names, and frame IDs.
  - Added MCAP sequence information support, including component metadata and recording URIs.
  - Returns clear not-found responses for unknown, mismatched, or unindexed sequences.
- **Tests**
  - Added coverage for successful MCAP summaries, channel metadata, timestamps, missing sequences, dataset mismatches, and non-MCAP components.


lightly_studio/src/lightly_studio/resolvers/mcap_group_sequence_resolver/get_by_recording_id.py
```diff
@@ -0,0 +1,28 @@
+"""Read the MCAP sequence a recording backs, if any."""
```
horatiualmasan: Is this a leftover form the previous implementation?

lightly_studio/src/lightly_studio/models/recording.py
```diff
@@ -38,3 +39,22 @@ class RecordingTable(RecordingBase, table=True):
 
     # The dataset the recording belongs to.
     dataset_id: UUID = Field(foreign_key="dataset.dataset_id", index=True)
+
+
+class MCAPSequenceDetails(BaseModel):
```
horatiualmasan: We could keep this as RecordingDetails

lightly_studio/src/lightly_studio/services/recording_service/get_summary.py
```diff
@@ -0,0 +1,45 @@
...
+    except ValueError:
+        return None
+
+    if info is None:
+        return None
+
+    if info.recording.dataset_id != dataset_id:
+        return None
+
+    start_log_time_ns = mcap_resolver.get_start_log_time_ns(
```
horatiualmasan, line 42: this will be used for the timeline? would it make sense to also get the end ts?

lightly_studio/src/lightly_studio/resolvers/mcap_group_sequence_resolver/get_info.py
```diff
@@ -0,0 +1,137 @@
...
+) -> list[McapGroupComponentDefinitionView]:
+    """Returns the MCAP slots of a GROUP collection, sorted by slot index.
+
+    Joins the generic slot naming/ordering to the MCAP-specific metadata by
+    `collection_id`. A child collection without an MCAP row (a classic IMAGE/VIDEO
+    slot) is not a component and is omitted.
+    """
+    component_collections_by_id = {
+        child.collection_id: child
+        for child in collection_resolver.get_group_components(
```
horatiualmasan: [optional improvement]
get_group_components lazy-loads each slot's group_component_definition individually, so the query count grows with the number of components: 11 queries per get_info call at 2 components, 17 at 8. We could improve by joining. One join across group_component_definition + mcap_group_component_definition + collection, with ORDER BY group_component_index, makes it 5 and flat. _get_group_collection has the same shape — sample → its collection → the lazy children load is three round trips that collapse into one join.

Measured A/B at 8 components (5 cameras, 3 lidars) on DuckDB, 1000 calls: 11.75 ms → 5.34 ms, ~2.2×. Both versions are flat in sequence count; the win is that it stops scaling with sensor count, which is what matters as rigs grow.

horatiualmasan (changes requested):
some minor comments, otherwise LGTM
The merge removed some unrelated files from embed/remote

lightly_studio/src/lightly_studio/services/recording_service/get_summary.py
ikondrat: It will be used to query initial details representation, I think we don't need end here

horatiualmasan (approved):

PR #2357, opened 2026-09-16
Add the remote embedding transport (remote embedder 1/5)

iunir:
## What has changed and why?

First of five stacked PRs adding a client for the LightlyStudio embedding protocol, so a customer can serve their own model over HTTP and LightlyStudio can query it.

This PR is the HTTP layer only — nothing uses it yet.

- `RemoteTransport` sends the protocol's requests through an injectable `httpx.Client` and reads the answers back into the wire models from `lightly-studio-serve`, so client and server cannot drift.
- `errors.py`: one exception per cause a caller can act on — unreachable, auth, capability, batch-too-large, protocol.
- A status becomes the exception that names it; a 429 or 503 is waited out for as long as `Retry-After` asks, capped.
- `httpx` becomes a direct dependency.

## Size

797 added lines, but only ~190 of them are logic:

| | lines |
| --- | --- |
| Source logic | 190 |
| Docstrings and comments | 180 |
| Blank lines | 96 |
| Tests | 330 |
| `pyproject.toml` | 1 |

`uv.lock` also changes.

## How has it been tested?

28 unit tests against `create_app` with a `TestClient`, plus `httpx.MockTransport` for the answers a conforming server cannot produce. Full backend suite and static checks green.

- [ ] Yes

## Summary by CodeRabbit

* **New Features**
  * Added support for remote text, image, and video embedding through the version 1 protocol.
  * Added clear error reporting for connection, authentication, capability, request-size, and protocol failures.
  * Added protocol compatibility checks between the client and remote server.

* **Reliability Improvements**
  * Automatically retries temporary rate-limit and service-unavailable responses.
  * Provides consistent handling for retry timing and unreachable servers.


lightly_studio/src/lightly_studio/embed/remote/transport.py
```diff
@@ -0,0 +1,408 @@
...
+        Raises:
+            RemoteEmbedderError: If the server gives no answer or answers a status that
+                carries no embeddings.
+        """
+        response = self._send(method=method, path=path, json=json, files=files)
+        for _ in range(_MAX_ATTEMPTS - 1):
+            if response.status_code not in _RETRY_STATUSES:
+                break
+            time.sleep(_retry_wait_seconds(response=response))
+            response = self._send(method=method, path=path, json=json, files=files)
```
iunir:
```suggestion
        attempt = 1
        while True:
            response = self._send(method=method, path=path, json=json, files=files)
            if response.status_code not in _RETRY_STATUSES or attempt >= _MAX_ATTEMPTS:
                break
            attempt += 1
            time.sleep(_retry_wait_seconds(response=response))                
```

Can it be simplified this way?

lightly_studio/src/lightly_studio/embed/remote/transport.py
```diff
@@ -0,0 +1,408 @@
...
+    Args:
+        items: The items to embed. An empty batch yields nothing, so it sends no request.
+        max_batch_size: The largest number of items in one request. Positive, because the
+            caller reads it from ``ServerLimits``.
+
+    Yields:
+        Each chunk together with the index of its first item in ``items``. That index
+        turns the ``kept_indices`` of a chunk back into indices of the whole batch.
+    """
+    assert max_batch_size > 0
```
iunir: Do not use assert outside of tests

lightly_studio/src/lightly_studio/embed/remote/transport.py
```diff
@@ -0,0 +1,408 @@
...
+                method=method, url=path, headers=self._headers, json=json, files=files
+            )
+        except httpx.HTTPError as error:
+            raise RemoteEmbedderUnreachableError(
+                f"The embedding server at {self._client.base_url} did not answer "
+                f"{method} {path}: {error}"
+            ) from error
+
+
+def split_batches(
```
iunir: Should it be a method of RemoteTransport since it knows max batch size?

lightly_studio/src/lightly_studio/embed/remote/transport.py
```diff
@@ -0,0 +1,408 @@
...
+    key press, so the answer is capped.
+
+    Returns:
+        The wait in seconds, between zero and ``_MAX_RETRY_WAIT_SECONDS``.
+    """
+    header = response.headers.get("Retry-After")
+    if header is None:
+        return _DEFAULT_RETRY_WAIT_SECONDS
+    seconds = _parse_retry_after(value=header.strip())
+    if seconds is None:
```
iunir:
```suggestion
    if not seconds:
```
To cover seconds == 0 since otherwise it will retry immediately

lightly_studio/pyproject.toml
```diff
@@ -62,6 +62,11 @@ dependencies = [
     # so one major runs everywhere rather than two that have to stay behaviourally identical.
     "posthog>=6.9,<7.0",
     "lightly-studio-serve>=0.1.0,<0.2.0",
+    # The client of the embedding protocol sends its requests with httpx: it separates the
+    # connect timeout from the read timeout, which a query under a key press needs, and it
+    # follows no redirect unless it is told to, which a request to a customer-supplied URL
+    # needs. Already resolved at runtime through `huggingface-hub`, so this adds no download.
```
iunir:
```suggestion
```
No need

lightly_studio/src/lightly_studio/embed/remote/transport.py
iunir: Applied as suggested — one `_send` call and one exit condition.

lightly_studio/src/lightly_studio/embed/remote/transport.py
iunir: Removed. It went away with `split_batches`: `ServerLimits` already validates `max_batch_size > 0`, so nothing is left to assert.

lightly_studio/src/lightly_studio/embed/remote/transport.py
iunir: Not quite — `RemoteTransport` deliberately holds no state of the server, so it does not know `max_batch_size`; `RemoteEmbedder` is what reads `/v1/describe` and keeps the limits. But the point stands that it should not be a free function here, so I moved it into `embedder.py` (PR 2/5) as a private helper, next to the merge that reverses it. That also puts this file back under the 400-line guideline.

lightly_studio/src/lightly_studio/embed/remote/transport.py
iunir: Applied, slightly widened: `if seconds is None or seconds <= 0`. Your version leaves a `Retry-After` date that has already passed truthy and negative, which clamps to 0 and retries immediately anyway — this covers both.

lightly_studio/pyproject.toml
iunir: Removed.

lightly_studio/src/lightly_studio/embed/remote/transport.py
```diff
@@ -0,0 +1,408 @@
...
+        Args:
+            client: The client to send with. Its ``base_url`` names the server. The caller
+                owns it and closes it.
+            api_key: The token to send as ``Authorization: Bearer``. ``None`` sends no
+                header, which is what an unauthenticated server needs.
+        """
+        self._client = client
+        # A header of the request, not of the client: the caller owns the client, and a
+        # transport must not put a token on a client that it was lent.
+        self._headers = {} if api_key is None else {"Authorization": f"Bearer {api_key}"}
```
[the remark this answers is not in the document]
iunir: we warning our clients to use HTTPs in -serve package

lightly_studio/src/lightly_studio/embed/remote/transport.py
```diff
@@ -0,0 +1,391 @@
+"""The HTTP half of the remote embedder: one request, one answer, one exception.
+
+``RemoteTransport`` sends the requests of version 1 of the LightlyStudio embedding
+protocol and reads the answers back into the wire models of ``lightly_studio_serve``. It
+holds no state of the server: ``RemoteEmbedder`` reads ``/v1/describe`` once and keeps the
+answer for the lifetime of the process.
+
+The module turns a status into the exception that names it and waits out a 429 or a 503
+for as long as the server asks.
```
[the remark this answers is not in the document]
iunir: Correct, thanks — fixed. The docstring now says the client sends a 429 or a 503 again a capped number of times, after a wait that it also caps, and names why (a server can ask for an hour; a text query sits under a key press). Both caps are real: three attempts and 30 seconds.

lightly_studio/tests/embed/remote/test_transport.py
```diff
@@ -0,0 +1,337 @@
...
+from fastapi.testclient import TestClient
+from lightly_studio_serve import protocol
+from lightly_studio_serve.embedder import (
+    Capability,
+    ImageBytesEmbedder,
+    TextEmbedder,
+    VideoBytesEmbedder,
+)
+from lightly_studio_serve.protocol import ServerLimits
+from lightly_studio_serve.server import create_app
```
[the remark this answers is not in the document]
iunir: Fixed — the tests now import the module (`from lightly_studio_serve import protocol, server`) and call `server.create_app(...)`.

lightly_studio/src/lightly_studio/embed/remote/transport.py
```diff
@@ -0,0 +1,408 @@
+"""The HTTP half of the remote embedder: one request, one answer, one exception.
```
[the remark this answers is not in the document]
iunir: Fixed — `transport.py` is 392 lines. `split_batches` moved into `RemoteEmbedder` (3/5 of this stack), where the `max_batch_size` it needs actually lives, and `_check_status` lost its branch ladder.

lightly_studio/src/lightly_studio/embed/remote/transport.py
```diff
@@ -0,0 +1,408 @@
...
+    """Raise the error that names the status of an answer. 200 is the only status with a body.
+
+    A 429 and a 503 arrive here only after the attempts of ``_request`` are spent, so the
+    server stayed busy for as long as the client waits.
+    """
+    status = response.status_code
+    if status == httpx.codes.OK:
+        return
+    where = _where(response=response)
+    detail = _detail(response=response)
```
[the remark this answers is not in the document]
iunir: Fixed — the branches are now a `_STATUS_ERRORS` mapping from status to (exception type, reason), and `_check_status` is 8 lines: return on 200, look up, raise.

lightly_studio/src/lightly_studio/embed/remote/transport.py
```diff
@@ -0,0 +1,408 @@
...
+
+        Raises:
+            RemoteEmbedderError: If the server gives no answer, refuses the request, or
+                answers a body that the protocol does not allow.
+        """
+        request = EmbedTextsRequest(texts=texts)
+        response = self._request(
+            method="POST", path=protocol.EMBED_TEXTS_PATH, json=request.model_dump()
+        )
+        return _parse(model=EmbeddingsResponse, response=response)
```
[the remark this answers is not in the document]
iunir: Real problem, handled one layer up rather than here. `EmbeddingsResponse` only enforces non-negative, ascending and unique, so `kept_indices: [n]` for an n-item request does pass the wire model.

`RemoteEmbedder._check_answer` rejects it (`any(index >= item_count ...)`), covered by `test_embed_text__kept_index_past_the_request`. That is 2/5 of this stack, so on this PR alone the check is genuinely absent.

Keeping it there rather than duplicating it here: the client adds exactly two checks on top of the wire model — the space identity against the cached `/v1/describe`, and the indices against what was sent — and the transport holds no `/describe`, so it cannot do the first. Splitting the pair across two layers costs more than it buys. The tradeoff is that `RemoteTransport` is public and returns a response that is validated by the wire model but not against the request, so a future caller that builds on the transport instead of on `RemoteEmbedder` inherits that check.

lightly_studio/src/lightly_studio/embed/remote/transport.py
```diff
@@ -0,0 +1,415 @@
+"""The HTTP half of the remote embedder: one request, one answer, one exception.
```
michal-lightly: Minor: The description is cryptic (feels like a marketing slogan).

lightly_studio/src/lightly_studio/embed/remote/transport.py
```diff
@@ -0,0 +1,415 @@
...
+class RemoteTransport:
+    """Sends the requests of the embedding protocol to one server.
+
+    The transport holds no address of its own. The ``base_url`` of the client names the
+    server, and every request goes to a relative path of the protocol. A test therefore
+    passes a client that drives an application in process, and production passes a client
+    that opens a socket.
+    """
+
+    def __init__(self, client: httpx.Client, api_key: str | None = None) -> None:
```
michal-lightly, line 100: While we are early in the implementation; I would seriously consider a pure function approach instead of a class. In python that tends to be the easier way:
- RemoteEmbedder can own the client and the api_key
- They can be transparently passed to each of the functions below
- Testing usually becomes much easier

What do you think, should we go the class-way or make it pure?

lightly_studio/src/lightly_studio/embed/remote/transport.py
```diff
@@ -0,0 +1,415 @@
...
+        response = self._request(
+            method="POST", path=protocol.EMBED_TEXTS_PATH, json=request.model_dump()
+        )
+        return _parse(model=EmbeddingsResponse, response=response)
+
+    def embed_image_bytes(self, images: list[bytes]) -> EmbeddingsResponse:
+        """Embed a batch of encoded images on the server.
+
+        Args:
+            images: The encoded images, at most the ``max_batch_size`` of the server.
```
michal-lightly: Minor: We should mention how the images are supposed to be encoded (I assume file bytes for a predetermined list of supported format). It can be mentioned only in the caller, if that's more convenient.

lightly_studio/src/lightly_studio/embed/remote/transport.py
```diff
@@ -0,0 +1,415 @@
...
+
+    def _post_files(self, path: str, items: list[bytes]) -> EmbeddingsResponse:
+        """Send a batch of items as multipart, one part for each item, in input order.
+
+        Raises:
+            ValueError: If ``items`` is empty. httpx then writes no body at all, so the
+                server cannot read the request and answers 400.
+        """
+        if not items:
+            raise ValueError(f"An empty batch cannot be sent to {path}. Return an empty result.")
```
michal-lightly, line 195: Optional: Consider making this a RemoteEmbedderError

lightly_studio/src/lightly_studio/embed/remote/transport.py
```diff
@@ -0,0 +1,415 @@
...
+    ),
+}
+
+# The content type of a multipart part. The server reads the real format from the header
+# of the data, so this value only has to be a type that is not text.
+_OCTET_STREAM = "application/octet-stream"
+
+# The longest message that an exception repeats from a server that this client does not
+# control.
+_MAX_DETAIL_CHARS = 200
```
michal-lightly: Optional: Could be larger.

lightly_studio/tests/embed/remote/test_transport.py
```diff
@@ -0,0 +1,330 @@
...
+        assert description.dimension == DIMENSION
+        assert description.ready
+        assert description.capabilities == [Capability.TEXT]
+
+    def test_describe__protocol_version_mismatch(self) -> None:
+        # A server of another major version is the one most likely to answer another
+        # shape. The version must still be what the message names, so the check runs
+        # before the model.
+        body = {"protocol_version": "2.0", "model": {"name": "acme"}}
+        client = _mock_client(handler=_answers(httpx.Response(status_code=200, json=body)))
```
michal-lightly: We have a guideline to keep tests DAMP (i.e. not DRY), and a related guideline to keep test setup in the test body. Although this cannot be always achieved (e.g. when setup is too complex), often it can.

For example, this test is difficult to read. First, I need to jump down to understand what `_mock_client` does - and that it is just one line. Then, I need to inspect how `handler` variable is used, and finally understand what `_answers` does.

So I'm not saying that everything (e.g. this) has to be rewritten, but in general the preference should be leaning towards clarity even if the tests repeat a little bit.

lightly_studio/tests/embed/remote/test_transport.py
```diff
@@ -0,0 +1,330 @@
...
+def _mock_client(handler: Callable[[httpx.Request], httpx.Response]) -> httpx.Client:
+    """A client that answers from ``handler`` instead of a socket.
+
+    It covers the answers that `create_app` cannot give, such as a wrong protocol version
+    or a server that stays busy. A conforming server is tested against `create_app`.
+    """
+    return httpx.Client(transport=httpx.MockTransport(handler), base_url=BASE_URL)
+
+
+def _answers(*responses: httpx.Response) -> Callable[[httpx.Request], httpx.Response]:
```
michal-lightly: `*responses` is unexpected; let's just type the argument as `list[httpx.Response]`.

Also I'd consider calling it `_yield_answers`.

lightly_studio/src/lightly_studio/embed/remote/transport.py
iunir: agreed to keep class

lightly_studio/src/lightly_studio/embed/remote/transport.py
iunir: rewritten

lightly_studio/src/lightly_studio/embed/remote/transport.py
iunir: increased

lightly_studio/tests/embed/remote/test_transport.py
```diff
@@ -0,0 +1,330 @@
...
+            httpx.Response(status_code=200, json=_embeddings_body()),
+        ]
+        attempts = 0
+
+        def answer(_request: httpx.Request) -> httpx.Response:
+            nonlocal attempts
+            attempts += 1
+            return answers[attempts - 1]
+
+        client = _mock_client(handler=answer)
```
michal-lightly: Why this does not use `_answers`?

lightly_studio/tests/embed/remote/test_transport.py
```diff
@@ -0,0 +1,330 @@
...
+            "kept_indices": [0, 1],
+            "embeddings": [[0.5, -0.5]],
+        }
+        client = _mock_client(handler=_answers(httpx.Response(status_code=200, json=body)))
+
+        with pytest.raises(RemoteEmbedderProtocolError, match="protocol does not allow"):
+            RemoteTransport(client=client).embed_texts(texts=["a dog", "a cat"])
+
+    def test_embed_texts__retries_while_busy(self, mocker: MockerFixture) -> None:
+        sleep = mocker.patch.object(time, "sleep")
```
michal-lightly: Minor: We have a naming convention:
```suggestion
        mock_sleep = mocker.patch.object(time, "sleep")
```

lightly_studio/tests/embed/remote/test_transport.py
```diff
@@ -0,0 +1,330 @@
...
+        assert description.space_key == SPACE_KEY
+        assert description.dimension == DIMENSION
+        assert description.ready
+        assert description.capabilities == [Capability.TEXT]
+
+    def test_describe__protocol_version_mismatch(self) -> None:
+        # A server of another major version is the one most likely to answer another
+        # shape. The version must still be what the message names, so the check runs
+        # before the model.
+        body = {"protocol_version": "2.0", "model": {"name": "acme"}}
```
michal-lightly, line 91: Side note: Do we have `model` in the protocol? We should call it `space_key` for consistency. The users should not know that the underlying table is called `EmbeddingModelTable` - it is not renamed just to deliver the embedding functionality faster.

Likely out of scope of this PR.

michal-lightly (commented):
Adding test comments, they are minor.

lightly_studio/tests/embed/remote/test_transport.py
iunir: done

lightly_studio/tests/embed/remote/test_transport.py
iunir: thanks fixed

lightly_studio/tests/embed/remote/test_transport.py
iunir: done

lightly_studio/tests/embed/remote/test_transport.py
iunir: Merged _mock_client and _answers into a single _client_answering

lightly_studio/tests/embed/remote/test_transport.py
iunir: Then we have to change defined protocol. definitely not in this PR and needs separate discussion

lightly_studio/src/lightly_studio/embed/remote/transport.py
iunir: done

lightly_studio/src/lightly_studio/embed/remote/transport.py
iunir: This is definitely not RemoteEmbedderError, this is wrong input issue

lightly_studio/tests/embed/remote/test_transport.py
michal-lightly: I checked the embed server protocol - model is not a field there, the comment here now explains it nicely. We don't need to change the protocol.

lightly_studio/tests/embed/remote/test_transport.py
```diff
@@ -0,0 +1,321 @@
...
+        assert description.ready
+        assert description.capabilities == [Capability.TEXT]
+
+    def test_describe__protocol_version_mismatch(self) -> None:
+        # A server of another major version is the one most likely to answer another
+        # shape. The version must still be what the message names, so the check runs
+        # before the model. `DescribeResponse` carries no `model` field: this body is
+        # deliberately one that version 1 cannot read.
+        body = {"protocol_version": "2.0", "model": {"name": "acme"}}
+        client = _client_answering(responses=[httpx.Response(status_code=200, json=body)])
```
michal-lightly, line 92: Nice!

michal-lightly (approved):

PR #2358, opened 2026-09-16
Add the RemoteEmbedder client

iunir:
## What has changed and why?

3/6 of the remote-embedder client stack.

- `RemoteEmbedder.connect(client)` reads `/v1/describe` and builds an embedder with exactly the advertised, routable capabilities.
- Requests follow the server's batch limits, merge chunk results, and validate the returned space and kept indices.
- The caller opens the client with `connection.build_client(url)` and owns it. The embedder manages no connection lifetime, because a registered embedder is never closed: `EmbedderRegistry` holds it for the lifetime of the process.
- Three smoke tests here; the exhaustive protocol and lifecycle matrix is #2388.

## How has it been tested?

Smoke tests for connection, embedding and registry resolution. Full remote suite and backend static checks pass at the top of the stack.

- [ ] Yes


lightly_studio/src/lightly_studio/embed/remote/embedder.py
```diff
@@ -0,0 +1,349 @@
...
+            An embedder that implements the interface of every capability that the server
+            advertises and this client routes to.
+
+        Raises:
+            RemoteEmbedderError: If the server gives no answer, rejects the token, answers
+                a description that the protocol does not allow, or advertises no
+                capability that LightlyStudio can use.
+        """
+        http_client = client if client is not None else _build_client(url=url)
+        remote = RemoteTransport(client=http_client, api_key=api_key)
```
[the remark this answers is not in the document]
iunir: Done in 2f22edff. `connect` now keeps the client it opened, `close()` and the context manager close it, and a failure from `describe` or the capability check closes it on the way out. A client passed in by the caller stays the caller's. The registry still holds a registered embedder for the life of the process and has no teardown hook, so that pool lives as long as the process — that part is now a TODO on `_build_client` rather than a silent leak.

lightly_studio/src/lightly_studio/embed/remote/embedder.py
```diff
@@ -0,0 +1,349 @@
...
+    ``EmbeddingResult``. Nothing else in LightlyStudio changes: a caller asks the registry
+    for the capability it needs and calls the method, whether the object behind it runs a
+    model in this process or speaks to a server.
+
+    ``ready`` keeps the default of ``True``. The answer of ``/v1/describe`` is read once,
+    at construction, so the property could only report a state that has passed. A server
+    whose model is still loading answers 503, and the transport waits that out.
+    """
+
+    def __init__(self, transport: RemoteTransport, description: DescribeResponse) -> None:
```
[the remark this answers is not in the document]
iunir: Done in 2f22edff. `__init__` now takes `spec: EmbeddingSpaceSpec` and `limits: ServerLimits`; `connect` reads `/v1/describe` and passes only those. Readiness and capabilities no longer reach the constructor.

lightly_studio/tests/embed/remote/test_embedder.py
```diff
@@ -0,0 +1,339 @@
...
+from fastapi.testclient import TestClient
+from lightly_studio_serve import protocol
+from lightly_studio_serve.embedder import (
+    Embedder,
+    ImageBytesEmbedder,
+    TextEmbedder,
+    VideoBytesEmbedder,
+)
+from lightly_studio_serve.protocol import ServerLimits
+from lightly_studio_serve.server import create_app
```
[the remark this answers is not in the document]
iunir: Already fixed before this review landed: the test imports `from lightly_studio_serve import protocol, server` and calls `server.create_app(...)`.

lightly_studio/src/lightly_studio/embed/remote/embedder.py
```diff
@@ -0,0 +1,370 @@
...
+        Returns:
+            An embedder that implements the interface of every capability that the server
+            advertises and this client routes to.
+
+        Raises:
+            RemoteEmbedderError: If the server gives no answer, rejects the token, answers
+                a description that the protocol does not allow, or advertises no
+                capability that LightlyStudio can use.
+        """
+        http_client = client if client is not None else _build_client(url=url)
```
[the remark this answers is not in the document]
iunir: Done in 2f22edff, along the lines you proposed. `connect` tracks the client it opened, `close()` and `__enter__`/`__exit__` close it and are idempotent, and a `try`/`except BaseException` closes it when `describe` or `_check_routable` raises. An injected client stays caller-owned.

Two differences from the diff: the ownership flag is the client itself (`owned_client = http_client if client is None else None`) rather than a separate bool, and `__init__` takes `spec`/`limits` instead of the whole `DescribeResponse` — the other review comment asked for that. Three tests cover it: the opened client closes, an injected one does not, and a failing `connect` closes what it opened.

lightly_studio/src/lightly_studio/embed/remote/embedder.py
```diff
@@ -0,0 +1,382 @@
...
+    def __init__(
+        self,
+        transport: RemoteTransport,
+        spec: EmbeddingSpaceSpec,
+        limits: ServerLimits,
+        owned_client: httpx.Client | None = None,
+    ) -> None:
+        """Embed through ``transport``, against a server of ``spec`` and ``limits``.
+
+        Use ``connect``.
```
michal-lightly: Minor: When and for what?

lightly_studio/src/lightly_studio/embed/remote/embedder.py
```diff
@@ -0,0 +1,382 @@
...
+            An embedder that implements the interface of every capability that the server
+            advertises and this client routes to.
+
+        Raises:
+            RemoteEmbedderError: If the server gives no answer, rejects the token, answers
+                a description that the protocol does not allow, or advertises no
+                capability that LightlyStudio can use.
+        """
+        http_client = client if client is not None else connection.build_client(url=url)
+        owned_client = http_client if client is None else None
```
michal-lightly: Why do we allow the client to be None? It seems cleaner to always inject it as a dependency. And remove all the `owned_client` complexity.

lightly_studio/src/lightly_studio/embed/remote/embedder.py
```diff
@@ -0,0 +1,382 @@
...
+        )
+        for offset, chunk in chunks:
+            response = self._send(send=send, chunk=chunk, capability=capability)
+            self._check_answer(response=response, item_count=len(chunk))
+            rows.extend(response.embeddings)
+            # A chunk counts its kept indices from its own start.
+            kept_indices.extend(offset + index for index in response.kept_indices)
+        return EmbeddingResult(embeddings=self._to_array(rows=rows), kept_indices=kept_indices)
+
+    def _check_open(self) -> None:
```
michal-lightly: Optional: I'd inline this. It is a simple guard.

lightly_studio/tests/embed/remote/test_embedder.py
```diff
@@ -0,0 +1,65 @@
...
+            assert isinstance(remote, TextEmbedder)
+            assert isinstance(remote, ImageBytesEmbedder)
+            result = remote.embed_text(texts=["first", "second"])
+
+        assert result.kept_indices == [0, 1]
+        np.testing.assert_array_equal(result.embeddings, np.array([[0.5, -0.5]] * 2))
+
+    def test_close__closes_owned_client(self, monkeypatch: pytest.MonkeyPatch) -> None:
+        client = TestClient(server.create_app(embedder=_TextImageEmbedder()))
+        monkeypatch.setattr(connection, "build_client", lambda **_kwargs: client)
```
michal-lightly: Minor: We usually use `mocker: MockerFixture` and `mocker.object.patch(...)`. Is there a reason for monkeypatch?

lightly_studio/tests/embed/remote/test_embedder.py
```diff
@@ -0,0 +1,65 @@
...
+        return EmbeddingSpaceSpec(space_key=SPACE_KEY, dimension=DIMENSION)
+
+    def embed_text(self, texts: list[str]) -> EmbeddingResult:
+        return _result(count=len(texts))
+
+    def embed_image_bytes(self, images: list[bytes]) -> EmbeddingResult:
+        return _result(count=len(images))
+
+
+class TestRemoteEmbedder:
```
michal-lightly, line 27: This is missing a test that calls the embed function (through the "extended class"). Is that planned for a follow-up?

lightly_studio/src/lightly_studio/embed/remote/connection.py
```diff
@@ -0,0 +1,51 @@
...
+
+    The embedder owns this client and closes it in ``RemoteEmbedder.close``. A registered
+    embedder is never closed, because ``EmbedderRegistry`` holds it for the lifetime of the
+    process and offers no teardown, so its pool lives as long as the process does.
+    """
+    # TODO(Iunir, 09/2026): Close a registered embedder when `EmbedderRegistry` gains a
+    # teardown hook.
+    return httpx.Client(
+        base_url=url, follow_redirects=False, timeout=httpx.Timeout(_DEFAULT_TIMEOUT_SECONDS)
+    )
```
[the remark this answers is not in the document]
michal-lightly: How do you plan to address this? Is it in the scope of this PR?

michal-lightly (changes requested):

lightly_studio/src/lightly_studio/embed/remote/embedder.py
iunir: Clarified — the docstring now says: "Use `connect`, which reads `/v1/describe` and passes what it learns here. This constructor takes the answer as given and validates nothing against the wire."

lightly_studio/src/lightly_studio/embed/remote/embedder.py
iunir: Done — inlined into `_embed`.

lightly_studio/tests/embed/remote/test_embedder.py
iunir: It's covered — test_connect_and_embed connects against a server advertising TEXT+IMAGE_BYTES, which composes RemoteEmbedder with both route classes, then calls embed_text on the composed instance end-to-end.

lightly_studio/tests/embed/remote/test_embedder.py
iunir: Keeping monkeypatch here — it's already the pattern this test file uses for patching connection.build_client, so switching just this one instance to mocker would split the style within the same file.

lightly_studio/src/lightly_studio/embed/remote/batching.py
```diff
@@ -0,0 +1,64 @@
...
+            chunk_bytes = 0
+        chunk.append(item)
+        chunk_bytes += item_bytes
+    if chunk:
+        yield offset, chunk
+
+
+def text_size(text: str) -> int:
+    """The bytes that one string puts in a JSON body, before it is escaped."""
+    return len(text.encode("utf-8"))
```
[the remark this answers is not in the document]
iunir: Already handled — text_size uses json.dumps(text, ensure_ascii=False)[1:-1] to account for JSON escaping, and there's a boundary test for it (test_text_size__escaped_characters).

lightly_studio/src/lightly_studio/embed/remote/embedder.py
iunir: Keeping the distinction: connect(url, api_key) opens and owns its client, and must close it on failure/close(); connect(url, client=...) lets a caller (mainly tests, via a TestClient/mock transport) inject one it keeps owning. Always closing regardless of origin would close a client the caller still needs. #2388 adds explicit coverage for both paths (test_close__closes_the_client_that_connect_opened vs test_close__keeps_an_injected_client_open), so the contract is at least pinned down now even if the shape stays.

lightly_studio/src/lightly_studio/embed/remote/connection.py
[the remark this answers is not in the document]
iunir: Tracked separately in #2359 (Check remote embedder URLs before connecting), kept out of this stack for a focused security review: strict mode requires HTTPS and rejects loopback/private/link-local/reserved/multicast/unspecified resolved addresses, with an opt-out (LIGHTLY_STUDIO_REMOTE_EMBEDDER_ALLOW_PRIVATE_URLS) for the normal self-hosted case.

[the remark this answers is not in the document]
iunir: I think this is how we designed. We should accept this.
This is user own data and user is responsible to set up proper URL + API key. Of course there is a risk is he would set some malicious URL and will leak his data there. We cannot add whitelists of acceptable URLs. I would prefer to not force user to use HTTPs, but in local -serve package we would warn him. Actually we can warn here as well.

If we will make https mandatory, then it will cause extra friction to try and test app.

I would reject this code rabbit request @michal-lightly

lightly_studio/src/lightly_studio/embed/remote/embedder.py
iunir: Agreed, done in 4a6da501. The `None` was doing two jobs, and `url` silently went unused whenever a client was passed, which is the worse half of it.

Now `connect(url, api_key)` opens its own client and owns it, `connect_with_client(client, api_key)` sends on yours and leaves it open, and a private `_from_client` holds the `describe()` call they share. Ownership follows from the entry point instead of from an argument being absent, and no call site passes a `url` that does nothing — that dropped 26 dead `url=BASE_URL` arguments from the tests in #2388.

`owned_client` survives as an internal field, since `close()` still has to know whether to close. Worth noting that in production it never matters: `EmbedderRegistry` has no teardown hook, so a registered embedder is never closed at all (TODO in `connection.py`). The two lifecycle tests in #2388 now line up one per factory.

lightly_studio/tests/embed/remote/test_embedder.py
iunir: No good reason — switched to `mocker` in 697a6ff8, and in the three places on #2388 too.

`mocker.patch.object(connection, "build_client", return_value=opened)` also drops the `lambda **_kwargs:` that the `setattr` form needed.

lightly_studio/src/lightly_studio/embed/remote/connection.py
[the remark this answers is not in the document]
michal-lightly: @iunir CodeRabbit was too eager to answer a question aimed at you :)

There are two raised issues:
* attacker can configure a local UI: Will be addressed with LIGHTLY_STUDIO_REMOTE_EMBEDDER_ALLOW_PRIVATE_URLS for the hosted version
* http: Let's not force https, give user a choice. Let's not address.

lightly_studio/src/lightly_studio/embed/remote/embedder.py
```diff
@@ -0,0 +1,380 @@
...
+        Returns:
+            An embedder that implements the interface of every capability that the server
+            advertises and this client routes to.
+
+        Raises:
+            RemoteEmbedderError: If the server gives no answer, rejects the token, answers
+                a description that the protocol does not allow, or advertises no
+                capability that LightlyStudio can use.
+        """
+        http_client = connection.build_client(url=url)
```
michal-lightly: Side note: I was suggesting to resolve the other way - always accept the client as non-optional (i.e. the caller must provide it). In that case, we'd need to provide a helper to the user in some public api. This also works (with more mocking in tests); I wanted to raise this though if the change was not intentional.

lightly_studio/src/lightly_studio/embed/remote/embedder.py
```diff
@@ -0,0 +1,380 @@
...
+        return cls._from_client(http_client=http_client, api_key=api_key, owned_client=http_client)
```
michal-lightly: Why does `_from_client` exist and it is not inlined here? For easier testing?

lightly_studio/src/lightly_studio/embed/remote/embedder.py
iunir: You were right, and the earlier answer was wrong. Done in fc8fba4f.

There is one `connect(client, api_key)` now. The caller does `connection.build_client(url)` and owns it, and `RemoteEmbedder` manages no connection lifetime at all — `connect_with_client`, `owned_client`, `close()`, the context manager and the closed guard are gone, 85 lines.

What settled it: nothing ever closes a remote embedder in production. `EmbedderRegistry` has no teardown and holds a registered embedder for the lifetime of the process, so all of that code served a path we never take.

On the testing argument — it was backwards. `test_embedder.py` now patches nothing at all; every test just passes its own client. The only mocking left in the package is `time.sleep` in the retry tests.

Intended use: nothing constructs these yet. The wiring lands in 5/6, where the registry builds one from the stored `remote_embedder_url`, and that caller owns the client.

michal-lightly (approved):

PR #2370, opened 2026-09-16
Add the Announce Release workflow

iunir:
## What has changed and why?

Part 2 of 3 for [LIG-10556](https://linear.app/lightly/issue/LIG-10556/public-repo-automate-the-studio-issues-and-feedback-changelog). Stacked on #2369, which renders the message. This adds the workflow that posts it, dispatchable on its own; #2371 calls it from Finish Release.

- Reads the changelog **at the tag** (not from `main`, which has moved on), renders it with the tooling from #2369, and posts it to `#studio-issues-and-feedback`.
- The tooling is checked out from `main`, never from the released tag — the same rule Draft Release states, and it matters more here because the job holds the bot token.
- Posts with `curl`, not the `slack_sdk` step the private `release_documentation.yml` uses. That step resolves a PyPI package at job time *inside the job holding the bot token*, in a public repo, in a tooling tree whose design is stdlib-only. Three things the port needs that the original hides: Slack answers a rejected message with `ok: false` and **HTTP 200**, so the status code alone reports a silent success; `unfurl_links`/`unfurl_media` off, or every announcement grows a GitHub preview card; and the message reaches the payload through `jq --rawfile`, so neither a `${{ }}` expansion nor shell quoting ever sees it.
- `SLACK_ACTION_BOT_TOKEN` goes in a **`slack` GitHub environment**, so it is unreadable from any job that does not declare it. Restricted to `main`, like `pypi` — both are reached from a run whose ref is the branch Finish Release was dispatched from, never the release tag, so the `v*` tag policy the ticket suggested is unreachable. No required reviewers: a gate there would pause a run that has already published to PyPI.
- The `workflow_dispatch` half is the re-post hatch and the only faithful rehearsal: pasting the text into Slack's composer proves nothing, because the composer treats `<url|label>` as literal text.

One accepted cost of living in the public repo: this file discloses the channel name and the message template. Neither is a secret in any meaningful sense.

## Before this can run

Manual, and not in this PR:

1. Slack app scoped to **`chat:write` only** — not `chat:write.public`; invite the bot instead, so a leaked token cannot post into arbitrary channels.
2. `/invite` the bot to `#studio-issues-and-feedback` and to a test channel.
3. Create the `slack` environment (selected branches → `main`, no required reviewers) and add `SLACK_ACTION_BOT_TOKEN`.

## How has it been tested?

- `actionlint` clean on this file and on `finish_release.yml`.
- Not yet dispatched — it needs the environment and the token first. The rehearsal plan once those exist: dispatch with `tag: v1.1.1` and `channel: <test channel>`, then with `tag: v1.1.0` for the ~4800-character worst case. Checks: the title renders as a link, `•` and indented `◦` bullets, no unfurl card, and `tag_depth > 1` shows a literal `>` rather than `&gt;` — that last one is the only rendering claim that cannot be checked offline. If Slack does not decode it inside a code span, the fix is to skip escaping inside backticks, one function and one test in #2369.

Alternative reviewer: @michal-lightly.

- [ ] Yes

## Summary by CodeRabbit

- **New Features**
  - Added automated Slack release announcements with package-specific prefixes.
  - Added support for rendering changelog entries into formatted Slack messages, including release links and character-limit handling.
  - Added a reusable and manually dispatchable release-announcement workflow with optional channel selection.

- **Bug Fixes**
  - Improved Slack formatting by escaping special characters and preserving bullet structure.

- **Tests**
  - Added coverage for Slack message rendering, release-tag parsing, truncation, escaping, and workflow command failures.


.github/workflows/announce_release.yml
```diff
@@ -0,0 +1,90 @@
...
+    inputs:
+      tag:
+        description: Tag of the published release, e.g. v1.2.3.
+        type: string
+        required: true
+
+permissions: {}
+
+env:
+  SLACK_CHANNEL: studio-issues-and-feedback
```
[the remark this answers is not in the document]
iunir: Addressed in 1c75b9f6.

`channel` is back as a `workflow_dispatch` input, defaulting to empty, and the post step resolves `${{ inputs.channel || env.SLACK_CHANNEL }}` — so a rehearsal names a test channel and a normal release still goes to `#studio-issues-and-feedback`.

It is declared on `workflow_dispatch` only, not on `workflow_call`: Finish Release never names a channel, and on a called run the reference is simply empty, so the fallback picks up the production channel.

.github/workflows/finish_release.yml
```diff
@@ -196,3 +196,20 @@ jobs:
             --draft=false \
             --notes-file "${RUNNER_TEMP}/release-notes.md"
           echo "::notice::Published ${TAG}."
+
+  # After `publish`, so nothing is announced that is not on PyPI and public - which is
```
mihnea-necsulescu: I think "After `publish`, so nothing is announced that is not on PyPI and public" should be enough

.github/workflows/announce_release.yml
```diff
@@ -0,0 +1,105 @@
+name: Announce Release
+
+# Posts a released changelog section to the Slack channel sales reads. Called by Finish
```
mihnea-necsulescu: the first sentence should be enough here as well

.github/workflows/announce_release.yml
```diff
@@ -0,0 +1,105 @@
...
+          PYTHONPATH: ${{ github.workspace }}/.github/scripts
+          REPO: ${{ github.repository }}
+          TAG: ${{ inputs.tag }}
+          CHANGELOG: ${{ steps.package.outputs.changelog }}
+          # Declared on workflow_dispatch only, so it is empty on a workflow_call run.
+          CHANNEL: ${{ inputs.channel || env.SLACK_CHANNEL }}
+        # The changelog is read at the tag, not from main, which has moved on since. The
+        # tooling writes the whole `chat.postMessage` request, so the message never passes
+        # through `${{ }}` expansion or shell quoting, and the request shape is covered by
+        # the tooling's tests rather than by this file. Its stdout is the rendered text.
```
mihnea-necsulescu: this is also kind of long

.github/workflows/announce_release.yml
```diff
@@ -0,0 +1,105 @@
...
+
+      - name: Post the announcement
+        # Undeclared on workflow_dispatch, so it is empty there and a manual re-post
+        # still fails loudly.
+        continue-on-error: ${{ inputs.tolerate-failure == true }}
+        # Pinned by digest: this is the one step that sees the bot token. `errors` is
+        # off by default, and Slack rejects a message with `ok: false` and HTTP 200, so
+        # without it a rejected post reports success. `retries` defaults to 5, which is
+        # the reason for using the action over a bare curl. `payload-file-path` rather
+        # than an inline `payload:`, so nothing expands the rendered changelog.
```
mihnea-necsulescu: same here

mihnea-necsulescu (approved):
LGTM, but the comments are very verbose and hard to follow - could you trim them?

.github/workflows/announce_release.yml
iunir: done

.github/workflows/announce_release.yml
iunir: thanks

.github/workflows/announce_release.yml
```diff
@@ -0,0 +1,94 @@
...
+          set -euo pipefail
+          gh api "repos/${REPO}/contents/${CHANGELOG}?ref=${TAG}" \
+            -H "Accept: application/vnd.github.raw" > "${RUNNER_TEMP}/CHANGELOG.md"
+          release_url="$(gh release view "${TAG}" --repo "${REPO}" --json url --jq .url)"
+          python3 -m prepare_release render-slack-message \
+            --changelog "${RUNNER_TEMP}/CHANGELOG.md" \
+            --tag "${TAG}" \
+            --release-url "${release_url}" \
+            --channel "${CHANNEL}" \
+            --output "${RUNNER_TEMP}/slack-payload.json" >> "$GITHUB_STEP_SUMMARY"
```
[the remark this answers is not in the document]
iunir: merged

PR #2376, opened 2026-09-16
Read the example dataset from `dataset_examples/` everywhere

gabrielfruet:
## What has changed and why?

One directory, two names. `make download-example-dataset` cloned into `lightly_studio/datasets` and every `EXAMPLES_*` path in `.env.example` pointed there, but `index_datasets.py` and `index_predictions.py` hardcode `dataset_examples/`, the default of `download_example_dataset()`. They found nothing. `example_activitynet.py` takes it as an env-var fallback, so it ran only with a `.env` in place. The e2e matrix runs none of the three, so the split went unnoticed.

This moves the Make target, the `EXAMPLES_*` paths, the remaining e2e scripts and CONTRIBUTING.md onto `dataset_examples/`. `download.py` is untouched. `lightly_studio/Makefile` keeps `datasets/mcap/...` for unrelated MCAP files. The e2e cache path and its manual key both change, so the first CI run repopulates the cache.

## How has it been tested?

Rename only. The e2e matrix covers the renamed scripts: `index_general.py`, `index_video.py`, `index_groups.py`, `index_evaluations.py`, `index_distribution.py`.

- [ ] Yes

## Summary by CodeRabbit

* **Documentation**
  * Updated setup and testing instructions to use the new example dataset directory.

* **Chores**
  * Updated example dataset downloads, configuration defaults, end-to-end examples, and CI caching to use the new dataset location.
  * Refreshed the CI cache key to ensure updated example data is used.


lightly_studio/src/lightly_studio/cli.py
```diff
@@ -110,3 +114,24 @@ def gui(
         properties={"launch_source": LaunchSource.GUI.value},
     )
     lightly_studio.start_gui(host=host, port=port)
+
+
+def _quickstart_download_dir() -> Path | None:
```
gabrielfruet: This here is not necessary.

PR #2378, opened 2026-09-16
perf: speed up video embedding by decoding videos in parallel

lorinczszabolcs:
`PerceptionEncoderEmbedder.embed_videos` decoded videos one at a time before the model forward, and video embedding is dominated by decode. This decodes videos on a bounded, order-preserving worker pool that overlaps decoding with the forward, using the same `thread_imap_lazy` helper and `get_media_worker_count()` as the image, frame, and crop embedding paths.

The output is unchanged. The pool preserves input order, and all `FileOutcomeReport` and `kept_indices` bookkeeping stays on the consumer thread in input order, so the embeddings and the skipped-file handling are identical to the sequential version.

This supersedes #2206. That PR sped up the old `PerceptionEncoderEmbeddingGenerator`, but the video path has since moved to `PerceptionEncoderEmbedder`, so the same speedup is applied to the new embedder here.

## How has it been tested?

- The existing `tests/embed/test_perception_encoder_embedder.py` video tests pass, including the interleaved broken and missing test that checks the `kept_indices` order through the pool.
- A local benchmark on 24 copies of a real 1080p clip (8 sampled frames per video, CUDA forward) confirmed the embeddings and `kept_indices` are byte-identical to the sequential version at 1, 2, 4, 8, and 16 workers (maximum absolute difference 0.0), and measured about 2.4x faster end to end at the default worker count (11.9s to 5.0s). The speedup scales with the number of videos up to the worker count.

- [x] Yes

## Summary by CodeRabbit

- **Performance**
  - Video embedding is now faster, with processing speeds improved by up to approximately 2.4×.
  - Videos are decoded in parallel while model processing is underway, reducing overall embedding time.

- **Reliability**
  - Results continue to be processed in the original input order.
  - Video processing continues to report file-related errors consistently, including missing or damaged video files.


lightly_studio/src/lightly_studio/embed/perception_encoder_embedder.py
```diff
@@ -210,6 +217,32 @@ def preprocessed_videos_iter() -> Iterator[torch.Tensor]:
...
+        preprocess: Transform applied to each sampled frame to produce a model input tensor.
+
+    Returns:
+        ``(index, filepath, frames, signal)`` where ``frames`` is the stacked model input on
+        success and ``None`` on a tolerated failure, and ``signal`` is the raised
+        ``MissingInputFileError`` or ``BrokenInputFileError`` on failure and ``None`` on success.
+    """
+    index, filepath = item
+    try:
+        frames = _load_video_frames(filepath, preprocess)
```
mrpositron: The Python guide requires keyword arguments for project functions. Use `_load_video_frames(filepath=filepath, preprocess=preprocess)`.

lightly_studio/src/lightly_studio/embed/perception_encoder_embedder.py
```diff
@@ -210,6 +217,32 @@ def preprocessed_videos_iter() -> Iterator[torch.Tensor]:
...
+
+    Touches no shared state, so the report and kept-index bookkeeping stay single-threaded and
+    in input order on the consumer.
+
+    Args:
+        item: The video's input index paired with its path or URL.
+        preprocess: Transform applied to each sampled frame to produce a model input tensor.
+
+    Returns:
+        ``(index, filepath, frames, signal)`` where ``frames`` is the stacked model input on
```
mrpositron: The Python guide requires tensor shapes in docstrings. Document `frames` as `(N, C, H, W)` and explain the dimensions, including N = VIDEO_FRAMES_PER_SAMPLE.

mrpositron (commented):
Thanks for your contribution! Everything looks good. Just left some code-style comments.

mrpositron (approved):
LGTM! 🚀

PR #2266, opened 2026-09-09
feat: read aggregate evaluation metrics from the Python API

lorinczszabolcs:
## What has changed and why?

The Python SDK can read an evaluation run's confusion matrix, but it cannot read the aggregate metrics. A user has to derive precision, recall, and F1 from the raw matrix by hand. This adds one read method to `ImageDatasetEvaluate`:

- `evaluate().metrics(run_id)` returns an `EvaluationMetrics` object: per-class precision, recall, F1, and support, plus micro-averaged precision, recall, and F1 over the pooled counts. Accuracy is set for classification runs and is `None` for object detection, where predictions and ground truths are matched rather than compared one to one. Object detection and classification are supported. Segmentation runs have no confusion matrix and raise `NotImplementedError`.

The metrics are derived from the run's stored confusion matrix, so this needs no schema change and no new dependency.

This is the second of three staged PRs for #2019, in the order confirmed on the issue (readback → confusion-derived aggregates → mean average precision).

The new `metrics()` method and the existing `confusion_matrix()` method share a small private helper, `_confusion_matrix_with_run()`, which fetches the run, checks that it belongs to this dataset, checks the task type, and returns the run with its matrix.

## How has it been tested?

- New unit tests for the pure aggregate function: classification metrics, object detection with false-positive and false-negative buckets, a zero-division case, and an empty matrix.
- New unit tests for the method: the happy path, run not found, an unsupported task type, and a run from another dataset.
- `ruff`, `ruff format`, and `mypy` pass.
- The evaluation, service, and route test suites pass.
- The docs build passes with `--strict`; the new `EvaluationMetrics` and `ClassMetrics` models render in the API reference.

## Did you update [CHANGELOG.md](../CHANGELOG.md)?

- [x] Yes
- [ ] Not needed (internal change)

## Summary by CodeRabbit

- **New Features**
  - Added an SDK method for retrieving aggregate evaluation metrics by run.
  - View per-class and micro-averaged precision, recall, and F1 scores.
  - Classification evaluations also include accuracy; accuracy is unavailable for object-detection evaluations.
  - Added aggregate metric output to evaluation examples.

- **Documentation**
  - Documented the metrics API, returned values, and usage examples.
  - Clarified support for classification and object-detection evaluations; segmentation remains unsupported.

PR #2381, opened 2026-09-17
feat: read mean average precision from the Python API

lorinczszabolcs:
## What has changed and why?

Follows #2266 and continues the evaluation read-back API from #2019. Adds `dataset.evaluate().mean_average_precision(run_id)`, so a script can read the per-class and mean average precision of a stored object-detection run without the GUI.

It re-matches the run's predictions and ground truths across IoU 0.50 to 0.95 in steps of 0.05, integrates the per-class precision-recall curve with 101-point interpolation, and returns the per-class AP and the mean. Only object detection is supported. Other task types raise a clear error.

Two design points for review:
- This is a separate, purpose-named method rather than folding into `metrics()`. Unlike the confusion-matrix aggregates that `metrics()` returns, mAP needs a full re-match over the IoU thresholds, so keeping it separate keeps `metrics()` cheap.
- The mAP follows COCO's greedy matching and 101-point interpolation over IoU 0.50–0.95, but is not a drop-in pycocotools replacement as it has no maxDets cap, area-range breakdown, or crowd/ignore handling, but I assumed these are acceptable deviations.

## How has it been tested?

- `ruff check`, `ruff format --check`, `mypy`, and strict `mkdocs build` pass.
- New `tests/evaluation/test_average_precision.py` covers the average-precision computation, including the empty-input and no-prediction edge cases, and new facade tests in `test_image_dataset_evaluate.py` cover the success path plus the unsupported-task and run-not-found errors. Run them with `uv run pytest tests/evaluation`.

- [x] Yes

## Summary by CodeRabbit

- **New Features**
  - Added Python SDK support for reading mean and per-class average precision for object-detection runs.
  - Results include per-class scores and the IoU thresholds used for evaluation.

- **Documentation**
  - Added API and workflow documentation with usage examples.
  - Documented the COCO-style evaluation methodology and supported thresholds.

- **Tests**
  - Added coverage for perfect detections, false positives, multiple classes, missing predictions, and invalid runs.


JonasWurst:
Sorry for the delayed response. Thank you so much for your contributions, it is highly appreciated!

TL;DR: In general, we definitely want to add average metrics support. But there are some open design questions we should first align on.

Following issues:
- The planning envisioned such metrics to be persisted in the DB. The reason is simple: the evaluation runs can be quite expensive for larger datasets. In contrast to the non averaged version (one of your previous PRs), we can not reuse the matched annotations, which makes this operation very expensive on re-run.
- You are proposing to add a new method. Internally, we did not yet decide on the API layout yet. Intuitively, I would have assumed that this would be returned by something like ` .metrics(averaged_values=True)`.

If you are interested, we can align on the details on how to make this happen. Alternatively, I can take it internally for the Lightly Team to look into this.

lorinczszabolcs:
No worries, and thanks for the feedback, @JonasWurst! I wanted to go through the evaluation code across task types before answering.

I would suggest persisting the per-class values once during `evaluate()` and recompute, in one generic table `evaluation_class_metric(evaluation_run_id, class_name, metric_name, value)` that mirrors `evaluation_sample_metric`. It would hold average precision for detection and instance segmentation and IoU for semantic segmentation, and mAP and mIoU are the means of its rows. Since the values are stored, I would return them from `.metrics()` by default instead of behind an `averaged_values=True` flag, but it's your call which one would you prefer.

In terms of PRs, I'm thinking
1. The `evaluation_class_metric` table
2. Detection mAP
3. Instance-segmentation mask mAP
4. Semantic-segmentation IoU and mIoU

Does this match what you had in mind?

PR #2383, opened 2026-09-17
Add register_default_embedder public API

michal-lightly:
## What has changed and why?

Adds `register_default_embedder`, the public replacement for the removed `set_default_embedding_model` (LIG-10708).

- Thin wrapper over `EmbedderRegistry.register`, exposed as `ls.register_default_embedder`.
- Passes through `bootstrap_for` to control which capabilities the embedder becomes the default for.

Examples and docs migrate in a follow-up.

## How has it been tested?

- New unit tests for delegation (default and explicit `bootstrap_for`).
- `make static-checks` and `make test` pass.

- [ ] Yes

## Summary by CodeRabbit

- **New Features**
  - Added a public API for registering embedders and optionally assigning them as defaults for selected capabilities.
  - Registered embedders can be used automatically when adding images to datasets.
  - The registration API is available directly from the main package.
  - Embedders can be registered without assigning default capabilities.
  - Registration validates supported capabilities and prevents incompatible embedding specifications.

- **Improvements**
  - Improved compatibility when assigning defaults using different set-like capability collections.
  - Renamed the capability-selection parameter to `for_capabilities` for clarity.


lightly_studio/src/lightly_studio/embed/public_api.py
```diff
@@ -0,0 +1,31 @@
...
+from __future__ import annotations
+
+from lightly_studio_serve.embedder import Capability, Embedder
+
+from lightly_studio.embed import embedder_registry
+
+
+def register_default_embedder(
+    embedder: Embedder,
+    bootstrap_for: set[Capability] | None = None,
```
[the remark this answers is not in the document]
michal-lightly: Addressed, widened the input type to `Set`.

lightly_studio/src/lightly_studio/embed/embedder_registry.py
```diff
@@ -8,6 +8,7 @@
 from __future__ import annotations
 
 import logging
+from collections.abc import Set
```
iunir, line 11: Why not to use built in lowercase set? Is it because of Python 3.9?

lightly_studio/src/lightly_studio/embed/public_api.py
iunir, line 1: Do we plane to add there anything else? File name sounds too generic

iunir (approved):

lightly_studio/src/lightly_studio/embed/embedder_registry.py
michal-lightly: The reason is to allow a broader set of values to be passed in - we do not rely on the `set` class, but just on a part of its interface. With `abc.Set`, users can call the function also e.g. by passing a `frozenset`.

lightly_studio/src/lightly_studio/embed/public_api.py
michal-lightly: I was thinking to put here functions that are exposed to the users - e.g. in the docs. The public interface that we eventually want to keep fixed.

I am not sure how the remote embedder will be registered, that function can be added also here (if it will be standalone).

I am open to change it though, do you have a suggestion?

lightly_studio/src/lightly_studio/embed/embedder_registry.py
iunir: I would prefer to keep things simple. frozenset is more exotic type rather than set and usually chosen when there is real need of it, let's say user want to use it as dict keys. But abc.Set also fine if you wish.

lightly_studio/src/lightly_studio/embed/public_api.py
iunir: I also don't have a strong opinion here it can be just a function of embedder_registry added to __init__.py

iunir (approved):

lightly_studio/src/lightly_studio/embed/embedder_registry.py
michal-lightly: This does not restrict the users in any way and it is the proper way to type things (we also have it in the guideline). Will keep.

lightly_studio/src/lightly_studio/embed/public_api.py
michal-lightly: I'd like to have it separate - it will also be easier to add it in docs I believe. I'll keep the name then, we can rename if we come up with something better.

PR #2385, opened 2026-09-17
Migrate examples to the register_default_embedder API

michal-lightly:
## What has changed and why?

Move the three embedding examples off the deleted `EmbeddingManager` / `set_default_embedding_model` API onto the capability-split `Embedder` interfaces and `register_default_embedder`.

- `example_custom_embedding_model.py` — embeds images, image crops and text into one shared space; batched `embed_text`.
- `example_load_existing_embeddings.py` — image-only; unsupported capabilities fall back to the built-in default instead of raising.
- `example_lightlytrain_explore.py` — vision-only, drops the text stub.

Removes the `# mypy: ignore-errors` shims and migration TODOs.

Stacked on `michal-lig-10708-register-default-embedder`.

## How has it been tested?

Examples are not part of the test suite. `ruff` and `mypy` pass on the three files.

- [ ] Yes

## Summary by CodeRabbit

- **Refactor**
  - Updated embedding examples to use the current embedder interfaces and default-embedder registration workflow.
  - Revised custom, LightlyTrain, and existing-embedding examples to return standardized embedding results.
  - Simplified image embedding examples with consistent path-based inputs and built-in progress handling.
  - Updated example terminology and usage guidance to match the current API.
  - The LightlyTrain and existing-embedding examples now focus on image-path embedding; text, crop, and in-memory image support is no longer shown.


lightly_studio/src/lightly_studio/examples/example_lightlytrain_explore.py
```diff
@@ -43,13 +35,13 @@
 IMAGENET_STD = (0.229, 0.224, 0.225)
 
 
-class LightlyTrainEmbeddingGenerator(ls.ImageEmbeddingGenerator):
+class LightlyTrainEmbedder(ImagePathEmbedder):
```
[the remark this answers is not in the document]
michal-lightly: The example only exercises the image path.

lightly_studio/src/lightly_studio/examples/example_lightlytrain_explore.py
```diff
@@ -91,40 +83,14 @@ def embedding_space_spec(self) -> ls.EmbeddingSpaceSpec:
...
-    def embed_images(self, filepaths: list[str], show_progress: bool = True) -> EmbeddingResult:
+    def embed_images(self, paths: list[str]) -> ls.EmbeddingResult:
         """Embed a batch of images, returning one row per readable input path."""
         return image_embedding.embed_image_files_batched(
-            filepaths=filepaths,
-            context=self._embedding_context(),
-            show_progress=show_progress,
-        )
-
-    def embed_image_crops(
```
JonasWurst, line 108: Don't we want to keep the crop embedding?

JonasWurst (changes requested):
I think the train example should keep its crop embedding

lightly_studio/src/lightly_studio/examples/example_lightlytrain_explore.py
michal-lightly: That depends on the intent of the tutorial. Because no annotation crops are ingested, it will never get called. Tagging @gabrielfruet.

gabrielfruet: I don't show crop embeddings in the tutorial at all

The problem is that the user would be unable to explore the crop embeddings, even though they can exist. So I think we should add this path in the embedder. It shouldn’t be much work.

michal-lightly: @gabrielfruet I'm not sure if I understand. Do you suggest to keep `embed_image_crops` in the example, even though it is not used in the tutorial?

JonasWurst (approved):
LGTM

PR #2386, opened 2026-09-17
Split remote embedding batches by server limits

iunir:
## What has changed and why?

1/6 of the active remote-embedder client stack. Base: `main`.

- Split input batches on both the server's item-count and request-byte limits.
- Preserve each chunk's offset so later result merging can restore batch-relative indices.
- Measure JSON text after escaping; non-ASCII and escaped characters therefore count correctly.

## How has it been tested?

9 focused unit tests cover count limits, byte limits, oversized single items, empty inputs, UTF-8, and JSON escaping. The complete remote test suite and backend static checks pass at the top of the stack.

- [ ] Yes
- [x] Not needed (internal helper)

## Summary by CodeRabbit

- **New Features**
  - Added batching support for embedding requests, splitting items according to item-count and request-size limits.
  - Oversized items are sent individually to keep requests within valid limits.
  - Added accurate UTF-8 size calculation for text content, including non-ASCII and escaped characters.

- **Tests**
  - Added coverage for batch limits, empty and exact-size inputs, oversized items, and text byte measurement.


lightly_studio/tests/embed/remote/test_batching.py
```diff
@@ -0,0 +1,70 @@
...
+def test_text_size__outside_ascii() -> None:
+    # The wire counts bytes, not characters.
+    assert batching.text_size(text="Grüezi") == 7
+
+
+def test_text_size__escaped_characters() -> None:
+    assert batching.text_size(text='a\n"b\\c') == 9
+
+
+def _split(
```
michal-lightly: Optional: I find it weird that the tested function is hidden in a helper. I would repeat it instead in every test body, so that all parameters are explicit.

michal-lightly (approved):

PR #2387, opened 2026-09-17
Compose remote embedders from advertised capabilities

iunir:
## What has changed and why?

2/6 of the active remote-embedder client stack. Base: `iunir-lig-10641-remote-batching` (#2386).

- Cache dynamically composed embedder classes so `isinstance` dispatch matches the capabilities advertised by `/v1/describe`.
- Reject capability sets that the client cannot route or LightlyStudio cannot resolve.
- Keep routing and registry policy as caller inputs rather than coupling this helper to either implementation.

## How has it been tested?

8 focused unit tests cover one and several capabilities, class reuse, unsupported capability sets, and readable capability names. The complete remote test suite and backend static checks pass at the top of the stack.

- [ ] Yes
- [x] Not needed (internal helper)

## Summary by CodeRabbit

* **New Features**
  * Added support for composing remote embedders from server-advertised capabilities.
  * Added validation for unsupported, unroutable, or unresolvable capabilities.
  * Ensured consistent naming and reuse for identical capability combinations.

* **Documentation**
  * Clarified error reporting when advertised capabilities cannot be resolved.

* **Tests**
  * Added coverage for routing, composition, reuse, unsupported capabilities, duplicate advertisements, and combined capabilities.


lightly_studio/src/lightly_studio/embed/remote/composition.py
```diff
@@ -0,0 +1,81 @@
...
+import functools
+from collections.abc import Sequence
+
+from lightly_studio_serve.embedder import Capability, Embedder
+
+from lightly_studio.embed.remote.errors import RemoteEmbedderCapabilityError
+
+
+@functools.cache
+def composed_class(bases: tuple[type[Embedder], ...], name: str) -> type[Embedder]:
```
michal-lightly: I suggest `name` not to be a parameter - it can be deduced from bases (ok to make it specific with RemoteXEmbedder in the name).

michal-lightly: Optional: I really think that the responsibility split here is poor (bad AI). My suggestion would be (if that's possible at all) to move here as much as possible from `_embedder_for` function:
```py
_RESOLVABLE_CAPABILITIES = ...

def compose_remote_embedder_class(
    capabilities: Set[Capabilities],
    capability_to_base: dict[Capability, type[Embedder]],
) -> type[Embedder]:
    ...
``` 

Then the caller can simplify to:
```py
def _embedder_for(
    transport: RemoteTransport,
    description: DescribeResponse,
    owned_client: httpx.Client | None,
) -> RemoteEmbedder:
    composed_class = cast(
        "type[RemoteEmbedder]",
        composition.compose_remote_embedder_class(
            capabilities=description.capabilities,
            capability_to_base=_CAPABILITY_TO_BASE,
        )
    )
    return composed_class(
        transport=transport,
        spec=EmbeddingSpaceSpec(space_key=description.space_key, dimension=description.dimension),
        limits=description.limits,
        owned_client=owned_client,
    )
```

The key is to get `mypy` happy. On the `composition.py` side we need to avoid any reference to `RemoteEmbedder`, and on the `_embedder_for` side we use `cast`.

michal-lightly: Afterwards, we can ask AI to simplify and polish namings (e.g. factory is a good term; `check_routable` would need to become private `_check_routable`, etc.)

lightly_studio/tests/embed/remote/test_composition.py
```diff
@@ -0,0 +1,119 @@
...
+    assert first is second
+
+
+def test_compose_remote_embedder_class__nothing_routable() -> None:
+    with pytest.raises(RemoteEmbedderCapabilityError) as error:
+        composition.compose_remote_embedder_class(
+            capabilities=[Capability.IMAGE_PATH], capability_to_base=_CAPABILITY_TO_BASE
+        )
+
+    assert "routes to text, image_bytes, video_bytes" in str(error.value)
```
michal-lightly: Optional: I believe `pytest.raises` has a `match=` parameter that checks the error message.

michal-lightly (approved):

michal-lightly:
@codex review

michal-lightly:
@coderabbit review

lightly_studio/src/lightly_studio/embed/remote/composition.py
iunir: Done — the composed name is now derived from `bases` inside `_composed_class`, not passed in.

iunir: Applied — added the public `compose_remote_embedder_class(capabilities, capability_to_base)` factory that does the validation + composition, kept `composition.py` free of any `RemoteEmbedder` reference (`capability_to_base: Mapping[Capability, type[Embedder]]`), and `_embedder_for` now just casts and constructs, per your snippet.

iunir: Done — `check_routable`/`names` are now private (`_check_routable`/`_names`), and `_RESOLVABLE_CAPABILITIES` moved into `composition.py` next to the check that uses it.

lightly_studio/tests/embed/remote/test_composition.py
iunir: Applied — switched to `pytest.raises(..., match=re.escape(...))`.

lightly_studio/src/lightly_studio/embed/remote/composition.py
```diff
@@ -0,0 +1,111 @@
...
+        RemoteEmbedderCapabilityError: If nothing advertised is routable, or if the only
+            routable capabilities are ones that the registry cannot resolve.
+    """
+    if not routable:
+        raise RemoteEmbedderCapabilityError(
+            f"The embedding server advertises {_names(capabilities=advertised)}. This client "
+            f"routes to {_names(capabilities=routes_to)}."
+        )
+    if not set(routable) & set(_RESOLVABLE_CAPABILITIES):
+        raise RemoteEmbedderCapabilityError(
```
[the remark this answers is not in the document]
iunir: Fixed — RemoteEmbedderCapabilityError's docstring in errors.py now documents the routable-but-unresolvable case too.

lightly_studio/src/lightly_studio/embed/remote/composition.py
```diff
@@ -0,0 +1,111 @@
...
+
+from lightly_studio_serve.embedder import Capability, Embedder
+
+from lightly_studio.embed.remote.errors import RemoteEmbedderCapabilityError
+
+# The capabilities that `EmbedderRegistry` can hand back. Its `_CAPABILITY_TO_TYPE` has no
+# `VIDEO_BYTES` entry, so a video-only server gives an embedder that nothing in
+# LightlyStudio ever asks for, and `register` refuses it with "implements no capability".
+# TODO(Iunir, 09/2026): Remove this constant when the registry gains a `VIDEO_BYTES` entry.
+_RESOLVABLE_CAPABILITIES = (Capability.TEXT, Capability.IMAGE_BYTES)
```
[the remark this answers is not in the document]
iunir: Valid concern, but keeping it as-is for now: Michal's earlier comment on this thread asked specifically to move this constant into composition.py and off the caller, to shrink _embedder_for to a cast + one call. Making it caller-supplied again would reopen that. The drift risk is bounded — the TODO above already flags this as temporary, tied to EmbedderRegistry gaining a VIDEO_BYTES entry, and it's the only place that needs updating when that happens. Deriving it from the registry directly would create an import-direction problem (embed.remote would depend on embedder_registry, which is the wrong way round). Happy to revisit if the registry's capability set grows more volatile.

Issue #2171, opened 2026-09-03
[FEAT] Add instance segmentation to dataset.evaluate()


lorinczszabolcs:
### 💡 Is your feature request related to a problem?

`dataset.evaluate()` compares an annotation source with the ground truth for object detection, classification, and semantic segmentation. It does not support instance segmentation. LightlyStudio does import segmentation-mask annotations, for example from COCO and Pascal VOC. The evaluation framework also defines an instance-segmentation task type. A user with instance-segmentation predictions still cannot measure them against the ground truth from the SDK or the GUI. For the other tasks the report gives per-image true and false positives, false negatives, the IoU of each matched mask, and a confusion matrix. Instance segmentation gives none of these today.

### 🧰 Describe the solution you'd like

Add `dataset.evaluate().instance_segmentation(...)`. It works like the other task methods. It matches each predicted mask to a ground-truth mask by mask IoU. It uses the same greedy matcher as object detection. It stores per-image true and false positives, false negatives, and the IoU of each match. It also adds instance segmentation to the evaluation dialog in the GUI and to the confusion matrix. The confusion matrix reuses the object-detection method that pairs each ground-truth label with a prediction label. The feature reuses the existing matcher and the metric-record builders, so it needs no new table and no migration.

### 🛠 Alternatives you've considered

- Evaluate the instance masks as semantic segmentation. This is not correct. Semantic segmentation measures a per-pixel overlap and does not match instances. So it loses the per-instance identity that instance segmentation needs.
- Leave instance segmentation out. This keeps the instance-segmentation task type unimplemented in the evaluation path. The annotation type itself already works for import and display.

### 📝 Additional context

The annotation type and the import path already work. Only the evaluation path is missing. The instance-segmentation task type is defined, but it is not connected to the evaluation config, the task dispatch, or a metric module. The confusion-matrix route also rejects the task type.

A working prototype is ready. It covers the backend, the Python API, and the GUI, with tests and docs. So this is a small change to accept, not new work to build. The work can go in small pull requests: (1) make the detection matcher reusable across tasks, (2) add the mask-IoU metric, (3) add the Python API method, and (4) add the GUI.

I would be happy to contribute, let me know your thoughts.

PushpakAg:
It looks like the current `dataset.evaluate()` function doesn't handle instance segmentation, which seems like a gap given the existing support for other segmentation types. I'd check how the current evaluation methods are structured to see where instance segmentation could fit in.

[...]

PR #2398, opened 2026-09-17
refactor: make the detection matcher reusable across tasks

lorinczszabolcs:
## What has changed and why?

This is the first of three stacked PRs that add instance-segmentation evaluation (#2171). It is a pure refactor of the object-detection metric code with no behavior change.

`match_with_iou_matrix` matched predictions to ground truths for object detection only. This change lets a second task reuse the same greedy matcher. It adds a small `Matchable` protocol (an annotation id and a confidence) and retypes `match_with_iou_matrix` to accept it, so `BoundingBox` and a later `InstanceMask` both satisfy it. It also promotes the two metric-record builders `get_sample_metric_records` and `get_annotation_metric_records` to public names, so the next PR can build the same per-sample and per-annotation records. No call site changes behavior.

## How has it been tested?

- `pytest tests/evaluation/test_object_detection.py` passes (31 tests). The change keeps the same behavior, so the existing detection tests cover it.

- [ ] Yes

## Summary by CodeRabbit

* **New Features**
  * Added support for matching both bounding boxes and instance masks using shared annotation and confidence information.
  * Exposed public helpers for generating sample-level and annotation-level object-detection metric records.


JonasWurst (approved):
LGTM 🙂

PR #2418, opened 2026-09-18
Don't overwrite PostHog identity for logged in users

mihnea-necsulescu:
## What has changed and why?

#2136 introduced a bug that caused logged in users to not be properly identified anymore when opening a dataset because their identity was overwritten with the unique installation id. This PR makes that overwrite conditional, i.e., it is skipped for enterprise users that are already logged in.

## How has it been tested?

New tests + manually 

- [ ] Yes

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserved existing PostHog identities for identified users during initialization.
  * Avoided replacing email-based identities with an installation ID.
  * Applied the current installation ID when no user identity exists or when an outdated installation ID is detected.


IgorSusmelj (approved):
LGTM

PR #2426, opened 2026-09-18
Improve Mcap file access

horatiualmasan:
## What has changed and why?

Serving a single camera frame from a recording in S3 fetched far more data than the frame
needed. The reader opened the file with the filesystem's default read cache, which is sized
for reading a whole recording end to end, so every read that missed the cache pulled a large
block — and serving a frame is a handful of small reads scattered across the file.

- `McapFileReader` now takes a `read_pattern`. `SEQUENTIAL` (the default) keeps the current
  behaviour, which is what indexing a whole recording wants. `RANDOM` opens the file with an
  fsspec `readahead` cache and a small block, for opening a remote file and reading a few
  messages from it.
- The camera-frame service opens its cached readers with `RANDOM`, because serving one frame
  reads only the summary and the chunk that holds the frame.

## How has it been tested?

New tests

- [ ] Yes

## Summary by CodeRabbit

* **New Features**
  * Added configurable sequential and random-access reading for MCAP files.
  * Random-access camera frame retrieval now uses read-ahead caching to improve remote data access.
  * Added a 64 KiB read-ahead cache for random access while preserving existing sequential behavior.


lightly_studio/src/lightly_studio/services/recording_service/get_camera_frame.py
```diff
@@ -62,7 +65,9 @@ def _get_cached_reader(uri: str) -> McapFileReader:
         endpoint_url = os.environ.get("AWS_ENDPOINT_URL")
         if endpoint_url:
             storage_options = {"client_kwargs": {"endpoint_url": endpoint_url}}
-    reader = McapFileReader(uri, storage_options=storage_options)
+    # A frame is a few reads spread over the file, so a small read cache keeps opening
+    # a remote recording fast. A large one would fetch tens of megabytes per read.
+    reader = McapFileReader(uri, storage_options=storage_options, read_pattern=ReadPattern.RANDOM)
```
horatiualmasan, line 70: We should see if this will affect the playback speed or if we need to increase the chunk size

ikondrat (approved):

